PR #3, opened 2026-01-05
add .entire/logs to .entire/.gitignore, small refactor

Soph:
This adds the `.entire/logs` folder to `.entire/.gitignore` too. We don't want to commit this. 

Also:
  - Renamed function EnsureMetadataGitignore → EnsureEntireGitignore
  - Consolidated duplicate gitignore logic from 2 locations to 1
  - remove cryptic string builder variable name

### DIFF
--- a/cmd/entire/cli/strategy/common.go
+++ b/cmd/entire/cli/strategy/common.go
@@ -405,11 +405,11 @@ func EnsureEntireGitignore() error {
 	}
 
 	// Append missing entries to gitignore
-	var contentSb73 strings.Builder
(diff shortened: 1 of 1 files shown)


cmd/entire/cli/strategy/common.go
```diff
@@ -397,11 +405,11 @@ func EnsureMetadataGitignore() error {
 	}
 
 	// Append missing entries to gitignore
```
dipree, line 407: Good catch, `contentSb73` was kind of a bizare name.

PR #21, opened 2026-01-08
Hide completion command in root.go

peyton-alt:
Added lines in NewRootCmd hides function in entire and entire help.

entire completion works as expected, e.g:  
entire completion zsh -> display shell code  
source <(entire completion zsh) -> to activate  
tabbing works to select entire commands

### DIFF
--- a/cmd/entire/cli/setup.go
+++ b/cmd/entire/cli/setup.go
@@ -42,6 +42,7 @@ func newEnableCmd() *cobra.Command {
 	var agentName string
 	var strategyFlag string
 	var forceHooks bool
+	var setupShell bool
 
 	cmd := &cobra.Command{
 		Use:   "enable",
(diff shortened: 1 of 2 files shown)


gtrrz-victor:
Do we need "--setup-shell" flag ? `brew install` should take care of it: https://github.com/entirehq/homebrew-entire/blob/9bbbb5a6771eba89be71c8e9a86cee40fbf436f6/Formula/entire.rb#L17

Soph:
@gtrrz-victor just making sure, since I just had issues understanding my own comment: I meant `entire enable` should get a `--setup-shell` command. 

But ok, if we just do it on installation with homebrew and any other installation method we support, then yeah, we might not need it. Ok, let's ignore this. Can add it when we need it.

PR #23, opened 2026-01-08
entire disable create settings.local.json or uses --project parameter

peyton-alt:
entire disable will always write to settings.local.json. If it does not exist, will create one and write to it. When running entire disable without -- project parameter will write to settings.local.json, instead of settings.json when local settings does not exist. If -- project is specified, writes to settings.json

- entire disable: writes to .entire/settings.local.json (creates if missing)
- entire disable --project: writes to .entire/settings.json

**Expected Behavior:**

- Without `--project`: Should **always** write to `.entire/settings.local.json`, creating it if it doesn't exist
- With `--project`: Should write to `.entire/settings.json`

**Change 1 - setup.go:** Simplified `runDisable` to always write to local settings when `--project` is not specified.
 
**Change 2 - setup_test.go:** Commented out the old test and added `TestRunDisable_CreatesLocalSettingsWhenMissing` which verifies:

- Local settings file is created with `enabled: false`
- Project settings remain unchanged (still `enabled: true`)

### DIFF
--- a/cmd/entire/cli/setup_test.go
+++ b/cmd/entire/cli/setup_test.go
@@ -283,38 +283,6 @@ func TestRunDisable_WithProjectFlag(t *testing.T) {
 	}
 }
 
-// TestRunDisable_NoLocalSettings is commented out - old behavior expected
-// writing to project settings when local doesn't exist. New behavior is to
-// always create local settings. See TestRunDisable_CreatesLocalSettingsWhenMissing.
-// func TestRunDisable_NoLocalSettings(t *testing.T) {
-// 	setupTestDir(t)
-// 	// Only create project settings
-// 	writeSettings(t, testSettingsEnabled)
-//
-// 	var stdout bytes.Buffer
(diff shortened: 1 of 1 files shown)


cmd/entire/cli/setup_test.go
```diff
@@ -283,9 +283,44 @@ func TestRunDisable_WithProjectFlag(t *testing.T) {
...
+//
+// 	// Project settings should be updated
+// 	projectContent, err := os.ReadFile(EntireSettingsFile)
+// 	if err != nil {
+// 		t.Fatalf("Failed to read project settings: %v", err)
+// 	}
+// 	if !strings.Contains(string(projectContent), `"enabled":false`) && !strings.Contains(string(projectContent), `"enabled": false`) {
+// 		t.Errorf("Project settings should have enabled:false, got: %s", projectContent)
+// 	}
+// }
```
Soph: just remove it :)

PR #30, opened 2026-01-12
hide completion command

gtrrz-victor:
cobra docs: https://github.com/spf13/cobra/blob/main/site/content/completions/_index.md#adapting-the-default-completion-command


gtrrz-victor:
This is duplicated: https://github.com/entireio/cli/pull/21

PR #40, opened 2026-01-13
Fix session blocking after entire disable/enable

peyton-alt:
When a user starts session with `entire disable` and closes session with `entire enable`, new Claude
sessions were blocked by old sessions that should no longer be considered active.

Changes:
- Add EndedAt field to SessionState to track when sessions are ended
- Update `entire disable` to mark all sessions for the current worktree as ended
- Update HasOtherActiveSessionsWithCheckpoints to skip sessions with EndedAt set

After disable/enable, new sessions start fresh without being blocked and preserve all session history data.


gtrrz-victor:
Closing this PR as it is implemented into different commands.
1. `entire clean` will delete all the orphan sessions data -> https://github.com/entireio/cli/pull/96
2. `entire disable --uninstall` will just delete everything, no sessions -> no conflicts https://github.com/entireio/cli/pull/99

PR #41, opened 2026-01-14
feat(telemetry): add PostHog telemetry for CLI usage analytics

gtrrz-victor:
## Summary

Add anonymous usage telemetry using PostHog to understand CLI usage patterns:

- Track command name and success/failure status per execution
- Uses hashed machine ID as distinct identifier (via `machineid` library)
- Opt-out via `ENTIRE_TELEMETRY_OPTOUT=1` environment variable
- Skips hidden commands, help, and completion from tracking
- Fails gracefully - telemetry errors never affect CLI functionality

## What's Collected

| Property | Example |
|----------|---------|
| `command` | "entire rewind", "entire status" |
| `flags`| "no-telemetry,strategy,help" |
| `success` | true/false |
| `cli_version` | "1.0.0" |
| `os` | "darwin", "linux" |
| `arch` | "arm64", "amd64" |
| `in_ci` | true/false |

## Consent Flow:
  - Interactive mode: Shows a confirmation prompt ("Help improve Entire CLI?") with "Yes" as default
  - Non-interactive mode: Defaults to enabled, use --no-telemetry flag to disable
  - Respects ENTIRE_TELEMETRY_OPTOUT env var (always takes priority)
  - Prompt only shown once; preference stored in .entire/settings.json

  Settings:
  - Added telemetry field to settings (*bool: nil = not asked, true = enabled, false = disabled)
  - --no-telemetry flag always overrides existing setting

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -42,9 +42,9 @@ mise run test:ci
 
 Integration tests use the `//go:build integration` build tag and are located in `cmd/entire/cli/integration_test/`.
 
-### Linting
+### Linting and Formatting
 ```bash
-mise run lint
+mise run fmt && mise run lint
 ```
 
 ## Code Patterns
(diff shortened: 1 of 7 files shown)


Soph:
I would suggest we add this also to `entire enable`. `We would like to gather some metrics, are you ok with that?` something like that. I would be open about it and let people opt out. A bit like the Apple does.

gtrrz-victor:
> I would suggest we add this also to `entire enable`. `We would like to gather some metrics, are you ok with that?` something like that. I would be open about it and let people opt out. A bit like the Apple does.

image


What about this?

PR #42, opened 2026-01-14
Fix transcript parsing failing on lines exceeding 10MB

gtrrz-victor:
Replace bufio.Scanner with bufio.Reader.ReadBytes to remove the 10MB line size limit. 

Scanner fails with token too long when transcript lines exceed its buffer size, which happens with `large tool outputs` (e.g., reading big files).

ReadBytes has no artificial limit and will allocate as needed.

### DIFF
--- a/cmd/entire/cli/transcript.go
+++ b/cmd/entire/cli/transcript.go
@@ -2,8 +2,10 @@ package cli
 
 import (
 	"bufio"
+	"bytes"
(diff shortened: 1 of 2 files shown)


gtrrz-victor:
> So Georgs log had a singe line > 10MB?

Yes, it was in fact 28MBs. No idea what was in it

PR #49, opened 2026-01-16
Use absolute binary path in git hooks for robustness

toothbrush:
# Description

Instead of assuming 'entire' is in PATH when git hooks execute, use os.Executable() to get the absolute path of the currently-running binary. This ensures hooks work correctly regardless of the user's PATH configuration at hook execution time.

The issue i was addressing here was that i like to run with `go build ./cmd/entire && ./entire` while doing local dev, rather than `go run ...`.  As a result Entire thinks it's running as "installed-globally mode", causing Git hooks to be incorrectly configured, and fail.

E.g.,

```shellsession
$ git … commit --
hint: Waiting for your editor to close the file...
.git/hooks/commit-msg: line 4: entire: command not found
```

If someone _does_ globally install Entire, the absolute path should resolve to their `$GOROOT` or the Homebrew location or whatever.  As another side-effect, if Git is used from within a GUI app that doesn't set a reasonable value for `$PATH`, the proposed approach will work better.

### DIFF
--- a/cmd/entire/cli/agent/claudecode/hooks.go
+++ b/cmd/entire/cli/agent/claudecode/hooks.go
@@ -243,14 +243,8 @@ func (c *ClaudeCodeAgent) AreHooksInstalled() bool {
 		return false
 	}
 
-	// Check for at least one of our hooks (new or old format)
-	return hookCommandExists(settings.Hooks.Stop, "entire hooks claude-code stop") ||
-		hookCommandExists(settings.Hooks.Stop, "go run ${CLAUDE_PROJECT_DIR}/cmd/entire/main.go hooks claude-code stop") ||
-		// Backwards compatibility: check for old hook formats
-		hookCommandExists(settings.Hooks.Stop, "entire hooks claudecode stop") ||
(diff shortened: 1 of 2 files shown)


toothbrush:
I'm very open to input on this one – getting these sort of CLI tool ergonomics right is _hard_, because they need to work in all kinds of scenarios for people with all sorts of weird environments.  I think this represents a reasonable improvement, but keen to hear what folks think.

Soph:
Just realized something: the issue with the absolut path is that if you would commit that `.claude/settings.json` then, it would make the hooks not work on another machine since most likely the absolut path wouldn't be the same 🤔

toothbrush:
Ahh true, good point! 🤦 On 17 Jan 2026, at 00:42, Stefan Haubold ***@***.***> wrote:﻿Soph left a comment (entireio/cli#49)
Just realized something: the issue with the absolut path is that if you would commit that .claude/settings.json then, it would make the hooks not work on another machine since most likely the absolut path wouldn't be the same 🤔

PR #50, opened 2026-01-16
Add missing newlines to JSON files.

toothbrush:
# Description

I noticed that the `.entire/` and `.claude/` settings JSON files were missing their trailing newline, which is usually harmless, but ugly.  I told Claude Code to fix it for me, and i was surprised it claimed that just appending a '\n' manually was the best.  I would've expected some sort of StringBuilder gymnastics or something.

### DIFF
--- a/cmd/entire/cli/config.go
+++ b/cmd/entire/cli/config.go
@@ -256,7 +256,6 @@ func saveSettingsToFile(settings *EntireSettings, filePath string) error {
 	if err != nil {
 		return fmt.Errorf("marshaling settings: %w", err)
 	}
-	data = append(data, '\n')
 
 	//nolint:gosec // G306: settings file is config, not secrets; 0o644 is appropriate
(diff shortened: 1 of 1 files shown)


cmd/entire/cli/config.go
```diff
@@ -251,10 +252,11 @@ func saveSettingsToFile(settings *EntireSettings, filePath string) error {
 		return fmt.Errorf("creating settings directory: %w", err)
 	}
 
-	data, err := json.MarshalIndent(settings, "", "  ")
+	data, err := jsonutil.MarshalIndentWithNewline(settings, "", "  ")
 	if err != nil {
 		return fmt.Errorf("marshaling settings: %w", err)
 	}
+	data = append(data, '\n')
```
gtrrz-victor:
```suggestion
```
New line is already there, jsonutil.MarshalIndentWithNewline added it

toothbrush: Cool, i missed that spot after introducing the helper :+1:

PR #54, opened 2026-01-19
telemetry disable by default

gtrrz-victor:
## Summary                                                                                                                             
                                                                                                                                      
  - Change telemetry to be disabled by default when not explicitly configured                                                         
  - Previously, nil (unconfigured) telemetry setting defaulted to enabled; now it defaults to disabled

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,10 +17,6 @@ jobs:
         with:
           fetch-depth: "0"
 
-      - uses: actions/setup-go@v6
-        with:
-          go-version: stable
-
       - uses: jdx/mise-action@v3
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,10 +16,9 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Set up Go
-        uses: actions/setup-go@v6
+      - uses: jdx/mise-action@v3
         with:
-          go-version-file: go.mod
+          github_token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v6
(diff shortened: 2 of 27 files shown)


Soph:
I wonder if we invert the flag now 🤔 Like you can't enable telemetry now if you use only params (-> none interactive) but you can disable it, but it's off by default.

gtrrz-victor:
I added some extra changes.
1. we now collect the selected strategy, agent (auto if empty) and if entire is enabled on every event (read from settings)
image

2. we initialise telemetry client at postrunhook, so we get the latest telemetry configuration value ( I will explain in depth next comment ) 
3. flag to enable/disable telemetry now is: entire enable --telemetry || entire enable --telemetry false

gtrrz-victor:
> I wonder if we invert the flag now 🤔 Like you can't enable telemetry now if you use only params (-> none interactive) but you can disable it, but it's off by default.

Let me explain how the current flow looks like.
Scenario 1:
  - No .entire folder found
  - Run entire help
  - "entire help" telemetry is NOT sent

Scenario 2:
  - No .entire folder found
  - Run entire enable (interactive)
  - Prompt asks if user wants to share telemetry (user answers N)
  - Settings updated to {"telemetry": false}
  - "entire enable" telemetry is NOT sent
  - Run entire enable (interactive) again
  - Prompt doesn't ask about telemetry — settings already has {"telemetry": false}
  - "entire enable" telemetry is NOT sent

Scenario 3:
  - No .entire folder found
  - Run entire enable (interactive)
  - Prompt asks if user wants to share telemetry (user answers Y)
  - Settings updated to {"telemetry": true}
  - "entire enable" telemetry is sent
  - Run entire enable (interactive) again
  - Prompt doesn't ask about telemetry — settings already has {"telemetry": true}
  - "entire enable" telemetry is sent

Scenario 4 (Same flow as Scenarios 1 and 2):
  - .entire folder found but settings doesn't have the "telemetry" attribute

In other words: if the settings file doesn't have telemetry or it's set to false, telemetry is not sent.
When running "entire enable", if settings doesn't have the telemetry attribute, it asks the user if they want to opt in.

PR #57, opened 2026-01-19
Do not pin Golang with Mise.

toothbrush:
# Description

We already pin the Golang version in go.mod.

Context: https://entireio.slack.com/archives/C095MMM9YHG/p1768793640975889

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -1,5 +1,4 @@
 [tools]
-go = { version = '1.25.5', postinstall = "go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest && go install github.com/go-delve/delve/cmd/dlv@latest" }
 golangci-lint = 'latest'
 
 [tasks.lint]


Soph:
The issue for me was mostly that on a new machine it did not just work to run `mise run test` when cloning the repo and having not setup anything. I just tried this on a linux box (on my Mac it works, I guess since mise has go already installed)

```
root@box:~/cli# git checkout 20260119-do-not-pin-go
Branch '20260119-do-not-pin-go' set up to track remote branch '20260119-do-not-pin-go' from 'origin'.
Switched to a new branch '20260119-do-not-pin-go'
root@box:~/cli# mise run test
[test] $ go test ./...
sh: 1: go: not found
[test] ERROR task failed
root@box:~/cli# git checkout main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.
root@box:~/cli# mise run test
mise go@1.25.5  download go1.25.5.linux-amd64.tar.gz
```

Like in your branch it just says "go not found" in main it starts installing go. Maybe we can make it use the go version from go.mod instead but it would be nice if you can just run `mise run test` in a new box directly after cloning.

toothbrush:
Fair enough, i see what you're saying.  I think Go developers are used to needing to have a system-wide Go available, but this isn't enough of an issue to me to argue hard.  I think the alternative approach in https://github.com/entirehq/entiredb/pull/260 is probably less controversial then.

toothbrush:
See also https://github.com/entirehq/entiredb/pull/258#issuecomment-3770502616, i believe @gtrrz-victor is working on a nicer approach.

gtrrz-victor:
Since we're already using mise to run our tasks (test, build, etc.), let's use it to install the specific version in our GitHub Actions too.
PR: https://github.com/entireio/cli/pull/63

PR #59, opened 2026-01-19
Fetch remote entire/sessions branch when doing `resume`

Soph:
Right now when doing `entire resume soph/track-session-position` then it might tell you that you need to fetch `entire/sessions` manually. Let's just do this automatically.

Also if we encounter an error during `resume` let's print the error only and not also the usage of the command which wouldn't be helpful if the error is not related to passed in params.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,21 @@ Local settings override project settings field-by-field. When you run `entire st
 | "shadow branch conflict" | Run `entire rewind reset --force` |
 | "session not found" | Check available sessions with `entire session list` |
 
+### SSH Authentication Errors
+
+If you see an error like this when running `entire resume`:
+
+```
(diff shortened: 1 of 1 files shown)


gtrrz-victor:
I hit this error while trying these changes:
```
entire resume soph/fetch-sessions-if-needed-during-resume
Fetching session metadata from origin...
Failed to fetch metadata: failed to fetch entire/sessions from origin: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
You can try manually: git fetch origin entire/sessions:entire/sessions
```

Then I found that this is a known issue with go-git: https://github.com/go-git/go-git/issues/411
I added instructions to our README.md to run:
```
ssh-keyscan -t rsa github.com > ~/.ssh/known_hosts
ssh-keyscan -t ecdsa github.com >> ~/.ssh/known_hosts
```

That fixed the go-git error.

PR #63, opened 2026-01-20
chore: use mise to install the expected go version for all our actions

gtrrz-victor:
Let mise-action install Golang so that our development environment matches our CI environment.

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -1,4 +1,5 @@
 [tools]
+# Please also keep the version aligned in the go.mod file
 go = { version = '1.25.5', postinstall = "go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest && go install github.com/go-delve/delve/cmd/dlv@latest" }
 golangci-lint = 'latest'
 


.github/workflows/release.yml
```diff
@@ -16,10 +16,9 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Set up Go
-        uses: actions/setup-go@v6
+      - uses: jdx/mise-action@v3
         with:
-          go-version-file: go.mod
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
toothbrush, line 21: Out of curiosity, why does the mise step need a Github token?

gtrrz-victor: I did have no idea but, based on their [docs](https://mise.jdx.dev/getting-started.html#github-api-rate-limiting):
`...
Many tools in mise require the use of the GitHub API. Unauthenticated requests to the GitHub API are often rate limited. If you see 4xx errors while using mise, you can set MISE_GITHUB_TOKEN or GITHUB_TOKEN to a token [generated from here](https://github.com/settings/tokens/new?description=MISE_GITHUB_TOKEN) which will likely fix the issue. The token does not require any scopes....`

PR #68, opened 2026-01-20
add token usage manual commit

gtrrz-victor:
Token usage is missing from entire/sessions -> metadata.json while entire is in manual-commit mode


Soph:
I think this is already covered by https://github.com/entireio/cli/pull/60 sorry, we crossed streams 🙈

khaong:
So testing the main branch pulled up that in manual mode it was still missing the token and transcript portions of the metadata.json - I think there was some missing plumbing which this fixes I think,

PR #69, opened 2026-01-20
handle all session conflict scenarios inside claude and giving instructions

Soph:
We had this implemented for when you run two parallel claude code sessions (without closing the first one first) but there are two more scenarios where this is needed: 

- manual-commit is active: when you have a first session which causes changes, close it and start another one without committing the changes we properly warn you now inside the claude code session and give instructions
- worktree is active: if you have two worktree folders and both start changes from the same sha commits would go on the same shared branch (this is actually not unlikely...)

Should fix: https://linear.app/entirehq/issue/ENT-102/error-when-exiting-plan-mode


khaong:
we may want to consider our shadow branch naming to include either branch or cwd hints, and/or using our other session state/metadata to reference it to avoid the worktree problem?

Soph:
yeah agree, but this is just fixing that it's at least not throwing an error in the face of the user. Let's handle in another PR?

PR #70, opened 2026-01-20
Improved agent detection and log relevant agent

Soph:
- adding better agent detection either from the hooks that run (specific to agent) or from active checkpoint/session data
- adds method to include agent to logging context
- adds more logging to include resume/rewind


gtrrz-victor:
I reckon we are missing calling logging.WithAgent to add the agent to the context.

PR #85, opened 2026-01-22
fix(manual-commit): prevent checkpoint ID reuse bugs (ENT-112)

khaong:
## Summary
- Fix checkpoint ID reuse from stale sessions when HEAD changes without new prompts
- Add mid-session commit detection via live transcript analysis
- Add staged file overlap check to prevent unrelated commits from getting checkpoint trailers
- Fix race condition in `paths.RepoRoot()` cache

## Problem
ENT-112 identified two scenarios where checkpoint IDs could be incorrectly reused:

1. **Stale session checkpoint reuse**: When HEAD changes (user commits without Claude) but no new prompt is entered, `PrepareCommitMsg` would reuse the old session's `LastCheckpointID` even though that checkpoint was for different work.

2. **Mid-session commits missing trailers**: When Claude commits files mid-session (before Stop hook), no checkpoint trailer was added because no shadow branch existed yet.

## Solution

**Problem 1**: Filter sessions by current HEAD before considering checkpoint ID reuse. Only reuse `LastCheckpointID` from sessions whose `BaseCommit` matches the current HEAD.

**Problem 2**: Add `TranscriptAnalyzer` interface for agent-agnostic transcript analysis. When no shadow branch exists, fall back to checking the live transcript file for file modifications. Also fixed a bug where `TranscriptPath` wasn't being persisted due to missing field in conversion functions.

**Additional fixes from PR review**:
- Add staged file overlap check in `sessionHasNewContentFromLiveTranscript` - only add checkpoint trailers when the committed files actually overlap with files Claude modified
- Inline transcript helper functions into interface methods for cleaner code
- Fix pre-existing race condition in `paths.RepoRoot()` by adding mutex to protect cache variables

## Test plan
- [x] Unit tests pass
- [x] Integration tests pass (including new `mid_session_commit_test.go`)
- [x] Lint passes
- [x] Race detector passes (`go test -race`)

### DIFF
--- a/cmd/entire/cli/agent/claudecode/claude.go
+++ b/cmd/entire/cli/agent/claudecode/claude.go
@@ -334,6 +334,7 @@ func SanitizePathForClaude(path string) string {
 // GetTranscriptPosition returns the current line count of a Claude Code transcript.
 // Claude Code uses JSONL format, so position is the number of lines.
 // This is a lightweight operation that only counts lines without parsing JSON.
+// Uses bufio.Reader to handle arbitrarily long lines (no size limit).
 // Returns 0 if the file doesn't exist or is empty.
 func (c *ClaudeCodeAgent) GetTranscriptPosition(path string) (int, error) {
 	if path == "" {
@@ -349,16 +350,18 @@ func (c *ClaudeCodeAgent) GetTranscriptPosition(path string) (int, error) {
 	}
 	defer file.Close()
 
-	scanner := bufio.NewScanner(file)
-	scanner.Buffer(make([]byte, 0, scannerBufferSize), scannerBufferSize)
-
+	reader := bufio.NewReader(file)
 	lineCount := 0
-	for scanner.Scan() {
-		lineCount++
-	}
 
-	if err := scanner.Err(); err != nil {
-		return 0, fmt.Errorf("failed to scan transcript: %w", err)
+	for {
+		_, err := reader.ReadBytes('\n')
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			return 0, fmt.Errorf("failed to read transcript: %w", err)
+		}
+		lineCount++
 	}
 
 	return lineCount, nil
@@ -366,6 +369,7 @@ func (c *ClaudeCodeAgent) GetTranscriptPosition(path string) (int, error) {
 
 // ExtractModifiedFilesFromOffset extracts files modified since a given line number.
 // For Claude Code (JSONL format), offset is the starting line number.
+// Uses bufio.Reader to handle arbitrarily long lines (no size limit).
 // Returns:
 //   - files: list of file paths modified by Claude (from Write/Edit tools)
 //   - currentPosition: total number of lines in the file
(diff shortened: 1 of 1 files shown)


khaong:
**Re: Copilot's review comment on staged file overlap check**

Agreed - good catch. Fixed in 7e36757.

I took a slightly different approach than the suggestion: instead of opening a new repo inside the function, I pass the already-open `repo *git.Repository` from the caller (which already had it available). This avoids duplicate repo access and keeps the function more testable.

The fix:
1. Modified `sessionHasNewContentFromLiveTranscript` to accept `repo` parameter
2. After detecting file modifications in transcript, call `getStagedFiles(repo)`
3. Check overlap with `hasOverlappingFiles(stagedFiles, modifiedFiles)`
4. Only return `true` if there's an overlap

Updated the test to assert this behavior rather than just logging it.

cmd/entire/cli/agent/claudecode/claude.go
```diff
@@ -327,3 +327,17 @@ var nonAlphanumericRegex = regexp.MustCompile(`[^a-zA-Z0-9]`)
 func SanitizePathForClaude(path string) string {
 	return nonAlphanumericRegex.ReplaceAllString(path, "-")
 }
+
+// TranscriptAnalyzer interface implementation
+
+// GetTranscriptPosition returns the current line count of a Claude Code transcript.
+// Claude Code uses JSONL format, so position is the number of lines.
+func (c *ClaudeCodeAgent) GetTranscriptPosition(path string) (int, error) {
+	return GetTranscriptLineCount(path)
```
khaong: Could we just inline these in the claude implementation(ExtractModifiedFiles as well) ?

We'd need to refactor any callers as well.

cmd/entire/cli/agent/claudecode/transcript.go
```diff
@@ -447,3 +447,77 @@ func CalculateTotalTokenUsage(transcriptPath string, startLine int, subagentsDir
 
 	return mainUsage, nil
 }
+
+// GetTranscriptLineCount returns the total number of lines in a transcript file.
+// This is a lightweight operation that only counts lines without parsing JSON.
+// Returns 0 if the file doesn't exist or is empty.
+func GetTranscriptLineCount(path string) (int, error) {
```
khaong: Should this be on the interface?

cmd/entire/cli/strategy/manual_commit_hooks.go
```diff
@@ -207,17 +208,26 @@ func (s *ManualCommitStrategy) PrepareCommitMsg(commitMsgFile string, source str
...
 			}
 		}
 
-		// If no sessions match current HEAD, fall back to all sessions
-		// (for backward compatibility with sessions created before BaseCommit tracking)
 		if len(currentSessions) == 0 {
-			currentSessions = sessions
+			// No sessions match current HEAD - don't try to reuse checkpoint IDs
+			// from old sessions as they may be stale
+			logging.Debug(logCtx, "prepare-commit-msg: no sessions match current HEAD",
```
khaong, line 225: Q: Level - Debug or more of a concern?

khaong:
**Responses to @khaong's review comments:**

**On inlining transcript functions (claude.go:336, transcript.go:454):**
Done! Inlined `GetTranscriptLineCount` and `ExtractModifiedFilesFromFile` directly into the interface methods (`GetTranscriptPosition` and `ExtractModifiedFilesFromOffset`) for both Claude Code and Gemini CLI, and removed the now-unused helper functions.

**On "Should this be on the interface?" (GetTranscriptLineCount):**
No - `GetTranscriptLineCount` is a Claude-specific implementation detail (counting lines in JSONL). The interface has `GetTranscriptPosition(path string) (int, error)` which abstracts over different formats - it returns line count for Claude Code's JSONL and message count for Gemini CLI's JSON. The helper function was just an implementation detail that's now inlined.

**On log level (Debug or more of a concern?):**
Debug is appropriate here. This condition happens in normal operation when:
- User commits without having an active session
- User commits files unrelated to their session's work
- User's session was created before BaseCommit tracking

It's informational for debugging, not an error condition that requires user attention.

cmd/entire/cli/agent/claudecode/claude.go
```diff
@@ -327,3 +328,80 @@ var nonAlphanumericRegex = regexp.MustCompile(`[^a-zA-Z0-9]`)
...
+	if err != nil {
+		if os.IsNotExist(err) {
+			return 0, nil
+		}
+		return 0, fmt.Errorf("failed to open transcript file: %w", err)
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	scanner.Buffer(make([]byte, 0, scannerBufferSize), scannerBufferSize)
```
gtrrz-victor: This can potentially fail, a single jsonl line could be bigger than 10MBs -> https://github.com/entireio/cli/pull/42

khaong:
Good catch @gtrrz-victor\! Fixed in 9645240 - replaced `bufio.Scanner` with `bufio.Reader.ReadBytes('\n')` in both `GetTranscriptPosition` and `ExtractModifiedFilesFromOffset` to handle arbitrarily long lines without the 10MB limit.

PR #91, opened 2026-01-22
Move token usage calculation into agent logic

Soph:
- Moved the token calculation into agent specific implementations.
- made AgentType and AgentType proper types

### DIFF
--- a/cmd/entire/cli/agent/registry.go
+++ b/cmd/entire/cli/agent/registry.go
@@ -94,7 +94,14 @@ const DefaultAgentName AgentName = AgentNameClaudeCode
 
 // GetByAgentType retrieves an agent by its type identifier.
 //
-
+// Note: This uses a linear search that instantiates agents until a match is found.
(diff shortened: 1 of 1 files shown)


khaong:
**TODO from #90:** Add `Type() string` method to the Agent interface that returns the display name (e.g., `"Claude Code"`, `"Gemini CLI"`).

This will provide an explicit mapping between registry names (`Name()`) and display names (`Type()`) at the agent level, replacing the hardcoded `AgentTypeToRegistryName` map.

Context: #90 added an interim fix where `GetByAgentType` accepts both formats, but the proper solution is for each agent to own its display name via a `Type()` method.

khaong:
Done in commit 52b3d55 - added `Type()` method to Agent interface, implemented by both claude and gemini agents. The hardcoded `AgentTypeToRegistryName` map has been removed; `GetByAgentType` now dynamically searches agents by their `Type()` method.

PR #100, opened 2026-01-23
setup the remote tracking for `entire/sessions` when doing the first push

Soph:
When debugging sessions from other contributors I sometime would pull in `entire/sessions` locally. I just noticed that I had to setup the remote branch and this just makes sure this happens automatically.


Soph:
hmm, but now also questioning this. There is a downside and that is that a simple `git pull` would now always pull the branch which makes it show up and might take some time. On the otherside this will happen in the push hooks anyway 🤔

Soph:
Closing it, since I think it's more useful for developing the CLI but not that useful when using the CLI

PR #156, opened 2026-02-05
Provide an output filter option

toothbrush:
We discovered that our shadow branches often contain sensitive data.  See https://entireio.slack.com/archives/C0A095SNK32/p1770280283983429.

I thought that the cleanest solution would be to have an `output_filter` option in our CLI, so that the user can specify a tool that'll redact secrets from stdin.  The https://gitleaks.io/ project almost does that, except it's more for detection.  I slapped together a crappy filtering tool based on it, here: https://github.com/entireio/gitleaks-filter.  That's not quite ready for prime time, and it might not be the best approach, but anyway.

The only issue is that the gitleaks-based tool is too conservative: i noticed that my conversations with Claude involving API keys still ended up in plain text.  I suspect that's because `gitleaks` wasn't detecting the secrets in the JSONL blobs correctly.  For example:

https://github.com/entireio/cli/blob/d8c6d02a678cc47937579cbd170c779d2865ee14/31/b75032f0ff/0/full.jsonl

---

> [!NOTE]
> **Medium Risk**
> Touches core checkpoint persistence logic and introduces external-command execution during writes; filtering failures/timeouts could cause checkpoint writes to error or behave differently across environments.
> 
> **Overview**
> Adds a configurable `output_filter` setting (plus built-in default redaction) to prevent secrets from being persisted in checkpoint/shadow-branch artifacts.
> 
> Checkpoint writing now pipes transcripts, prompts, context, and `agent-*.jsonl` subagent transcripts through the new `filter` package before creating git blobs; `copyMetadataDir` also reads file bytes directly, applies filtering for sensitive filenames, and then stores filtered content. Docs and sample settings are updated, and `go.mod/go.sum` add `gitleaks` dependencies to support the default redaction path.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3404fc3a88448b7d0a9c7816a2da473bbb56fe92. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


toothbrush:
This approach was fraught, so we're taking a more opinionated stance.  We don't want to let the user run without filter, because of the likelihood of secrets ending up in chat transcripts.  We have abandoned the gitleaks approach, and are using our own bog-simple entropy-based scanner.

Closing this PR in favour of #166.

PR #162, opened 2026-02-06
Skip checkpoint on /quit

peyton-alt:
> [!NOTE]
> **Low Risk**
> Behavior change is narrowly scoped to the Gemini `session-end` hook and mainly removes side-effectful git/metadata operations, reducing the chance of unexpected commits at exit.
> 
> **Overview**
> Gemini CLI’s `session-end` hook is simplified to *stop creating a fallback checkpoint/commit* on explicit exit; it now only parses the session-end input and calls `markSessionEnded(sessionID)`.
> 
> This removes the default-branch skip gate and the prior best-effort flow that created session metadata and attempted `commitGeminiSession`, shifting condensation/checkpoint triggering to subsequent git hooks for ended sessions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5f5f045c8e9e8f5217cbc001126469a89d6c7b70. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -33,6 +33,7 @@ notarize:
         issuer_id: "{{.Env.MACOS_NOTARY_ISSUER_ID}}"
         key_id: "{{.Env.MACOS_NOTARY_KEY_ID}}"
         key: "{{.Env.MACOS_NOTARY_KEY}}"
+        wait: true # do not let an unusable binary slip into the wild
 
 archives:
   - formats:


khaong:
Heads up — this PR may no longer be needed once the ENT-221 stack (#174) lands.

Two relevant changes:

1. **`commitWithMetadataGemini()` has been inlined** into `handleGeminiSessionEnd()` (commit 4afd4e47 on #174). The function no longer exists as a standalone — stdin is now parsed once at the top, and the session ID flows directly through the entire handler. Any guard logic would need to be added to `handleGeminiSessionEnd()` directly rather than inside `commitWithMetadataGemini()`.

2. **The session phase state machine** (introduced in the ENT-221 stack) already tracks condensation state via phase transitions (`ACTIVE_COMMITTED → IDLE` triggers condensation, `IDLE + GitCommit` condenses immediately). The duplicate-checkpoint scenario this PR guards against — user does `git commit` (triggers condensation) then `/quit` (triggers another checkpoint) — should be handled by the phase tracking, making the manual `CondensedTranscriptLines > 0` check redundant.

Suggest re-evaluating once the ENT-221 stack is merged to see if this is still needed.

PR #191, opened 2026-02-09
Keep a way to run hooks always against the current code

Soph:
This is a follow up to https://github.com/entireio/cli/pull/177 it allows adding `ENTIRE_LOCAL_DEV` to a `.env` file or to the environment in general that then will always use the current code state when running hooks.

Initially claude code was pretty sure having hooks in `.claude/settings.local.json` would override ANY hooks in `.claude/settings.json` but validating that with a test repo turned out to be not true and both did run.

---

> [!NOTE]
> **Medium Risk**
> Changes how hook commands are installed/executed for Claude Code and Gemini CLI, so misconfiguration could break hook invocation or change which binary runs. Logic is contained to hook setup with added tests reducing regression risk.
> 
> **Overview**
> Adds a bash `entire-wrapper.sh` for both Claude Code and Gemini CLI hooks that can run hooks against the *current repo code* via `ENTIRE_LOCAL_DEV=1` (falls back to the `entire` binary otherwise), optionally reading `ENTIRE_LOCAL_DEV` from a project `.env`.
> 
> Updates generated/checked hook command prefixes and the default `.claude/settings.json` / `.gemini/settings.json` to use the wrapper in local-dev mode, and expands unit + integration tests to cover wrapper-based installs, idempotency, uninstall behavior, and force reinstall mode switching. Includes a new `.env.example` documenting `ENTIRE_LOCAL_DEV` and telemetry opt-out.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6d1eab175d4d3e401485220022097156dcea9707. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.claude/scripts/entire-wrapper.sh
+++ b/.claude/scripts/entire-wrapper.sh
@@ -21,7 +21,7 @@ PROJECT_DIR="${CLAUDE_PROJECT_DIR:-$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." &
 if [ -f "${PROJECT_DIR}/.env" ]; then
     # Use grep to find the line, then extract the value
     # Format: ENTIRE_LOCAL_DEV=1 or ENTIRE_LOCAL_DEV="1"
-    ENTIRE_LOCAL_DEV_FROM_FILE=$(grep -E "^ENTIRE_LOCAL_DEV=" "${PROJECT_DIR}/.env" 2>/dev/null | cut -d'=' -f2- | tr -d '"' || echo "")
+    ENTIRE_LOCAL_DEV_FROM_FILE=$(grep -E "^ENTIRE_LOCAL_DEV=" "${PROJECT_DIR}/.env" 2>/dev/null | cut -d'=' -f2- | tr -d '"\r' || echo "")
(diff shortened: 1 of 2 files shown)


toothbrush:
I agree that it's probably valuable to run our local code so we shake out bugs early, rather than relying on whatever the dev has installed.  However, IMHO this is really complicated for what it achieves (lots of Coffee-per-LoC ☕, non-essential and duplicated machinery).

Could we rather revert https://github.com/entireio/cli/pull/177?  Or would that not get what you're after?

gtrrz-victor:
> I agree that it's probably valuable to run our local code so we shake out bugs early, rather than relying on whatever the dev has installed. However, IMHO this is really complicated for what it achieves (lots of Coffee-per-LoC ☕, non-essential and duplicated machinery).
> 
> Could we rather revert #177? Or would that not get what you're after?

Agree, I would rather reverting #177 and not adding this extra complexity to go around what #177 changed

gtrrz-victor:
Let's revert it: https://github.com/entireio/cli/pull/199

PR #222, opened 2026-02-10
[WIP] feat(resume): add --run/-r to auto-start restored session

jaydenfyi:
Adds a `--run`/`-r` flag and `resumeAutoRun` config option so calling resume will automatically start the session for the given coding agent. 

WIP

---

> [!NOTE]
> **Medium Risk**
> Changes alter CLI behavior by executing a subprocess based on the agent-provided resume command, which could impact user environments and requires careful handling of command parsing/execution.
> 
> **Overview**
> `entire resume` now supports **auto-starting the restored agent session** via a new `--run`/`-r` flag, and can default to this behavior using the new `autoRunResume` setting (with explicit CLI flags overriding config).
> 
> Implementation updates thread an `autoRun` parameter through the resume flow, add helpers to print-or-exec the formatted resume command (`exec.Command`), and extend settings loading/merging plus tests to cover the new setting and flag behavior. README is updated with the new usage and configuration option.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6248e461f2a6af9df67d6b6b6a679f30e1268a66. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,8 @@ Multiple AI sessions can run on the same commit. If you start a second session w
 | `--force`, `-f`  | Resume from older checkpoint without prompt |
 | `--run`, `-r`    | Start the restored session immediately      |
 
+Set `"autoRunResume": true` in `.entire/settings.json` (or `.entire/settings.local.json`) to make `--run` the default.
+
 ### `entire enable` Flags
 
 | Flag                   | Description                                                        |
@@ -238,6 +240,7 @@ Personal overrides, gitignored by default:
 |--------------------------------------|----------------------------------|------------------------------------------------------|
 | `enabled`                            | `true`, `false`                  | Enable/disable Entire                                |
 | `log_level`                          | `debug`, `info`, `warn`, `error` | Logging verbosity                                    |
+| `autoRunResume`                      | `true`, `false`                  | Auto-run restored session for `entire resume`        |
 | `strategy`                           | `manual-commit`, `auto-commit`   | Session capture strategy                             |
 | `strategy_options.push_sessions`     | `true`, `false`                  | Auto-push `entire/checkpoints/v1` branch on git push |
 | `strategy_options.summarize.enabled` | `true`, `false`                  | Auto-generate AI summaries at commit time            |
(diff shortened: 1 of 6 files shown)


Soph:
Thanks for proposing this. We discussed the idea before to just start the agent as a subprocess. But given that the parent process could intercept all inputs and outputs, we felt this is a feature that might not be too well received. Like, yes, the CLI is open source, but still. But I guess making this an optional feature would allow anyone to pick what he feels comfortable with. 

We will discuss this internally, and I would also love to hear what others think.

jaydenfyi:
Yeah fair enough. I guess if there's an interactive CLI to pick a session which you mentioned [here](https://github.com/entireio/cli/pull/224#pullrequestreview-3781472166) it might not be as important, although it feels a bit less ergonomic/convenient if selecting a session only outputs the command to run. I'd  probably at least want to be able to have a config option to start the session in the subprocess.

Also assuming the `entire resume` is intended to open a session picker, would it make sense for it to default to the current branch rather than needing to specify it? If that were the case, maybe the arg could instead be a session ID (i.e. `entire resume <session_id>`) as a shortcut to jump into the TUI for whichever coding agent that session might have happened in.

Maybe I haven't fully understood the expected usage yet either, but it also feels like it doesn't necessarily need be restricted to only the first checkpoint inside a given branch as far as what sessions you can resume. There might be previous sessions inside an earlier checkpoint of the current branch, or even sessions inside checkpoints of a different branch that I'd want to resume.

e.g. maybe I had a discussion in my `feature-a` branch that I'd like to continue inside of `feature-b` branch.

Maybe I want a `--all` flag or something that finds sessions across all branches and checkpoints, not just the current one.

Happy to have a go at implementing the picker, although would be keen to understand the desired behaviour a bit more first.

Soph:
Basically the idea is to have `entire checkpoints` that gives you a list of checkpoints in the context of the current branch you are on, or if you specify a branch as a parameter then those from that branch. You can then either `explain` that checkpoint, or `resume` (get the command to get into the session) or `rewind` (basically get to the repo state at the end of that session - for example `git checkout <sha>` + actual session resume command). This was the PR https://github.com/entireio/cli/pull/37 we will pick this up again in the near future.

jaydenfyi:
Happy to close this one for now if you'd like @Soph :) 

Can also capture/summarize some of these ideas in a discussion or issue if thats helpful. Let me know.

jaydenfyi:
Closing given staleness :)

PR #241, opened 2026-02-11
Unify DetectChangedFiles and ComputeFileChanges into DetectFileChanges.

toothbrush:
# Description

Both functions opened a git repo, got worktree status, filtered infrastructure paths, and iterated the status map. The only difference was how they categorized untracked vs modified files. Merge them into a single DetectFileChanges(previouslyUntracked) returning a FileChanges struct with Modified, New, and Deleted slices.

This is a followup cleanup after working on #218.

### DIFF
--- a/cmd/entire/cli/state.go
+++ b/cmd/entire/cli/state.go
@@ -38,7 +38,7 @@ type PrePromptState struct {
 }
 
 // PreUntrackedFiles returns the untracked files list, or nil if the receiver is nil.
-// This nil-vs-empty distinction lets ComputeFileChanges know whether to skip new-file detection.
+// This nil-vs-empty distinction lets DetectFileChanges know whether to skip new-file detection.
 // When the receiver is non-nil but UntrackedFiles is nil (e.g., old state files deserialized with
 // "untracked_files": null), returns an empty non-nil slice so that all current untracked files
 // are correctly treated as new.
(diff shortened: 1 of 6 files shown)


cmd/entire/cli/state.go
```diff
@@ -223,58 +223,71 @@ func CleanupPrePromptState(sessionID string) error {
...
+				changes.New = append(changes.New, file)
 			}
-		case st.Worktree == git.Deleted && st.Staging != git.Deleted:
-			// Deleted file (tracked file that was deleted, not yet staged)
-			deletedFiles = append(deletedFiles, file)
+		case st.Worktree == git.Deleted || st.Staging == git.Deleted:
+			changes.Deleted = append(changes.Deleted, file)
+		case st.Worktree == git.Modified || st.Staging == git.Modified ||
+			st.Worktree == git.Added || st.Staging == git.Added:
+			changes.Modified = append(changes.Modified, file)
```
toothbrush, line 286: I was suspicious of this changed logic, but the computer assuaged my fears:

On origin/main the two functions had different logic:                                                                       

- ComputeFileChanges (line 271): Worktree == Deleted && Staging != Deleted — only unstaged deletions                               
- DetectChangedFiles (line 528): Worktree == Deleted || Staging == Deleted — both staged and unstaged deletions                    

The plan explicitly called out broadening to the || version, and I noted it was "more correct and harmless." But let me reconsider whether that's actually true.                                                                                                      

The && guard (Staging != Deleted) existed to avoid double-counting: if a file is git rm'd (staged deletion), Worktree == Deleted is true and Staging == Deleted is true. The old ComputeFileChanges excluded that case. But there's no double-counting risk here since we only append once per file regardless — the || just means we catch both git rm (staged) and plain rm (unstaged) deletions.

The || version is correct. With the old && logic, a file that was git rm'd (staged for deletion) would be missed entirely — Worktree == Deleted is true but Staging == Deleted is also true, so the && condition fails. That was arguably a bug in ComputeFileChanges. The callers (stop hook, post-task) use deleted files to record what changed during the session, so missing staged deletions would be wrong.

The || version matches what DetectChangedFiles already had and catches all deletions regardless of staging state. No caller is harmed by this since they all just pass the list into SaveContext.DeletedFiles for checkpoint metadata.

TL;DR: Yes, it's intentionally broadened. The old && version in ComputeFileChanges was a bug that silently missed git rm'd files. The || from DetectChangedFiles is correct.

gtrrz-victor (approved):

Issue #253, opened 2026-02-11
AWS Access Key IDs are not redacted despite being sensitive credentials


hi120ki:
### Problem

The redaction package currently fails to detect AWS Access Key IDs (format: `AKIA[A-Z0-9]{16}`), leaving them exposed in logs and session transcripts. While AWS Secret Access Keys are successfully redacted due to their high Shannon entropy (4.66 > threshold 4.5), Access Key IDs have lower entropy (3.68 < 4.5) due to their fixed prefix and limited character set.

### Impact

AWS Access Key IDs are sensitive credentials that, when combined with Secret Access Keys, grant full access to AWS resources. Exposing these IDs in logs creates a security vulnerability.

### Current Behavior

```go
input := "export AWS_ACCESS_KEY_ID=AKIAIOSFODNN7EXAMPLE"
result := redact.String(input)
// Result: "export AWS_ACCESS_KEY_ID=AKIAIOSFODNN7EXAMPLE" (unchanged)
```

### Expected Behavior

```go
input := "export AWS_ACCESS_KEY_ID=AKIAIOSFODNN7EXAMPLE"
result := redact.String(input)
// Result: "export REDACTED" (redacted)
```

### Root Cause

The current implementation relies solely on Shannon entropy for detection:

```go
func isSecret(match string) bool {
    return shannonEntropy(match) > entropyThreshold // 4.5
}
```

AWS Access Key IDs fail this check:
- Entropy: 3.68 (below threshold)
- Reason: Fixed "AKIA" prefix and uppercase-only character set reduce entropy

### Proposed Solution

Add pattern-based detection for structured secrets with distinctive formats but low entropy:

1. Introduce `knownSecretPatterns` slice containing regex patterns for well-known secret formats
2. Extend `isSecret()` to check both entropy and pattern matching
3. Add AWS Access Key ID pattern: `\bAKIA[A-Z0-9]{16}\b`

This approach:
- ✅ Maintains backward compatibility (existing entropy-based detection unchanged)
- ✅ Adds minimal code (~20 lines)
- ✅ Uses word boundaries to avoid false positives
- ✅ Easily extensible for other structured secrets

### Test Case

```go
func TestAWSAccessKeyID_PatternDetection(t *testing.T) {
    awsKeyID := "AKIA" + "IOSFODNN7EXAMPLE"

    if !isSecret(awsKeyID) {
        t.Error("AWS Access Key ID should be detected")
    }

[...]

PR #254, opened 2026-02-11
Add pattern-based detection for AWS Access Key IDs

hi120ki:
Extends redaction to detect AWS Access Key IDs (AKIA[A-Z0-9]{16}) which have low Shannon entropy due to fixed prefix but are sensitive credentials that must be redacted from logs and session transcripts.

Fixes https://github.com/entireio/cli/issues/253


hi120ki:
Close due to https://github.com/entireio/cli/pull/280

PR #259, opened 2026-02-11
Copy checkpoints from forks on PR merge via actions

Soph:
This allows running an action on merge of a PR from a fork to then pull in the matching checkpoint commits from `entire/sessions/v1`.

Need to figure out permissions for it to work.


Soph:
this is not really an option

PR #263, opened 2026-02-11
Fix exit code propagation from subcommands

### LINKED ISSUE #256: CLI always exits with status 1 on command errors (does not preserve underlying exit codes)
### What happened?

main.go currently exits with status 1 for any command error, so command-specific exit codes are lost. Is this intentional?
If not, would you accept a centralized error type / exit-code mapping so subcommands can return specific codes when needed (while defaulting to 1)?

### Steps to reproduce

1. Inspect `cmd/entire/main.go`.
2. Any command that returns an error from `RunE` is handled by `main.go`, which always exits with `os.Exit(1)`.

### Expected behavior
Commands should be able to return command-specific non-zero exit codes when needed.

### Actual behavior
All command errors are collapsed to exit code `1`.

### Entire CLI version

4f09f97-dirty (4f09f97)

### OS and architecture

Darwin/ arm64

### Agent

Not specific

### Strategy

Not sure

### Terminal

zsh vscode

### Logs / debug output

```bash

```

### Additional context

This is a top-level CLI behavior issue, not tied to any specific agent integration.

lubluniky:
## Summary
- Adds `ExitCodeError` type that allows subcommands to specify exit codes
- Updates `main.go` to propagate exit codes instead of always using `os.Exit(1)`
- Supports both direct `ExitCodeError` and `SilentError` wrapping `ExitCodeError`

Closes #256

## Test plan
- [x] Unit tests for `ExitCodeError` type (wrapping, unwrapping, errors.As)
- [x] Verify `SilentError` wrapping `ExitCodeError` propagates code
- [x] Verify `ExitCodeError` wrapping `SilentError` is detectable
- [x] Verify `fmt.Errorf` wrapping preserves `ExitCodeError`
- [x] `gofmt -s -w .` passes
- [x] `golangci-lint run` passes (no new issues)
- [x] `go test ./cmd/entire/cli/` passes

### DIFF
--- a/cmd/entire/cli/errors.go
+++ b/cmd/entire/cli/errors.go
@@ -20,3 +20,25 @@ func (e *SilentError) Unwrap() error {
 func NewSilentError(err error) *SilentError {
 	return &SilentError{Err: err}
 }
+
+// ExitCodeError wraps an error with a specific process exit code.
+// Use this when a subcommand needs to exit with a code other than 1.
+// main.go checks for this type and uses ExitCode instead of the default 1.
+type ExitCodeError struct {
+	Err      error
+	ExitCode int
+}
+
+func (e *ExitCodeError) Error() string {
+	return e.Err.Error()
+}
+
+func (e *ExitCodeError) Unwrap() error {
+	return e.Err
+}
+
+// NewExitCodeError creates an ExitCodeError wrapping the given error
+// with the specified exit code.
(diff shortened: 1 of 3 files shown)


gtrrz-victor:
Thanks for the PR! We'll implement multi exit codes once we have a concrete use case for them. So far this is scaffolding without a real need underneath. Closing for now, but happy to revisit when the need comes up!

PR #268, opened 2026-02-11
fixed wrong reusage of checkpoint id for multiple commits

Soph:
This fixes a bug that could lead to creating multiple commits with the same checkpoint id. A checkpoint id should only be used in one commit.


khaong:
bugbot run

Soph:
superseeded by: https://github.com/entireio/cli/pull/325

Issue #261, opened 2026-02-11
entire overrides all user set git hooks


SchutteJan:
### What happened?

If you have a custom git hook in your repo, `entire enable` will remove them without warning.
Running `entire disable` does not remove the entire hooks, thereby not disabling entire in the repo.

Example
```bash
❯ git commit -m "test"
this is my custom hook
[main 71e18a8] test
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 test.md

testing-dir main
❯ entire enable
Agent: Claude Code (use --agent to change)

✓ Hooks installed
✓ Project configured (.entire/settings.json)

✓ Created orphan branch 'entire/checkpoints/v1' for session metadata

Ready.

testing-dir main* 3s
❯ touch test2.md

testing-dir main*
❯ git add test2.md

testing-dir main*
❯ git commit -m "test 2"
[main c9f9fee] test 2
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 test2.md

testing-dir main*
❯ cat .git/hooks/commit-msg
#!/bin/sh
# Entire CLI hooks
# Commit-msg hook: strip trailer if no user content (allows aborting empty commits)
entire hooks git commit-msg "$1" || exit 1

testing-dir main*
❯ entire disable
Entire is now disabled.

testing-dir main*
❯ cat .git/hooks/commit-msg
#!/bin/sh
# Entire CLI hooks
# Commit-msg hook: strip trailer if no user content (allows aborting empty commits)
entire hooks git commit-msg "$1" || exit 1
```

### Steps to reproduce

1. `entire enable && entire disable` leave `.git/hooks/*` files intact

### Entire CLI version

Entire CLI dev (unknown)

### OS and architecture

Linux Manjaro

### Agent

N/A

### Strategy

manual-commit (default)

### Terminal

Ghostty

### Logs / debug output

```bash

```

eulerkochy:
I guess your expectation is that custom hooks are kept as it is, on top of it additional `enable` hooks run?

SchutteJan:
At least disable the entire hooks on `entire disable`. Might be tricky to have a system where you run the users custom hooks and your hooks, but that would be nice.

nicolewhite-msft:
+1 you should definitely warn and ask for confirmation when `entire enable` is going to override a pre-existing hook. As-is, it wipes away any git hooks you might've had already:

```
$ cat .git/hooks/post-commit
#!/bin/sh
[...]

PR #284, opened 2026-02-12
Added self heal for third party overwriting Entire's githooks

suyashkumar2409:
#261 raised that Entire overwrites existing hooks. While #272 fixes Entire overwriting existing git hooks, Entire's hooks can still be lost after `entire enable` if a third-party tool (e.g., [lefthook](https://github.com/evilmartians/lefthook)) is installed later and overwrites the git hooks directory. 

This adds self-healing: on each session  
  - Detection via IsGitHookInstalled() — returns false if any managed hook is missing or lacks our marker                                                                                                  
  - Reinstall via InstallGitHook(true) — backs up the third-party hook to .pre-entire and chains to it, preserving existing backups from the original install                                              
  - Warnings printed to stderr so the user knows hooks were repaired                                                                                                                                       

  Followup needed:
  - This only heals on session start — hooks overwritten between sessions without an agent running won't be caught until the next session

### DIFF
--- a/cmd/entire/cli/strategy/hooks_test.go
+++ b/cmd/entire/cli/strategy/hooks_test.go
@@ -906,8 +906,6 @@ func TestRemoveGitHook_PermissionDenied(t *testing.T) {
 	}
 }
 
-// Self-heal: reinstalling hooks after a third-party tool overwrites or deletes them.
-
 func TestInstallGitHook_SelfHealAfterThirdPartyOverwrite(t *testing.T) {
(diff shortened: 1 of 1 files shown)


Soph:
Hey, thanks for doing this. But we already had a way to ensure git hooks are there. I just had a look at the logic today and did some minor improvements: https://github.com/entireio/cli/pull/298

suyashkumar2409:
No worries! I really like the project and will be on the lookout for more ways to contribute!

PR #293, opened 2026-02-12
use local settings when local flag is used

### LINKED ISSUE #279: entire enable --local is writing to settings.json
### What happened?

.claude/settings.json was modified even if I run `entire enable --local`. Only claude/settings.local.json should be modified.

### Steps to reproduce

1. entire enable --local
2. git status
3. .claude/settings.json is modified

### Entire CLI version

Entire CLI 0.4.2 (2a9998e)

### OS and architecture

OS/Arch: darwin/arm64

### Agent

Claude Code

### Strategy

manual-commit (default)

### Terminal

Apple Terminal Version 2.14 (455.1)

### Logs / debug output

```bash

```

### Additional context

I had previously enabled it without `--local`, then disabled it, restored all changes using git restore, then enabled it with `--local`

suyashkumar2409:
Fixes #279                                                                                                                                                                                                
                                                                                                                                                                                                           
  entire enable --local was writing Claude Code hooks to .claude/settings.json instead of .claude/settings.local.json. The --local flag only controlled which .entire/settings* file was written to, but
  the agent hook installation was using the harcoded value `settings.json`.

  - Extract InstallHooksTo on ClaudeCodeAgent that accepts a target settings file name
  - setupClaudeCodeHook now receives the useLocal flag and passes the appropriate file name
  - determineSettingsTarget is called before agent hook setup so the result is available

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,37 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/),
+and this project adheres to [Semantic Versioning](https://semver.org/).
+
+## [0.4.3] - 2026-02-12
+
+### Added
+
+- Layered secret detection using gitleaks patterns alongside entropy-based scanning ([#280](https://github.com/entireio/cli/pull/280))
+- Multi-agent rewind and resume support for Gemini CLI sessions ([#214](https://github.com/entireio/cli/pull/214))
+
+### Changed
+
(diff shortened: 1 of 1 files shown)


squishykid:
Thanks for raising this PR! You can see I spent some time playing with the idea you presented here, but for now we are going to take this as expected behavior- docs have been updated accordingly. Agents have different hierarchies of config (e.g. gemini does not support `.gemini/settings.local.json`) and hopefully we will have a better story around handling this in the future!

Issue #308, opened 2026-02-12
entire enable silently drops unknown Claude Code hook types (e.g. PreCompact)


TyceHerrman:
### Preamble

Written with Claude Code but reviewed by me.

### What happened?

Running `entire enable` silently removes any Claude Code hook types not defined in Entire's `ClaudeHooks` Go struct. In my case, my `PreCompact` hook was permanently deleted from `.claude/settings.json` without warning.

Claude Code supports 14 hook event types, but the struct only defines 6. The other 8 are silently discarded on every `entire enable` run:

| Dropped hook type | Purpose |
|---|---|
| `PreCompact` | Before context compaction |
| `Notification` | When Claude sends a notification |
| `SubagentStart` | When a subagent is spawned |
| `SubagentStop` | When a subagent finishes |
| `PermissionRequest` | When a permission dialog appears |
| `PostToolUseFailure` | After a tool call fails |
| `TeammateIdle` | When a teammate goes idle |
| `TaskCompleted` | When a task is marked completed |

This is also a forward-compatibility problem — any new hook types added by Claude Code in the future will be silently dropped too.

### Root cause

The `ClaudeHooks` struct in `cmd/entire/cli/agent/claudecode/types.go` only defines 6 hook types:

```go
type ClaudeHooks struct {
    SessionStart     []ClaudeHookMatcher `json:"SessionStart,omitempty"`
    SessionEnd       []ClaudeHookMatcher `json:"SessionEnd,omitempty"`
    UserPromptSubmit []ClaudeHookMatcher `json:"UserPromptSubmit,omitempty"`
    Stop             []ClaudeHookMatcher `json:"Stop,omitempty"`
    PreToolUse       []ClaudeHookMatcher `json:"PreToolUse,omitempty"`
    PostToolUse      []ClaudeHookMatcher `json:"PostToolUse,omitempty"`
}
```

In `hooks.go`, `InstallHooks` unmarshals the existing hooks JSON into this struct — Go's `json.Unmarshal` silently discards any fields without a corresponding struct field. When the struct is marshaled back and written to disk, the unknown hooks are gone:

```go
// Step 1: unmarshal — PreCompact (and 7 others) silently dropped here
if err := json.Unmarshal(hooksRaw, &settings.Hooks); err != nil { ... }

// Step 2: modify — Entire's hooks are added to the struct

// Step 3: marshal back — only the 6 struct fields are written
hooksJSON, err := json.Marshal(settings.Hooks)
[...]

PR #314, opened 2026-02-12
Safer agent hook parsing, make sure we preserve unknown hook types

Soph:
Fixes https://github.com/entireio/cli/issues/308

The agent settings parsing was too strict and removed any hook types that weren't defined / known in the CLI repo. This handles this safer and prevents removing hook types the CLI doesn't know about.


toothbrush (approved):
Looks reasonable to me.  I wonder if this kind of logic can/should be extracted out of the individual Gemini & Claude hook definitions, so that it'll be easier to reuse for the N+1st agent we add?  Something we might want to bear in mind.

Soph:
@toothbrush yes, it should. I also thought about this, but given we are having an agent refactor coming up I decided to leave it like that for now.

Issue #266, opened 2026-02-11
Opencode support


jaxondk:
### Problem or use case

Integrate [opencode](https://github.com/anomalyco/opencode) as a supported agent, like claude code and google gemini CLI.

Opencode is very popular, open source, and well architected.

### Desired behavior

```shell

```

khaong:
Hi, thanks for raising this - we currently have an issue open to make new agent integrations easier #281 - this one would need to come after 🙏

Soph:
Hey, with the release 0.4.6 we shipped OpenCode support in preview. Please give it a test and let us now if you find any issues.

Please open new issues for bugs or missing features. Thanks!

Issue #281, opened 2026-02-12
Refactor Agent interface to allow simple(r) Agent integration contributions


khaong:
### Problem or use case

We need to restructure the hooks, transcript parsing and lifecycle actions so that it's easy to contribute a new agent.

Currently the boundaries are not fully defined as to what is the framework and what is Agent-specific.

This will get picked up by the core maintainer team - we ask that new agent support please come after this to limit rework.

Acceptance criteria
===============
- Documentation describing
  - how the hooks, lifecycle events and actions work together
  - what is the responsibility of the framework vs the agent implementation
- a clear interface for mandatory agent concerns
- clarity around optional actions/hooks if necessary (documentation or code-based)

### Desired behavior
New agent support can be added without changing 'framework' code.

### Proposed solution

- define 'framework' and 'agent' responsibilities in the lifecycle
- Refactor Agent Interface to reflect the above
- Move/create framework code where necessary
- Refactor existing Claude and Gemini agents

Soph:
We have finished this now and also added OpenCode support based on this. 

The docs (useful for agents too) are here: 

- The longer Agent Guide gives an overview of how the integration should look like and a lot of details https://github.com/entireio/cli/blob/main/docs/architecture/agent-guide.md 
- The Agent Integration Checklist is shorter but has a few key points, running an agent / llm against the guide and your implementation before submitting would be nice https://github.com/entireio/cli/blob/main/docs/architecture/agent-integration-checklist.md

PR #257, opened 2026-02-11
Add OpenCode and Codex CLI agent support

Avyukth:
Closes #245
Closes #266
Relates to #281

## Summary

Adds two new AI coding agent integrations to the Entire CLI: **OpenCode** and **Codex CLI**. Both agents support full session checkpointing via hooks — capturing transcripts, file changes, prompts, and metadata on shadow branches.

As part of this work, **duplicated framework code was extracted into a shared commit pipeline** ("$(hooks_common_session.go)"), directly addressing issue #281's goal of separating framework from agent-specific logic. The three agent handlers (Codex, OpenCode, Gemini) that previously duplicated ~250 lines of commit logic are now thin wrappers calling a shared "$(commitAgentSession())" function.

**~3,000 lines across 16 files** (all new code from main's perspective).

---

### Framework Extraction (addresses #281)

The shared "$(hooks_common_session.go)" consolidates duplicated logic from Codex, OpenCode, and Gemini handlers:

| Extracted Function | Replaces | Purpose |
|---|---|---|
| "$(commitAgentSession())" | "$(commitCodexSession)", "$(commitOpenCodeSession)", "$(commitGeminiSession)" | Core commit pipeline: load state → detect changes → save via strategy |
| "$(createContextFile())" | "$(createContextFileForCodex)", "$(createContextFileForOpenCode)", "$(createContextFileForGemini)" | Unified context.md generation (single + multi-prompt) |
| "$(transitionSessionTurnEnd())" | Was in Claude Code handlers, used by all agents | Session phase state machine transition |
| "$(logFileChanges())" | Was in Gemini handlers, used by all agents | Stderr logging of modified/new/deleted files |

Each agent handler is now a thin wrapper that populates an "$(agentCommitContext)" struct:
- **Codex**: provides transcript-extracted modified files, single prompt
- **OpenCode**: sets "$(useGitStatusForModified: true)" (no transcript parsing), single prompt
- **Gemini**: adds token usage calculation, multi-prompt array, transcript step tracking

This reduces the barrier for adding new agents — a new handler only needs to populate the context struct and call "$(commitAgentSession())".

**Not yet extracted** (future #281 work): Claude Code's "$(commitWithMetadata)" has unique subagent/summarization/incremental-checkpoint logic that requires its own pattern. The "$(Agent)" interface itself is also unchanged — broader interface refactoring is deferred to the core team per #281.

---

### OpenCode Agent
[...]

PR #220, opened 2026-02-10
Add OpenCode agent integration

ammarateya:
Relates to #266

Adds native OpenCode support with transcript reconstruction from ~/.local/share/opencode/storage/

---

> [!NOTE]
> **Medium Risk**
> Adds a new hook-driven integration that shells out to `opencode export` and writes session metadata/transcripts, which can impact checkpoint creation and has some operational risk (timeouts, missing binaries, malformed exports). Changes to manual-commit condensation detection also affect when sessions are considered to have new content.
> 
> **Overview**
> Adds a new **OpenCode agent integration** that installs a repo-local Bun plugin (`.opencode/plugin/entire.js`) to bridge OpenCode events into `entire hooks opencode` commands, including session detection and validated `session_id` parsing.
> 
> On `stop`, the new handlers export the OpenCode session (`opencode export` with timeout), convert it into Entire-compatible JSONL (including timestamps), write transcript/prompt metadata, infer modified files from tool calls, and compute per-turn token usage for checkpoint metadata.
> 
> Also updates session condensation logic to handle agents without transcripts (e.g., OpenCode) by treating `FilesTouched`/`StepCount` as new content when no transcript file exists, and wires OpenCode into the hook registry/uninstall flow; minor cleanup removes `ireturn` nolint annotations from strategy constructors/registry.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d5395ffed41e191e0bae72345d226dd21b08fdb6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #315, opened 2026-02-12
Add OpenCode agent support and allow enable without hook installation

MementoMori123:
Relates to #266

## Summary

Adds OpenCode as a supported agent so Entire can be used in OpenCode projects. OpenCode is detected via `.opencode/` or `opencode.json`. Because OpenCode is expected to integrate via its own plugin (calling `entire hooks ...`), this agent does not install or manage hooks; the setup flow is updated so `entire enable` can succeed for such agents.

## Changes

- **New agent: OpenCode** (`cmd/entire/cli/agent/opencode/`)
  - Implements the `Agent` interface: detection, session ID transform, session dir resolution, protected dirs (`.opencode`), and resume command.
  - `SupportsHooks()` returns `false`; hook input parsing is unimplemented (returns a clear error if used).
  - Session read/write is implemented as a passthrough of OpenCode’s on-disk format so future hook/plugin integration can be added without changing this contract.

- **Setup behavior**
  - Agents that do not support hooks (e.g. OpenCode) can complete `entire enable` without installing hooks. Entire’s git hooks and strategy setup still run; only agent-specific hook installation is skipped.
  - Message when no hooks are installed is updated to: “Hooks for … already installed or managed externally” so it’s accurate for both “already installed” and “managed by the agent/plugin elsewhere.”

- **Registration**
  - OpenCode is registered in the agent registry and wired in `config.go` and `hooks_cmd.go` so it is discoverable and usable when OpenCode (or its plugin) invokes the CLI.

## Testing

- `mise run test` (unit tests including new registry test for `opencode`).
- Manual: `entire enable` in a repo with `.opencode/` or `opencode.json` completes successfully and does not require hook support.

### DIFF
--- a/cmd/entire/cli/agent/opencode/opencode.go
+++ b/cmd/entire/cli/agent/opencode/opencode.go
@@ -1,0 +1,235 @@
+package opencode
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/entireio/cli/cmd/entire/cli/agent"
+	"github.com/entireio/cli/cmd/entire/cli/paths"
+	"github.com/entireio/cli/cmd/entire/cli/sessionid"
+)
+
+//nolint:gochecknoinits // Agent self-registration is the intended pattern
+func init() {
+	agent.Register(agent.AgentNameOpencode, NewOpenCodeAgent)
+}
+
+// OpenCodeAgent implements the Agent interface for OpenCode.
+//
+//nolint:revive // OpenCodeAgent is clearer than Agent in this context
+type OpenCodeAgent struct{}
+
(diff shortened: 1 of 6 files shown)


khaong:
Thanks for this! We're currently reworking the agent interface (#281) so we're holding agent PRs until that lands — otherwise the work would need to be redone against the new interface.

Heads up there are a few other OpenCode PRs in flight (#220, #257, #341) — we'll evaluate them together once #281 is done. We'll update #281 when it's ready for contributions.

Soph:
We've now merged our OpenCode integration. While we went with our own implementation, your PRs were valuable in helping us validate our design choices and ensure we covered the right scenarios. We appreciate the effort you put into this!

PR #324, opened 2026-02-13
feat: add import-session command for manual Claude Code session import

### LINKED ISSUE #287: Manually Importing Claude Code sessions
### Problem or use case

There's two use cases where it would be very helpful to be able to manually import Claude Code sessions

1. Because of some bug in `entire` the a session was not properly registered as checkpoint

2. When I want to start using `entire` with an existing repository that already has a wealth of sessions associated with it 

### Desired behavior

```shell
# Importing into head commit or adding to the existing checkpoint
$ entire import-session path/to/session.jsonl [path/to/session2.jsonl ...]

Imported agent session to new checkpoint .... on HEAD

# Specifying past commit
$ entire import-session path/to/session.jsonl [path/to/session2.jsonl ...] --commit efgd111

Imported agent session to new checkpoint .... on efgd111
# Maybe some kind of warning in case a checkpoint marker was added to a past commit
Caution: Your git history has now diverged from origin you will need to force push to share it with contributors
```

### Proposed solution

_No response_

### Alternatives or workarounds

Maybe instead of rewriting the history maybe there's also a way to somehow leave markers on future commits which assign checkpoints to past commits to keep the git history clean? But since commit sha's are not guaranteed to be stable that is probably more fragile than just biting the bullet and force pushing 😬

fahd04:
## Summary
Adds `entire import-session` command to manually import Claude Code session transcripts into Entire checkpoints.

## Motivation
Closes #287

Two main use cases:
1. Recover sessions that weren't properly checkpointed due to bugs
2. Onboard a repository with existing Claude Code sessions

## Changes
- **New command:** `entire import-session <path> [paths...] [--commit <hash>]`
- Reads Claude Code JSONL transcripts, extracts modified files and prompts
- Writes checkpoint metadata to `entire/checkpoints/v1` via existing `WriteCommitted` API
- Supports multiple sessions per call (stored as multi-session checkpoint)
- Prints instructions for adding `Entire-Checkpoint` trailer; warns about history rewrite when targeting past commits

## Testing
- Integration tests: `TestImportSession_ImportToHEAD`, `TestImportSession_MultipleSessions`, `TestImportSession_RequiresEnabled`
- Integration test setup now sets `enabled: true` in settings so commands run as expected

### DIFF
--- a/cmd/entire/cli/import_session.go
+++ b/cmd/entire/cli/import_session.go
@@ -1,0 +1,207 @@
+package cli
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/entireio/cli/cmd/entire/cli/agent"
+	"github.com/entireio/cli/cmd/entire/cli/agent/claudecode"
+	"github.com/entireio/cli/cmd/entire/cli/checkpoint"
+	"github.com/entireio/cli/cmd/entire/cli/checkpoint/id"
+	"github.com/entireio/cli/cmd/entire/cli/paths"
+	"github.com/entireio/cli/cmd/entire/cli/settings"
+	"github.com/entireio/cli/cmd/entire/cli/strategy"
+	"github.com/entireio/cli/cmd/entire/cli/trailers"
+
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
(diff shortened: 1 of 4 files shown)


Soph:
Thanks for your contribution. We are handling this as part of https://github.com/entireio/cli/pull/688 now.

PR #339, opened 2026-02-15
fix: enable --agent respects settings.local.json overrides

adrianmg:
## Summary

- `entire enable --agent <name>` always wrote to `settings.json`, ignoring a stale `enabled: false` in `settings.local.json` left by a prior `entire disable`
- This caused `entire status` to show "Disabled" immediately after a successful enable

image

## Root Cause

`runDisable()` writes `enabled: false` to `settings.local.json`. But `setupAgentHooksNonInteractive()` (the `--agent` code path) only wrote `enabled: true` to `settings.json`. Since `settings.Load()` merges local over project, the stale `false` in `settings.local.json` always won.

The interactive enable and `--strategy` paths already handled this correctly via `determineSettingsTarget()` — the `--agent` path was the only one missing it.

## Fix

After writing `enabled: true` to `settings.json`, check if `settings.local.json` exists with a stale `enabled: false` and clear it. No signature changes, no new params — just 4 lines of targeted cleanup.

## Repro (before fix)

```
entire enable --agent gemini     # writes enabled:true → settings.json
entire disable                   # writes enabled:false → settings.local.json
entire enable --agent gemini     # writes enabled:true → settings.json (local untouched)
entire status                    # → "Disabled (manual-commit)"
```

## Affects

All agents using the \`--agent\` flag — not specific to any one agent.

### DIFF
--- a/cmd/entire/cli/integration_test/hooks.go
+++ b/cmd/entire/cli/integration_test/hooks.go
@@ -455,6 +455,29 @@ func (env *TestEnv) GetSessionState(sessionID string) (*strategy.SessionState, e
 	return &state, nil
 }
 
+// WriteSessionState writes the session state for the given session ID.
+// This is useful for tests that need to manipulate session state directly.
+func (env *TestEnv) WriteSessionState(sessionID string, state *strategy.SessionState) error {
+	env.T.Helper()
+
+	stateDir := filepath.Join(env.RepoDir, ".git", "entire-sessions")
+	if err := os.MkdirAll(stateDir, 0o755); err != nil {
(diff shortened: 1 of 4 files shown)


khaong:
Thanks for the contribution! We ask that all PRs have a corresponding GitHub issue so we can discuss the approach first. Could you open an issue describing the problem this solves? Once we've aligned there, we'll prioritize the review. This helps make sure your effort doesn't go to waste.

Soph:
Thanks again for your contribution. The issue you have described is also handled by https://github.com/entireio/cli/pull/678 so I'll close this one.

PR #343, opened 2026-02-15
Make E2E pass and add some more tests

Soph:
Follow up to https://github.com/entireio/cli/pull/336 and https://github.com/entireio/cli/pull/325

- E2E suite passes now
- Added more tests
- Updated how we consider a change to belong to the last session after partial commits
- E2E suite also working for Gemini


Soph:
cursor review

PR #367, opened 2026-02-17
fix: delete extra shadow branches

gtrrz-victor:
Bug fix: prevent extra shadow branches after carry-forward

Two behavioral fixes:

1. state.FilesTouched updated before carry-forward (manual_commit_hooks.go:699) — Previously set inside carryForwardToNewShadowBranch, which only runs when remaining files > 0. When all files were committed, FilesTouched kept its stale value, making the session look like it still had work — causing extra shadow
branches on subsequent commits.
2. extractFilesFromLiveTranscript uses CheckpointTranscriptStart offset (manual_commit_hooks.go:1077) — Previously used offset 0, re-extracting already-committed files and inflating the carry-forward list.


Soph:
@cursor review

Soph:
@cursor review again

PR #371, opened 2026-02-17
Avoid showing new version available warning with completion command

danhper:
The completion command is typically run non-interactively when the shell starts, and evaluated with `source`.
The version check outputs the warning to stdout, which makes `source` fail.

This PR skips version check when running the completion command.

### DIFF
--- a/cmd/entire/cli/root.go
+++ b/cmd/entire/cli/root.go
@@ -39,11 +39,8 @@ func NewRootCmd() *cobra.Command {
 			HiddenDefaultCmd: true,
 		},
 		PersistentPostRun: func(cmd *cobra.Command, _ []string) {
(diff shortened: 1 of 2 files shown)


gtrrz-victor:
First of all, thanks for contributing to this project @danhper , we really appreciate it.

Quick question: which version of Entire are you currently using?
This problem should have been fixed after merging this PR -> https://github.com/entireio/cli/pull/238 and it was included in v0.4.3 -> https://github.com/entireio/cli/releases/tag/v0.4.3

Let me know if the bug hasn't been fixed yet.

Thanks.

danhper:
Hi @gtrrz-victor and thanks for the reply.
I was indeed using 0.4.2, so I probably just missed the fix and wrongly assumed it was still an ongoing issue.
Sorry for the noise, I'll close this!

PR #372, opened 2026-02-17
feat: add interactive TUI dashboard command

yasunogithub:
## Summary

- Add `entire dashboard` command with four tabs: Sessions, Checkpoints, Active Sessions, and Settings
- Built with bubbletea/lipgloss TUI framework with accessible text-mode fallback (`ACCESSIBLE=1`)
- Includes checkpoint rewind with `PreviewRewind` safety check for untracked file deletion warnings
- Rune-safe filter input handling for multi-byte character support
- Header-pinning scroll in Active tab for consistent UX across all tabs

## Test plan

- [ ] Run `mise run fmt && mise run lint && mise run test:ci` — all pass
- [ ] Run `entire dashboard` and navigate all four tabs
- [ ] Test filter input with multi-byte characters (e.g., Japanese)
- [ ] Test rewind from Checkpoints tab confirms before proceeding
- [ ] Test `ACCESSIBLE=1 entire dashboard` for accessible text mode
- [ ] Verify scroll behavior in Active tab with many sessions (headers stay pinned)

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -108,6 +108,7 @@ linters:
         - github.com/go-git/go-git/v6/storage.Storer
         - github.com/go-git/go-git/v6/plumbing/storer.EncodedObjectIter
         - github.com/go-git/go-billy/v6.Filesystem
+        - github.com/entireio/cli/cmd/entire/cli/strategy.Strategy
     nolintlint:
       require-explanation: true
       require-specific: true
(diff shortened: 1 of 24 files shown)


Soph:
Thanks for opening this! The cli iterated quite heavily since this was opened and we added `entire activity` which is  a dashboard as the one on entire.io.

I still like your approach and our suggestion would be to add an external command instead: https://github.com/entireio/entire-plugin-template

PR #401, opened 2026-02-18
skip version check for dev builds

gtrrz-victor:
## Summary

- Dev builds now use `{tag}-dev-{commit}` version format (e.g. `v0.4.5-dev-755fbfde`)
- Version check skips versions containing `-dev` in the prerelease suffix
- Normal prereleases like `-rc1` still trigger the update notification
- Removed `mise build:all` task — cross-arch builds are handled by the release GitHub Action, so a local build:all serves no purpose

This gives us a simple way to mark dev builds and skip version checks on them, without affecting real prerelease versions.

### DIFF
--- a/mise-tasks/build
+++ b/mise-tasks/build
@@ -1,6 +1,8 @@
 #!/bin/sh
 #MISE description="Build the CLI"
 
-VERSION=$(git describe --tags --always --dirty 2>/dev/null || echo "dev")
+TAG=$(git describe --tags --abbrev=0 || echo "v0.0.0")
 COMMIT=$(git rev-parse --short HEAD 2>/dev/null || echo "unknown")
+VERSION="${TAG}-dev-${COMMIT}"
+
 go build -ldflags "-X github.com/entireio/cli/cmd/entire/cli/versioninfo.Version=${VERSION} -X github.com/entireio/cli/cmd/entire/cli/versioninfo.Commit=${COMMIT}" -o entire ./cmd/entire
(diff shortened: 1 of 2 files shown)


cmd/entire/cli/versioncheck/versioncheck_test.go
```diff
@@ -35,8 +35,15 @@ func TestIsOutdated(t *testing.T) {
 		{"1.0.0", "v1.0.1", true, "mixed v prefix reversed"},
 
 		// Pre-release versions (semver uses hyphen)
-		{"1.0.0-rc1", "1.0.0", true, "prerelease in current"},
+		{"1.0.0-rc1", "1.0.0", false, "prerelease in current"},
```
toothbrush:
```suggestion
		{"1.0.0-rc1", "1.0.0", true, "prerelease in current"},
```


I might be misunderstanding, but 1.0.0 is "newer" than 1.0.0-rc1, because really a release candidate shouldn't actually be released as mainline yet?  I.e., a random person doing `curl|bash` should not get a rc version?

toothbrush: I see the same thing with "1.0.0" vs "1.0.1-rc1" – i would only expect an end user to get prompted to update once 1.0.1 is available?  Not the rc?

gtrrz-victor: 100% with you

PR #413, opened 2026-02-18
fix: Gemini e2e tests stall

peyton-alt:


gtrrz-victor:
I will close it as we added this other one: https://github.com/entireio/cli/pull/474

PR #436, opened 2026-02-20
Status optimization

evisdren:
Replaces shelling out to a git process by using a pure go function to find the HEAD branch name. Shaves another 10ms off `entire status`.

image

### DIFF
--- a/cmd/entire/cli/status.go
+++ b/cmd/entire/cli/status.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"io/fs"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"sort"
 	"strings"
@@ -178,7 +179,10 @@ type worktreeGroup struct {
 	sessions []*session.State
 }
 
-const unknownPlaceholder = "(unknown)"
+const (
+	unknownPlaceholder  = "(unknown)"
+	detachedHEADDisplay = "HEAD"
+)
 
 // writeActiveSessions writes active session information grouped by worktree.
 func writeActiveSessions(w io.Writer) {
@@ -326,9 +330,32 @@ func resolveWorktreeBranch(worktreePath string) string {
 
 	// Symbolic ref: "ref: refs/heads/<branch>"
 	if strings.HasPrefix(ref, "ref: refs/heads/") {
-		return strings.TrimPrefix(ref, "ref: refs/heads/")
+		branch := strings.TrimPrefix(ref, "ref: refs/heads/")
+		// Reftable ref storage uses "ref: refs/heads/.invalid" as a dummy HEAD stub.
+		// Fall back to git to resolve the actual branch in that case.
+		if branch == ".invalid" {
+			return resolveWorktreeBranchGit(worktreePath)
+		}
+		return branch
 	}
 
 	// Detached HEAD or other ref type
-	return "HEAD"
+	return detachedHEADDisplay
+}
+
+// resolveWorktreeBranchGit resolves the branch name by shelling out to git.
+// Used as a fallback for reftable ref storage where .git/HEAD is a stub.
+func resolveWorktreeBranchGit(worktreePath string) string {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	cmd := exec.CommandContext(ctx, "git", "-C", worktreePath, "rev-parse", "--symbolic-full-name", "HEAD")
+	out, err := cmd.Output()
+	if err != nil {
+		return detachedHEADDisplay
+	}
+	ref := strings.TrimSpace(string(out))
+	if strings.HasPrefix(ref, "refs/heads/") {
+		return strings.TrimPrefix(ref, "refs/heads/")
+	}
+	return detachedHEADDisplay
 }
--- a/cmd/entire/cli/status_test.go
+++ b/cmd/entire/cli/status_test.go
@@ -171,6 +171,26 @@ func TestResolveWorktreeBranch_NotARepo(t *testing.T) {
 	}
 }
 
+func TestResolveWorktreeBranch_ReftableStub(t *testing.T) {
+	t.Parallel()
+
+	// Simulate a reftable repo where .git/HEAD contains "ref: refs/heads/.invalid"
+	dir := t.TempDir()
+	gitDir := filepath.Join(dir, ".git")
+	if err := os.MkdirAll(gitDir, 0o755); err != nil {
+		t.Fatalf("mkdir .git: %v", err)
+	}
+	if err := os.WriteFile(filepath.Join(gitDir, "HEAD"), []byte("ref: refs/heads/.invalid\n"), 0o644); err != nil {
+		t.Fatalf("write HEAD: %v", err)
+	}
+
+	branch := resolveWorktreeBranch(dir)
+	// Should fall back to git, which will fail on this fake repo and return "HEAD"
+	if branch != "HEAD" {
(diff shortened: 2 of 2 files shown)


evisdren:
> ## PR Summary
> **Medium Risk** Touches git-path resolution and branch detection used by `status`; filesystem parsing and caching could mis-handle unusual worktree/HEAD formats or stale cwd-based cache in edge cases.
> 
> **Overview** Improves `entire status` performance by **caching `git rev-parse --git-common-dir`** in `session.getGitCommonDir()` and by replacing `resolveWorktreeBranch`’s `git rev-parse` call with direct reads of `.git/HEAD` (including worktree `.git` file support).
> 
> Adds a new `benchutil` fixture package plus benchmarks for `status` and checkpoint hot paths (`WriteTemporary`, `WriteCommitted`, tree helpers), and introduces `mise` tasks (`bench`, `bench:cpu`, `bench:mem`) to run/profiling benchmarks. Updates tests to cover the new caching and HEAD-parsing behavior, and updates the Discord invite link in `CONTRIBUTING.md`.
> 
> Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit [f413d34](https://github.com/entireio/cli/commit/f413d34be360402257b66341f7c371f1e73e1b10). Configure [here](https://cursor.com/dashboard?tab=bugbot).

Mainly for posterity:"

```
  Risk 1: Unusual HEAD/worktree formats in resolveWorktreeBranch
                                                                                                    
  Known HEAD file formats:                                                                        
                                                                                                  
  Format: Branch ref                              
  Example: ref: refs/heads/main
  Our handling: Returns "main"
  git rev-parse handling: Returns "main"
  ────────────────────────────────────────
  Format: Detached (hash)
  Example: abc123...
  Our handling: Returns "HEAD"
  git rev-parse handling: Returns "HEAD"
  ────────────────────────────────────────
  Format: Tag ref
  Example: ref: refs/tags/v1.0
  Our handling: Returns "HEAD"
  git rev-parse handling: Returns "tags/v1.0"
  ────────────────────────────────────────
  Format: Remote ref
  Example: ref: refs/remotes/origin/main
  Our handling: Returns "HEAD"
  git rev-parse handling: Returns "remotes/origin/main"

  Known .git formats:

  ┌──────────────────────┬──────────────────────────┬──────────────────────────────────────────────┐
  │        Format        │         Scenario         │                 Our handling                 │
  ├──────────────────────┼──────────────────────────┼──────────────────────────────────────────────┤
  │ Directory            │ Regular repo             │ Read .git/HEAD directly                      │
  ├──────────────────────┼──────────────────────────┼──────────────────────────────────────────────┤
  │ File (gitdir: <abs>) │ Worktree (absolute path) │ Follow path, read HEAD                       │
  ├──────────────────────┼──────────────────────────┼──────────────────────────────────────────────┤
  │ File (gitdir: <rel>) │ Worktree (relative path) │ Resolve relative to worktree dir, read HEAD  │
  ├──────────────────────┼──────────────────────────┼──────────────────────────────────────────────┤
  │ Symlink              │ Some container setups    │ os.Stat follows symlinks, works as directory │
  └──────────────────────┴──────────────────────────┴──────────────────────────────────────────────┘

  Behavioral difference: tag/remote refs — we return "HEAD" while the git subprocess returns the
  ref's short name. In practice this doesn't matter because HEAD only points to non-branch refs in
  extremely unusual manual operations, and the output is display-only (it shows in entire status as
  the branch label next to a worktree path). An incorrect label has zero functional impact — no data
   is written, no decisions are made from it.

  Unreachable cases: Bare repos never reach this function because runStatus exits at line 42
  (paths.RepoRoot() fails for bare repos). Submodules work correctly — their .git is a file with
  gitdir: just like worktrees.

  Degradation behavior: Every error path returns "", same as the old implementation on failure. The
  caller at line 248 just skips showing the branch name. No crash, no wrong action taken.

  Test coverage: 6 tests covering regular repo, detached HEAD, worktree (absolute path), worktree
  (relative path), non-existent path, not-a-repo.

  Risk level: Low. The formats are stable (unchanged for 15+ years), we handle all common cases
  identically, the one edge case difference (tag/remote refs) is display-only and functionally
  irrelevant, and all errors degrade gracefully.

  ---
  Risk 2: Stale cwd-based cache in getGitCommonDir

  How the cache works: Stores the result keyed by os.Getwd(). On subsequent calls, if cwd matches
  the cached key, returns the cached value. If cwd changed, falls through to the git subprocess.

  What could make it stale:
  1. cwd changes mid-process — Cache invalidates automatically (different key). Safe.
  2. Git repo structure changes while process runs (e.g., converting repo to worktree) — Cache would
   be stale. But this can't happen during a single entire status invocation (milliseconds), and even
   if it did, the git common dir doesn't change unless you restructure the .git directory itself.
  3. os.Getwd() fails — We use "" as the key. If two calls both fail, we'd get a stale hit. But if
  Getwd() fails, we're in a broken environment anyway.

  This is the exact same pattern as paths.RepoRoot() (lines 60-92 in paths.go), which has been in
  production. The cache is strictly additive — on a miss, it falls through to the real subprocess.
  The worst case is a redundant subprocess call, never a wrong answer (since the subprocess is the
  canonical source).

  Test coverage: 5 tests covering valid return, cache hit, cache clear, cwd change invalidation,
  error outside repo.

  Risk level: Very low. The pattern is proven (RepoRoot uses it), CLI processes are short-lived
  (cache barely matters per-process — it mainly helps when getGitCommonDir is called multiple times
  within one invocation), and the fallback is always the real subprocess.

  ---
  Summary

  ┌─────────────────┬──────────────────────┬───────────────────────────┬──────────────────────────┐
  │      Risk       │       Severity       │        Likelihood         │        Mitigation        │
  ├─────────────────┼──────────────────────┼───────────────────────────┼──────────────────────────┤
  │ Unusual HEAD    │ Display-only         │ Very rare                 │ Graceful degradation to  │
  │ format          │ cosmetic diff        │                           │ ""                       │
  ├─────────────────┼──────────────────────┼───────────────────────────┼──────────────────────────┤
  │ Stale cwd-based │ Returns old          │ Nearly impossible in      │ Same proven pattern as   │
  │  cache          │ git-common-dir       │ short-lived CLI           │ RepoRoot()               │
  ├─────────────────┼──────────────────────┼───────────────────────────┼──────────────────────────┤
  │ Corrupt .git    │ Would return ""      │ External problem          │ Same behavior as         │
  │ file            │                      │                           │ subprocess failure       │
  └─────────────────┴──────────────────────┴───────────────────────────┴──────────────────────────┘

  The risk assessment is accurate in identifying the right areas but overstates the practical
  impact. Both changes are display/performance-only with no write-path consequences, and every
  failure mode degrades to the same behavior as the old code on error.
  
  ```

File: cmd/entire/cli/status.go [hunk 1]
```diff
@@ -286,12 +286,49 @@ func writeActiveSessions(w io.Writer) {
...
+
+	data, err := os.ReadFile(headPath) //nolint:gosec // path constructed from .git/HEAD
+	if err != nil {
+		return ""
+	}
+
+	ref := strings.TrimSpace(string(data))
+
+	// Symbolic ref: "ref: refs/heads/<branch>"
+	if strings.HasPrefix(ref, "ref: refs/heads/") {
```
pjbgf, line 332: If the Git repository is using `refStorage=reftable`, `HEAD` will point to `ref: refs/heads/.invalid`, instead of the correct branch. In that specific case, it may be worth shelling out to `git`.

Refs:
- https://github.com/git/git/blob/73fd77805fc6406f31c36212846d9e2541d19321/Documentation/technical/reftable.adoc?plain=1#L911

evisdren: good call out - updated to support this case

PR #440, opened 2026-02-20
setup: iterate agents instead of hard-coding each one

squishykid:
## Summary
- Refactors uninstall logic to dynamically iterate over registered agents instead of hard-coding Claude Code, Gemini CLI, and OpenCode individually
- Replaces three separate `check*HooksInstalled()` functions with a single `installedAgentHooks()` that uses `agent.List()` and the `HookSupport` interface

## Test plan
- [x] Run `mise run test:ci` to verify all tests pass
- [x] Test `entire enable --uninstall` with various agents installed
- [x] Verify new agents added to the registry are automatically picked up without code changes

### DIFF
--- a/cmd/entire/cli/setup_test.go
+++ b/cmd/entire/cli/setup_test.go
@@ -1238,6 +1238,32 @@ func TestDetectOrSelectAgent_ReRun_NoTTY_KeepsInstalled(t *testing.T) {
 	}
 }
 
+// checkClaudeCodeHooksInstalled checks if Claude Code hooks are installed.
+func checkClaudeCodeHooksInstalled() bool {
+	ag, err := agent.Get(agent.AgentNameClaudeCode)
+	if err != nil {
+		return false
+	}
+	hookAgent, ok := ag.(agent.HookSupport)
+	if !ok {
+		return false
+	}
+	return hookAgent.AreHooksInstalled()
+}
+
+// checkGeminiCLIHooksInstalled checks if Gemini CLI hooks are installed.
+func checkGeminiCLIHooksInstalled() bool {
+	ag, err := agent.Get(agent.AgentNameGemini)
+	if err != nil {
+		return false
+	}
+	hookAgent, ok := ag.(agent.HookSupport)
+	if !ok {
+		return false
+	}
+	return hookAgent.AreHooksInstalled()
+}
+
 func TestUninstallDeselectedAgentHooks(t *testing.T) {
 	// Cannot use t.Parallel() because we use t.Chdir
 	setupTestRepo(t)


cmd/entire/cli/setup.go
```diff
@@ -1245,48 +1194,24 @@ func checkEntireDirExists() bool {
 }
 
 // removeAgentHooks removes hooks from all agents that support hooks.
-func removeAgentHooks(w io.Writer) error {
+// It takes a list of agents to process, so it only removes hooks for the agents that were previously listed.
+func removeAgentHooks(w io.Writer, agents []agent.AgentName) error {
```
Soph: Hmm, why would we not remove all hooks? Like it's not removing any hook, but just the entire hooks.

squishykid: We print the list of hooks we detected above, and then my thinking was to remove exactly that list later- perhaps something happens in the environment which causes hooks to be installed for another agent.

But since this is within the scope of “remove entire” it makes sense to just remove everything.

squishykid: I’ll do a follow up

cmd/entire/cli/setup.go
```diff
@@ -1247,46 +1202,21 @@ func checkEntireDirExists() bool {
...
-			if err := hookAgent.UninstallHooks(); err != nil {
-				errs = append(errs, err)
-			} else if wasInstalled {
-				fmt.Fprintln(w, "  Removed OpenCode hooks")
-			}
+		if err := hs.UninstallHooks(); err != nil {
+			errs = append(errs, err)
+		} else {
+			fmt.Fprintf(w, "  Removed %s hooks\n", ag.Type())
 		}
```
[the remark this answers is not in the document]
squishykid: Missed this again :/

cmd/entire/cli/setup.go
```diff
@@ -1095,18 +1093,14 @@ func runUninstall(w, errW io.Writer, force bool) error {
...
-		if len(agentNames) > 0 {
-			fmt.Fprintf(w, "  - Agent hooks (%s)\n", strings.Join(agentNames, ", "))
+		if len(agentsWithInstalledHooks) > 0 {
+			displayNames := make([]string, 0, len(agentsWithInstalledHooks))
+			for _, name := range agentsWithInstalledHooks {
+				if ag, err := agent.Get(name); err == nil {
+					displayNames = append(displayNames, string(ag.Type()))
+				}
+			}
+			fmt.Fprintf(w, "  - Agent hooks (%s)\n", strings.Join(displayNames, ", "))
```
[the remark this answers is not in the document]
squishykid: ignoring this

PR #446, opened 2026-02-20
Prevent argument injection in git CLI calls

pfleidi:
Entire-Checkpoint: ac699460bd01

  - CheckoutBranch() and performGitResetHard() pass string arguments directly to git subprocesses. A ref starting with - would be interpreted as a git flag instead of a ref, causing unintended behavior (e.g., git checkout -b evil creates a
  branch, git reset --hard -q silently resets to HEAD).
  - Add input validation rejecting refs/hashes that start with - before they reach git, with tests demonstrating both attack vectors.

  ## Test plan

  - TestCheckoutBranch/rejects_ref_starting_with_dash_to_prevent_argument_injection — verifies CheckoutBranch("-b evil") returns an error instead of creating a branch
  - TestPerformGitResetHard_RejectsArgumentInjection — verifies performGitResetHard("-q") returns an error instead of silently resetting to HEAD
  - Full CI suite passes (mise run fmt && mise run lint && mise run test:ci)


cmd/entire/cli/git_operations.go
```diff
@@ -291,6 +291,9 @@ func BranchExistsLocally(branchName string) (bool, error) {
 // Should be switched back to go-git once we upgrade to go-git v6
 // Returns an error if the ref doesn't exist or checkout fails.
 func CheckoutBranch(ref string) error {
+	if strings.HasPrefix(ref, "-") {
+		return fmt.Errorf("checkout failed: invalid ref %q", ref)
+	}
```
evisdren, line 296: This is probably okay in the short term. 

But I wonder if we should actually split this out into SwitchBranch and CheckoutBranch functions. CheckoutBranch is a bit overloaded. It seems like the intention of the function is just to switch to a local branch or If you go with switch branch, then you can pass in `-- <branch>` which will treat everything after the `--` as a branch name and fail if a `-` is passed as an evil flag. 

The nuance here is that `git switch` has to handle both commit hashes and branch names. For commit hashes you have to pass in `git switch --detach <commithash>`, which is actually clearer in my opinion. The challenge is then determining the difference between a commit hash and a branch. You could shell out to a git sub process to do that (most reliable), but that seems wasteful. 

The actual better long term strategy here would be to split up the swtich into SwitchBranchWithName() and SwitchBranchWithCommit() and pass in the right argument. That would fully remove the argument injection problem since the argument would always come after a `--` and be treated as a literal and make it clearer in the codebase what argument we're passing.

cmd/entire/cli/rewind.go
```diff
@@ -1179,6 +1179,9 @@ func countCommitsBetween(repo *git.Repository, ancestor, descendant plumbing.Has
 // Uses the git CLI instead of go-git because go-git's HardReset incorrectly
 // deletes untracked directories (like .entire/) even when they're in .gitignore.
 func performGitResetHard(commitHash string) error {
+	if strings.HasPrefix(commitHash, "-") {
+		return fmt.Errorf("reset failed: invalid commit hash %q", commitHash)
+	}
```
evisdren, line 1184: pretty much same comment as above

cmd/entire/cli/git_operations.go
pfleidi: Yeah, `SwitchBranch` would also address the case where legitimate branch names beginning with a `-` would work and not confuse git unnecessarily.

 I like the idea but that might be a change for down the road. What do you think?

PR #454, opened 2026-02-22
Entire Enable performance optimzations

evisdren:
Some optimizations to the `entire enable ` command:

image

### DIFF
--- a/mise-tasks/bench/compare
+++ b/mise-tasks/bench/compare
@@ -6,6 +6,7 @@ BENCH_PATTERN="${BENCH_PATTERN:-.}"
 BENCH_PKG="${BENCH_PKG:-./...}"
 BENCH_COUNT="${BENCH_COUNT:-6}"
 BENCH_TIMEOUT="${BENCH_TIMEOUT:-10m}"
+BENCH_TAGS="${BENCH_TAGS:-}"
 BASE_REF="${BASE_REF:-main}"
 
 if ! command -v benchstat &>/dev/null; then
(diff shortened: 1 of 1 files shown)


cmd/entire/cli/strategy/hooks.go
```diff
@@ -176,7 +177,7 @@ func InstallGitHook(silent bool) (int, error) {
 		return 0, fmt.Errorf("failed to create hooks directory: %w", err)
 	}
 
-	specs := buildHookSpecs(hookCmdPrefix())
+	specs := buildHookSpecs(hookCmdPrefix(localDev))
 	installedCount := 0
```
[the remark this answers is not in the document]
evisdren: resolved

cpu_enable.prof
Soph, line 1: I think this shouldn't be here, and we might want to add `*.prof` to .gitignore?

mise.toml
```diff
@@ -122,65 +122,92 @@ echo "Profile saved to mem.prof. View with: go tool pprof -http=:8080 mem.prof"
 """
 
 [tasks."bench:compare"]
-description = "Compare benchmarks between current branch and main"
+description = "Compare benchmarks between current branch and base ref (requires benchstat: go install golang.org/x/perf/cmd/benchstat@latest)"
 run = """
```
Soph: Maybe this should be put into `mise-tasks` as script? (sorry late to the party, but just saw it in this PR)

PR #458, opened 2026-02-22
update to go 1.26 (debug/buildinfo is a new default package)

Soph:
- Update Go from 1.25.6 to 1.26.0
- Update golangci-lint from 2.8.0 to 2.10.1 (required for Go 1.26 support)
- Rename buildinfo package to versioninfo (Go 1.26 added debug/buildinfo to stdlib, causing linter conflict)
- Add gosec exclusions for CLI-appropriate rules with hybrid approach:
  - Global: G204/G702 (subprocess), G705 (XSS) - always irrelevant for CLI
  - Targeted //nolint comments for G115, G703, G704 - preserves future detection

### DIFF
--- a/cmd/entire/cli/status_style.go
+++ b/cmd/entire/cli/status_style.go
@@ -66,7 +66,7 @@ func shouldUseColor(w io.Writer) bool {
 		return false
 	}
 	if f, ok := w.(*os.File); ok {
-		return term.IsTerminal(int(f.Fd()))
+		return term.IsTerminal(int(f.Fd())) //nolint:gosec // G115: uintptr->int is safe for fd
 	}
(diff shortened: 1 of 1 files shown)


gtrrz-victor:
> Looks reasonable to me! Copilot's comment seems sensible but you haven't changed the behaviour in this PR so up to you whether you want to address it now.
> 
> As an aside, there are so many `context.Background()` around – we should do a sweep and fix that up.

> I will merge this and create another PR setting those timeouts

We have already a 2 seconds timeout. Is it too long ?

gtrrz-victor (dismissed):

gtrrz-victor (approved):

Issue #460, opened 2026-02-22
Global git hooks create .entire/logs/ directory in repos where Entire is not enabled


knowsuchagency:
## Description

After installing Entire CLI and running `entire enable` in one project, the global git hooks that Entire installs (`~/.config/git/hooks/`) create a `.entire/logs/` directory in **every** git repository on any git operation (commit, push, etc.) — even repos where Entire has never been enabled.

## Steps to reproduce

1. Install Entire CLI
2. Run `entire enable` in project A
3. Go to project B (where Entire is **not** enabled — `entire status` shows "not set up")
4. Make a git commit
5. Observe that `.entire/logs/entire.log` is created in project B

## Expected behavior

The global hooks should check whether Entire is enabled in the current repository before creating any files or directories. If `entire status` reports "not set up", the hooks should be a no-op with no side effects.

## Actual behavior

Every hook invocation (`entire hooks git post-commit`, `commit-msg`, `prepare-commit-msg`, `pre-push`) eagerly creates `.entire/logs/` regardless of whether the repo has Entire enabled.

## Environment

- Entire CLI installed via Homebrew (Linuxbrew)
- Global git hooks path: `~/.config/git/hooks/`
- OS: Ubuntu 24.04

Soph:
Thanks for reporting this. By default the entire cli is not installing hooks in the global git hooks path. It's using `git rev-parse --git-path hooks` to figure out what path is configured for the repo and then use that. My assumption is, that the repo where you installed it had `core.hooksPath` set to `~/.config/git/hooks`

Can you confirm this?

We will take a look if we can make all hooks first check if the repo they are called in actually has `entire` enabled (by looking for `.entire/settings.json` and if it has not `enabled: false` in it.

knowsuchagency:
Thank you! I can confirm the repo where I observed this behavior did have `core.hooksPath` set to `~/.config/git/hooks`

PR #467, opened 2026-02-23
Make git hooks a noop if entire is not enabled in the repo

Soph:
In the case of git hooks being installed but entire not enabled in the repository this will make any git hook call a noop. 

https://github.com/entireio/cli/issues/460


toothbrush (approved):
Great idea :+1: Thanks!

PR #494, opened 2026-02-25
refactor: CalculateTokenUsage accepts bytes instead of file path

squishykid:
## Summary
- Changes the `TokenCalculator` interface to accept `[]byte` data instead of a `sessionRef string` (file path), so callers own file I/O and agents only parse
- Each agent's `CalculateTokenUsage` now operates on in-memory bytes rather than reading files internally
- Replaces hardcoded agent-type switch in `manual_commit_condensation.go` with the `TokenCalculator` interface via the agent registry

## Test plan
- [x] Existing unit tests updated to pass bytes directly
- [ ] `mise run test:ci` passes

### DIFF
--- a/cmd/entire/cli/agent/agent.go
+++ b/cmd/entire/cli/agent/agent.go
@@ -165,8 +165,8 @@ type TranscriptPreparer interface {
 type TokenCalculator interface {
 	Agent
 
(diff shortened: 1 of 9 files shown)


squishykid:
use https://github.com/entireio/cli/pull/509/ instead

Issue #489, opened 2026-02-25
Git hooks fail with `entire: command not found` when committing from Xcode


petrenk0n:
### What happened?

The git hooks installed by `entire enable` call `entire` by name without an absolute path. When committing from Xcode (which launches git with a minimal environment that doesn't include `~/.local/bin` on PATH), the hook fails: `.git/hooks/commit-msg: line 4: entire: command not found`

Commits from the terminal work fine because the shell profile adds `~/.local/bin` to PATH. Xcode doesn't source shell profiles, so the hook can't find the binary.

### Steps to reproduce

1. entire enable in any git repo
2. Open the project in Xcode and commit via Xcode's source control UI
3. Commit fails with .git/hooks/commit-msg: line 4: entire: command not found

### Entire CLI version

Entire CLI 0.4.7 (80695ef9)

### OS and architecture

Darwin 25.3.0 arm64

### Agent

Claude Code

### Strategy

manual-commit (default)

### Terminal

Ghostty

### Logs / debug output

```bash

```

Soph:
Thanks for reporting. What would be your expectation, that we figure out where `entire` is in your path when you run `entire enable` and then use that path to configure the git hooks?

petrenk0n:
Yes, when `entire enable` installs the hooks, it already knows its own binary path (or can resolve it). Writing the absolute path into the hook would make it work in any environment (Xcode, GUI git clients, IDEs, cron, etc.) instead of relying on the invoking process having the right PATH.

Something like:
```sh
`/Users/me/.local/bin/entire hooks git commit-msg "$1" || exit 1`

instead of:

`entire hooks git commit-msg "$1" || exit 1`

Soph:
@petrenk0n with v0.4.9 we shipped a new `--absolute-git-hook-path` flag you can use when running `entire enable` that then uses absolute paths when setting up the git hooks, can you give it a try?

petrenk0n:
@Soph It works now, thanks!

mvanhorn:
I submitted #664 to make absolute paths the default behavior for git hooks.

The `--absolute-git-hook-path` flag (v0.4.9) fixed this for users who knew about it, but the default still used bare `entire` which breaks in GUI git clients. This PR changes `hookCmdPrefix()` to always resolve the absolute binary path via `os.Executable()`, with a graceful fallback to bare `entire` if resolution fails. Users no longer need to pass `--absolute-git-hook-path` - it just works.

[...]

PR #495, opened 2026-02-25
add --absolute-git-hook-path to entire enable, this will setup your hooks with absolute paths to the entire binary

Soph:
Addresses https://github.com/entireio/cli/issues/489

This adds an optional setting to setup git hooks with absolute paths. I considered doing this always but I feel there is a risk path changes and then it just stops working when the most common use case is you have a full path available. 

Entire-Checkpoint: 43cdf4753c06


cmd/entire/cli/settings/settings.go
```diff
@@ -41,6 +41,11 @@ type EntireSettings struct {
 	// StrategyOptions contains strategy-specific configuration
 	StrategyOptions map[string]any `json:"strategy_options,omitempty"`
 
+	// AbsoluteGitHookPath embeds the full binary path in git hooks instead of
+	// bare "entire". This is needed for GUI git clients (Xcode, Tower, etc.)
+	// that don't source shell profiles and can't find "entire" on PATH.
+	AbsoluteGitHookPath bool `json:"absolute_git_hook_path,omitempty"`
```
toothbrush, line 48: We don't really have a concept of "user settings" yet, right?  Like, `$HOME/.config/entire/settings.json`?  I was thinking last night that i would actually like such a thing.  Firstly, that would be a good place to put a setting like `absolute_git_hook_path`, because maybe i use Xcode but my colleague uses Vim in the terminal, so we have different needs for the same repo?  At the moment we could put that in `$PROJECT/.entire/settings.local.json`, but then as an Xcode user i have to do that for every single `entire enable`, right?  That's inconvenient.

The other reason i like the idea of a "user global" settings file is that i have my dotfiles in version control by default.  That would allow me to also version `$HOME/.config/entire/settings.json`, which might allow me to store `entire enable/disable` status of a particular project where i can see it, rather than in a git-ignored file inside the given project.  That would also allow it to persist past delete-and-re-clone.  

I guess there are two thoughts here, and only one of them is a strong argument IMHO.  What do you think?  Would a "user global" entire settings be useful?

We could add that independently of this PR and have settings cascade as usual, "command flag > ENV > local settings > project settings > user global settings".  Or something like that.

toothbrush (commented):
I think this looks good!  Only concern i have is the flag's name.  The setting influences both Git hooks as well as agent hooks, so should we call it `--absolute-hook-paths`?

cmd/entire/cli/settings/settings.go
gtrrz-victor: I agree on moving absolute_git_hook_path into the global user settings (we already have it). The problem with doing it for the agent hooks too is that the agent config gets pushed to the repo, and not everyone would have the same absolute path.

gtrrz-victor: I will merge this one, and then create another PR to move settings into global user settings:
- enable
- telemetry
- auto push sessions
- absolute_git_hook_path

Soph:
@toothbrush no this is really only changing the git hooks. I felt for the agent hooks this doesn't make sense since relative paths are always better I think and the agent is started from an interactive shell so paths should be i.O. for these hooks.

cmd/entire/cli/settings/settings.go
Soph: good idea! thanks :)

toothbrush:
> @toothbrush no this is really only changing the git hooks. 

Ah yes, i looked again, carefully :+1:

> I felt for the agent hooks this doesn't make sense since relative paths are always better I think and the agent is started from an interactive shell so paths should be i.O. for these hooks.

That surprises me, is that also true for e.g. Cursor IDE?  Typically (in macOS at least) GUI apps have pretty bare-bones environments.

Soph:
@toothbrush yes, cursor IDE works, and if we run into an issue with another IDE I would change it only there.

PR #503, opened 2026-02-25
Replace FlattenTree+BuildTreeFromEntries with ApplyTreeChanges in sha…

evisdren:
update shadowbranch write process to not read every tree entry and insetad be selective in what is needed to be read. Create an array to track changes and then apply those. 

Entire-Checkpoint: f4c624aac28e


evisdren:
bugbot run

PR #513, opened 2026-02-26
fix: isolate integration tests from user git config

toothbrush:
The TestShadow_UntrackedFilePreservation test was failing because the user's global gitignore was leaking into the test environment.

Root cause: The user's ~/.gitconfig sets core.excludesFile to ~/.gitignore-global, which contains a `*.log` pattern. During rewind, collectUntrackedFiles() calls `git ls-files --others --exclude-standard`, and the --exclude-standard flag reads the global gitignore. This caused `temp-debug.log` (an untracked file the test expected to be deleted on rewind) to be treated as gitignored and therefore invisible to the rewind cleanup logic — only `build-artifacts.zip` was deleted.

Fix: Set GIT_CONFIG_GLOBAL=/dev/null and GIT_CONFIG_SYSTEM=/dev/null in cliEnv() so all CLI subprocess invocations in integration tests are isolated from user and system git configuration. This prevents global gitignore patterns, aliases, and other config from affecting test behavior.

Entire-Checkpoint: 752661e73b09

### DIFF
--- a/cmd/entire/cli/integration_test/testenv.go
+++ b/cmd/entire/cli/integration_test/testenv.go
@@ -107,14 +107,25 @@ func (env *TestEnv) Cleanup() {
 	// No-op - temp dirs are cleaned up by t.TempDir()
 }
 
-// gitIsolatedEnv returns os.Environ() with git isolation variables appended.
+// gitIsolatedEnv returns os.Environ() with git isolation variables set.
 // This prevents user/system git config (global gitignore, aliases, etc.) from
 // affecting test behavior. Use this for any exec.Command that runs git or the
 // CLI binary in integration tests.
 //
 // See https://git-scm.com/docs/git#Documentation/git.txt-GITCONFIGGLOBAL
+//
+// Existing GIT_CONFIG_GLOBAL/GIT_CONFIG_SYSTEM entries are filtered out before
+// appending overrides to ensure they take effect regardless of parent env.
 func gitIsolatedEnv() []string {
-	return append(os.Environ(),
+	env := os.Environ()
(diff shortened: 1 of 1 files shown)


cmd/entire/cli/integration_test/testenv.go
```diff
@@ -107,10 +107,22 @@ func (env *TestEnv) Cleanup() {
 	// No-op - temp dirs are cleaned up by t.TempDir()
 }
 
+// gitIsolatedEnv returns os.Environ() with git isolation variables appended.
+// This prevents user/system git config (global gitignore, aliases, etc.) from
+// affecting test behavior. Use this for any exec.Command that runs git or the
+// CLI binary in integration tests.
```
toothbrush, line 113:
```suggestion
// affecting test behavior. Use this for any exec.Command that runs git or the
// CLI binary in integration tests.
//
// See https://git-scm.com/docs/git#Documentation/git.txt-GITCONFIGGLOBAL
```

Soph (approved):

squishykid:
perhaps similar to? https://github.com/entireio/cli/issues/261

toothbrush:
> perhaps similar to? #261

I don't think so.  The scenario here is that my global gitignore file had `*.log` as an entry, but the test suite expected `git status` to show `something.log` as untracked (roughly).  This patch makes it so a user's `~/.gitconfig` or `~/.gitignore-global` or whatever they have configured, are not active, so it's testing "default out of the box Git". 

The linked issue on the other hand is talking about the fact that we clobber(ed) `.git/hooks/..` in a user's repo, IIUC.

PR #516, opened 2026-02-26
fix: update empty local metadata branch from remote on enable

dvydra:
## Problem

Users who already ran `entire enable` on old code have an empty orphan `entire/checkpoints/v1` branch. When they update to the new code (from #511) and run `enable` again, it sees the local branch exists and does nothing — leaving `entire explain` broken.

This is the recovery path for existing installs. PR #511 prevents the problem for fresh clones; this PR fixes repos already in the broken state.

**Discussion:** https://entireio.slack.com/archives/C0A095SNK32/p1772083910982219?thread_ts=1772052543.300359&cid=C0A095SNK32

## Fix

When the local `entire/checkpoints/v1` exists, check if it's an empty orphan (single commit with empty tree). If so, and the remote has real data, update the local branch to match the remote.

Non-empty local branches are never overwritten.

## Test plan

- [x] `TestEnsureMetadataBranch/updates_empty_orphan_from_remote` — detects empty orphan, updates from remote
- [x] Existing tests still pass (fresh clone, no remote)
- [x] `mise run fmt && mise run lint && mise run test:ci` clean

### DIFF
--- a/cmd/entire/cli/strategy/common.go
+++ b/cmd/entire/cli/strategy/common.go
@@ -285,15 +285,35 @@ func resolveAgentType(ctxAgentType agent.AgentType, state *SessionState) agent.A
 	return DefaultAgentType
 }
 
-// EnsureMetadataBranch creates the local entire/checkpoints/v1 branch if it doesn't exist.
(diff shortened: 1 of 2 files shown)


dvydra:
superceded by #533

PR #518, opened 2026-02-26
fix: reconcile disconnected local/remote metadata branches on enable

dvydra:
## Problem

When a user hit the empty-orphan bug (#511) and then worked normally, the CLI recorded checkpoints onto the local orphan. Now the local `entire/checkpoints/v1` has real data on a completely disconnected history from the remote — no common ancestor, both containing real checkpoint data.

Neither #511 (fresh clone fix) nor #516 (empty orphan update) handle this case because the local branch isn't empty.

**Discussion:** https://entireio.slack.com/archives/C0A095SNK32/p1772083910982219?thread_ts=1772052543.300359&cid=C0A095SNK32

## Fix

`EnsureMetadataBranch()` now detects when local and remote metadata branches have real data but differ, and uses `git merge-base --is-ancestor` to determine the relationship:

1. **Remote is ancestor of local** → already synced, nothing to do
2. **Local is ancestor of remote** → fast-forward local to remote
3. **Neither (disconnected/diverged)** → merge both trees into a merge commit

The merge is safe because checkpoint data is sharded by random ID (`<id[:2]>/<id[2:]>/`) with no shared index or manifest — different checkpoints never write to the same paths.

Uses the same tree-merge pattern as `fetchAndMergeSessionsCommon()` in `push_common.go`: flatten both trees via `FlattenTree()`, combine entries, rebuild via `BuildTreeFromEntries()`, create merge commit with both parents.

## Test plan

- [x] `TestEnsureMetadataBranch_ReconcilesDisconnectedBranches` — creates disconnected orphan with local data, verifies merge contains both sides
- [x] `TestEnsureMetadataBranch_FastForwardsWhenBehind` — local behind remote, verifies fast-forward
- [x] Existing tests still pass (fresh clone, empty orphan, no remote)
- [x] `mise run fmt && mise run lint && mise run test:ci` clean

### DIFF
--- a/cmd/entire/cli/explain.go
+++ b/cmd/entire/cli/explain.go
@@ -886,6 +886,11 @@ func walkFirstParentCommits(repo *git.Repository, from plumbing.Hash, limit int,
 //   - On default branch (main/master): show all checkpoints in history (up to limit)
 //   - Includes both committed checkpoints (entire/checkpoints/v1) and temporary checkpoints (shadow branches)
 func getBranchCheckpoints(ctx context.Context, repo *git.Repository, limit int) ([]strategy.RewindPoint, error) {
+	// Ensure disconnected metadata branches are reconciled before reading
+	if err := strategy.EnsureMetadataReconciled(repo); err != nil {
+		logging.Warn(ctx, "metadata reconciliation failed", "error", err)
+	}
+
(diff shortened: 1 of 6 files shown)


dvydra:
superceded by #533

PR #534, opened 2026-02-26
Support resuming from squash merge commits with multiple checkpoints

pfleidi:
## Summary

- Support `entire resume` on squash merge commits that contain multiple `Entire-Checkpoint` trailers
- Fix checkpoint ordering bug where git CLI squash merges list trailers newest-first, causing the oldest transcript to overwrite the newest on disk
- Deduplicate sessions by `SessionID` across checkpoints, keeping the most recent transcript

## Problem

When a feature branch with multiple commits (each with its own `Entire-Checkpoint` trailer) is squash-merged, all trailers end up in a single commit message. `entire resume` previously only parsed the first trailer, losing all other sessions.

Additionally, GitHub squash merges list trailers chronologically (oldest first), but git CLI squash merges list them in reverse order (newest first). Since `RestoreLogsOnly` writes session files eagerly, the last checkpoint processed wins on disk — meaning reverse-ordered trailers caused the oldest transcript to overwrite the newest.

## Changes

**`cmd/entire/cli/trailers/trailers.go`** — Add `ParseAllCheckpoints()` to extract all `Entire-Checkpoint` trailers from a commit message (not just the first).

**`cmd/entire/cli/resume.go`** — 
- Switch from `ParseCheckpoint` (single) to `ParseAllCheckpoints` (multi) throughout `findBranchCheckpoint` and `findCheckpointInHistory`
- Change `branchCheckpointResult.checkpointID` → `checkpointIDs []CheckpointID`
- Add `resumeMultipleCheckpoints`: reads metadata for all checkpoint IDs, sorts by `CreatedAt` ascending, then restores in order so the newest checkpoint always writes last
- Add `deduplicateSessions`: merges sessions across checkpoints, keeping the one with the latest `CreatedAt` when a `SessionID` appears in multiple checkpoints
- Extract `displayRestoredSessions` to share session display logic between single and multi-checkpoint paths

**`cmd/entire/cli/resume_test.go`** — Add unit tests for `deduplicateSessions` (including three-occurrence staleness edge case), `findCheckpointInHistory` with multiple trailers, `findBranchCheckpoint` with squash merge, and checkpoint timestamp sorting.

**`cmd/entire/cli/integration_test/resume_test.go`** — Add integration test simulating a full squash merge workflow: two sessions on a feature branch, squash merge to main, then `entire resume` restoring both sessions.

**`cmd/entire/cli/trailers/trailers_test.go`** — Add tests for `ParseAllCheckpoints`.

## Caveat: Optimization

I thought about optimizing this process by only restoring the very latest checkpoint, but realized that this wouldn't work as expected:

A committed checkpoint only contains the sessions that were condensed at that specific commit. When a feature branch is squash merged, Git/GitHub preserves all Entire-Checkpoint trailers from the original commits in the squash commit body. Each trailer points to a different checkpoint on entire/checkpoints/v1, and different checkpoints may contain different sessions.

Since there is no single checkpoint that aggregates all sessions from the branch, the resume command must iterate over every checkpoint ID found in the squash commit and restore each one. Sessions are de-duplicated by ID, keeping the entry with the latest CreatedAt.

Optimizing by only restoring the latest checkpoint would silently drop any session whose last activity was on an earlier commit.

## Test plan

- [x] `mise run test:ci` passes (unit + integration)
- [x] Integration test `TestResume_SquashMergeMultipleCheckpoints` covers the end-to-end squash merge flow
- [x] Unit test `TestResumeMultipleCheckpoints_SortsByCreatedAt` verifies reverse-ordered checkpoints are sorted correctly
- [x] Unit tests for `deduplicateSessions` cover no-duplicates, newer-wins, older-loses, three-occurrence, and mixed scenarios

### DIFF
--- a/e2e/tests/resume_test.go
+++ b/e2e/tests/resume_test.go
@@ -1,0 +1,239 @@
+//go:build e2e
+
+package tests
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/entireio/cli/e2e/entire"
+	"github.com/entireio/cli/e2e/testutil"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestResumeFromFeatureBranch: agent creates a file on a feature branch and
+// user commits, then switches back to main and runs `entire resume feature`.
+// Verifies the branch is switched and the session is restored.
+func TestResumeFromFeatureBranch(t *testing.T) {
+	testutil.ForEachAgent(t, 3*time.Minute, func(t *testing.T, s *testutil.RepoState, ctx context.Context) {
+		mainBranch := testutil.GitOutput(t, s.Dir, "branch", "--show-current")
+
+		// Commit files from `entire enable` so main has a clean working tree
+		// for branch switching (mirrors a real repo where .gitignore is tracked).
+		s.Git(t, "add", ".")
+		s.Git(t, "commit", "-m", "Enable entire")
+
+		// Do agent work on a feature branch.
+		s.Git(t, "checkout", "-b", "feature")
+
+		_, err := s.RunPrompt(t, ctx,
+			"create a file at docs/hello.md with a paragraph about greetings. Do not ask for confirmation, just make the change.")
+		if err != nil {
+			t.Fatalf("agent failed: %v", err)
+		}
+
+		s.Git(t, "add", ".")
+		s.Git(t, "commit", "-m", "Add hello doc")
+		testutil.WaitForCheckpoint(t, s, 15*time.Second)
+
+		// Switch back to main and resume the feature branch.
+		s.Git(t, "checkout", mainBranch)
+
+		out, err := entire.Resume(s.Dir, "feature")
+		require.NoError(t, err, "entire resume failed: %s", out)
+
+		current := testutil.GitOutput(t, s.Dir, "branch", "--show-current")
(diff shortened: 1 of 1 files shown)


toothbrush:
As someone who prefers¹ squash-merge to messy merge-tennis with careless commit messages, this effort really pleases me.

1. Out of an abundance of caution i want to add that doesn't mean it's my favouritest merge.

cmd/entire/cli/resume.go
```diff
@@ -420,51 +547,45 @@ func resumeSession(ctx context.Context, sessionID string, checkpointID id.Checkp
 		return resumeSingleSession(ctx, ag, sessionID, checkpointID, repoRoot, force)
 	}
 
-	// Sort sessions by CreatedAt so the most recent is last (for display).
-	// This fixes ordering when subdirectory index doesn't reflect activity order.
-	sort.Slice(sessions, func(i, j int) bool {
```
pfleidi, line 425: This happens earlier in the process now.

cmd/entire/cli/resume.go
```diff
@@ -131,7 +132,7 @@ func resumeFromCurrentBranch(ctx context.Context, branchName string, force bool)
 	if err != nil {
 		return err
 	}
-	if result.checkpointID.IsEmpty() {
+	if len(result.checkpointIDs) == 0 {
```
pfleidi, line 135: `result.checkpointID` changed to `result.checkpointIDs` which is a different type.

pfleidi:
@cursor review

gtrrz-victor (approved):

Issue #490, opened 2026-02-25
OpenCode checkpoints have empty sessions in the UI


petrenk0n:
### What happened?

OpenCode checkpoints are visible in the UI but are missing sessions. Sessions are present when using `entire status`

### Entire CLI version

Entire CLI 0.4.7 (80695ef9)

### Agent

OpenCode

### Strategy

manual-commit (default)

### Terminal

Ghostty

### Logs / debug output

```bash

```

gtrrz-victor:
Thanks for reporting this! Could you share a bit more context so I can help?

Specifically:
- The output of `entire status`
- The logs located at `<repoRootFolder>/.entire/logs/entire.log`

petrenk0n:
● Enabled · manual-commit · branch <Redacted>

── Active Sessions ────────────────────────────────────────────────────────────

OpenCode · ses_372
> <Redacted>
started 12h ago · active 8h ago · tokens 0

───────────────────────────────────────────────────────────────────────────────
1 session

petrenk0n:
```
{"time":"2026-02-24T22:02:42.151908-05:00","level":"INFO","msg":"turn-start","session_id":"<REDACTED_SESSION_ID>","component":"lifecycle","agent":"opencode","event":"TurnStart","session_ref":"/Users/<REDACTED_USER>/Desktop/<REDACTED_FOLDER>/<REDACTED_APP>/.entire/tmp/<REDACTED_SESSION_ID>.json"}
{"time":"2026-02-24T22:04:05.349521-05:00","level":"INFO","msg":"turn-start","session_id":"<REDACTED_SESSION_ID>","component":"lifecycle","agent":"opencode","event":"TurnStart","session_ref":"/Users/<REDACTED_USER>/Desktop/<REDACTED_FOLDER>/<REDACTED_APP>/.entire/tmp/<REDACTED_SESSION_ID>.json"}
{"time":"2026-02-24T22:05:15.870885-05:00","level":"INFO","msg":"turn-start","session_id":"<REDACTED_SESSION_ID>","component":"lifecycle","agent":"opencode","event":"TurnStart","session_ref":"/Users/<REDACTED_USER>/Desktop/<REDACTED_FOLDER>/<REDACTED_APP>/.entire/tmp/<REDACTED_SESSION_ID>.json"}
[...]

Issue #488, opened 2026-02-25
Incorrect badge for OpenCode in the UI


petrenk0n:
### What happened?

Checkpoints made with OpenCode have Claude Code badges in the UI

### Entire CLI version

Entire CLI 0.4.7 (80695ef9)

### Agent

OpenCode

### Strategy

manual-commit (default)

### Terminal

Ghostty

### Logs / debug output

```bash

```

peyton-alt:
Thanks for reporting! To debug, could you run this for any checkpoint showing the wrong badge?

Using the checkpoint ID from the UI, split it as <first 2 chars>/<remaining chars>:

```
gh api "repos/YOUR_ORG/YOUR_REPO/contents/ab/1234567890/0/metadata.json?ref=entire/checkpoints/v1" \
  -H "Accept: application/vnd.github.raw"
```

Look for the `agent` field in the output, we want to see if it says "OpenCode", "Claude Code", or is missing entirely.

Also, could you confirm: did you run `entire enable --agent opencode` (or select OpenCode from the list of available agents during entire enable)? Thanks again!

petrenk0n:
I selected OpenCode from the list of agents using `entire enable`

gh command throws 404 because `entire/checkpoints/v1` is empty

peyton-alt:
This is the same root cause as #490, since entire/checkpoints/v1 is empty, no metadata was written during post-commit, so the UI defaults the badge to "Claude Code." We're tracking the fix in #490.

PR #541, opened 2026-02-27
fix turn-end hook chain for opencode checkpoints

peyton-alt:
OpenCode checkpoints weren't being created because the turn-end hook chain failed in two places:

1. `ParseHookEvent` ran `opencode export` as a subprocess — if it failed (binary not on PATH, DB locked), the hook died
2. `fileExists` was checked before `PrepareTranscript` had a chance to create the file

No checkpoints meant empty sessions (#490) and wrong badge (#488).

## Changes

- Defer `opencode export` from `ParseHookEvent` to `PrepareTranscript`
- Call `PrepareTranscript` before `fileExists` check
- Add session ID validation against path traversal
- Fall back to tracked `currentSessionID` in plugin when event omits it

### DIFF
--- a/cmd/entire/cli/agent/opencode/lifecycle.go
+++ b/cmd/entire/cli/agent/opencode/lifecycle.go
@@ -142,7 +142,7 @@ func (a *OpenCodeAgent) PrepareTranscript(ctx context.Context, sessionRef string
 // sessionTranscriptPath validates the session ID and returns the expected transcript path.
 func sessionTranscriptPath(ctx context.Context, sessionID string) (string, error) {
 	if err := validation.ValidateSessionID(sessionID); err != nil {
-		return "", err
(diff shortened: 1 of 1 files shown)


cmd/entire/cli/agent/opencode/entire_plugin.ts
```diff
@@ -127,6 +127,17 @@ export const EntirePlugin: Plugin = async ({ $, directory }) => {
           })
           break
         }
+
+        case "session.idle": {
```
Soph: When I looked at this last week, I found that `session.idle` is deprecated and it also didn't fire when directly running a prompt via the CLI. OpenCode code base said you should look for `session.status` with `type`  equal `idle`

Soph:
@cursor review

PR #550, opened 2026-02-27
don't update LastInteraction when only git hooks were triggered

Soph:
I noticed a lot of sessions still sitting around locally. And what jumped out was that all of them were constantly updated. Turns out that the commit hooks kept looking at older sessions - which is fine - but also kept bumping `LastInteractionTime` which made them never going to be cleaned up. Now a commit isn't really an interaction in a session, so let's not do this in this case.


khaong (approved):
LGTM. Git commit hooks only ever self-transition on phase, so removing ActionUpdateLastInteraction from EventGitCommit paths is correct and complete. Also pushed a small fix for the stale comment on LastInteractionTime in state.go.

PR #551, opened 2026-02-27
fix: resolve `entire explain` hanging on repos with many checkpoints

dvydra:
## Summary

`entire explain` (and any code path calling `getBranchCheckpoints`) hangs indefinitely on repos that have accumulated hundreds of checkpoints. This repo has 236 checkpoints on main and the command never returns — requires Ctrl-Z + kill -9.

**Root cause:** Two O(N) go-git operations that don't scale:

### 1. Full transcript parsing for list view prompts

`collectCheckpoint` called `ReadLatestSessionContent` + `extractPromptsFromTranscript` for **every checkpoint** — reading multi-MB JSONL transcript blobs from git and parsing them as JSON, just to extract the first user prompt for the list view display.

**Introduced in:** `0e7a05f1` (Feb 2, PR #130) — replaced the cheap `ReadSessionPromptFromTree` (reads tiny `prompt.txt`) with full transcript parsing.

**Fix:** New `readPromptFromMetadataTree` reads `prompt.txt` directly from the metadata branch tree. The metadata tree is fetched once and reused for all checkpoints — O(1) per checkpoint instead of O(transcript_size).

### 2. Full tree diff for temporary checkpoint filtering

`convertTemporaryCheckpoint` called `hasCodeChanges` which does `parentTree.Diff(commitTree)` via go-git for every temporary checkpoint. Each diff resolves hundreds of tree/blob objects through go-git's pure-Go SHA1 packfile reader (no ARM64 assembly), making each call very slow.

**Fix:** New `hasAnyChanges` compares tree hashes in O(1). For the list view, showing a few metadata-only checkpoints is acceptable — the full `hasCodeChanges` diff is preserved for contexts that need precise filtering.

### Performance

| Metric | Before | After |
|--------|--------|-------|
| `entire explain` | Hangs indefinitely | **344ms** |
| Transcript reads | 236 × multi-MB blobs | 0 (reads prompt.txt) |
| Tree diffs | 32 × full diffs | 32 × hash comparisons |

## Test plan

- [x] `mise run fmt` — clean
- [x] `mise run lint` — 0 issues  
- [x] `mise run test` — all pass
- [x] `mise run test:integration` — all pass
- [x] Manual test: `entire explain --no-pager` completes in 344ms
- [ ] Verify prompt text still shows correctly in list view
- [x] Verify `entire explain --checkpoint <id>` detail view still works (uses different code path)

@pfleidi — this was a gradual degradation (code introduced Feb 2, grew worse as checkpoints accumulated + large packfile created Feb 25). The fix is surgical — only changes the list view's data fetching strategy, doesn't touch the detail view or any other code paths.

### DIFF
--- a/cmd/entire/cli/explain_test.go
+++ b/cmd/entire/cli/explain_test.go
@@ -3626,11 +3626,6 @@ func TestGetBranchCheckpoints_ReadsPromptFromCommittedCheckpoint(t *testing.T) {
 		t.Fatalf("failed to create commit with checkpoint trailer: %v", err)
 	}
 
-	// Create .entire directory (needed for repo detection)
-	if err := os.MkdirAll(filepath.Join(tmpDir, ".entire"), 0o750); err != nil {
-		t.Fatalf("failed to create .entire dir: %v", err)
-	}
-
 	// Call getBranchCheckpoints and verify prompt is populated
 	points, err := getBranchCheckpoints(context.Background(), repo, 10)
 	if err != nil {


dvydra:
The second fix definitely fixed the problem. The first one might be a red herring
> Full transcript parsing for list view prompts

Check if reverting that bit makes a difference to timing.

gtrrz-victor (approved):

PR #560, opened 2026-02-28
fix(agent): use AgentConfigDir for local configuration priority and preserve hooks fallback

Gribesh:
### What happened?

Entire initializes correctly when the coding agent is started from the root directory of a project. However, when the coding agent is initiated from within a subdirectory of the same project, Entire was not properly evaluating the target directory (`.claude`, `.gemini`) dynamically favoring local project folder structures.

### How it's handled and the intuition behind it
1. Started by prioritizing the configuration lookup of the agents into the local subdirectory CWD. Previously, the logic incorrectly forced `WorktreeRoot()` universally at the root repository, stripping configuration tracking when users try to isolate hooks inside subdirectories.
2. Abstracted this cleanly into a new shared function: `paths.AgentConfigDir(ctx, folderName)`. By explicitly running an `os.Stat()` against the CWD first, it appropriately prioritizes localized subdirectories if the user instantiated them. If no configuration exists inside the CWD, it optimally falls back up to `WorktreeRoot()`.
3. In `--local-dev` mode, the generated hook commands now aggressively calculate the absolute file path deterministically (e.g. `go run /path.../cmd/entire/main.go hooks`). However, if WorkTree parsing fails (non-git repo context edge cases), we conservatively fall back onto parsing the classic runtime variables (e.g., `${CLAUDE_PROJECT_DIR}`, `${GEMINI_PROJECT_DIR}`) so that execution robustness is securely preserved and backward compatible.

### DIFF
--- a/cmd/entire/cli/bench_enable_test.go
+++ b/cmd/entire/cli/bench_enable_test.go
@@ -35,7 +35,7 @@ func BenchmarkEnableCommand(b *testing.B) {
 			b.StartTimer()
 
 			w := &bytes.Buffer{}
-			if err := setupAgentHooksNonInteractive(context.Background(), w, ag, true, false, false, false); err != nil {
+			if err := setupAgentHooksNonInteractive(context.Background(), w, ag, EnableOptions{LocalDev: true}); err != nil {
 				b.Fatalf("setupAgentHooksNonInteractive: %v", err)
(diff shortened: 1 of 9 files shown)


Gribesh:
Removing commits and closing as it couldn't fix the issue

PR #561, opened 2026-02-28
Fix entire cli hanging on `prepare-commit-msg` git hook

sbalay:
## Problem

[Reported this in Discord earlier](https://discord.com/channels/1468014954689855716/1476959081594880052).  After setting up the entire cli in my project, I wasnt able to commit anymore. `git commit` would just hang.

After some quick research, I found that the root cause for the hang was the entire cli hook: `entire hooks git prepare-commit-msg`, it just hangs.

## Proposed solution

**Disclaimer: This solution was written by claude code. I think it makes sense, that's why I'm sending this PR but my knowledge around how the entire cli works is limited**

- Fix prepare-commit-msg hook hanging on large repos by replacing go-git's worktree.Status() with native git diff --cached --name-only in getStagedFiles()
   - worktree.Status() scans the entire working tree (O(all files)) — computing hashes, reading .gitignore, comparing with the index — all in pure Go. On repos with deep node_modules or many packages (like stilla-server), this takes 20+ seconds or hangs indefinitely
   - git diff --cached --name-only only lists staged files (O(staged files)), completing in under a second

The codebase already uses exec.CommandContext(ctx, "git", ...) extensively, so this fix reuses and existing pattern, it is not creating a new one:
  - paths.WorktreeRoot() — git rev-parse --show-toplevel
  - HardResetWithProtection() in common.go — git reset --hard
  - CheckoutBranch() in git_operations.go — git checkout
  - getUntrackedFiles() in common.go — git ls-files --others
  - HasUncommittedChanges() in git_operations.go — git status --porcelain

## Test plan

- All existing TestPrepareCommitMsg tests pass
- Full test suite passes (go test ./...)
- Manual test: I built the CLI locally and tested it in my project. It no longer hangs and completes in less than a second

### DIFF
--- a/cmd/entire/cli/strategy/manual_commit_hooks.go
+++ b/cmd/entire/cli/strategy/manual_commit_hooks.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


gtrrz-victor:
First of all, thanks for jumping in and proposing this fix. 

I'm going to close this PR as we should have addressed those errors in the following PRs, which will be shipped in the next release.

- https://github.com/entireio/cli/pull/551
- https://github.com/entireio/cli/pull/553

PR #567, opened 2026-03-01
feat: add configurable branch_prefix and commit_prefix settings

moisei:
## Intro

Many organizations enforce strict policies on git branch naming conventions and commit message formats. With the current hardcoded `entire/` prefix and `Checkpoint` commit subject, Entire's branches and commits fail these checks out of the box - effectively blocking adoption in those environments. This PR addresses the issue by allowing users to configure custom branch name prefixes and commit message prefixes.
  
## Summary

- Adds two new settings to `.entire/settings.json` for customizing branch names and commit message prefixes
- `branch_prefix` (default `"entire/"`) — controls shadow branch prefix and metadata branch name (e.g. `"jfrog/"` → `jfrog/checkpoints/v1`)
- `commit_prefix` (default `"Checkpoint"`) — controls checkpoint commit subject prefix (e.g. `"JFrog Checkpoint"` → `JFrog Checkpoint: a3b2c4d5e6f7`)
- Enables running Entire on internal GitHub Enterprise without colliding with public conventions

## Changes

- **settings.go**: new `branch_prefix` / `commit_prefix` fields with getters, validation, and merge logic
- **paths.go**: `MetadataBranchName` changed from const to configurable var with `ConfigureBranchPrefix()`
- **temporary.go**: `ShadowBranchPrefix` changed from const to configurable var with `ConfigureBranchPrefix()`/`ConfigureCommitPrefix()`
- **committed.go**: commit subject uses configurable prefix instead of hardcoded `"Checkpoint"`
- **cleanup.go**: `IsShadowBranch()` uses configurable prefix instead of hardcoded regex
- **root.go, hooks_git_cmd.go, hook_registry.go**: configure prefixes from settings during CLI/hook startup
- **settings_test.go**: comprehensive tests for new fields (getters, loading, validation, merge)

## Test plan

- [x] All existing unit tests pass (`go test ./cmd/entire/cli/...`)
- [x] All integration tests pass (`go test -tags integration ./cmd/entire/cli/...`)
- [x] New settings tests cover: default values, explicit values, metadata branch derivation, loading from file, validation (missing `/`, `..`, invalid chars), local override merge
- [ ] Manual test with custom settings to verify branch names and commit subjects use configured prefixes


Soph:
Thanks for making the PR, could you help us understand more what's the use case is or what problem you are trying to solve? The issue with this approach is, that everything using the commits would need to know what the right prefix is to be used for lookup. Especially on entire.io this would be an issue and adds complexity.

moisei:
Hi @Soph , I've added this Intro to PR. Please lemme know if it is more clear now

Many organizations enforce strict policies on git branch naming conventions and commit message formats. With the current hardcoded `entire/` prefix and `Checkpoint` commit subject, Entire's branches and commits fail these checks out of the box - effectively blocking adoption in those environments. This PR addresses the issue by allowing users to configure custom branch name prefixes and commit message prefixes.

moisei:
Hi @Soph , is there anything else missing from my side to have it merged?

Soph:
@moisei sorry, we haven't made a decision yet. The issue is a bit that we need to support those trailers in all places. Inside the CLI it might be a small change but the entire.io would also need to support it - even for example if you wouldn't want to use our UI anyway because of the same compliance rules - but if we add it, someone might and then now the UI needs to read the settings of a repo to figure out which trailer is now the right one or which branch it is which adds a tone of duplication.

Soph:
Thanks again for your contribution! Our plan is to move to refs instead of branches which should help with your branch problem. For the commit trailers supporting this at this point in time put's another layer of complexity on our cli/backend architecture we don't want to invest in right now. Sorry!

PR #575, opened 2026-03-02
fix: fail-open overlap check for ACTIVE sessions in PostCommit

khaong:
## Problem

When an agent commits a file mid-turn before `SaveStep` records it in `FilesTouched`, the overlap check in `shouldCondenseWithOverlapCheck` finds no intersection between committed files and `FilesTouched`, returning `false`. This orphans the checkpoint trailer added by `PrepareCommitMsg` — metadata never gets written to `entire/checkpoints/v1`.

**Root cause:** Mismatch between `PrepareCommitMsg` (unconditionally adds trailer for ACTIVE no-TTY sessions) and `PostCommit` (requires file overlap to condense). `FilesTouched` is stale mid-turn because it's only updated via `SaveStep` (Stop hook) or `SaveTaskStep`.

Observed 3 times in a single session (`2335990a`): checkpoints `7a17709559b6`, `291f49fbcf7b`, and `720e4558e8e2` all orphaned.

## Fix

When `committedTouchedFiles` is empty for an ACTIVE session, return `true` (fail-open) instead of `false`. This matches the existing fail-open behavior when `filesTouchedBefore` is empty entirely (line 658-659).

```diff
 if len(committedTouchedFiles) == 0 {
-    return false
+    return isActive
 }
```

## Tradeoff

Stale ACTIVE sessions (agent killed without Stop hook) may now be condensed into unrelated commits as a side effect, attaching extra metadata. This is acceptable because **orphaned checkpoints (data loss) are worse than extra metadata (data duplication)**.

### Alternatives considered

| Approach | Why not |
|---|---|
| Time-based staleness (`LastInteractionTime`) | Arbitrary threshold, fragile |
| Transcript parsing at PostCommit time | Complex, user rejected |
| Make PrepareCommitMsg smarter | Can't reliably attribute commits to sessions |
| Defer trailer to PostCommit (amend) | Amending commits is risky |
| PostCommit removes orphaned trailers | Same amend issues |
| Track staged files in PrepareCommitMsg | Same attribution problem |

## Discussion points

1. **Is the stale-session false positive acceptable?** Extra metadata linked to wrong commit vs. no metadata at all.
2. **Should we explore a hybrid approach?** e.g., fail-open for ACTIVE but add a heuristic to reduce false positives later.
3. **PrepareCommitMsg/PostCommit contract:** Should we formalize the expectation that PostCommit always honors trailers from ACTIVE sessions?

## Test plan

- [x] `TestPostCommit_ActiveSession_CommitsFileNotInFilesTouched_Condenses` — new test for the bug
- [x] `TestPostCommit_StaleActiveSession_CondensedAlongside` — updated: stale session now condensed (was `_NotCondensed`)
- [x] Full test suite passes (`mise run fmt && mise run lint && mise run test:ci`)

### DIFF
--- a/cmd/entire/cli/strategy/manual_commit_hooks.go
+++ b/cmd/entire/cli/strategy/manual_commit_hooks.go
@@ -638,19 +638,20 @@ func (h *postCommitActionHandler) HandleCondenseIfFilesTouched(state *session.St
 // into this commit. Requires both that hasNew is true AND that the session's files
 // overlap with the committed files with matching content.
 //
-// This prevents stale sessions (ACTIVE sessions where the agent was killed, or
-// ENDED/IDLE sessions with carry-forward files) from being condensed into every
-// unrelated commit.
+// This prevents IDLE/ENDED sessions with carry-forward files from being condensed
+// into every unrelated commit.
+//
+// For ACTIVE sessions, we fail-open when file overlap cannot be confirmed. This is
+// because the agent may commit files mid-turn before SaveStep records them in
+// FilesTouched, causing the overlap check to see stale data. Failing closed in this
+// case orphans the checkpoint trailer (metadata never written to entire/checkpoints/v1).
+// The tradeoff: stale ACTIVE sessions (agent killed without Stop hook) may be
+// condensed into unrelated commits as a side effect. This is acceptable because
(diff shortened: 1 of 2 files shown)


khaong:
closed in favour of https://github.com/entireio/cli/pull/578

PR #581, opened 2026-03-02
feat: track and display LLM model name in session info

peyton-alt:


khaong:
bugbot run

khaong:
I think it's a good idea to keep the model name for claude and gemini as well - i.e. `opus-4.6` or `gemini-flash-3.1-preview`

I'll see if it's easy to add on or will put in a follow up

squishykid:
> I think it's a good idea to keep the model name for claude and gemini as well - i.e. `opus-4.6` or `gemini-flash-3.1-preview`
> 
> I'll see if it's easy to add on or will put in a follow up

I think we don't need to extract gemini, because the model name is included in the transcript.

peyton-alt:
> > I think it's a good idea to keep the model name for claude and gemini as well - i.e. opus-4.6 or gemini-flash-3.1-preview
> 
> I'll see if it's easy to add on or will put in a follow up
> 
> I think we don't need to extract gemini, because the model name is included in the transcript.

Same with claude code, has model in transcript on every assistant message. And looks like copilot has in transcript also. 

There isn't anywhere in UI where we see it, but we can consider if we'd want this anywhere.  Unless we still want this additional visibility in json and `entire status`, for example.

PR #582, opened 2026-03-02
test: Disable commit signing in tests

erezrokah:
Going through the contributing guide and noticed that if you have commit signing enabled (I do) the tests fail.
This PR fixes the issue by disabling it for the relevant tests.

I could also move the logic to a centralized place but that seems premature

### DIFF
--- a/cmd/entire/cli/integration_test/deferred_finalization_test.go
+++ b/cmd/entire/cli/integration_test/deferred_finalization_test.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/entireio/cli/cmd/entire/cli/paths"
 	"github.com/entireio/cli/cmd/entire/cli/session"
+	"github.com/entireio/cli/cmd/entire/cli/testutil"
 	"github.com/go-git/go-git/v5"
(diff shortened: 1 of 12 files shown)


erezrokah:
> Hey @erezrokah, thanks for raising a PR to improve the CLI!
> 
> Rather than poking at a specific config flag that breaks tests in this particular situation, i think a better approach would be to ensure Git uses a "clean" config by ignoring any user-configured flags on the system, something like this:
> 
> #513
> 
> What do you think?

Thanks for the review @toothbrush your suggestion is a much better approach. I've updated the PR.
I duplicated the existing `gitIsolatedEnv` to avoid changing too many files in this PR. I can follow up with a separate PR to use the `testutil` version, or update this PR as well if you prefer

toothbrush:
Cool, this is great, glad you liked the suggestion.  I made another change on your branch – let's just use one implementation of the isolated Git env.  I will merge this once the tests pass.

Thanks @erezrokah for your contribution!

toothbrush (approved):

PR #586, opened 2026-03-03
Fix phantom file carry-forward causing lingering shadow branches

khaong:
## Summary

- Fixes E2E failures (`TestMixedNewAndModifiedFiles`, `TestPartialStaging`) caused by phantom file paths in `filesTouched` creating infinite carry-forward loops
- Phantom files are paths captured from agent transcript tool calls that were never actually created on disk (e.g., Gemini writes `src/types.go` in a tool call but actually creates `src/types/types.go` for Go package structure)
- `filesWithRemainingAgentChanges` now checks shadow tree existence before carrying forward non-committed files — the shadow tree is the ground truth since `buildTreeWithChanges` already skips non-existent files

## Failing E2E runs

- [TestMixedNewAndModifiedFiles (gemini-cli)](https://github.com/entireio/cli/actions/runs/14091927913/job/39462990764)
- [TestPartialStaging (gemini-cli)](https://github.com/entireio/cli/actions/runs/14065073498/job/39384447651)

## Test plan

- [x] Added `TestFilesWithRemainingAgentChanges_PhantomFile` unit test
- [x] All existing tests pass (`mise run test:ci`)
- [x] Lint clean (`mise run lint`)
- [x] E2E: `TestMixedNewAndModifiedFiles` and `TestPartialStaging` should stop failing for gemini-cli

### DIFF
--- a/cmd/entire/cli/explain_test.go
+++ b/cmd/entire/cli/explain_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/entireio/cli/cmd/entire/cli/checkpoint/id"
 	"github.com/entireio/cli/cmd/entire/cli/paths"
 	"github.com/entireio/cli/cmd/entire/cli/strategy"
+	"github.com/entireio/cli/cmd/entire/cli/testutil"
 	"github.com/entireio/cli/cmd/entire/cli/trailers"
 	"github.com/entireio/cli/cmd/entire/cli/transcript"
 	"github.com/go-git/go-git/v5"
@@ -2493,6 +2494,7 @@ func TestRunExplain_SessionFlagFiltersListView(t *testing.T) {
 	} {
 		cmd := exec.CommandContext(context.Background(), "git", args...)
 		cmd.Dir = tmp
+		cmd.Env = testutil.GitIsolatedEnv()
 		if out, err := cmd.CombinedOutput(); err != nil {
 			t.Fatalf("git %v: %v\n%s", args, err, out)
 		}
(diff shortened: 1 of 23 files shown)


gtrrz-victor (approved):

cmd/entire/cli/strategy/content_overlap.go
gtrrz-victor: wouldn't be the same as moving this check on top, before `if _, wasCommitted := committedFiles[filePath]; !wasCommitted {`

khaong: Good call — you're right, both branches (committed and uncommitted) skip files absent from the shadow tree, so a single early guard at the top of the loop is equivalent and simpler. Refactored in 6d6e6ea0.

cmd/entire/cli/strategy/content_overlap.go
```diff
@@ -434,8 +434,19 @@ func filesWithRemainingAgentChanges(
...
+		// creates src/types/types.go instead). buildTreeWithChanges skips non-existent
+		// files, so the shadow tree is the ground truth. Without this check, phantom
+		// paths cause infinite carry-forward loops.
 		if _, wasCommitted := committedFiles[filePath]; !wasCommitted {
+			if _, err := shadowTree.File(filePath); err != nil {
+				logging.Debug(logCtx, "filesWithRemainingAgentChanges: file not committed and not in shadow tree, skipping",
+					slog.String("file", filePath),
+				)
+				continue
+			}
```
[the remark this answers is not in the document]
khaong: Investigated the deletion scenario — it's a non-issue in practice. When an agent deletes a file, `buildTreeWithChanges` excludes it from the shadow tree (no content on disk). Even if we carried it forward, the next `WriteTemporary` would see the file is missing and record another deletion no-op, then it'd be dropped on the following cycle anyway.

Added `TestFilesWithRemainingAgentChanges_UncommittedDeletion` in 6d6e6ea0 to document this as intentional behavior. Also simplified the check per Victor's suggestion — moved the shadow tree guard to the top of the loop since both branches (committed and uncommitted) skip files absent from the shadow tree.

khaong:
bugbot run

PR #602, opened 2026-03-04
Only consider latest Checkpoint when resuming a branch which latest commit is a squash

Soph:
The concept of `entire resume` is to remove the latest work in a branch. This means for squashed commits we also should only look at the latest checkpoint. Otherwise to much information (potentially a long list of sessions) is presented to the person that resumes but he has not always enough context to understand what the last work was and seeing multiple session doesn't help with that. 

In this example (each checkpoint was a single commit): 
```
Restored 3 sessions. To continue, run:
  claude -r 697037a7-3a77-474e-92a0-bf4f5e401632  # can you add a ruby script that returns random letters
  claude -r 90ace089-fb0e-4558-a0d8-8d908e09d030  # can you add a ruby script that returns a random number
  claude -r e15dc7ed-e938-4740-ae4a-81f4c9b3fa72  # can you make the output of the script yellow (most recent)
```
becomes now: 
```
Restored 1 sessions. To continue, run:
  claude -r e15dc7ed-e938-4740-ae4a-81f4c9b3fa72  # can you make the output of the script yellow
```

### DIFF
--- a/cmd/entire/cli/resume.go
+++ b/cmd/entire/cli/resume.go
@@ -191,7 +191,7 @@ func resumeFromCurrentBranch(ctx context.Context, branchName string, force bool)
 		return checkRemoteMetadata(ctx, repo, checkpointID)
 	}
 
-	return resumeSession(ctx, metadata.SessionID, checkpointID, force)
+	return resumeSession(ctx, metadata, force)
 }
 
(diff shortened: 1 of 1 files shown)


pfleidi (approved):
Thanks for simplifying this. I was operating under the assumption that we should try to restore all possible sessions here. If that gets too confusing, I'm all for simplifying this code!

PR #604, opened 2026-03-04
Add the ability to integrate external agents using plugins

nodo:
### Summary

This pull request introduces support for external agent binaries in the CLI, allowing agents implemented as standalone executables (following a defined protocol) to be discovered, registered, and used alongside built-in agents. The changes include a minimal reference implementation, protocol definitions, dynamic discovery and registration logic, and capability-based interface wrapping.

### For the reviewer

A few things I would like the reviewer to focus on:
- Discovery mechanism, right now is done by scanning the executable in path, which is not ideal and should probably done lazily. I think it could work for the first version given this feature is feature-flagged, but I would love feedback.
- Capabilities, build-in agents use type assertion, while external agents will have capabilities based on what they declare. I didn't want to change the build-in agents just yet, but I guess they could converge at some point.

### DIFF
--- a/cmd/entire/cli/hooks_cmd.go
+++ b/cmd/entire/cli/hooks_cmd.go
@@ -6,13 +6,18 @@ import (
 	"time"
 
 	"github.com/entireio/cli/cmd/entire/cli/agent"
-	_ "github.com/entireio/cli/cmd/entire/cli/agent/claudecode" // register agent
-	_ "github.com/entireio/cli/cmd/entire/cli/agent/cursor"     // register agent
-	"github.com/entireio/cli/cmd/entire/cli/agent/external"
-	_ "github.com/entireio/cli/cmd/entire/cli/agent/factoryaidroid" // register agent
-	_ "github.com/entireio/cli/cmd/entire/cli/agent/geminicli"      // register agent
-	_ "github.com/entireio/cli/cmd/entire/cli/agent/opencode"       // register agent
 	"github.com/entireio/cli/cmd/entire/cli/agent/types"
+
+	// Import agents to ensure they are registered before we iterate
+	_ "github.com/entireio/cli/cmd/entire/cli/agent/claudecode"
+	_ "github.com/entireio/cli/cmd/entire/cli/agent/cursor"
+	_ "github.com/entireio/cli/cmd/entire/cli/agent/factoryaidroid"
+	_ "github.com/entireio/cli/cmd/entire/cli/agent/geminicli"
+	_ "github.com/entireio/cli/cmd/entire/cli/agent/opencode"
+	_ "github.com/entireio/cli/cmd/entire/cli/agent/vogon"
+
+	// support external agents
+	"github.com/entireio/cli/cmd/entire/cli/agent/external"
 
 	"github.com/spf13/cobra"
 )
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ go.work.sum
 
 # Binary output (only in root)
 /entire
+/vogon
 /testreport
 
 # Build output directory
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -46,7 +46,22 @@ mise run test:integration
 mise run test:ci
 ```
 
-Integration tests use the `//go:build integration` build tag and are located in `cmd/entire/cli/integration_test/`.
+This runs unit tests, integration tests, and the E2E canary (Vogon agent) in sequence. Integration tests use the `//go:build integration` build tag and are located in `cmd/entire/cli/integration_test/`.
+
+### Running E2E Canary Tests (Vogon Agent)
+
+The Vogon agent is a deterministic fake agent that exercises the full E2E test suite without making any API calls. Named after the Vogons from The Hitchhiker's Guide to the Galaxy — bureaucratic, procedural, and deterministic to a fault.
+
+```bash
+mise run test:e2e:canary           # Run all E2E tests with the Vogon agent
+mise run test:e2e:canary TestFoo   # Run a specific test
+```
+
+- **Runs as part of `test:ci`** — canary failures block merges
+- **No API calls, no cost** — safe to run freely, unlike real agent E2E tests
+- **If a canary test fails, the bug is in the CLI or test infrastructure**, not in an agent
+- Located in `e2e/vogon/` (binary) and `cmd/entire/cli/agent/vogon/` (Agent interface)
+- The binary parses prompts via regex, creates/modifies/deletes files, and fires lifecycle hooks
 
 ### Running E2E Tests (Only When Explicitly Requested)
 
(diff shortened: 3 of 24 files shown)


cmd/entire/cli/hooks_cmd.go
```diff
@@ -23,6 +25,9 @@ func newHooksCmd() *cobra.Command {
 	// Git hooks are strategy-level (not agent-specific)
 	cmd.AddCommand(newHooksGitCmd())
 
+	// Discover external agent plugins from PATH before iterating
+	external.DiscoverAndRegister(context.Background())
```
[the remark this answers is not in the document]
nodo: Same as the built-in agents ?

nodo:
Just as an example, https://github.com/entireio/cli/commit/9678525ca1a9bff1e3b6a5232f2e6d3710c7132b is an example of migrating cursor to external

cmd/entire/cli/settings/settings.go
```diff
@@ -63,6 +63,10 @@ type EntireSettings struct {
 	// Defaults to "prompt" (preserves existing user behavior).
 	CommitLinking string `json:"commit_linking,omitempty"`
 
+	// ExternalAgents enables discovery and registration of external agent
+	// plugins (entire-agent-* binaries on $PATH). Defaults to false.
+	ExternalAgents bool `json:"external_agents,omitempty"`
```
pjbgf, line 68: It would be nice if we showed a warning the first time this was enabled, confirming with the user they are happy to proceed regardless of the security implications - e.g. trusting external binaries. That could also be a good time to ask them nicely to ensure they only use external agents from trusted sources. 👼

cmd/entire/cli/agent/external/capabilities.go
```diff
@@ -0,0 +1,137 @@
...
+// Wrap returns the Agent wrapped as an agent.Agent that implements ALL optional
+// interfaces (forwarding to the underlying external agent) plus CapabilityDeclarer.
+// The As* helpers in the agent package use DeclaredCapabilities() to gate access,
+// so callers only see capabilities the external binary actually declared.
+func Wrap(ea *Agent) agent.Agent {
+	return &wrappedAgent{
+		ea:   ea,
+		caps: ea.info.Capabilities,
+	}
+}
```
pjbgf: If agent is `nil` most (if not all) subsequent calls will fail. It may be worth blocking that here:

```suggestion
func Wrap(ea *Agent) (agent.Agent, error) {
	if ea == nil {
		return nil, errors.New("unable to wrap nil agent")
	}
	
	return &wrappedAgent{
		ea:   ea,
		caps: ea.info.Capabilities,
	}, nil
}
```

cmd/entire/cli/agent/external/discovery.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	seen := make(map[string]bool) // deduplicate binaries across PATH dirs
+	for _, dir := range filepath.SplitList(pathEnv) {
+		entries, err := os.ReadDir(dir)
+		if err != nil {
+			continue // skip unreadable directories
+		}
+		for _, entry := range entries {
+			name := entry.Name()
```
pjbgf: It would be more efficient to use `glob` on a per `dir` basis, as opposed to open each dir and go through the dir entries:

```golang
	dirs := strings.Split(pathEnv, ":")

	for _, dir := range dirs {
		matches, _ := filepath.Glob(filepath.Join(dir, binaryPrefix+"*"))

		for _, m := range matches {
			fmt.Println(m)
		}
	}
```

cmd/entire/cli/agent/capabilities.go
```diff
@@ -0,0 +1,117 @@
...
+	TokenCalculator        bool `json:"token_calculator"`
+	TextGenerator          bool `json:"text_generator"`
+	HookResponseWriter     bool `json:"hook_response_writer"`
+	SubagentAwareExtractor bool `json:"subagent_aware_extractor"`
+}
+
+// AsHookSupport returns the agent as HookSupport if it both implements the
+// interface and (for CapabilityDeclarer agents) has declared the capability.
+func AsHookSupport(ag Agent) (HookSupport, bool) { //nolint:ireturn // type-assertion helper must return interface
+	hs, ok := ag.(HookSupport)
```
pjbgf, line 34: We could add a `nil` check across these methods, short-circuiting to `return nil, false`.

cmd/entire/cli/agent/capabilities.go
```diff
@@ -0,0 +1,117 @@
...
+// at registration time (e.g., external plugin agents). The As* helper functions
+// below use this interface to gate capability access: an agent must both implement
+// the optional interface AND declare the capability as true.
+//
+// Built-in agents (Claude Code, Gemini CLI, etc.) do NOT implement this interface.
+// For those agents, the As* helpers fall through to a direct type assertion,
+// preserving existing behavior.
+type CapabilityDeclarer interface {
+	DeclaredCapabilities() DeclaredCaps
+}
```
pjbgf, line 13: Not sure about this `interface`. A more idiomatic Go approach would lean on composable interfaces. Like a `agent.Agent` with the absolute minimal. Then have capability specific interfaces. 

This way, when we expand we don't break users, as the current approach every time we expand on `DeclaredCaps` we are breaking our external API.

cmd/entire/cli/agent/capabilities.go
```diff
@@ -0,0 +1,117 @@
+package agent
+
+// CapabilityDeclarer is implemented by agents that declare their capabilities
+// at registration time (e.g., external plugin agents). The As* helper functions
+// below use this interface to gate capability access: an agent must both implement
+// the optional interface AND declare the capability as true.
+//
+// Built-in agents (Claude Code, Gemini CLI, etc.) do NOT implement this interface.
```
pjbgf, line 8: Potentially this pattern could be also applied to the built-in agents, as they would be the key source to see new emergent patterns/capabilities that would then need to become new interfaces. This also decreases the amount of changes needed to support a new capability - as the core logic across internal/external would be mostly shared.

nodo: That was my thinking too, I will do it in a separate PR.

cmd/entire/cli/agent/capabilities.go
nodo: I agree that would be more idiomatic, but it wouldn't work for external agents because we don't know which capabilities they implement until we try (or they tell us). Let me know if I am missing something, I have gone back and forth a lot on this point.

cmd/entire/cli/agent/capabilities.go
nodo: Good point, fixed in https://github.com/entireio/cli/pull/604/commits/7ae2cb3ac790cdc899b36c4425b23c9cabe79455

cmd/entire/cli/agent/external/capabilities.go
nodo: Makes sense! Fixed in https://github.com/entireio/cli/pull/604/commits/9198516acb0825136a16239dfc05d39a1135c76f

cmd/entire/cli/settings/settings.go
nodo: 🤔 that's a good point, I am thinking this could be a separate PR maybe? We would still need add docs etc, so maybe we can do it in that context.

cmd/entire/cli/agent/external/discovery.go
nodo: Fixed in https://github.com/entireio/cli/pull/604/changes/b91a4a36835af71050b0e27e96af2e9bc908535a

cmd/entire/cli/settings/settings.go
Soph: Agree with nodo, given this is disabled now and only way to turn it on is manually setting it I feel this is good to merge now. Need to build a better experience anyhow.

cmd/entire/cli/agent/capabilities.go
nodo: Resolving for now to merge - but let's chat about it and address it in a separate PR if needed.

PR #605, opened 2026-03-04
Setup factory ai droid env properly for interactive tests to work too

Soph:
Turn out the `--model` and `--skip-permissions-unsafe` params only work for `droid exec` not when running it in interactive mode. This made them pass locally if everything was setup correctly, but didn't work in CI.


Soph:
@alishakawaguchi beat me to it, and I was ripping my hair out :)

PR #613, opened 2026-03-05
feat(cursor): capture session metrics (duration, turns, model, attrib…

peyton-alt:
…ution) via hooks


peyton-alt:
@cursor review

PR #618, opened 2026-03-05
fix: filter out test sessions from state store

peyton-alt:


peyton-alt:
@cursor review

Soph:
@peyton-alt I know I brought this up yesterday but I also now let it analyze the code base and it should not happen. We have all tests isolated and nothing should write in the regular session folder for the cli repo anymore. So I feel this is to edge case to be merged.

khaong:
I believe the root cause is now fixed in #625 , and I did a cleanup in #627 . I'm going to close this, thanks!

PR #623, opened 2026-03-05
Add hk to detected hook managers

TyceHerrman:
Detect hk (jdx's git hook manager) config files during `entire enable`. Checks hk.pkl, .config/hk.pkl, hk.local.pkl, and .config/hk.local.pkl with precedence-ordered dedup matching hk's own discovery order.

Entire-Checkpoint: e75e9ecebc00


Soph:
Thanks for your effort, sorry merged #623 today which implements the same!

PR #629, opened 2026-03-06
fix: propagate PATH into tmux sessions for E2E interactive tests

khaong:
## Summary

- **Fix tmux PATH propagation**: Interactive E2E tests (using `StartSession` / tmux) were failing because the tmux session inherited the tmux server's stale `PATH` instead of the test runner's. The freshly built `entire` binary wasn't found, so agent lifecycle hooks (session-start, user-prompt-submit, stop) silently failed, leaving `prepare-commit-msg` with "no active sessions".
- **Rename interactive tests**: Prefix the 3 interactive attribution tests with `Interactive` to match the existing convention, making it clear from test names which tests use interactive mode.

## Fix

`NewTmuxSession` now wraps the command with `env PATH=...` to explicitly propagate the current process's PATH into the tmux session, ensuring all binaries (including freshly built ones) are resolvable.

## Test plan

- [x] All 5 previously-failing interactive tests now pass
- [x] Full canary suite: 43/43 passing
- [x] `mise run fmt && mise run lint` clean

### DIFF
--- a/e2e/tests/attribution_test.go
+++ b/e2e/tests/attribution_test.go
@@ -44,10 +44,10 @@ func TestLineAttributionReasonable(t *testing.T) {
 	})
 }
 
-// TestAttributionOnAgentCommit: agent creates a file and commits it in the
+// TestInteractiveAttributionOnAgentCommit: agent creates a file and commits it in the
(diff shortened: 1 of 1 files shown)


khaong:
bugbot run

PR #652, opened 2026-03-07
Add `entire trace` command to find slow performance hooks etc.

pfleidi:
## `entire trace` command

Adds a new `entire trace` CLI command that displays hook performance traces, and extends the `perf/span` package with loop iteration tracking.

### New command: `entire trace`

Parses structured perf log entries from `.entire/logs/entire.log` and renders them as formatted timing tables with hierarchical sub-step breakdowns.

- `entire trace` — show the most recent hook trace
- `entire trace --last 5` — show the last 5 traces
- `entire trace --hook post-commit` — filter by hook type

Traces require `ENTIRE_LOG_LEVEL=DEBUG` to be enabled.

**Implementation** (`cmd/entire/cli/trace.go`, `trace_cmd.go`):

- JSONL log parser that extracts `steps.*_ms` and `steps.*_err` keys into a hierarchical step tree
- Sub-step detection: keys like `foo.0`, `foo.1` are nested under their parent `foo` when the parent exists
- ASCII tree rendering (`├─`/`└─`) with manual padding to avoid UTF-8 multi-byte alignment issues
- Input validation on `--last` flag (must be >= 1)

Example output:

```
❯ entire trace --hook post-commit --last 2
post-commit  123ms  2026-03-06T16:35:48-08:00

  STEP                        DURATION
  cleanup_shadow_branches         12ms
  find_sessions_for_worktree       0ms
  open_repository_and_head         0ms
  process_sessions               100ms
    ├─ process_sessions.0          3ms
    ├─ process_sessions.1          0ms
    ├─ process_sessions.2         95ms
    └─ process_sessions.3          0ms
  resolve_commit_trees             5ms
```

### `perf/span`: LoopSpan and child deduplication

Extends the span package to support tracking loop iterations as grouped sub-spans.

- **`LoopSpan`** — wraps a grouping span; `Iteration()` creates child spans, `End()` auto-ends unfinished iterations
- **`childStepKey()`** — deduplicates child span names (`.1`, `.2` suffixes) so repeated loop iterations don't overwrite each other in log output
- **Grandchild emission** — group spans emit their children as `steps.<name>.<index>_ms` entries with 0-based indexing

### Hook integration

`PostCommit` in `manual_commit_hooks.go` now wraps the session processing loop with `StartLoop`/`Iteration`/`End` to emit per-session timing in traces.

### DIFF
--- a/perf/span.go
+++ b/perf/span.go
@@ -83,7 +83,8 @@ func (s *Span) End() {
 	// to prevent later values from overwriting earlier ones in JSON output.
 	//
 	// Group spans (children that have their own children) also emit grandchildren
-	// with 0-based indexing: steps.<name>.0_ms, steps.<name>.1_ms, etc.
+	// with 0-based indexing: steps.<name>.0_ms, steps.<name>.1_ms, etc. The ~N
+	// suffix avoids collisions with this .0, .1, ... iteration indexing.
 	seen := make(map[string]int, len(s.children))
 	for _, child := range s.children {
 		// Auto-end children that were not explicitly ended


cmd/entire/cli/strategy/common_test.go
```diff
@@ -1432,8 +1432,8 @@ func TestReadLatestSessionPromptFromCommittedTree(t *testing.T) {
 		// Session 1 (latest) has no prompt.txt, session 0 does.
 		// This happens when a test session gets condensed alongside a real one.
 		tree := buildCommittedTree(t, map[string]string{
-			"a3/b2c4d5e6f7/0/prompt.txt":      "Real session prompt",
-			"a3/b2c4d5e6f7/1/metadata.json":    `{"session_id":"test"}`,
+			"a3/b2c4d5e6f7/0/prompt.txt":    "Real session prompt",
```
pfleidi, line 1437: These kept coming up for some reason. 🤷

cmd/entire/cli/strategy/manual_commit_hooks.go
```diff
@@ -857,16 +857,20 @@ func (s *ManualCommitStrategy) PostCommit(ctx context.Context) error { //nolint:
...
 
+	loopCtx, processSessionsLoop := perf.StartLoop(ctx, "process_sessions")
 	for _, state := range sessions {
 		// Skip fully-condensed ended sessions — no work remains.
 		// These sessions only persist for LastCheckpointID (amend trailer reuse).
 		if state.FullyCondensed && state.Phase == session.PhaseEnded {
 			continue
 		}
-		s.postCommitProcessSession(ctx, repo, state, &transitionCtx, checkpointID,
+		iterCtx, iterSpan := processSessionsLoop.Iteration(loopCtx)
```
pfleidi, line 867: Tracking cumulative timings for looped steps didn't really work with the existing package. This adds a new abstraction for loops and other code that calls the the same operation multiple times. It will group it hierarchically under the `processSessionsLoop` span and render it as a tree in the `entire perf` command.

pfleidi:
Bugbot run

cmd/entire/cli/strategy/common_test.go
pjbgf: We could add `golangci-lint fmt` in mise to ensure we keep it to a specific format.

cmd/entire/cli/trace.go
```diff
@@ -0,0 +1,313 @@
...
+		if errors.Is(err, os.ErrNotExist) {
+			return nil, nil
+		}
+		return nil, fmt.Errorf("opening perf log: %w", err)
+	}
+	defer f.Close()
+
+	var entries []perfEntry
+
+	scanner := bufio.NewScanner(f)
```
khaong, line 207: **Must fix: `bufio.Scanner` default 64KB buffer limit.**

The default scanner fails with `bufio.ErrTooLong` if any log line exceeds 64KB. Since `entire.log` is a shared file containing *all* log output (not just perf entries), a single large non-perf line would kill the entire `entire perf` command.

Other files in the codebase already handle this — e.g. `copilotcli/transcript.go` and `testreport/main.go` use `scanner.Buffer(...)`. One-line fix:

```go
scanner := bufio.NewScanner(f)
scanner.Buffer(make([]byte, 0, bufio.MaxScanTokenSize), 1024*1024)
```

perf/span.go
```diff
@@ -74,18 +74,38 @@ func (s *Span) End() {
...
 	}
 
-	// Add child step durations (and error flags) as flat keys
+	// Add child step durations (and error flags) as flat keys.
+	// Disambiguate duplicate child names (from loops) with .1, .2, etc. suffixes
+	// to prevent later values from overwriting earlier ones in JSON output.
+	//
+	// Group spans (children that have their own children) also emit grandchildren
+	// with 0-based indexing: steps.<name>.0_ms, steps.<name>.1_ms, etc.
+	seen := make(map[string]int, len(s.children))
```
khaong, line 88: **Should fix: `childStepKey` dedup naming collides with grandchild indexing.**

`childStepKey` disambiguates duplicate children as `name`, `name.1`, `name.2`. Grandchild (loop iteration) keys are emitted as `name.0`, `name.1`, `name.2` (lines 97-108). If two children share a name and the first has grandchildren, `steps.foo.1_ms` is ambiguous — it could be grandchild index 1 of the first child OR the second child's own duration.

Doesn't trigger today (the one `LoopSpan` usage has a unique name), but it's a latent collision. Now that `LoopSpan` exists specifically to handle the loop case, consider whether `childStepKey` dedup is still needed — or use a different separator (e.g. `~1`) to avoid the collision.

perf/span.go
```diff
@@ -95,3 +115,51 @@ func (s *Span) End() {
...
+// Iteration creates a child span for a single loop iteration. The caller must
+// call End() on the returned span when the iteration completes.
+func (l *LoopSpan) Iteration(ctx context.Context) (context.Context, *Span) {
+	return Start(ctx, l.span.name)
+}
+
+// End completes the loop span by ending the underlying span.
+// Note: this does not itself iterate over and end unended iteration spans;
+// any automatic cleanup of child spans occurs when the root span is ended.
+func (l *LoopSpan) End() {
```
khaong, line 170: **Should fix: `LoopSpan.End()` doesn't auto-end iteration children.**

This delegates to `l.span.End()`, which returns early for child spans (line 63-64) without iterating children. Unended iteration spans are only cleaned up later by the root span's `End()`, at which point their duration is `time.Since(start)` — potentially wrong if the root ends much later than the loop.

Current usage in `manual_commit_hooks.go` is safe (calls `iterSpan.End()` explicitly each iteration), but future callers may reasonably expect `LoopSpan.End()` to clean up its children, matching how root `End()` behaves.

Suggested fix:
```go
func (l *LoopSpan) End() {
	for _, child := range l.span.children {
		if !child.ended {
			child.End()
		}
	}
	l.span.End()
}
```

cmd/entire/cli/trace.go
```diff
@@ -0,0 +1,313 @@
...
+	if msgRaw, ok := raw["msg"]; !ok {
+		return nil
+	} else if err := json.Unmarshal(msgRaw, &msg); err != nil || msg != "perf" {
+		return nil
+	}
+
+	entry := &perfEntry{}
+
+	// Extract op
+	if opRaw, ok := raw["op"]; ok {
```
khaong, line 62: **Should fix: Confirmed perf entries silently discarded on field parse failure.**

After passing the `msg == "perf"` gate (line 44-48), a type mismatch on `op`, `duration_ms`, or `error` returns `nil` — the entire entry is dropped. The user sees "No perf entries found. Set ENTIRE_LOG_LEVEL=DEBUG..." when entries exist but have a schema issue.

Since these fields are already optional (missing is fine — the entry is kept with zero values), a parse *failure* should be treated the same way. Use best-effort extraction after the `msg` gate:

```go
if opRaw, ok := raw["op"]; ok {
	_ = json.Unmarshal(opRaw, &entry.Op) // best-effort; Op stays "" on failure
}
```

Same pattern for `duration_ms` and `error`.

cmd/entire/cli/trace.go
pfleidi: I just addressed this. 👍

PR #654, opened 2026-03-07
Fix cursor-cli E2E flakiness

alishakawaguchi:
## Summary

Two targeted fixes discovered during E2E triage:

- **Cursor CLI E2E: isolate config dir and fix wait pattern** — Parallel cursor-cli E2E tests raced on the shared `~/.config/cursor/cli-config.json` file, causing `ENOENT` errors. Each session now gets an isolated `XDG_CONFIG_HOME` with a pre-seeded `cli-config.json`. Also switches `WaitFor` from the `/ commands` prompt pattern (always visible in the status bar, even during "Thinking") to the `Add a follow-up` text that only appears after the agent finishes.


alishakawaguchi:
@cursor review

PR #656, opened 2026-03-07
Fix: Disable agent hooks when Entire is not active

pfleidi:
### Problem

When Entire was disabled via `.entire/settings.json` (`"enabled": false`) or never set up in a repository, **agent hooks** (Claude Code, Gemini CLI, etc.) still initialized logging on every invocation. Git hooks already had an early-exit guard via `settings.IsSetUpAndEnabled()` + a `gitHooksDisabled` flag, but agent hooks were missing an equivalent check in their `PersistentPreRunE`.

Additionally, `IsEnabled()` (used by `executeAgentHook`) returns `true` by default when no settings file exists — meaning agent hooks could proceed past the enabled check in repos where Entire was never set up.

### Changes

**`cmd/entire/cli/hook_registry.go`**

Added a `settings.IsSetUpAndEnabled(ctx)` guard in the `PersistentPreRunE` of `newAgentHooksCmd`. When Entire is not set up or is disabled, the function returns `nil` early — skipping `initHookLogging()` and avoiding unnecessary file I/O and log initialization. The child command's `RunE` (`executeAgentHook`) still runs but bails out via its own `IsEnabled()` check.

### DIFF
--- a/cmd/entire/cli/strategy/common_test.go
+++ b/cmd/entire/cli/strategy/common_test.go
@@ -1432,8 +1432,8 @@ func TestReadLatestSessionPromptFromCommittedTree(t *testing.T) {
 		// Session 1 (latest) has no prompt.txt, session 0 does.
 		// This happens when a test session gets condensed alongside a real one.
 		tree := buildCommittedTree(t, map[string]string{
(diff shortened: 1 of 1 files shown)


pfleidi:
Closing this for now.

PR #660, opened 2026-03-08
fix(agent): scan full transcript for spawned agent IDs

### LINKED ISSUE #329: ExtractSpawnedAgentIDs misses subagents spawned before startLine offset
## Problem

`ExtractAllModifiedFiles` and `CalculateTotalTokenUsage` both extract spawned agent IDs from the transcript slice starting at `startLine`. When `startLine > 0` (i.e., after a checkpoint), subagents spawned in earlier turns are not discovered.

If those earlier-spawned subagents continue writing to their transcript files in later turns, their file modifications will be missed.

## Impact

After a checkpoint is saved (advancing `CheckpointTranscriptStart`), modifications by previously-spawned subagents become invisible to both file extraction and token usage calculation.

## Suggested Fix

Extract agent IDs from the full transcript (`startLine=0`) while still limiting main-agent modified file extraction to `startLine`. This ensures all subagent transcripts are scanned regardless of when the agent was spawned.

## Context

Identified during review of #323 (ENT-297 fix). This is a pre-existing limitation shared by `CalculateTotalTokenUsage`.

mvanhorn:
## Summary

Closes #329

Fixes a bug where subagent token usage and modified files were undercounted when a checkpoint boundary (`startLine`) was set.

### The Problem

`CalculateTotalTokenUsage` and `ExtractAllModifiedFiles` both:
1. Slice the transcript from `startLine` to get the current turn's data
2. Extract spawned agent IDs from that sliced data
3. Read subagent transcript files and sum their usage/files

The issue: subagents spawned **before** `startLine` (in a previous turn) may still be actively writing to their transcript files during the current turn. By only scanning from `startLine`, those agent IDs were never discovered, so their token usage and file modifications were silently dropped.

### The Fix

Agent ID extraction now scans the **full transcript** from the beginning (line 0), while file/token extraction still respects the `startLine` boundary. This ensures all subagents are discovered regardless of when they were spawned.

Added `extractAllSpawnedAgentIDs()` helper that parses the full transcript bytes independently of the startLine-sliced data.

## Changes

**Claude Code agent (`claudecode/transcript.go`):**
- `CalculateTotalTokenUsage` - Uses `extractAllSpawnedAgentIDs(transcriptData)` instead of `ExtractSpawnedAgentIDs(parsed)` 
- `ExtractAllModifiedFiles` - Same fix
- Added `extractAllSpawnedAgentIDs()` helper

**Factory AI Droid agent (`factoryaidroid/transcript.go`):**
- `CalculateTotalTokenUsageFromBytes` / `ExtractAllModifiedFilesFromBytes` - Added `extractAllSpawnedAgentIDsFromBytes()` helper with same full-scan approach
- `CalculateTotalTokenUsageFromTranscript` / `ExtractAllModifiedFilesFromTranscript` - Re-parses full transcript file at line 0 for agent ID discovery

## Testing

- All existing transcript tests continue to pass
- The fix is conservative - it only broadens agent ID discovery, not file/token extraction boundaries

---

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/cmd/entire/cli/agent/claudecode/transcript.go
+++ b/cmd/entire/cli/agent/claudecode/transcript.go
@@ -264,6 +264,17 @@ func ExtractSpawnedAgentIDs(transcript []TranscriptLine) map[string]string {
 	return agentIDs
 }
 
+// extractAllSpawnedAgentIDs parses the full transcript from the beginning to find
+// all spawned agent IDs, regardless of startLine offset. This ensures subagents
+// spawned before a checkpoint boundary are still discovered.
+func extractAllSpawnedAgentIDs(transcriptData []byte) map[string]string {
+	fullParsed, err := transcript.ParseFromBytes(transcriptData)
+	if err != nil {
+		return nil
+	}
+	return ExtractSpawnedAgentIDs(fullParsed)
+}
+
 // extractAgentIDFromText extracts an agent ID from text containing "agentId: <id>".
 func extractAgentIDFromText(text string) string {
 	const prefix = "agentId: "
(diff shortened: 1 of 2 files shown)


suhaanthayyil:
Thanks for this, @mvanhorn — and for the precise write-up of the bug (extracting spawned agent IDs from the `startLine`-sliced window instead of the full transcript, so subagents spawned before the checkpoint boundary were undercounted).

#329 has since been resolved by #1710, which takes the same approach you identified: scanning the full transcript for spawned agent IDs before summing subagent token usage and modified files. Closing this as superseded — but genuinely appreciate you diagnosing it so clearly and sending a fix. 🙏

PR #665, opened 2026-03-09
fix: resolve repo root for initialization when started from subdirectory

### LINKED ISSUE #559: Entire not initialized when coding agent is started from subdirectory
### What happened?

Entire initializes correctly when the coding agent is started from the root directory of a project. However, when the coding agent is initiated from within a subdirectory of the same project, Entire is not detected and does not initialize (e.g., hooks are not added to the coding agent session).

### Steps to reproduce

1. Navigate to the root of a project that has Entire configured.
2. Start the coding agent → ✅ Entire initializes correctly (hooks are added).

Image

3. Navigate to a subdirectory inside the same project.
4. Start the coding agent from that subdirectory → ❌ Entire does not initialize (hooks are missing).

Image

### Entire CLI version

Entire CLI 0.4.8 (81ddee25) Go version: go1.26.0 OS/Arch: darwin/arm64

### OS and architecture

Darwin 25.3.0 arm64

### Agent

gemini

### Terminal

zsh

### Logs / debug output

```bash

```

### Additional context

_No response_

mvanhorn:
## Summary
Fixes #559

When `entire enable` is run from a subdirectory (e.g., `cd repo/frontend && entire enable`), the agent hook config files (`.gemini/settings.json`, `.claude/settings.json`) must be created at the repo root for agents to find them. While internal path resolution via `paths.AbsPath()` already uses `git rev-parse --show-toplevel`, some agent config lookups are CWD-relative. This fix explicitly changes to the repo root before the enable flow begins.

## Changes
- Detect when `entire enable` is run from a subdirectory (CWD != repo root)
- Change to the repo root before proceeding with setup
- Clear the worktree root cache after changing directory
- Print a note so the user knows the repo root is being used

## Testing
- Run `entire enable` from a subdirectory within a git repo and verify it succeeds
- Verify `.entire/settings.json` and agent config files are created at the repo root, not the subdirectory
- Run `entire enable` from the repo root and verify no extra note is printed

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/cmd/entire/cli/setup.go
+++ b/cmd/entire/cli/setup.go
@@ -388,15 +388,11 @@ If Entire is already configured but disabled, this re-enables it.`,
 				return NewSilentError(errors.New("not a git repository"))
 			}
 
-			// Change to repo root so all paths resolve correctly.
-			// This ensures agent hooks (e.g., .gemini/settings.json, .claude/settings.json)
-			// are created at the repo root even when running from a subdirectory.
+			// Notify the user when running from a subdirectory.
+			// All path resolution uses paths.WorktreeRoot(ctx) which returns the
+			// correct repo root regardless of CWD, so no directory change is needed.
 			cwd, err := os.Getwd() //nolint:forbidigo // Need CWD to detect subdirectory
 			if err == nil && cwd != repoRoot {
-				if err := os.Chdir(repoRoot); err != nil {
-					return fmt.Errorf("failed to change to repository root %s: %w", repoRoot, err)
-				}
-				paths.ClearWorktreeRootCache()
 				fmt.Fprintf(cmd.ErrOrStderr(), "Note: Running from repository root (%s)\n\n", repoRoot)
 			}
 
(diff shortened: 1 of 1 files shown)


Soph (dismissed):

cmd/entire/cli/setup.go
```diff
@@ -382,11 +382,24 @@ If Entire is already configured but disabled, this re-enables it.`,
...
 				fmt.Fprintln(cmd.ErrOrStderr(), "Not a git repository. Please run 'entire enable' from within a git repository.")
 				return NewSilentError(errors.New("not a git repository"))
 			}
 
+			// Change to repo root so all paths resolve correctly.
+			// This ensures agent hooks (e.g., .gemini/settings.json, .claude/settings.json)
+			// are created at the repo root even when running from a subdirectory.
+			cwd, err := os.Getwd() //nolint:forbidigo // Need CWD to detect subdirectory
+			if err == nil && cwd != repoRoot {
+				if err := os.Chdir(repoRoot); err != nil {
```
pjbgf: The use of `os.Chdir` can come with side effects, as we are changing the current path for the running process. Instead, we could continue to identify the correct Worktree root path (`L385`), open an `os.Root` for it and then pass that on to any funcs that need to act within that location.

mvanhorn: Good call. Removed the `os.Chdir` - all downstream functions resolve paths through `paths.WorktreeRoot(ctx)` which calls `git rev-parse --show-toplevel`, so the Chdir was unnecessary global state mutation. Pushed the fix.

Soph:
Sorry, this got lost somehow and looking at it, it only looks an error now and I don't think it's worth picking that up now and adapt to all the change son setup.go.

PR #671, opened 2026-03-10
fix: filter system-reminder messages in OpenCode transcript parser

### LINKED ISSUE #564: [OpenCode] <system-reminder> messages from oh-my-opencode is tracked as user message instead of opencode
### What happened?

Image

### Steps to reproduce

using opencode with oh-my-opencode which added the feature of `<system-reminder>` (for TODO continuation) when it involves multi-agent task orchestration

This might be tricky as those messages are sent with `user` role instead of `assistant`, and IIUC, they currently use `x-initiator: agent` and some internal message marker to help identify that as internal message, but those info may not be exposed to plugin API yet 

### Entire CLI version

Entire CLI 0.4.8 (81ddee25)

### OS and architecture

OS/Arch: darwin/arm64

### Agent

opencode

### Terminal

Ghostty

### Logs / debug output

```bash

```

### Additional context

_No response_

mvanhorn:
## Summary

Fixes #564

Filters out `<system-reminder>` messages that are injected by tools like oh-my-opencode with `user` role. These messages were incorrectly counted as user prompts in attribution and checkpoint data.

## Changes

- `cmd/entire/cli/agent/opencode/transcript.go` - Added `isSystemReminderOnly()` and `stripSystemReminders()` functions; updated `ExtractAllUserPrompts()` to skip or strip system-reminder content
- `cmd/entire/cli/agent/opencode/transcript_test.go` - Added 6 test functions covering pure system-reminder messages, mixed content, and whitespace handling

## Approach

Content-based heuristic: messages entirely wrapped in `<system-reminder>...</system-reminder>` tags are skipped. Mixed messages have the reminder tags stripped. This is the best approach given that OpenCode doesn't expose `x-initiator` metadata.

## Test plan

- [x] System-reminder-only messages excluded from prompts
- [x] Mixed messages have reminder sections stripped
- [x] Normal user messages unaffected
- [x] `gofmt` and `go vet` clean

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/cmd/entire/cli/agent/opencode/transcript.go
+++ b/cmd/entire/cli/agent/opencode/transcript.go
@@ -209,20 +209,22 @@ func ExtractTextFromParts(parts []Part) string {
 	return strings.Join(texts, "\n")
 }
 
-// systemReminderOpen is the opening tag for system-reminder messages injected by
-// oh-my-opencode and similar orchestration tools. These arrive with role "user"
-// but are not actual user prompts.
-const systemReminderOpen = "<system-reminder>"
+// Tags used by oh-my-opencode and similar orchestration tools to inject
+// context into the conversation with role "user" — not actual user prompts.
+const (
+	systemReminderOpen  = "<system-reminder>"
+	systemReminderClose = "</system-reminder>"
+)
 
-// systemReminderClose is the corresponding closing tag.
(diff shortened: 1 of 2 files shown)


mvanhorn:
Thanks for the review!

PR #678, opened 2026-03-10
Improve initial flow and enabling/disabling

Soph:
This changes a few things with entire: 

- running `entire`, `entire enable` or `entire configure` in a repo that has no `.entire` folder yet, will bring you to the setup wizard
- running `entire enable` on a repo that is setup earlier already it only tells you that entire is enabled
- running `entire disable` then `entire enable` will disable -> enable entire now
- running `entire configure` let's you add more agents later but not remove them, for that you need to use `entire configure --remove <agent>` to actually remove it 

Happy to hear thoughts.

### DIFF
--- a/cmd/entire/cli/settings/settings.go
+++ b/cmd/entire/cli/settings/settings.go
@@ -286,7 +286,7 @@ func IsSetUpAny(ctx context.Context) bool {
 	if err != nil {
 		return false
 	}
-	_, err = os.Stat(localFileAbs)
+	_, err = os.Lstat(localFileAbs)
 	return err == nil
 }
 


pjbgf (approved):

cmd/entire/cli/settings/settings.go
```diff
@@ -275,6 +275,21 @@ func IsSetUp(ctx context.Context) bool {
...
+// Use this to detect any prior setup, even if only local settings exist.
+func IsSetUpAny(ctx context.Context) bool {
+	if IsSetUp(ctx) {
+		return true
+	}
+	localFileAbs, err := paths.AbsPath(ctx, EntireSettingsLocalFile)
+	if err != nil {
+		return false
+	}
+	_, err = os.Stat(localFileAbs)
```
pjbgf: Assuming we don't want to follow a `settings.json` which is a symlink:
```suggestion
	_, err = os.Lstat(localFileAbs)
```

Soph: Hmm. good question if we would want or not 🤔

pjbgf: The safest approach would be to only support it to be a regular file. Symlinks could be leveraged to display contents of sensitive files across the disk, or corrupt such file.

We are better off not supporting symlinks for settings until we have a legit feature that requires it.

PR #682, opened 2026-03-11
Fix test failures from GPG signing in CI environments

nodo:
Tests in strategy/common_test.go, strategy/metadata_reconcile_test.go,
and trail/store_test.go were failing because they created git repos
without disabling GPG commit signing. Added commit.gpgsign=false config
and --no-gpg-sign flags to match the pattern used by testutil.InitRepo.

https://claude.ai/code/session_01Prt6rAFudswZp7AmUEebxa


toothbrush:
I'm probably misunderstanding something, but my understanding was that `GitIsolatedEnv` was supposed to take care of making the git config "vanilla".  Is that not working here?


https://github.com/entireio/cli/blob/9375083dc197cf9583e13301add95e16518b5be0/cmd/entire/cli/testutil/testutil.go#L294-L306

PR #692, opened 2026-03-13
feat: add CLI device auth login flow

dipree:
## Summary
- add CLI login support for Entire using the device authorization flow
- open the browser approval URL, poll for completion, and persist the returned auth token locally
- cover the flow with integration tests plus base URL and client protocol tests

## Testing
- go test ./cmd/entire/cli/...
- go test -tags=integration ./cmd/entire/cli/integration_test

### DIFF
--- a/cmd/entire/cli/api/base_url.go
+++ b/cmd/entire/cli/api/base_url.go
@@ -9,7 +9,7 @@ import (
 )
 
 // ErrInsecureHTTP is returned when the base URL uses HTTP without an explicit opt-in.
-var ErrInsecureHTTP = errors.New("refusing to use insecure http:// base URL for authentication (set ENTIRE_API_BASE_URL to override)")
+var ErrInsecureHTTP = errors.New("refusing to use insecure http:// base URL for authentication (use --insecure-http-auth to override)")
 
 const (
 	// DefaultBaseURL is the production Entire API origin.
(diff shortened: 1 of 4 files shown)


cmd/entire/cli/auth/store.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// NewStoreWithService returns a Store with a custom keyring service name (for testing).
+func NewStoreWithService(service string) *Store {
+	return &Store{service: service}
+}
+
+// SaveToken persists an access token for the given base URL.
+func (s *Store) SaveToken(baseURL, token string) error {
+	if strings.TrimSpace(token) == "" {
```
pjbgf: We trim spaces from the token to check whether it is empty, but we save the untrimmed version in the keyring. It may be worth just updating the token instead:

```suggestion
	token = strings.TrimSpace(token)
	if token == "" {
```

That way we don't store tokens with leading/trailing spaces.

cmd/entire/cli/api/base_url.go
```diff
@@ -0,0 +1,55 @@
...
+
+// ResolveURLFromBase joins an API-relative path against an explicit base URL.
+// Only http and https schemes are accepted.
+func ResolveURLFromBase(baseURL, path string) (string, error) {
+	base, err := url.Parse(baseURL)
+	if err != nil {
+		return "", fmt.Errorf("parse base URL: %w", err)
+	}
+
+	if base.Scheme != "http" && base.Scheme != "https" {
```
pjbgf, line 45: Given how sensitive the data transmitted is, ideally the `http` scheme would be blocked by an explicit opt-in flag.

A good practice is prefixing it with `insecure` or `unsafe` so that users think twice before using it: i.e. `--insecure-http-auth`.

cmd/entire/cli/auth/store.go
Soph: added!

cmd/entire/cli/api/base_url.go
Soph: Good point. Changed it so you would need to use `--insecure-http-auth`.

dipree:
Removed the `--insecure-http-auth` flag. HTTP is now allowed automatically when `ENTIRE_API_BASE_URL` is set — setting the env var already implies a dev/testing context, so the flag was redundant and leaked a dev concern into the release CLI.

dipree:
Reverted the env-var approach — good call that `ENTIRE_API_BASE_URL` can be injected. Instead: `--insecure-http-auth` is kept as an explicit opt-in but hidden from `entire login --help` via `MarkHidden`. The error message still advertises the flag when you hit the HTTP check.

PR #699, opened 2026-03-16
go 1.26.1 and golangci-lint 2.11.3

Soph:
Bump to go 1.26.1 and golangci-lint 2.11.3 and fixing / silencing some of the new things.

### DIFF
--- a/cmd/entire/cli/utils.go
+++ b/cmd/entire/cli/utils.go
@@ -58,7 +58,7 @@ func copyFile(src, dst string) error {
 		return err
 	}
 
-	input, err := os.ReadFile(src) //nolint:gosec // src validated by callers from checkpoint storage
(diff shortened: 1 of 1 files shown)


cmd/entire/cli/utils.go
```diff
@@ -38,14 +43,75 @@ func fileExists(path string) bool {
 	return err == nil
 }
 
-// copyFile copies a file from src to dst
+// copyFile copies a file from src to dst.
+// Both paths must be absolute after cleaning; dst must reside under either the
+// repo worktree root, the git common dir, or the user's home directory (for
```
pjbgf, line 48: These types of checks are better enforced via [traversal-resistant primitives](https://go.dev/blog/osroot), but we can do that as a follow-up change.

PR #700, opened 2026-03-16
Add e2e tests for external agents

nodo:
Add some simple tests to our e2e suite using our test external agent https://github.com/entireio/roger-roger

### DIFF
--- a/.github/workflows/e2e-isolated.yml
+++ b/.github/workflows/e2e-isolated.yml
@@ -42,9 +42,7 @@ jobs:
             factoryai-droid) curl -fsSL https://app.factory.ai/cli | sh ;;
             copilot-cli)  npm install -g @github/copilot  ;;
             roger-roger)
-              git clone --depth 1 -q https://github.com/entireio/roger-roger.git /tmp/roger-roger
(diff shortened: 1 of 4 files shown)


e2e/testutil/repo.go
```diff
@@ -66,6 +66,19 @@ func SetupRepo(t *testing.T, agent agents.Agent) *RepoState {
...
+	if ea, ok := agent.(agents.ExternalAgent); ok && ea.IsExternalAgent() {
+		entireDir := filepath.Join(dir, ".entire")
+		if err := os.MkdirAll(entireDir, 0o755); err != nil {
+			t.Fatalf("create .entire for external agent: %v", err)
+		}
+		if err := os.WriteFile(filepath.Join(entireDir, "settings.json"),
+			[]byte("{\"external_agents\": true}\n"), 0o644); err != nil {
+			t.Fatalf("write external_agents setting: %v", err)
+		}
+	}
```
Soph, line 80: I wonder if it's worth adding a flag to `entire enable` and use it here instead of doing this. Since that will break once we ever change `external_agents` but on the otherside we will figure out because the tests will fail ¯\_(ツ)_/¯

nodo: yeah not super robust but also I felt probably ok for now?

nodo:
> Should we also add it to e2e:canary ?

I am going back and forth with this, let's discuss separately.

PR #714, opened 2026-03-18
fix(gemini): strip non-array values from hooks to fix Gemini CLI 0.33+

peyton-alt:
validation

Entire-Checkpoint: 332b287f14ea


peyton-alt:
@bugbot review

PR #716, opened 2026-03-18
External Agent: discover external agent plugins before hook execution

keyu98:
Ensure external agent plugins are discovered and registered during the git hook pre-run phase, so GetByAgentType resolves correctly during condensation (e.g. post-commit). Without this, token usage and other agent-specific data would be missing from metadata.json.


keyu98:
@nodo When integrating our agent, we found that even after implementing the token_calculator capability, token usage was still not being written to metadata.json. We later tracked down the root cause — would appreciate it if you could review this when you have a chance, nodo. Thanks!

PR #729, opened 2026-03-19
External Agent: Add Windows-compatible agent name derivation

keyu98:
## Summary

External agent discovery currently fails to correctly derive agent names on Windows because:

1. Windows executables have file extensions (`.exe`, `.bat`, `.cmd`) that get included in the agent name.
2. The Unix executable-bit check (`mode & 0o111`) rejects all binaries on Windows, since Windows doesn't set POSIX execute permission bits.

## Changes

- **Strip Windows executable extensions** before deriving the agent name via a new `stripExeExt()` helper. This removes `.exe`, `.bat`, and `.cmd` suffixes so that `entire-agent-foo.exe` correctly resolves to agent name `foo`. On Unix this is a no-op.
- **Skip the executable-bit check on Windows** (`runtime.GOOS != "windows"`), since Windows filesystems do not expose POSIX permission bits.

## Files changed

- `cmd/entire/cli/agent/external/discovery.go`

### DIFF
--- a/cmd/entire/cli/agent/external/discovery.go
+++ b/cmd/entire/cli/agent/external/discovery.go
@@ -14,7 +14,10 @@ import (
 	"github.com/entireio/cli/cmd/entire/cli/settings"
 )
 
-const binaryPrefix = "entire-agent-"
+const (
+	binaryPrefix = "entire-agent-"
+	osWindows    = "windows"
+)
 
 // DiscoverAndRegister scans $PATH for executables matching "entire-agent-<name>",
 // calls their "info" subcommand, and registers them in the agent registry.
@@ -66,7 +69,7 @@ func DiscoverAndRegister(ctx context.Context) {
 				continue
 			}
 			// Check executable bit (on Unix; Windows doesn't set execute bits)
-			if runtime.GOOS != "windows" && finfo.Mode()&0o111 == 0 {
+			if runtime.GOOS != osWindows && finfo.Mode()&0o111 == 0 {
 				continue
 			}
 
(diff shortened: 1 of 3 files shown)


cmd/entire/cli/agent/external/discovery.go
```diff
@@ -94,3 +98,16 @@ func DiscoverAndRegister(ctx context.Context) {
...
 	}
 }
+
+// stripExeExt removes Windows executable extensions (.exe, .bat, .cmd) from a
+// file name so that the agent name derived from the binary matches on all platforms.
+// On Unix this is effectively a no-op because binaries have no extension.
+func stripExeExt(name string) string {
+	lower := strings.ToLower(name)
+	for _, ext := range []string{".exe", ".bat", ".cmd"} {
+		if strings.HasSuffix(lower, ext) {
```
nodo: Instead of checking the suffix we can use `filepath.Ext`, something like:

```
  func stripExeExt(name string) string {
        switch strings.ToLower(filepath.Ext(name)) {
        case ".exe", ".bat", ".cmd":
                return strings.TrimSuffix(name, filepath.Ext(name))
        }
        return name
  }
```

keyu98:
> Thank you for the contribution @keyu98 ❤️ I have added a suggestion to simplify `stripExeExt`, also do you mind adding tests for it? The idea is great, just a couple of minor points.

Thanks for the review! I've simplified stripExeExt as suggested and added tests for it in the latest commit.

nodo:
Thank you @keyu98 !

PR #728, opened 2026-03-19
Use cleaner base URL in login output

dipree:
## Summary

- Prefer `verification_uri` (base URL) over `verification_uri_complete` (URL with code query param) in `entire login` output
- The device code is already printed on its own line, so repeating it in the URL is redundant noise

### Before

```
Device code: RB3X-56HU
Press Enter to open https://entire.io/cli/auth?code=RB3X-56HU in your browser...
```

### After

```
Device code: RB3X-56HU
Press Enter to open https://entire.io/cli/auth in your browser...
```

## Test plan

- [x] `mise run fmt && mise run lint` passes
- [x] `mise run test:ci` passes (unit + integration + canary)
- [ ] Manual test: run `entire login` and verify the URL no longer includes the code query param

> **Note:** Depends on #723 — merge that first.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Low risk: only changes which verification URL is printed/opened during device login, with a simple fallback if the preferred field is empty.
> 
> **Overview**
> Updates `entire login` to prefer `verification_uri` (base URL) over `verification_uri_complete` when displaying/opening the approval link, falling back to the complete URL only when the base URL is missing. This reduces redundant noise since the device code is already printed separately.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0e016d62cae9543bf051717b16b8283715114683. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #740, opened 2026-03-20
Fix login URL handling, add Ctrl+C support, add logout command

matthiaswenz:
## Summary

- **Only use `verification_uri`**: Never open `verification_uri_complete` — showing one URL but opening another is a phishing vector (an attacker who can tamper with the device auth response could set `verification_uri_complete` to a lookalike domain). Fixes regression from #728.
- **Handle Ctrl+C**: Make the "Press Enter to open..." prompt context-aware so Ctrl+C exits cleanly instead of leaving the process stuck on a blocking tty read
- **Fix double-close bug**: `waitForEnter` called `tty.Close()` on cancellation while a `defer tty.Close()` also ran on return — after the first close releases the FD, another goroutine could reuse that FD number and the deferred close would silently close an unrelated file
- **Browser fallback URL**: When the browser fails to open, print the approval URL so the user can copy-paste it
- **Add `entire logout`**: Removes stored credentials from the OS keyring, fully reversing `entire login`

## Test plan

- [x] Unit tests pass (`mise run test:ci`)
- [ ] Manual: run `entire login`, verify the same URL is shown and opened in browser
- [ ] Manual: run `entire login`, hit Ctrl+C at the prompt — verify clean exit
- [ ] Manual: run `entire login` with browser unavailable — verify URL is printed for copy-paste
- [ ] Manual: run `entire logout` — verify "Logged out." message and token removed from keyring

### DIFF
--- a/.claude/skills/e2e/implement.md
+++ b/.claude/skills/e2e/implement.md
@@ -2,6 +2,10 @@
 
 Apply fixes for E2E test failures, verify with scoped E2E tests.
 
+> **Before implementing any fixes, enter plan mode by invoking /plan.**
+> Analyze the findings (Steps 1-2 below), produce a complete fix plan with
+> specific file paths and code changes, and get user approval before executing.
+
 > **IMPORTANT: Running real E2E tests is a HARD REQUIREMENT of this procedure.**
 > Every fix MUST be verified with real E2E tests before the summary step.
 > Canary tests use the Vogon fake agent and cannot catch agent-specific issues.
--- a/.gemini/settings.json
+++ b/.gemini/settings.json
@@ -9,7 +9,7 @@
           {
             "name": "entire-session-start",
             "type": "command",
-            "command": "go run ${GEMINI_PROJECT_DIR}/cmd/entire/main.go hooks gemini session-start"
+            "command": "go run \"$(git rev-parse --show-toplevel)\"/cmd/entire/main.go hooks gemini session-start"
           }
         ]
       }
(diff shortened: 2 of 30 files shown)


cmd/entire/cli/login.go
```diff
@@ -67,25 +67,34 @@ func runLogin(ctx context.Context, outW, errW io.Writer, client deviceAuthClient
...
 
 	fmt.Fprintf(outW, "Device code: %s\n", start.UserCode)
-	approvalURL := start.VerificationURI
-	if approvalURL == "" {
-		approvalURL = start.VerificationURIComplete
+
+	// Use the clean URL (without code) for display, but open the complete URL
+	// (with code pre-filled) in the browser for a smoother experience.
+	displayURL := start.VerificationURI
+	browserURL := start.VerificationURIComplete
```
Soph: I'd actually would like to completely remove this. Basically I can send you a shortened url via iMessage, you click it and then it's just one click to disaster. I kind of assumed this was the case but I miss read daniels PR.

Yes it's less convenient but it's more secure.

Soph: Thinking more: The page asks to compare the code so if you just click that through, it's probably not really our fault.

cmd/entire/cli/login.go
```diff
@@ -170,16 +179,28 @@ func waitForApproval(ctx context.Context, poller deviceAuthClient, deviceCode st
...
+	go func() {
+		reader := bufio.NewReader(tty)
+		_, err := reader.ReadString('\n')
+		done <- err
+	}()
+
+	select {
+	case <-ctx.Done():
+		// Close tty to unblock the reading goroutine.
+		_ = tty.Close()
```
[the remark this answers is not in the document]
matthiaswenz: This shouldn't be a real issue since in this case we're exiting anyway.

PR #759, opened 2026-03-23
feat: v2 dual-write for checkpoints behind feature flag

pfleidi:
## Summary

Introduces checkpoints v2 dual-write support. When `checkpoints_v2` is enabled in strategy options, the CLI writes checkpoint data to both v1 (`entire/checkpoints/v1` branch) and v2 custom refs (`refs/entire/checkpoints/v2/main` and `refs/entire/checkpoints/v2/full/current`).

- **`V2GitStore`** — new store type for v2 ref operations, separate from `GitStore` (v1) to simplify future v1 removal
- **`/main` ref** — receives metadata, prompts, session metadata (no raw transcript or content hash). Compact `transcript.jsonl` will be added here once compaction (workstream B) is ready
- **`/full/current` ref** — accumulates raw transcripts (`full.jsonl`) + `content_hash.txt` across checkpoints via tree splicing (same pattern as v1). Generation rotation will archive and reset this ref when it reaches capacity (future work)
- **Dual-write wiring** — `CondenseSession` and stop-time finalization both dual-write, gated by `settings.IsCheckpointsV2Enabled()`. V2 failures are best-effort (logged as warnings, never block v1)
- **Stop-time optimization** — v2 feature flag is evaluated once before the finalization loop (not per-checkpoint) to avoid repeated settings I/O

### What's NOT in scope (future steps per design spec)

- Task checkpoint support on `/main`
- Push logic for v2 refs
- Read paths (`entire status`, `entire explain`, `entire resume`)
- Generation rotation for `/full/current`
- Compact transcript format / compaction (workstream B)
- v1 → v2 migration tool

## Test plan

- [x] 20 unit tests for `V2GitStore` (ref management, write/update for both refs, multi-session, accumulation, edge cases)
- [x] 2 strategy-level tests (v2 enabled vs disabled in `CondenseSession`)
- [x] 3 integration tests (full workflow, disabled path, stop-time finalization)
- [x] `mise run fmt && mise run lint` — 0 issues
- [x] `mise run test:ci` — all unit, integration, and E2E canary tests pass

### DIFF
--- a/cmd/entire/cli/checkpoint/v2_committed.go
+++ b/cmd/entire/cli/checkpoint/v2_committed.go
@@ -281,10 +281,10 @@ func (s *V2GitStore) writeMainCheckpointEntries(ctx context.Context, opts WriteC
 	return sessionIndex, nil
 }
 
-// writeMainSessionToSubdirectory writes a single session's metadata, prompts, and
-// content hash to a session subdirectory (0/, 1/, 2/, … indexed by session order
-// within the checkpoint). Unlike the v1 equivalent, this does NOT write the raw
-// transcript (full.jsonl) — that goes to /full/current.
+// writeMainSessionToSubdirectory writes a single session's metadata and prompts
+// to a session subdirectory (0/, 1/, 2/, … indexed by session order within the
+// checkpoint). Unlike the v1 equivalent, this does NOT write the raw transcript
+// (full.jsonl) or content hash (content_hash.txt) — those go to /full/current.
 func (s *V2GitStore) writeMainSessionToSubdirectory(opts WriteCommittedOptions, sessionPath string, entries map[string]object.TreeEntry) (SessionFilePaths, error) {
 	filePaths := SessionFilePaths{}
 
@@ -326,7 +326,6 @@ func (s *V2GitStore) writeMainSessionToSubdirectory(opts WriteCommittedOptions, 
 		ToolUseID:                   opts.ToolUseID,
 		TranscriptIdentifierAtStart: opts.TranscriptIdentifierAtStart,
 		CheckpointTranscriptStart:   opts.CheckpointTranscriptStart,
-		TranscriptLinesAtStart:      opts.CheckpointTranscriptStart,
 		TokenUsage:                  opts.TokenUsage,
 		SessionMetrics:              opts.SessionMetrics,
 		InitialAttribution:          opts.InitialAttribution,
@@ -402,10 +401,17 @@ func (s *V2GitStore) writeCommittedFullTranscript(ctx context.Context, opts Writ
 	checkpointPath := opts.CheckpointID.Path()
 	sessionPath := fmt.Sprintf("%s%d/", basePath, sessionIndex)
 
-	// Read existing entries at this checkpoint's shard path (accumulate, don't replace)
+	// Read existing entries at this checkpoint's shard path
 	entries, err := s.gs.flattenCheckpointEntries(rootTreeHash, checkpointPath)
 	if err != nil {
 		return err
+	}
+
+	// Clear existing entries at this session path before writing new ones
+	for key := range entries {
+		if strings.HasPrefix(key, sessionPath) {
+			delete(entries, key)
+		}
 	}
 
 	redactedTranscript, err := s.writeTranscriptBlobs(ctx, transcript, opts.Agent, sessionPath, entries)


cmd/entire/cli/checkpoint/checkpoint.go
```diff
@@ -419,8 +419,8 @@ func (m CommittedMetadata) GetTranscriptStart() int {
 // Used in CheckpointSummary.Sessions to map session IDs to their file locations.
 type SessionFilePaths struct {
 	Metadata    string `json:"metadata"`
-	Transcript  string `json:"transcript"`
-	ContentHash string `json:"content_hash"`
+	Transcript  string `json:"transcript,omitempty"`
+	ContentHash string `json:"content_hash,omitempty"`
```
pfleidi: These attributes might not be set going forward. Omitting them in such a case made sense to me.

cmd/entire/cli/checkpoint/v2_store.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"fmt"
+
+	"github.com/go-git/go-git/v6"
+	"github.com/go-git/go-git/v6/plumbing"
+	"github.com/go-git/go-git/v6/plumbing/object"
+)
+
+// V2GitStore provides checkpoint storage operations for the v2 ref layout.
```
computermode, line 11: Re: all the `V2` renames.. what are your thoughts on nesting GitStore in v1 and v2 packages? Since we'd need to support both for a while depending on how/when the migration is complete.

checkpoint/v1/store.go  -> v1.GitStore
checkpoint/v2/store.go -> v2.GitStore

During migration we can try to use the v2 store, and if that doesn't work, fall back to v1. Then once we switch over and verify nothing is using v1 anymore, we can delete v1 package and fallback.

cmd/entire/cli/strategy/manual_commit_condensation.go
```diff
@@ -265,10 +265,16 @@ func (s *ManualCommitStrategy) CondenseSession(ctx context.Context, repo *git.Re
...
-	}); err != nil {
+	}
+
+	// Write checkpoint metadata to v1 branch
+	if err := store.WriteCommitted(ctx, writeOpts); err != nil {
 		return nil, fmt.Errorf("failed to write checkpoint metadata: %w", err)
 	}
 
+	// Dual-write to v2 refs when enabled
+	writeCommittedV2IfEnabled(ctx, repo, writeOpts)
```
computermode, line 275: I think this could be another example of a case where we can check if v2 is enabled and optionally call the v2 or v1 package versions of `WriteCommitted` depending on which one is enabled

cmd/entire/cli/checkpoint/checkpoint.go
computermode: Why would Transcript be missing but Prompt isn't? Should Prompt be omitempty too? (Also, is this for the sub-agent case? Or is there some other case you had in mind?)

pfleidi: The original thought was that by moving the full transcript to the `/full/*` refs we wouldn't need this anymore unless a checkpoint was pinned. Given that these attributes were only used for the full transcripts we should be okay omitting them unless checkpoints are pinned down the road.

Your comment brought up a few more questions, though:

- Do we want this attribute to point to the compacted`transcript.jsonl` file instead?
- Do we want to introduce new attributes for compacted transcripts and their hashes?
- Do we want a way to refer to objects in the `/full/*` namespace to somehow make the lookup simpler?

Let's take a look at those a bit more closely tomorrow.

cmd/entire/cli/checkpoint/v2_store.go
pfleidi: I thought about that but there might be a lot of code we'd have to duplicate in order to make it work. I'm not opposed to the idea, though.

cmd/entire/cli/checkpoint/v2_committed.go
```diff
@@ -0,0 +1,461 @@
...
+	if err := validation.ValidateSessionID(opts.SessionID); err != nil {
+		return fmt.Errorf("invalid checkpoint options: %w", err)
+	}
+	if err := validation.ValidateToolUseID(opts.ToolUseID); err != nil {
+		return fmt.Errorf("invalid checkpoint options: %w", err)
+	}
+	if err := validation.ValidateAgentID(opts.AgentID); err != nil {
+		return fmt.Errorf("invalid checkpoint options: %w", err)
+	}
+	return nil
```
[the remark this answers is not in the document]
pfleidi: This is done on purpose to not impact v1 code.

cmd/entire/cli/strategy/manual_commit_condensation.go
pfleidi: The basic idea here was to do dual writes until we're confident that the v2 implementation works. Does that work for you?

computermode: Yes what I meant was in the world where we have the `v1` and `v2` packages. It's the same behavior just the code is written differently to support that

cmd/entire/cli/checkpoint/v2_committed.go
```diff
@@ -0,0 +1,480 @@
...
+
+// writeCommittedFullTranscript writes the raw transcript to the /full/current ref.
+// Transcripts accumulate across checkpoints — each write splices into the existing
+// tree. When the ref reaches capacity, generation rotation (future work) archives
+// the current ref and starts a fresh one.
+//
+// sessionIndex is the session slot (0-based), determined by the caller to stay
+// consistent with the /main ref's session numbering.
+// This is a no-op if opts.Transcript is empty (and opts.TranscriptPath is unset).
+func (s *V2GitStore) writeCommittedFullTranscript(ctx context.Context, opts WriteCommittedOptions, sessionIndex int) error {
```
Soph, line 377: >  updateCommittedFullTranscript clears existing entries at the session path before writing new blobs, but writeCommittedFullTranscript does not. If a transcript is ever written twice without going through the update path, stale blobs could persist. Adding the same clear-before-write loop would make it defensive and consistent.

cmd/entire/cli/checkpoint/v2_committed.go
```diff
@@ -0,0 +1,480 @@
...
+		CheckpointsCount:            opts.CheckpointsCount,
+		FilesTouched:                opts.FilesTouched,
+		Agent:                       opts.Agent,
+		Model:                       opts.Model,
+		TurnID:                      opts.TurnID,
+		IsTask:                      opts.IsTask,
+		ToolUseID:                   opts.ToolUseID,
+		TranscriptIdentifierAtStart: opts.TranscriptIdentifierAtStart,
+		CheckpointTranscriptStart:   opts.CheckpointTranscriptStart,
+		TranscriptLinesAtStart:      opts.CheckpointTranscriptStart,
```
Soph: `TranscriptLinesAtStart` is deprecated, let's remove it in v2

cmd/entire/cli/checkpoint/v2_committed.go
pfleidi: I think this should be addressed by https://github.com/entireio/cli/pull/759/changes/3560d7b349fdb7d18c552ce2050894276e84cf90

PR #766, opened 2026-03-25
Windows Support

dvydra:
## Summary

Adds **Windows support** across the CLI and test harness.

- Cross-compile Windows binaries via goreleaser (build, zip archive, Scoop bucket) and `mise` tasks
- Normalize git output/path handling for Windows (CRLF-safe parsing, `more` as default pager)
- Switch signal handling to `os.Interrupt` for cross-platform compatibility
- OS-specific process-group/cancellation helpers (`procattr_unix.go`, `procattr_windows.go`)
- Windows-compatible detached telemetry spawning with proper `DETACHED_PROCESS` flag
- **Skip interactive E2E tests on Windows** (tmux not available) — `StartSession` returns nil, tests auto-skip
- Add Windows E2E as parallel job in the main E2E workflow with Slack failure notifications
- `WINDOWS.md` guide for build, test, and architecture docs

## Test plan

- [x] `mise run fmt && mise run lint` passes
- [x] `mise run test:ci` passes (unit + integration + canary)
- [x] Windows E2E runs non-interactive (RunPrompt) tests via CI
- [x] Interactive tests auto-skip on Windows (`session == nil`)
- [ ] Verify Slack notifications fire on Windows E2E failure

### DIFF
--- a/cmd/entire/cli/integration_test/procattr_windows.go
+++ b/cmd/entire/cli/integration_test/procattr_windows.go
@@ -5,12 +5,14 @@ package integration
 import (
 	"os/exec"
 	"syscall"
+
+	"golang.org/x/sys/windows"
 )
 
 // detachFromTerminal configures the command to not inherit the parent console,
 // preventing interactive prompts from hanging tests.
 func detachFromTerminal(cmd *exec.Cmd) {
 	cmd.SysProcAttr = &syscall.SysProcAttr{
-		CreationFlags: syscall.CREATE_NEW_PROCESS_GROUP,
+		CreationFlags: windows.CREATE_NEW_PROCESS_GROUP | windows.DETACHED_PROCESS,
 	}
 }


pjbgf:
Perhaps we could also expand on the [goreleaser settings](https://github.com/entireio/cli/blob/fda9b31e8470d19417a8e86124783c60fb6f841d/.goreleaser.yaml#L15-L17) to add windows as well.

cmd/entire/main.go
```diff
@@ -19,7 +18,7 @@ func main() {
 
 	// Handle interrupt signals
 	sigChan := make(chan os.Signal, 1)
-	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
+	signal.Notify(sigChan, os.Interrupt)
 	go func() {
```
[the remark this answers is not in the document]
pjbgf: +1 on this.

e2e/agents/claude.go
```diff
@@ -221,3 +193,27 @@ func (c *Claude) StartSession(ctx context.Context, dir string) (Session, error)
...
+	dst, err := os.MkdirTemp("", "claude-config-*")
+	if err != nil {
+		return "", err
+	}
+
+	if os.Getenv("CI") != "" {
+		if home, err := os.UserHomeDir(); err == nil {
+			src := filepath.Join(home, ".claude", ".claude.json")
+			if _, err := os.Stat(src); err == nil {
+				_ = os.Symlink(src, filepath.Join(dst, ".claude.json"))
```
Soph: Is symlinks a thing on windows? Quick google mention it needs admin rights and is not always guranteed to work?

pjbgf: It works, but not in the same way as Unix as [per docs](https://cs.opensource.google/go/go/+/refs/tags/go1.26.1:src/os/file_unix.go;l=415-417). It also requires admin access [before Windows 10](https://github.com/golang/go/blob/d247ed00e498e9717fb7c80d126bee5a8afdb4e8/src/os/file_windows.go#L377). The latter should be OK due to the pre-reqs we set. The former we should be able to work around - ensure we always link to existing targets.

gtrrz-victor:
> Small nit on the sigterm, apart from that LGTM.

Fixed

cmd/entire/main.go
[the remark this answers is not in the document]
gtrrz-victor: Fixed

e2e/agents/claude.go
gtrrz-victor: I'll replace symlinks with simple file writes so we make sure it works with Windows versions older than 10.

gtrrz-victor: done!

PR #771, opened 2026-03-25
fix(opencode): handle session.idle event for turn-end detection

dipree:
## Summary

- OpenCode ≥1.3.x changed the session event model: `session.status` now only carries `"busy"`/`"retry"` (never `"idle"`), and a separate `session.idle` event fires when the agent finishes a turn
- The plugin only matched `session.status` with `status.type === "idle"`, so `turn-end` hooks stopped firing entirely after upgrading from 1.2.x to 1.3.x
- Added `session.idle` as a fall-through case alongside `session.status` for cross-version compatibility

## Root cause

Confirmed via diagnostic plugin logging. After the OpenCode 1.2.x → 1.3.x upgrade, the plugin's event handler receives:

```jsonl
// Many of these during a turn — status is always "busy", never "idle":
{"type":"session.status","props":{"sessionID":"ses_...","status":{"type":"busy"}}}
{"type":"session.status","props":{"sessionID":"ses_...","status":{"type":"busy"}}}

// One of these when the agent finishes — this is what the plugin needs to trigger turn-end:
{"type":"session.idle","props":{"sessionID":"ses_..."}}
```

The plugin checked `props?.status?.type !== "idle"` which always evaluated to `true` (since type is `"busy"`), so the handler broke out of every `session.status` event. The `session.idle` event that actually signals turn completion was unhandled.

## Impact chain

1. `turn-end` hook never fires → `PrepareTranscript` never runs
2. `opencode export <sessionID>` never called → transcript file (`.entire/tmp/<sessionID>.json`) never created
3. Post-commit condensation fails silently: `"transcript not found at .entire/tmp/<sessionID>.json: file does not exist"`
4. `Entire-Checkpoint` trailer is still added to commits, but no session data is written to `entire/checkpoints/v1`
5. Checkpoint directories on GitHub are empty

## Fix

Handle both event types as a fall-through case. `session.idle` has `properties.sessionID` directly, while `session.status` requires checking `properties.status.type` for `"idle"` (for ≤1.2.x backwards compatibility).

## Test plan

- [x] Diagnostic logging confirmed `session.idle` is emitted by OpenCode 1.3.2
- [x] Diagnostic logging confirmed `session.status` only carries `"busy"` in 1.3.2
- [ ] Start an OpenCode session with the updated plugin
- [ ] Make changes and commit — verify `turn-end` appears in `.entire/logs/entire.log`
- [ ] Verify transcript file is created at `.entire/tmp/<sessionID>.json`
- [ ] Verify `session condensed` log entry with non-zero `transcript_lines`
- [ ] Push and verify checkpoint data appears on `entire/checkpoints/v1` branch

### DIFF
--- a/cmd/entire/cli/agent/opencode/entire_plugin.ts
+++ b/cmd/entire/cli/agent/opencode/entire_plugin.ts
@@ -117,9 +117,10 @@ export const EntirePlugin: Plugin = async ({ directory }) => {
 
           case "session.idle":
           case "session.status": {
-            // session.status fires with status.type "idle" when the agent finishes.
-            // session.idle is a separate event that some OpenCode versions emit instead.
-            // Handle both for compatibility across OpenCode versions.
+            // OpenCode ≤1.2.x: session.status with status.type "idle" signals turn end.
+            // OpenCode ≥1.3.x: session.status only carries "busy"/"retry"; a separate
+            //   session.idle event fires when the agent finishes.
+            // Handle both for cross-version compatibility.
             let sessionID: string | null = null
             if (event.type === "session.idle") {
               sessionID = (event as any).properties?.sessionID ?? currentSessionID


Soph:
this isn't the right fix... `session.idle` is deprecated and also where ever `session.idle` fires `session.status` should fire too, closing this for now

PR #775, opened 2026-03-25
rewrite opencode plugin if content differs

Soph:
Entire-Checkpoint: f92b8ef4844d

---

> [!NOTE]
> **Low Risk**
> Low risk behavioral tweak limited to OpenCode hook installation; main change is that `InstallHooks` will now overwrite stale plugin content instead of treating any file with the marker as already installed.
> 
> **Overview**
> Makes `opencode` hook installation truly idempotent by comparing the on-disk `.opencode/plugins/entire.ts` contents to the newly generated template output.
> 
> `InstallHooks` now returns no-op only when the file is *exactly* up-to-date (unless `force`), and will rewrite the plugin when content differs (e.g., `localDev` vs released command prefix).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 13238aa91dc8dea31bdf5c82fe69ac8780c1574e. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


cmd/entire/cli/agent/opencode/hooks.go
```diff
@@ -68,6 +58,15 @@ func (a *OpenCodeAgent) InstallHooks(ctx context.Context, localDev bool, force b
 	// Generate plugin content from template
 	content := strings.ReplaceAll(pluginTemplate, entireCmdPlaceholder, cmdPrefix)
 
+	// Check if already installed with identical content (idempotent) unless force
+	if !force {
+		if existing, readErr := os.ReadFile(pluginPath); readErr == nil { //nolint:gosec // Path constructed from repo root
+			if string(existing) == content {
```
pfleidi, line 65: I'm assuming this is an optimization to avoid having to overwrite the same file over and over again, correct?

pfleidi (dismissed):

cmd/entire/cli/agent/opencode/hooks.go
Soph: As it was before we did not update it at all but just checked if it's there so I went with this. But I guess on the other side since this happens only in the context of `entire enable` we could also just always write it? 🤔

pfleidi: I'm fine, either way. I was mostly curious about the reason to validate the content.

gtrrz-victor (approved):
Tested and working!

Issue #768, opened 2026-03-25
Entire enabled but commits never get Entire-Checkpoint trailer (hooks run, no linkage)


dayhaysoos:
### What happened?

Hello! I'm seeing a reproducible issue where Entire is enabled andhooks are installed, but new commits do not receive Entire-Checkpoint trailers.

### Steps to reproduce

1. Make sure Entire is enabled (I'm on the latest version 0.5.1)

```shell
entire enable --force --absolute-git-hook-path --agent opencode
entire status
ls -l .git/hooks/prepare-commit-msg .git/hooks/commit-msg .git/hooks/post-commit
```

2. Stage any file change
3. Commit with debug:

`ENTIRE_LOG_LEVEL=DEBUG git commit -m "checkpoint debug test"`

4. Check commit linkage

```shell
git show -s --format=%B HEAD
entire explain --commit HEAD --no-pager
entire trace
```

What happens:

- Commit succeeds, but commit message has no trailer:
- no Entire-Checkpoint: ...
- entire explain --commit HEAD --no-pager returns:
    - No associated Entire checkpoint
    - Commit <sha> does not have an Entire-Checkpoint trailer.
- entire trace only shows post-commit:
    - post-commit ...
    - step: open_repository_and_head
entire doctor reports metadata branches OK (no obvious repair action suggested).

### Entire CLI version

Entire CLI 0.5.1 (d46fdc24)

### OS and architecture

Darwin 25.3.0 arm64

### Agent

GPT-5.3 Codex

### Terminal

OpenCode Desktop terminal (I suppose it's Ghostty?)

### Logs / debug output

```bash
# Config confirms Entire is enabled with always-linking
.entire/settings.json:1-6
{
  "enabled": true,
  "absolute_git_hook_path": true,
  "telemetry": true,
  "commit_linking": "always"
}


# Commit attempt 2 (same result)
.entire/logs/entire.log:6329-6374
{"time":"2026-03-24T23:24:34.756062-04:00","level":"DEBUG","msg":"prepare-commit-msg hook invoked","session_id":"ses_2e7798cf4ffe51qGR1tS3y3sM7","component":"hooks","hook":"prepare-commit-msg","hook_type":"git","strategy":"manual-commit","source":"message"}
[...]

PR #779, opened 2026-03-25
Fix: Normalize FilesTouched paths to prevent missing checkpoint trail…

Adityakk9031:
Issue
Commits never receive Entire-Checkpoint trailers despite hooks running correctly. The prepare-commit-msg hook finds sessions but logs "no content to link" for all of them.

The root cause is a path format mismatch in 
stagedFilesOverlapWithContent
. 
getStagedFiles()
 returns repo-relative paths (e.g., src/screens/SessionDetail.tsx) from git diff --cached --name-only, but state.FilesTouched stores the exact paths the agent provides — often absolute (e.g., /Users/alex/project/src/screens/SessionDetail.tsx). The touchedSet[stagedPath] lookup silently fails, so no file ever reaches the content comparison logic.

A contributing factor is that opencode export returns invalid JSON, which prevents the active session's live transcript fallback from detecting content.

Solution
Normalize at write time: 
mergeFilesTouched
 now accepts a worktreePath parameter and normalizes all incoming paths to repo-relative form via a new 
normalizeToRepoRelative
 helper before storing them in state.FilesTouched.
Basename fallback at read time: 
stagedFilesOverlapWithContent
 now builds a secondary touchedByBase map keyed by filepath.Base(). When the exact path lookup fails, it falls back to basename matching to handle stale sessions from older CLI versions.
Debug logging: A new log line ("path format mismatch, matched by basename") makes this class of silent failure visible in future.

### DIFF
--- a/cmd/entire/cli/strategy/content_overlap.go
+++ b/cmd/entire/cli/strategy/content_overlap.go
@@ -186,10 +186,14 @@ func filesOverlapWithContent(ctx context.Context, repo *git.Repository, shadowBr
 func stagedFilesOverlapWithContent(ctx context.Context, repo *git.Repository, shadowTree *object.Tree, stagedFiles, filesTouched []string) bool {
 	logCtx := logging.WithComponent(ctx, "checkpoint")
 
-	// Build set of filesTouched for quick lookup
+	// Build set of filesTouched for quick lookup.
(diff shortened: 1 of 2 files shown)


Soph:
Thanks for opening this PR. The issue #768 was closed by #832 and hardened by #785. Sorry that your change didn't made it.

PR #783, opened 2026-03-26
Add authenticated API client for trail commands

matthiaswenz:
## Summary
- Adds a bearer-token HTTP client (`api/client.go`) that injects the CLI login token into all outgoing API requests via a custom `http.RoundTripper` (currently only used for `/api/trails` routes)
- Adds support for private trails API with typed request/response structs for it (`api/trail_types.go`)
- Migrates all four trail commands (`list`, `create`, `show`, `update`) from local git storage (`trail.Store`) to the remote trails API at `/api/v1/trails/:host/:owner/:repo`
- Exports `strategy.ResolveRemoteRepo()` to extract host/owner/repo from git remotes to figure out the proper parameters for use with the trails API
- Removes git-based trail storage from hooks and push:
  - Removes checkpoint-trail linking from post-commit hook (API links implicitly by branch name)
  - Removes trails branch push from `PrePush` and `PushTrailsBranch()`
  - Keeps `TrailsBranchName` constant and `IsShadowBranch` exclusion as safety net for existing repos

## Test plan
- [ ] Run `entire login` then `entire trail list` against local dev server
- [ ] Verify trail CRUD operations work via API (`create`, `update`, show by branch)
- [ ] Verify `git push` no longer attempts to push `entire/trails/v1`
- [ ] Run `mise run fmt && mise run lint && mise run test:ci` (passing locally)

### DIFF
--- a/cmd/entire/cli/api_client.go
+++ b/cmd/entire/cli/api_client.go
@@ -10,7 +10,8 @@ import (
 
 // NewAuthenticatedAPIClient creates an API client using the bearer token
 // from the CLI login flow. Returns an error if the user is not logged in.
-func NewAuthenticatedAPIClient() (*api.Client, error) {
+// Pass insecureHTTP=true to allow plain HTTP base URLs (for local development).
+func NewAuthenticatedAPIClient(insecureHTTP bool) (*api.Client, error) {
 	token, err := auth.LookupCurrentToken()
 	if err != nil {
 		return nil, fmt.Errorf("lookup auth token: %w", err)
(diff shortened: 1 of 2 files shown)


cmd/entire/cli/api/client.go
```diff
@@ -0,0 +1,175 @@
+package apiurl
```
Soph: why `apiurl` maybe `api` is better` but `apiurl` feels strange 🤔  ?

cmd/entire/cli/api_client.go
```diff
@@ -0,0 +1,23 @@
...
+// from the CLI login flow. Returns an error if the user is not logged in.
+func NewAuthenticatedAPIClient() (*apiurl.Client, error) {
+	token, err := auth.LookupCurrentToken()
+	if err != nil {
+		return nil, fmt.Errorf("lookup auth token: %w", err)
+	}
+	if token == "" {
+		return nil, errors.New("not logged in (run 'entire login' first)")
+	}
+
```
Soph, line 22:
```suggestion

  if err := apiurl.RequireSecureURL(apiurl.BaseURL()); err != nil {
    return nil, err
  }
```

Let's make sure nobody can sneak in a http url here. If you want that for local testing we should do the same as with login and allow `insecure-http-auth` as a param to be passed through somehow.

cmd/entire/cli/api/client.go
matthiaswenz: Good callout, will change this

PR #795, opened 2026-03-27
feat: add v2 checkpoint read path for entire resume

pfleidi:
## Summary

- Add v2 read methods to `V2GitStore` (`ReadCommitted`, `ReadSessionContent`, `GetSessionLog`) that read metadata from `/main` ref and raw transcripts from `/full/*` refs with archived generation search
- Add `ResolveTranscript` (v2→v1 fallback) and `GetV2MetadataTree` (fetch with fallback) as shared helpers for other commands to reuse
- Add v2 fetch helpers (`FetchV2MainTreeOnly`, `FetchV2MainRef`) using explicit refspecs for `refs/entire/` namespace
- Wire v2-first resolution into `entire resume`: `resumeFromCurrentBranch`, `resolveLatestCheckpoint`, `resumeSingleSession`, and `checkRemoteMetadata` all try v2 first when `IsCheckpointsV2Enabled`, falling back to v1
- Wire v2-first reads into `RestoreLogsOnly` for multi-session restore
- All v2 code paths are gated behind `settings.IsCheckpointsV2Enabled(ctx)` — no behavior change when disabled

## Test plan

- [x] Unit tests for `ReadCommitted`, `ReadSessionContent` (including archived generation search and missing transcript)
- [x] Unit tests for `ResolveTranscript` (v2 found, v2 disabled falls back to v1, v2 miss falls back to v1)
- [x] Integration tests: v2 resume across branches, same-branch resume, v1 fallback after upgrade
- [x] `mise run fmt && mise run lint && mise run test:ci` all pass

---

> [!NOTE]
> **Medium Risk**
> Changes the `resume`/log-restore code paths to prefer v2 custom refs and adds new git fetch behaviors; mistakes could prevent session restoration or cause unexpected network/ref resolution issues. Feature-flag gating reduces blast radius when `checkpoints_v2` is disabled.
> 
> **Overview**
> Adds a **v2 checkpoint read path**: `V2GitStore` can now read checkpoint metadata from the v2 `/main` ref and restore **raw transcripts** from `/full/current` or archived `/full/*` generations (including chunk reassembly), returning `ErrNoTranscript` when transcripts are missing.
> 
> Updates `entire resume` and `RestoreLogsOnly` to **try v2 first when `settings.IsCheckpointsV2Enabled(ctx)` is true**, with clean fallback to v1 reads when v2 data/refs aren’t available. Also introduces v2 ref resolution + fetch helpers (treeless and full fetch, plus checkpoint-remote support) so resume can obtain v2 metadata even when refs aren’t present locally.
> 
> Adds unit tests for v2 committed/session reads (missing transcript, archived generations, chunked transcripts), tests for v2 metadata-tree resolution, checkpoint-remote fetch tests for v2 `/main`, and new integration tests covering v2 resume, same-branch resume, and v1 fallback after enabling v2.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 87ce7070734f2d7c6b5cf94cfa63385b9680a895. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/strategy/manual_commit_rewind.go
+++ b/cmd/entire/cli/strategy/manual_commit_rewind.go
@@ -633,7 +633,9 @@ func (s *ManualCommitStrategy) RestoreLogsOnly(ctx context.Context, w, errW io.W
 
 	// Resolve which store has this checkpoint. Try v2 first when enabled.
 	// The chosen reader is used for all subsequent reads (summary + session content)
-	// to avoid mixed v1/v2 reads.
+	// to avoid mixed v1/v2 reads. No per-session fallback to v1: during dual-write,
+	// both stores receive the same data, so if v2 has the summary it also has the
+	// transcripts on /full/* refs.
 	var reader committedReader
 	var summary *cpkg.CheckpointSummary
 


pfleidi:
Bugbot run

cmd/entire/cli/checkpoint/v2_read.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	return &summary, nil
+}
+
+// ReadSessionContent reads a session's metadata and prompts from the v2 /main ref,
+// and the raw transcript from /full/* refs (current + archived generations).
+// If the transcript is not found in any /full/* ref, the returned SessionContent
+// has an empty Transcript field — metadata and prompts are still populated.
+func (s *V2GitStore) ReadSessionContent(ctx context.Context, checkpointID id.CheckpointID, sessionIndex int) (*SessionContent, error) {
```
computermode, line 66: Is this function intended only to be used for the resume logic / anything that requires the full.jsonl file? If so, maybe `ReadRawSessionContent` or `ReadFullSessionContent` is clearer since we'll have two read paths as I understand it: one for metadata + the compact transcript.jsonl file, and another for the raw transcript + metadata from `/full/*`.


```
// If the transcript is not found in any /full/* ref, the returned SessionContent
// has an empty Transcript field — metadata and prompts are still populated.
```

And if this is only for the resume logic, do we still want to return these if the transcript is empty? I don't think `resume` would work in that case, maybe we should error? 🤔

computermode:
I think it's worth checking against the changes in https://github.com/entireio/cli/pull/796 to make sure we're still supporting the separate repo path / not reintroducing any of the bugs that PR fixes here.

cmd/entire/cli/checkpoint/v2_read.go
pfleidi: Yes, this is used to read the `full.jsonl` file for the resume logic. The main reason why I decided to keep the `ReadSessionContent` is that it's part of the [Store interface](https://github.com/entireio/cli/blob/ed2931c5e786c9c5cfdee65a4c0ae6f570b80efb/cmd/entire/cli/checkpoint/checkpoint.go#L97) which I don't want to change since it'll also impact the already existing [GitStore](https://github.com/entireio/cli/blob/ed2931c5e786c9c5cfdee65a4c0ae6f570b80efb/cmd/entire/cli/checkpoint/committed.go#L777).

I'm going to update the comment and the error handling part.

pfleidi:
Bugbot run

cmd/entire/cli/checkpoint/committed.go
```diff
@@ -824,6 +825,10 @@ func (s *GitStore) ReadSessionContent(ctx context.Context, checkpointID id.Check
 		}
 	}
 
+	if len(result.Transcript) == 0 {
```
pfleidi, line 828: This was added for compatibility with checkpoints v2 which is more specific about error states for debug logging and such.

cmd/entire/cli/strategy/manual_commit_rewind.go
```diff
@@ -621,16 +631,38 @@ func (s *ManualCommitStrategy) RestoreLogsOnly(ctx context.Context, w, errW io.W
...
+	if summary == nil {
+		v1Store, err := s.getCheckpointStore()
+		if err != nil {
+			return nil, fmt.Errorf("failed to get checkpoint store: %w", err)
+		}
+		summary, err = v1Store.ReadCommitted(ctx, point.CheckpointID)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read checkpoint: %w", err)
+		}
+		reader = v1Store
```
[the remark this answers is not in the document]
pfleidi: During dual writing, both data should be available locally. Given that enabling is a global on/off switch not something that happens on a per session basis. I've added a comment describing it.

cmd/entire/cli/checkpoint/v2_read.go
```diff
@@ -0,0 +1,256 @@
...
+		if err == nil {
+			content, contentErr := file.Contents()
+			if contentErr == nil {
+				return []byte(content), nil
+			}
+		}
+	}
+
+	return nil, nil
+}
```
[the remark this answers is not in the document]
pfleidi: Since v1 is going to essentially be legacy code, I'd rather live with that duplication than trying to find an abstraction that would make it harder for us to move forward.

PR #805, opened 2026-03-30
Fix checkpoint_remote silently falling back to origin on owner mismatch

### LINKED ISSUE #800: remote checkpoint broken
### What happened?

I want conversations to go to a different repo, never my working repo. I have this:

```
{
  "enabled": true,
  "strategy": "manual-commit",
  "strategy_options": {
    "checkpoint_remote": {
      "provider": "github",
      "repo": "org/repo"
    },
    "push_sessions": false,
    "summarize": {
      "enabled": true
    }
  },
  "telemetry": false
}
```

I've tried some variations of push_sessions to true, removing push_sessions entirely, etc. Disabled and cleared the projects .entire as well and deleted the branches and tried to re-initialize but for the life of me I *cannot* get entire to push to a remote checkpoint.

I also don't notice any summaries or what to look out for to see if the summaries are working, but I don't see anything that screams "this is a summary" to me in the UI

### Steps to reproduce

_No response_

### Entire CLI version

Entire CLI 0.5.1 (d46fdc24)

### OS and architecture

Darwin 24.6.0 arm64

### Agent

2.1.87 (Claude Code)

### Terminal

iTerm2 Build 3.6.9

### Logs / debug output

```bash

```

### Additional context

_No response_

blackgirlbytes:
## Summary

- **Removed fork detection skip** in `resolvePushSettings()` — when a user explicitly configures `checkpoint_remote`, that choice is now respected regardless of owner mismatch between push remote and checkpoint remote
- **Added user-visible stderr warnings** for all fallback paths (remote URL lookup failure, URL parse failure, URL derivation failure) so users know when checkpoint_remote is being skipped

## Context

When `checkpoint_remote` was configured (e.g., `{"provider": "github", "repo": "alice/checkpoints"}`) but the origin remote had a different owner (e.g., `org/work-repo`), fork detection silently skipped the checkpoint remote and pushed to origin instead. All failure messages went to debug/warn logs in `.entire/logs/` — the user had no idea their config was being ignored.

Reproduced with: origin = `hydeparksda/invite-automation`, checkpoint_remote = `blackgirlbytes/genuary2026`.

Fixes #800

## Test plan

- [x] Unit tests pass (`mise run test`)
- [x] Integration tests pass (`mise run test:integration`)
- [x] Lint passes (`mise run lint`)
- [x] Updated `TestResolvePushSettings_ForkDetection` to expect checkpoint URL is set even with owner mismatch
- [ ] Manual verification: configure checkpoint_remote with different owner than origin, push, confirm checkpoints land on checkpoint remote

### DIFF
--- a/cmd/entire/cli/strategy/checkpoint_remote.go
+++ b/cmd/entire/cli/strategy/checkpoint_remote.go
@@ -76,9 +76,11 @@ func resolvePushSettings(ctx context.Context, pushRemoteName string) pushSetting
 		return ps
 	}
 
-	// Get the push remote URL for protocol detection and fork detection
+	// Get the push remote URL for protocol detection
 	pushRemoteURL, err := getRemoteURL(ctx, pushRemoteName)
 	if err != nil {
+		fmt.Fprintf(os.Stderr, "[entire] Warning: checkpoint_remote is configured (%s) but could not get push remote URL: %v\n", config.Repo, err)
+		fmt.Fprintln(os.Stderr, "[entire] Checkpoints will be pushed to origin instead. Check your git remote configuration.")
 		logging.Debug(ctx, "checkpoint-remote: could not get push remote URL, skipping",
 			slog.String("remote", pushRemoteName),
 			slog.String("error", err.Error()),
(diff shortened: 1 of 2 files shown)


Soph:
The issue is that the CLI repo has a remote source setup but if you fork it, then you get that setup but have for sure no access to the remote and then the push would also fail. So while this fixes one thing, it makes another thing worse. 

I wonder if a better fix is: 

- Skip fork detection if `checkpoint_remote` is set in `settings.local.json` (user explicitly configured it)
- Keep fork detection if it only comes from `settings.json` (inherited from the repo)

This could be done by checking whether `settings.local.json` has `checkpoint_remote` set (using `LoadFromFile` which already exists), and only applying fork detection when it doesn't. That way the user's explicit local config always wins.

rublev:
Stoked for this to get merged! Great job and thanks for tackling it quickly 👏

Soph:
closing this one, it got obsolete with #1898

PR #821, opened 2026-03-31
feat: push logic for checkpoints v2 refs

pfleidi:
## Summary

Implements push logic for checkpoints v2 refs under `refs/entire/`. This is task A7 from the [checkpoints v2 design spec](docs/superpowers/specs/2026-03-19-checkpoints-v2-design.md).

When both `checkpoints_v2` and `push_v2_refs` are enabled in strategy options, the pre-push hook pushes v2 refs alongside the existing v1 branch using explicit refspecs:

```
refs/entire/checkpoints/v2/main:refs/entire/checkpoints/v2/main
refs/entire/checkpoints/v2/full/current:refs/entire/checkpoints/v2/full/current
refs/entire/checkpoints/v2/full/0000000000001:refs/entire/checkpoints/v2/full/0000000000001
```

### How it works

**Push gating:** Both settings must be true:
```json
{
  "strategy_options": {
    "checkpoints_v2": true,
    "push_v2_refs": true
  }
}
```

**What gets pushed:** `/main`, `/full/current`, and the latest archived generation (older archives are immutable and were pushed when created).

**Merge recovery for `/main` and `/full/current`:** When a push fails (non-fast-forward), the CLI fetches the remote ref, flattens both trees, combines entries, and creates a merge commit — same strategy as v1. This works conflict-free because both refs contain only sharded checkpoint directories (`<id[:2]>/<id[2:]>/`) with no root-level files.

**Rotation conflict recovery:** When `/full/current` push fails and the remote has an archived generation we don't have locally, another machine rotated. Instead of merging two divergent `/full/current` refs (which would duplicate data across generations), the CLI merges local data into the latest archived generation:

```
Starting state (synced): /full/current has 95 checkpoints

User A: adds 7 → 102 → rotation triggers
  /full/0000000000001: 102 checkpoints (95 shared + 7 from A)
  /full/current: fresh orphan
  Pushed to remote.

User B: adds 3 → 98 locally, pushes /full/current → non-fast-forward
  Recovery:
    1. ls-remote → discovers /full/0000000000001 (not local)
    2. Fetch /full/0000000000001
    3. Tree-merge local /full/current into /full/0000000000001
       → 95 shared deduplicated by git content-addressing, result: 105
    4. Update generation.json timestamps, push /full/0000000000001
    5. Adopt remote /full/current as local

Final state:
  /full/0000000000001: 105 checkpoints (no duplication)
  /full/current: fresh orphan
```

This preserves the key GC property: each checkpoint's raw transcript exists in exactly one generation, so deleting a generation ref makes its unique blobs unreachable.

### generation.json changes

`generation.json` is no longer written to `/full/current` on every checkpoint write. It's written only at archive time (during rotation), containing just timestamps:

```json
{
  "oldest_checkpoint_at": "2026-01-01T00:00:00Z",
  "newest_checkpoint_at": "2026-02-15T00:00:00Z"
}
```

The `checkpoints` list field has been removed. Checkpoint count for rotation is determined by walking shard directories in the tree. This keeps `/full/current` free of root-level files, making all tree merges conflict-free.

### Fetch-on-demand

- **Checkpoint remote:** When `push_v2_refs` is enabled and a checkpoint remote is configured, the v2 `/main` ref is fetched from the checkpoint remote URL on first push (same one-time pattern as v1).
- **`entire resume`:** When a transcript isn't found locally, `readTranscriptFromFullRefs` discovers and fetches remote `/full/*` refs before giving up. Respects `checkpoint_remote` configuration.
- All `V2GitStore` instances receive the resolved fetch remote at construction time — no hardcoded `"origin"`.

### Other improvements

- Extracted `WalkCheckpointShards` helper for reusable two-level shard iteration (used by `CountCheckpointsInTree` and `ListCommitted`)
- Exported `GenerationRefPattern`, `GetRefState`, `ListArchivedGenerations`, `AddGenerationJSONToTree` for cross-package use

## Test plan

- [x] Unit tests for all push/merge/rotation functions (`push_v2_test.go`)
- [x] Integration tests for full push cycle and disabled-push gating (`v2_push_test.go`)
- [x] Updated generation tests for archive-time-only `generation.json`
- [x] `mise run fmt && mise run lint && mise run test:ci` passes

---

> [!NOTE]
> **Medium Risk**
> Adds new git push/fetch/merge paths for custom v2 refs (including rotation-conflict handling) and changes generation rotation metadata, which could impact checkpoint durability and resume/restore behavior if edge cases are missed.
> 
> **Overview**
> Enables **optional pushing of v2 checkpoint refs** (under `refs/entire/checkpoints/v2/*`) from the pre-push hook when both `checkpoints_v2` and `push_v2_refs` are set, including fetch+tree-merge recovery on non-fast-forward and special handling when `/full/current` was rotated remotely.
> 
> Changes v2 generation management so `/full/current` no longer writes `generation.json` on each write; rotation now **counts checkpoints by walking shard dirs**, writes `generation.json` only when archiving, and resets `/full/current` to an empty orphan commit.
> 
> Adds v2 read/fetch plumbing: `V2GitStore` can read committed metadata and raw transcripts across `/full/current` + archived generations (with on-demand remote fetch), `resume`/`RestoreLogsOnly` prefer v2 when enabled with v1 fallback, and checkpoint-remote setup can fetch the v2 `/main` ref when v2 pushing is enabled.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit edca0a90161e18f5b1e86877c4041e6185b66735. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/resume.go
+++ b/cmd/entire/cli/resume.go
@@ -907,11 +907,7 @@ func resumeSingleSession(ctx context.Context, w, errW io.Writer, ag agent.Agent,
 	if settings.IsCheckpointsV2Enabled(ctx) {
 		repo, repoErr := openRepository(ctx)
 		if repoErr == nil {
-			fetchRemote := strategy.ResolveCheckpointURL(ctx, "origin")
-			if fetchRemote == "" {
-				fetchRemote = "origin"
-			}
-			v2Store := checkpoint.NewV2GitStore(repo, fetchRemote)
+			v2Store := checkpoint.NewV2GitStore(repo, strategy.ResolveCheckpointURL(ctx, "origin"))
 			var v2Err error
 			logContent, _, v2Err = v2Store.GetSessionLog(ctx, checkpointID)
 			if v2Err != nil {
--- a/cmd/entire/cli/strategy/checkpoint_remote.go
+++ b/cmd/entire/cli/strategy/checkpoint_remote.go
@@ -171,15 +171,6 @@ func ResolveCheckpointURL(ctx context.Context, pushRemoteName string) string {
 	return url
 }
 
-// resolveV2FetchRemote returns the remote to use for v2 fetch operations.
-// Returns the checkpoint remote URL if configured, otherwise "origin".
-func resolveV2FetchRemote(ctx context.Context) string {
-	if url := ResolveCheckpointURL(ctx, "origin"); url != "" {
-		return url
-	}
-	return "origin"
-}
-
 // ResolveRemoteRepo returns the host, owner, and repo name for the given git remote.
 // It parses the remote URL (SSH or HTTPS) and extracts the components.
 // For example, git@github.com:org/my-repo.git returns ("github.com", "org", "my-repo").
--- a/cmd/entire/cli/strategy/manual_commit.go
+++ b/cmd/entire/cli/strategy/manual_commit.go
@@ -79,7 +79,7 @@ func (s *ManualCommitStrategy) getV2CheckpointStore(ctx context.Context) (*check
 			s.v2CheckpointStoreErr = fmt.Errorf("failed to open repository: %w", err)
 			return
 		}
-		s.v2CheckpointStore = checkpoint.NewV2GitStore(repo, resolveV2FetchRemote(ctx))
+		s.v2CheckpointStore = checkpoint.NewV2GitStore(repo, ResolveCheckpointURL(ctx, "origin"))
 	})
 	return s.v2CheckpointStore, s.v2CheckpointStoreErr
 }
--- a/cmd/entire/cli/strategy/manual_commit_condensation.go
+++ b/cmd/entire/cli/strategy/manual_commit_condensation.go
@@ -1088,7 +1088,7 @@ func writeCommittedV2IfEnabled(ctx context.Context, repo *git.Repository, opts c
 		return
 	}
 
-	v2Store := cpkg.NewV2GitStore(repo, resolveV2FetchRemote(ctx))
+	v2Store := cpkg.NewV2GitStore(repo, ResolveCheckpointURL(ctx, "origin"))
 	if err := v2Store.WriteCommitted(ctx, opts); err != nil {
 		logging.Warn(ctx, "v2 dual-write failed",
 			slog.String("checkpoint_id", opts.CheckpointID.String()),
--- a/cmd/entire/cli/strategy/manual_commit_hooks.go
+++ b/cmd/entire/cli/strategy/manual_commit_hooks.go
@@ -2337,7 +2337,7 @@ func (s *ManualCommitStrategy) finalizeAllTurnCheckpoints(ctx context.Context, s
 	// Evaluate v2 flag once before the loop to avoid re-reading settings per checkpoint
 	var v2Store *checkpoint.V2GitStore
 	if settings.IsCheckpointsV2Enabled(logCtx) {
-		v2Store = checkpoint.NewV2GitStore(repo, resolveV2FetchRemote(logCtx))
+		v2Store = checkpoint.NewV2GitStore(repo, ResolveCheckpointURL(logCtx, "origin"))
 	}
 
 	// Update each checkpoint with the full transcript
--- a/cmd/entire/cli/strategy/push_v2.go
+++ b/cmd/entire/cli/strategy/push_v2.go
@@ -398,7 +398,7 @@ func pushV2Refs(ctx context.Context, target string) {
 	if err != nil {
 		return
 	}
-	store := checkpoint.NewV2GitStore(repo, resolveV2FetchRemote(ctx))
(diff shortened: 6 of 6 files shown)


cmd/entire/cli/checkpoint/committed.go
```diff
@@ -889,76 +886,49 @@ func (s *GitStore) ListCommitted(ctx context.Context) ([]CommittedInfo, error) {
...
 
 	// Scan sharded structure: <2-char-prefix>/<remaining-id>/metadata.json
-	for _, bucketEntry := range tree.Entries {
-		if bucketEntry.Mode != filemode.Dir {
-			continue
-		}
-		// Bucket should be 2 hex chars
-		if len(bucketEntry.Name) != 2 {
-			continue
+	_ = WalkCheckpointShards(s.repo, tree, func(checkpointID id.CheckpointID, cpTreeHash plumbing.Hash) error {
```
pfleidi: We were doing this tree walk in three different places. This change here is essentially just using a common helper instead of re-inventing the wheel.

cmd/entire/cli/checkpoint/v2_generation.go
```diff
@@ -168,15 +149,54 @@ func (s *V2GitStore) addGenerationToRootTree(rootTreeHash plumbing.Hash, gen Gen
 		UpdateSubtreeOptions{MergeMode: MergeKeepExisting})
 }
 
+// computeGenerationTimestamps derives timestamps for a generation being archived.
+// Uses the commit history of the /full/current ref: oldest = first commit time,
+// newest = latest commit time. Falls back to time.Now() if the ref has no history.
+// Note: /full/* trees don't contain session metadata (that's on /main), so we
+// derive timestamps from git commit times rather than walking the tree.
+func (s *V2GitStore) computeGenerationTimestamps() GenerationMetadata {
+	now := time.Now().UTC()
```
computermode, line 158: I notice we're using `time.Now().UTC()` here which I think makes sense, but elsewhere in the code we have a lot of instances of just local `time.Now()` and I'm not sure if it will cause issues for us here or not. Non-blocking for this PR (if we're not relying on any conflicting `time.Now()` timestamps elsewhere in the existing code for this logic to work correctly), but I think ultimately we'll want to standardize on UTC.

pfleidi: I did some digging and it looks like `UTC` is the default in most places but we seem to have some inconsistencies that should be rectified in a different PR. 👍

cmd/entire/cli/resume.go
```diff
@@ -907,7 +907,11 @@ func resumeSingleSession(ctx context.Context, w, errW io.Writer, ag agent.Agent,
 	if settings.IsCheckpointsV2Enabled(ctx) {
 		repo, repoErr := openRepository(ctx)
 		if repoErr == nil {
-			v2Store := checkpoint.NewV2GitStore(repo)
+			fetchRemote := strategy.ResolveCheckpointURL(ctx, "origin")
+			if fetchRemote == "" {
+				fetchRemote = "origin"
+			}
```
computermode: Maybe this is a nitpick, but calling out that we seem to be duplicating logic here as what's in `resolveV2FetchRemote` in `checkpoint_remote`.. maybe we want to export that and reuse it here so we don't drift / have to maintain it in two places

cmd/entire/cli/checkpoint/v2_store.go
```diff
@@ -35,10 +40,16 @@ func (s *V2GitStore) maxCheckpoints() int {
 }
 
 // NewV2GitStore creates a new v2 checkpoint store backed by the given git repository.
-func NewV2GitStore(repo *git.Repository) *V2GitStore {
+// fetchRemote is the git remote used for fetch-on-demand operations (e.g., fetching
+// /full/* refs during entire resume). Pass "origin" or the checkpoint remote URL.
+func NewV2GitStore(repo *git.Repository, fetchRemote string) *V2GitStore {
+	if fetchRemote == "" {
+		fetchRemote = "origin"
+	}
```
computermode, line 48: Same comment here, re: duplicating some logic

computermode: (I'm finding these because I wanted to make sure we were still supporting external repos for storing the checkpoints.. looks like we are if I am understanding the logic correctly)

pfleidi: Yes, we should be supporting external repos with this 👍

computermode (approved):

pfleidi:
Okay, validated that the push works locally:

```bash
pfleidi-test-repo on  main [?⇡] on ☁️
❯ git ls-remote origin 'refs/entire/checkpoints/v2/*'
b12c390bbf1a96e416e07daa4c12cb07d38f6593        refs/entire/checkpoints/v2/full/current
9682b6f922a59a40d7330314d2ffbcf4db46f6a9        refs/entire/checkpoints/v2/main
```

PR #825, opened 2026-04-01
feat: add nightly releases via GoReleaser and Homebrew tap

gtrrz-victor:
## Summary
- Add nightly cron workflow (`.github/workflows/nightly.yml`) that creates daily prerelease tags when new commits exist
- Nightly tags bump the patch version: stable `v0.5.4` → nightly `v0.5.5-nightly.202604061200.<commit>`
- Add `entire@nightly` Homebrew cask with conditional `skip_upload` using GoReleaser's `.Prerelease` template variable
- Add `conflicts` between `entire` and `entire@nightly` casks to prevent both being installed
- Add `release.prerelease: auto` so nightly GitHub releases are marked as prereleases
- Update release workflow to generate release notes from git log for prerelease tags (stable releases still require CHANGELOG entry)
- Skip Discord announcements for nightly releases
- Add independent version check channels: nightly users get notified about newer nightlies, stable users about newer stable releases

### How it works
- Nightly workflow creates a tag → existing release workflow picks it up automatically
- `.Prerelease` routes uploads: stable tags → `entire` cask, nightly tags → `entire@nightly` cask
- Casks conflict with each other — users must uninstall one before installing the other
- Version check uses `/releases/latest` for stable, `/releases` (filtered) for nightly

```
brew install --cask entireio/tap/entire          # stable
brew install --cask entireio/tap/entire@nightly   # nightly
```

### Version scheme
```
Stable:  v0.5.4                                      → entire cask
Nightly: v0.5.5-nightly.YYYYMMDDhhmm.<commit>        → entire@nightly cask
```

### Note
The `scripts/install.sh` curl-based installer does not support the nightly channel.

## Test plan
- [x] Verify goreleaser config parses: `goreleaser release --snapshot --clean`
- [x] Test nightly tag creation with manual `workflow_dispatch`
- [x] Verify stable release still works (skip_upload doesn't affect non-prerelease tags)
- [x] Confirm nightly cask appears in `entireio/homebrew-tap` after first nightly run
- [x] Run version check tests: `go test ./cmd/entire/cli/versioncheck/ -v`

### DIFF
--- a/cmd/entire/cli/versioncheck/versioncheck_test.go
+++ b/cmd/entire/cli/versioncheck/versioncheck_test.go
@@ -42,9 +42,9 @@ func TestIsOutdated(t *testing.T) {
 		{"1.0.0-dev-xxx", "1.0.1", false, "dev build skips version check"},
 
 		// Nightly-vs-nightly comparisons (same channel)
-		{"0.5.3-nightly.20260405.abc1234", "0.5.3-nightly.20260406.def5678", true, "older nightly is outdated by newer nightly"},
-		{"0.5.3-nightly.20260406.def5678", "0.5.3-nightly.20260405.abc1234", false, "newer nightly is not outdated by older nightly"},
-		{"0.5.3-nightly.20260406.abc1234", "0.5.3-nightly.20260406.abc1234", false, "same nightly is not outdated"},
+		{"0.5.3-nightly.202604051159.abc1234", "0.5.3-nightly.202604061200.def5678", true, "older nightly is outdated by newer nightly"},
+		{"0.5.3-nightly.202604061200.def5678", "0.5.3-nightly.202604051159.abc1234", false, "newer nightly is not outdated by older nightly"},
+		{"0.5.3-nightly.202604061200.abc1234", "0.5.3-nightly.202604061200.abc1234", false, "same nightly is not outdated"},
 	}
 
 	for _, tt := range tests {
@@ -63,8 +63,8 @@ func TestIsNightly(t *testing.T) {
 		version string
 		want    bool
 	}{
-		{"0.5.3-nightly.20260406.abc1234", true},
-		{"v0.5.3-nightly.20260406.abc1234", true},
+		{"0.5.3-nightly.202604061200.abc1234", true},
+		{"v0.5.3-nightly.202604061200.abc1234", true},
 		{"1.0.0", false},
 		{"v1.0.0", false},
 		{"1.0.0-rc1", false},
@@ -85,8 +85,8 @@ func TestFetchLatestNightlyVersion(t *testing.T) {
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
 		releases := []GitHubRelease{
 			{TagName: "v0.5.4", Prerelease: false},
-			{TagName: "v0.5.4-nightly.20260406.abc1234", Prerelease: true},
-			{TagName: "v0.5.4-nightly.20260405.def5678", Prerelease: true},
+			{TagName: "v0.5.4-nightly.202604061200.abc1234", Prerelease: true},
+			{TagName: "v0.5.4-nightly.202604051159.def5678", Prerelease: true},
 		}
 		w.Header().Set("Content-Type", "application/json")
 		//nolint:errcheck // test helper
@@ -102,8 +102,8 @@ func TestFetchLatestNightlyVersion(t *testing.T) {
 	if err != nil {
 		t.Fatalf("fetchLatestNightlyVersion() error = %v", err)
 	}
-	if version != "v0.5.4-nightly.20260406.abc1234" {
-		t.Errorf("fetchLatestNightlyVersion() = %q, want v0.5.4-nightly.20260406.abc1234", version)
+	if version != "v0.5.4-nightly.202604061200.abc1234" {
+		t.Errorf("fetchLatestNightlyVersion() = %q, want v0.5.4-nightly.202604061200.abc1234", version)
 	}
 }
 
--- a/scripts/create-nightly-tag.sh
+++ b/scripts/create-nightly-tag.sh
@@ -4,7 +4,7 @@ set -euo pipefail
 # Calculate the next nightly release tag based on the latest stable tag and HEAD.
 # Outputs the tag name to stdout, or exits 0 silently if a nightly already exists for HEAD.
 #
-# Tag format: v<major>.<minor>.<patch+1>-nightly.<YYYYMMDD>.<short-commit>
+# Tag format: v<major>.<minor>.<patch+1>-nightly.<YYYYMMDDhhmm>.<short-commit>
 # Usage: scripts/create-nightly-tag.sh
 
 SHORT_COMMIT=$(git rev-parse --short HEAD)
@@ -29,7 +29,7 @@ PATCH=$(echo "$LATEST_STABLE" | sed 's/^v//' | cut -d. -f3)
 NEXT_PATCH=$((PATCH + 1))
 NIGHTLY_VERSION="v${MAJOR}.${MINOR}.${NEXT_PATCH}"
 
-DATE=$(date -u +%Y%m%d)
+DATE=$(TZ=UTC0 date +%Y%m%d%H%M)
 TAG="${NIGHTLY_VERSION}-nightly.${DATE}.${SHORT_COMMIT}"
 
 # Skip if this exact tag already exists


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,46 @@
...
+          fi
+
+          LATEST_STABLE=$(git describe --tags --abbrev=0 --match 'v[0-9]*' --exclude '*-*' 2>/dev/null)
+          if [ -z "$LATEST_STABLE" ]; then
+            echo "::error::No stable version tag found"
+            exit 1
+          fi
+
+          DATE=$(date -u +%Y%m%d)
+          TAG="${LATEST_STABLE}-nightly.${DATE}"
```
pjbgf: Shall we also add the short hash to it?

```suggestion
          TAG="${LATEST_STABLE}-nightly.${DATE}.${SHORT_COMMIT_ID}"
```

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,46 @@
...
+          LAST_NIGHTLY=$(git tag -l 'v*-nightly.*' --sort=-v:refname | head -1)
+          if [ -n "$LAST_NIGHTLY" ]; then
+            COMMITS=$(git rev-list "${LAST_NIGHTLY}..HEAD" --count)
+            if [ "$COMMITS" -eq 0 ]; then
+              echo "No new commits since ${LAST_NIGHTLY}, skipping."
+              exit 0
+            fi
+          fi
+
+          LATEST_STABLE=$(git describe --tags --abbrev=0 --match 'v[0-9]*' --exclude '*-*' 2>/dev/null)
```
pjbgf: Not sure if we want current stable or current stable +1 patch. Because:

`v0.5.0-nightly < v0.5.0 < v0.5.1`

So if the current release is `v0.5.0`, semver wise I think the nightly would be `v0.5.1-nightly...` so that:

`v0.5.0 < v0.5.1-nightly < v0.5.1`

.goreleaser.yaml
pjbgf, line 4: I think for nightly:

```
prerelease: '{{ if .Prerelease }}true{{ end }}'
```

gtrrz-victor: this instead ? 
`prerelease: auto`

pjbgf: SGTM. :+1:

gtrrz-victor:
> **Note:** The `scripts/install.sh` curl-based installer does not support the nightly channel. Nightly builds are only distributed via Homebrew cask (`brew install --cask entireio/tap/entire@nightly`). If we want install.sh support for nightlies in the future, it would need to be updated to resolve nightly tags.

.github/workflows/nightly.yml
gtrrz-victor: Bump patch version on nightly, so.
Last stable: 0.5.3
Next nightly: 0.5.4-nightly......

gtrrz-victor:
**Fix: nightly tag push won't trigger release workflow**

- The default `GITHUB_TOKEN` cannot trigger other workflows (GitHub anti-recursion policy)
- Added GitHub App token generation (`actions/create-github-app-token`) to `nightly.yml`
- Passed the token to `actions/checkout` so `git push` uses the App identity
- Tag push from the App token will now correctly trigger `release.yml`

> Note: the App needs `contents: write` on the `cli` repo — verify this is configured.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,46 @@
...
+          TAG="${LATEST_STABLE}-nightly.${DATE}"
+
+          if git rev-parse "${TAG}" >/dev/null 2>&1; then
+            echo "Tag ${TAG} already exists, skipping."
+            exit 0
+          fi
+
+          echo "Creating nightly tag: ${TAG}"
+          git tag "${TAG}"
+          git push origin "${TAG}"
```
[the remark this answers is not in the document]
gtrrz-victor: Good catch. Generating token with write access.

gtrrz-victor:
**Refactor: split release notes into separate steps**

- Stable and nightly release notes are now generated in independent GitHub Actions steps with `if:` conditionals
- Nightly changelog only shows commits since the last nightly tag (falls back to latest stable if first nightly)
- Fixed `git describe` to use `HEAD` instead of `HEAD^` per review feedback

gtrrz-victor:
**Added `mise run release vX.X.X` task**

- Creates stable tag + nightly tag on the same commit using `scripts/create-nightly-tag.sh`
- Skips nightly gracefully if one already exists for the commit
- Validates version format and clean working tree before tagging

gtrrz-victor:
**Addressed PR review findings (6 fixes)**

Critical:
- Stable tag now pushed before nightly in `mise-tasks/release` to avoid race
- Removed `2>/dev/null` stderr suppression on nightly script call
- Simplified error handling in `nightly.yml` (`$?` instead of `|| EXIT_CODE=$?`)

Important:
- Removed unnecessary `fetchErr` intermediate in `versioncheck.go`
- Fixed tag glob to `v*-nightly.*.${SHORT_COMMIT}` (anchored on dot)
- Changed `--sort=-v:refname` to `--sort=-creatordate` in `release.yml`
- Added `t.Parallel()` to `TestIsNightly` (fetch tests can't be parallel — they mutate global vars)

cmd/entire/cli/versioncheck/versioncheck_test.go
```diff
@@ -40,6 +40,11 @@ func TestIsOutdated(t *testing.T) {
 		{"1.0.0-rc1", "1.0.0", true, "prerelease in current"},
 		{"1.0.0", "1.0.1-rc1", true, "prerelease in latest is still newer"},
 		{"1.0.0-dev-xxx", "1.0.1", false, "dev build skips version check"},
+
+		// Nightly-vs-nightly comparisons (same channel)
+		{"0.5.3-nightly.20260405.abc1234", "0.5.3-nightly.20260406.def5678", true, "older nightly is outdated by newer nightly"},
+		{"0.5.3-nightly.20260406.def5678", "0.5.3-nightly.20260405.abc1234", false, "newer nightly is not outdated by older nightly"},
+		{"0.5.3-nightly.20260406.abc1234", "0.5.3-nightly.20260406.abc1234", false, "same nightly is not outdated"},
```
pjbgf: Maybe we need an additional case whereby date is the same, but commit isn't - e.g. automated release + dispatched release on the same day.

cmd/entire/cli/versioncheck/versioncheck.go
```diff
@@ -200,6 +205,57 @@ func fetchLatestVersion(ctx context.Context) (string, error) {
...
+	ctx, cancel := context.WithTimeout(ctx, httpTimeout)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, githubReleasesURL+"?per_page=20", nil)
+	if err != nil {
+		return "", fmt.Errorf("creating request: %w", err)
+	}
+
+	req.Header.Set("Accept", "application/vnd.github+json")
+	req.Header.Set("User-Agent", "entire-cli")
```
pjbgf: Maybe?
```suggestion
	req.Header.Set("User-Agent", "entire-cli/"+currentVersion)
```

mise-tasks/release
```diff
@@ -0,0 +1,45 @@
...
+NIGHTLY_TAG=$(scripts/create-nightly-tag.sh) || true
+
+git push origin "$VERSION"
+echo "Pushed tag ${VERSION}"
+
+if [ -n "$NIGHTLY_TAG" ]; then
+  git tag "$NIGHTLY_TAG"
+  echo "Created tag ${NIGHTLY_TAG}"
+  git push origin "$NIGHTLY_TAG"
+fi
```
pjbgf, line 43: I'd probably segregate the official release from the nightly from a mise task perspective. Ideally the nightly would be its own thing.

```suggestion
```

cmd/entire/cli/versioncheck/versioncheck_test.go
gtrrz-victor: uhm, you are right! Let's add more precision to the time stamp. What about 202604062459 (YYYYMMDDHHmm) ? So we support up to one version per minute

mise-tasks/release
gtrrz-victor: Let's release both stable and nightly when we cut a stable release.

cmd/entire/cli/versioncheck/versioncheck_test.go
gtrrz-victor: Done

PR #826, opened 2026-04-01
Avoid dangling manual checkpoints when content detection fails

sanogueralorenzo:
## Summary
- skip manual commit linking when session content detection errors instead of failing open
- add a regression test covering a corrupt shadow-branch ref during prepare-commit-msg

## Problem
`prepare-commit-msg` currently adds an `Entire-Checkpoint` trailer whenever `filterSessionsWithNewContent` hits an error checking a session. That fail-open path means a stale or corrupt session can still mark a manual commit as linked even though post-commit condensation cannot resolve the session into checkpoint metadata. The result is a commit that appears linked on entire.io but opens to an empty checkpoint page.

## Fix
This change makes content detection fail closed for manual commit linking. If `sessionHasNewContent` errors for a session, we log it and skip that session instead of attaching a trailer. That preserves the safer invariant for terminal commits: only link the commit when the session can actually be resolved into checkpoint content.

## Testing
- added `TestShadowStrategy_PrepareCommitMsg_SkipsSessionWhenContentCheckFails`
- not run locally in this environment (`go`/`gofmt` are not installed here)

### DIFF
--- a/cmd/entire/cli/agent/opencode/hooks_test.go
+++ b/cmd/entire/cli/agent/opencode/hooks_test.go
@@ -134,7 +134,7 @@ func TestInstallHooks_SessionStartIsGuardedBySessionSwitch(t *testing.T) {
 	if assignIdx == -1 {
 		t.Fatalf("plugin file missing current session assignment %q", currentSessionAssignment)
 	}
-	if !(guardIdx < hookIdx && hookIdx < assignIdx) {
+	if guardIdx >= hookIdx || hookIdx >= assignIdx {
 		t.Fatalf("expected guarded session-start call before session assignment, got guard=%d hook=%d assignment=%d",
 			guardIdx, hookIdx, assignIdx)
 	}


peyton-alt:
Thanks for your contribution! The fail-closed change is the right fix and OpenCode plugin fix also looks correct.

One issue: the test uses `assert.False` and `assert.Equal` but the file only imports `testify/require`, not `testify/assert`. This causes a build failure. Either add the import or switch to `require.False/require.Equal`.

We can fix this on our end if you'd prefer, let us know!

Also for future contributions, please open an issue first before creating a PR so we can discuss the approach. Appreciate it!

sanogueralorenzo:
Thanks @peyton-alt , I opened the PR to provide more context on a possible solution but will open issues first moving forward

cmd/entire/cli/strategy/manual_commit_hooks.go
```diff
@@ -1283,12 +1283,10 @@ func (s *ManualCommitStrategy) filterSessionsWithNewContent(ctx context.Context,
 		}
 		hasNew, err := s.sessionHasNewContent(ctx, repo, state, contentCheckOpts{stagedFiles: stagedFiles})
 		if err != nil {
-			logging.Debug(logCtx, "filterSessionsWithNewContent: error checking session, including it (fail-open)",
+			logging.Debug(logCtx, "filterSessionsWithNewContent: error checking session, skipping it",
 				slog.String("session_id", state.SessionID),
 				slog.String("error", err.Error()),
 			)
-			// On error, include the session (fail open for hooks)
-			result = append(result, state)
```
sanogueralorenzo, line 1291: @peyton-alt this was a bit confusing, seems that the initial decision of leaving the sessions open for hooks on fail was intentional, isn't my change reverting some previous decision?

peyton-alt: Hey, yeah it is a bit. It wasn't really intentional, just cautious default. Your fix is correct, better to skip the trailer than attach one that points to nothing.

peyton-alt (approved):
Approved

PR #832, opened 2026-04-01
fix: make OpenCode transcript export resilient to stdout truncation

pfleidi:
## Summary
- fix OpenCode transcript refresh to export directly to a temp file instead of capturing `opencode export` stdout in-memory
- keep JSON validation in place by reading the written temp file and rejecting invalid export content with debug context
- add a regression test that simulates invalid JSON when output is piped and valid JSON when redirected, proving the integration now uses the redirected path

## Problem
OpenCode sessions with large exports were intermittently producing truncated JSON (often at 65536 bytes) when consumed via subprocess stdout capture.

In our integration, this caused transcript refresh failures during condensation (`opencode export returned invalid JSON`), which then prevented transcripts from being attached/updated on checkpoints.

## Root Cause
`fetchAndCacheExport` called `opencode export` and read stdout directly (`cmd.Output` path). That path is vulnerable to the upstream truncation behavior tracked in anomalyco/opencode issues (e.g. #14948), while shell redirection to a file is stable.

## Fix
- replace stdout capture for OpenCode export with direct redirection to the expected temp transcript file:
  - added `runOpenCodeExportToFile(ctx, sessionID, outputPath)`
  - updated `fetchAndCacheExport` to always use that path
- validate the on-disk export JSON before returning so callers keep the same safety checks

## Validation
- `go test ./cmd/entire/cli/agent/opencode -count=1`
- `go build ./...`
- `go vet ./...`
- `mise run test:e2e --agent opencode` (49 passed, 1 expected skip)

## Notes

This change intentionally avoids relying on piped stdout capture for OpenCode export in our integration, matching the behavior that remains reliable in upstream reports (`opencode export ... > file`).

Potential fix for https://github.com/entireio/cli/issues/768

---

> [!NOTE]
> **Medium Risk**
> Changes how OpenCode transcripts are exported and cached (switching from in-memory stdout capture to on-disk redirection), which could affect transcript refresh behavior and error handling across sessions.
> 
> **Overview**
> Fixes intermittent truncated OpenCode transcript exports by **redirecting `opencode export` stdout directly to the `.entire/tmp/<session>.json` file** instead of capturing stdout in memory.
> 
> `fetchAndCacheExport` now ensures the temp directory exists, calls the new `runOpenCodeExportToFile` (injectable via `runOpenCodeExportToFileFn` for tests), then reads the file back to **validate JSON** and logs debug metadata (byte count + path) on invalid output. Adds a unit test that stubs the export-to-file function and asserts the cached transcript is written and valid JSON.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dac4b4f9c5c4ef8a865947b08851778aafc4d8d1. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/integration_test/testenv.go
+++ b/cmd/entire/cli/integration_test/testenv.go
@@ -987,11 +987,11 @@ func (env *TestEnv) gitCommitWithShadowHooks(message string, simulateTTY bool, f
 	if simulateTTY {
 		// Simulate human at terminal: ENTIRE_TEST_TTY=1 makes hasTTY() return true
 		// and askConfirmTTY() return defaultYes without reading from /dev/tty.
-		prepCmd.Env = append(testutil.GitIsolatedEnv(), "ENTIRE_TEST_TTY=1")
+		prepCmd.Env = env.gitHookEnv("ENTIRE_TEST_TTY=1")
 	} else {
 		// Simulate agent: ENTIRE_TEST_TTY=0 makes hasTTY() return false,
 		// triggering the fast path that adds trailers for ACTIVE sessions.
-		prepCmd.Env = append(testutil.GitIsolatedEnv(), "ENTIRE_TEST_TTY=0")
+		prepCmd.Env = env.gitHookEnv("ENTIRE_TEST_TTY=0")
 	}
 	if output, err := prepCmd.CombinedOutput(); err != nil {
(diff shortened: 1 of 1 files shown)


pfleidi:
Bugbot run

PR #840, opened 2026-04-03
fix: add tree hash and patch ID checkpoint linkage fallback

peyton-alt:
## Summary
Store a narrow, Git-native linkage block in checkpoint metadata so the web has limited fallback signals when history rewrites strip the `Entire-Checkpoint` trailer.

- add checkpoint-level `linkage` metadata with `tree_hash` and `patch_id`
- keep the existing active-session revert/cherry-pick trailer behavior, but only reuse an existing `LastCheckpointID`
- avoid generating new checkpoint IDs during sequence operations, since those commits are not condensed immediately

## Scope
This PR is intentionally narrow.

It does **not** implement full automatic relinking across all rewrite scenarios, and it does **not** add custom derived signals like `files_changed_hash` or `session_files_hash`.

The supported fallback signals are:
- `tree_hash` for message-only rewrites where the tree is unchanged
- `patch_id` for clean rebases or cherry-picks where the diff content is unchanged

## Why this shape
The broader multi-signal design implied server-side indexing and reconciliation work that this PR does not provide.

This branch keeps only the cheap, Git-native fallback signals that are defensible today, while the real active-session rewrite tracking work moves to the CLI (`post-rewrite`) in a separate follow-up.

## Sequence-operation behavior
`prepare-commit-msg` can still add a trailer during agent-driven revert/cherry-pick when a session is ACTIVE, but it now only reuses an existing `LastCheckpointID`.

That matters because `post-commit` still skips condensation while the sequence operation is in progress. Reusing an existing checkpoint ID preserves a valid link; minting a fresh one would leave a dangling `Entire-Checkpoint` trailer on the commit.

## Testing
Passed:
- `GOCACHE=/tmp/go-build go test ./cmd/entire/cli/strategy -run 'TestShadowStrategy_(PrepareCommitMsg_(AgentRevertReusesLastCheckpointID|AgentRevertWithoutCheckpointIDSkipped|UserRevertSkipped)|PostCommit_LinkagePopulated)' -count=1`
- `GOCACHE=/tmp/go-build go test ./cmd/entire/cli/checkpoint ./cmd/entire/cli/gitops -count=1`
- `GOCACHE=/tmp/go-build GOLANGCI_LINT_CACHE=/tmp/golangci-lint golangci-lint run ./cmd/entire/cli/strategy ./cmd/entire/cli/checkpoint ./cmd/entire/cli/gitops`

Included coverage verifies:
- `linkage` is written at the checkpoint summary level
- both `tree_hash` and `patch_id` are populated in the condensation path
- clean rebases keep a stable `patch_id`
- agent-driven revert reuses an existing checkpoint trailer when available
- agent-driven revert skips trailer injection when no existing checkpoint ID can be reused
- user-driven revert still skips trailer injection


Soph (commented):
Claude Review Suggestions:

1. **Misleading field name `IsRebaseInProgress`** at `cmd/entire/cli/strategy/manual_commit_hooks.go:1098` — now also covers cherry-pick and revert. Rename to `IsSequenceOperationInProgress` or add a comment.
2. **PatchID unstable under conflict-resolved rebases** — `git patch-id --stable` changes if the user resolves conflicts during rebase. Add a one-line doc note on `PatchID` in `LinkageMetadata`.
3. **Missing cherry-pick test** for `PrepareCommitMsg` with an active session (revert is covered thoroughly, cherry-pick is symmetric).
4. **Duplicate helpers** in `cmd/entire/cli/gitops/diff_test.go` (`revParse`, `gitCheckout`, `gitRebase`) shadow the new `testutil` helpers added in this branch — maybe replace them?
5. **Document "same ID from two sessions = ambiguous"** in `singleReusableCheckpointID` — the behavior is correct, just non-obvious.

PR #913, opened 2026-04-10
Handle autocrlf in carry-forward detection

Soph:
Fixed a false carry-forward bug in manual-commit shadow branch cleanup when Git line-ending normalization is enabled.

In the failing Codex E2E path, the second edit left `src/main.go` in a CRLF-normalized working-tree form while the committed blob was LF-normalized under `core.autocrlf=true`. Our carry-forward logic compared raw on-disk bytes to the commit blob hash, treated that as a dirty file, and incorrectly recreated a shadow branch after the user had already committed the full change.

This change updates `workingTreeMatchesCommit` to ask Git whether the path is clean before falling back to raw blob hashing. That makes the decision respect Git’s clean/smudge behavior and avoids phantom carry-forward on line-ending-only differences. It also adds a regression test covering the `autocrlf` normalization case.

---

> [!NOTE]
> **Medium Risk**
> Changes carry-forward detection to shell out to `git diff` for cleanliness checks, which can affect behavior across environments (git availability/config) and could alter which files are carried forward in edge cases.
> 
> **Overview**
> Fixes a false-positive carry-forward case by updating `workingTreeMatchesCommit` to first ask Git (`git diff --quiet`) whether a path is clean, instead of relying solely on hashing raw on-disk bytes (which can disagree under `core.autocrlf` and other clean/smudge filters).
> 
> Adds a regression test that enables `core.autocrlf=true` and verifies line-ending-only normalization does not cause files to be treated as having remaining agent changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6f18f100f7cb9d89d7264b6814a7657d5f558cf0. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/strategy/content_overlap.go
+++ b/cmd/entire/cli/strategy/content_overlap.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"time"
 
 	"github.com/entireio/cli/cmd/entire/cli/logging"
 	"github.com/go-git/go-git/v6"
@@ -494,7 +495,7 @@ func filesWithRemainingAgentChanges(
 		// Committed content differs from shadow. Check whether the working tree
 		// still has changes — if clean, the user intentionally replaced the content
 		// and there's nothing left to carry forward.
-		if worktreeRoot != "" && workingTreeMatchesCommit(worktreeRoot, filePath, commitFile.Hash) {
+		if worktreeRoot != "" && workingTreeMatchesCommit(ctx, worktreeRoot, filePath, commitFile.Hash) {
 			logging.Debug(logCtx, "filesWithRemainingAgentChanges: content differs from shadow but working tree is clean, skipping",
 				slog.String("file", filePath),
 				slog.String("commit_hash", commitFile.Hash.String()[:7]),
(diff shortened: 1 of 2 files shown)


gtrrz-victor:
**Follow-up: add 5s timeout to `git diff` in carry-forward detection**

- `workingTreeMatchesCommit` was using `context.Background()` with no timeout — the entire call chain from `PostCommit` passes an unbounded context
- Now threads the caller's `ctx` and wraps it with `context.WithTimeout(ctx, 5*time.Second)`, matching the pattern used by `resolveWorktreeBranchGit` for similar local git operations
- If `git diff` hangs (e.g., misbehaving smudge filter), it will be killed after 5s and fall through to the existing hash comparison

gtrrz-victor:
**Follow-up: distinguish `git diff` exit codes in carry-forward detection**

- Exit code 1 from `git diff --exit-code` means "file is dirty" — now returns `false` immediately instead of falling through to raw blob hashing, which could re-introduce the autocrlf mismatch this PR fixes
- Only falls back to blob hash comparison on actual git failures (exit 128+, timeout, etc.)
- Follows the existing `errors.As(*exec.ExitError)` + `ExitCode()` pattern from `isDisconnected` and `DeleteBranchCLI`

gtrrz-victor (approved):

Soph:
superseded by #2336

PR #918, opened 2026-04-10
Fix GoReleaser using wrong tag for concurrent releases

gtrrz-victor:
## Summary
- Resolve the release tag once as workflow `env.RELEASE_TAG` (manual `workflow_dispatch` `version` input or `github.ref_name` on tag push) so checkout, GoReleaser, and failure notifications all use the same value
- Set `GORELEASER_CURRENT_TAG` to `${{ env.RELEASE_TAG }}` so GoReleaser does not auto-detect the wrong tag when nightly and stable releases are created from the same commit
- Support `workflow_dispatch` with a required `version` input for manual releases

## Test plan
- [ ] Push a nightly and stable tag on the same commit and verify each release uses the correct version
- [ ] Run the workflow manually with `version` set to an existing tag and confirm checkout and artifacts match that tag

---

> [!NOTE]
> **Medium Risk**
> Touches the release pipeline by overriding GoReleaser tag detection, which could impact versioning/published artifacts if misconfigured, but the change is small and isolated to CI.
> 
> **Overview**
> Ensures concurrent tag-based releases (e.g., nightly and stable from the same commit) use the intended version by explicitly setting `GORELEASER_CURRENT_TAG` to `github.ref_name` in the GitHub Actions `release.yml` GoReleaser step.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63945011c288a3c359ab8a02157c12766f24952a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -80,6 +80,7 @@ jobs:
           version: latest
           args: release --clean --release-notes=${{ runner.temp }}/release_notes.md
         env:
+          GORELEASER_CURRENT_TAG: ${{ github.ref_name }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           TAP_GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
           GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,6 +2,11 @@ name: Release
 
 on:
   workflow_dispatch:
+    inputs:
+      version:
+        description: "Release tag (semver with v prefix, e.g. v1.2.3)"
+        required: true
+        type: string
   push:
     tags:
       - "v*"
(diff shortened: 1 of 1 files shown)


pjbgf (dismissed):

gtrrz-victor:
Added `workflow_dispatch` support with a `version` input for manual releases. Unified all tag references through a `RELEASE_TAG` job-level env var that resolves correctly for both push and manual dispatch triggers. Slack failure notifications also updated.

.github/workflows/release.yml
```diff
@@ -110,15 +118,15 @@ jobs:
                   "type": "section",
                   "text": {
                     "type": "mrkdwn",
-                    "text": ":red_circle: *Release Failed* for `${{ github.ref_name }}`\n\n<${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|View run details>"
+                    "text": ":red_circle: *Release Failed* for `${{ github.event_name == 'workflow_dispatch' && github.event.inputs.version || github.ref_name }}`\n\n<${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|View run details>"
```
pjbgf: Maybe:

```suggestion
                    "text": ":red_circle: *Release Failed* for `${{ env.RELEASE_TAG }}`\n\n<${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|View run details>"
```

gtrrz-victor: +1

gtrrz-victor: done

pjbgf (approved):
In the future we could add some validation - if the tag does not align with a given pattern it stops early. 

I think we have some sort of validation on the script, but it would be good a fail safe earlier.

PR #922, opened 2026-04-10
Clean up condensation logic: type-enforced redaction boundaries

pfleidi:
## Problem

The condensation path in `CondenseSession()` had three issues:

1. **Redundant redaction** — The same transcript was redacted up to 3 times during a v2 condensation: once in `CondenseSession` for the compact transcript input, once in `GitStore.writeTranscript` for v1, and once in `V2GitStore.writeTranscriptBlobs` for v2 full transcript.

2. **Scattered v2 checks** — `settings.IsCheckpointsV2Enabled(ctx)` was checked in `CondenseSession`, again redundantly inside `compactTranscriptForV2`, and in `writeCommittedV2IfEnabled`.

3. **Implicit redaction contracts** — The `compact` package and checkpoint stores accepted raw `[]byte` transcripts with no indication of whether redaction had been applied. The expectation that data should be redacted was implicit and undocumented.

## Solution

Introduce `redact.RedactedBytes` — an opaque struct type (unexported `data` field) that can only be produced by `redact.JSONLBytes()` or the escape-hatch `redact.AlreadyRedacted()` for trusted sources. This establishes compile-time contracts at package boundaries:

- **`compact.Compact`** accepts `RedactedBytes`, enforcing that only redacted transcripts are compacted
- **`WriteCommittedOptions.Transcript`** and **`UpdateCommittedOptions.Transcript`** use `RedactedBytes`, documenting the store contract
- **`replaceTranscript`** and **`writeTranscriptBlobs`** accept `RedactedBytes` at the function parameter level, preventing unredacted writes to permanent storage

### Key changes

- **`CondenseSession`** redacts the transcript once early, then passes the single `RedactedBytes` result to both the compact path and the checkpoint stores — eliminating 2 redundant redaction passes
- **Store write paths** (`writeTranscript`, `writeTranscriptBlobs`, `replaceTranscript`, `UpdateCommitted`) no longer redact internally for in-memory data. The `TranscriptPath` file-read fallback retains its own redaction since data from disk is untrusted
- **`attach.go`** now explicitly redacts before calling `WriteCommitted` (previously relied on store-internal redaction)
- **`compactTranscriptForV2`** accepts `RedactedBytes` and drops the redundant `IsCheckpointsV2Enabled` guard (both callers already check)
- **Hooks finalize path** propagates `RedactedBytes` end-to-end instead of stripping to `[]byte` and re-wrapping

### Safety

If transcript redaction fails in `CondenseSession`, the transcript is dropped rather than persisting potentially sensitive content. The checkpoint metadata (files touched, attributions, etc.) is still written — only the transcript is omitted.

## Test plan

- [x] `mise run fmt && mise run lint` — zero issues
- [x] `mise run test:ci` — all unit + integration tests pass
- [x] `mise run test:e2e:canary` — Vogon E2E tests pass
- [ ] Verify `redact.JSONLBytes` calls in store write paths exist only in `TranscriptPath` fallback blocks
- [ ] Verify `IsCheckpointsV2Enabled` no longer appears in `compactTranscriptForV2`

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes core checkpoint persistence APIs (transcript types and redaction responsibilities) and could affect what transcript data is stored if any call site fails to redact or mishandles the new contract.
> 
> **Overview**
> **Enforces explicit redaction boundaries for checkpoint transcripts** by introducing `redact.RedactedBytes` and requiring callers to pass *pre-redacted* transcript data into checkpoint write/update flows.
> 
> Redaction is moved out of store internals for in-memory transcripts: `WriteCommittedOptions.Transcript`/`UpdateCommittedOptions.Transcript`, transcript replacement, and v2 full-transcript blob writing now operate on `RedactedBytes` as-is, while file-based fallbacks (`TranscriptPath`, subagent transcripts) still redact on read because disk content is treated as untrusted.
> 
> Condensation/finalization paths now redact once (with perf spans) and reuse the result for v1 storage, v2 compact transcript generation, and AI summarization; on redaction failure they **drop the transcript but still write metadata**. Tests/benchmarks/migration/explain/summarize/compact are updated to wrap trusted fixtures and persisted transcripts via `redact.AlreadyRedacted` and to validate the new “store preserves redacted transcript” behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 09c40402e7b32a326913f190e206e6378527c6d9. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pfleidi:
Bugbot run

pfleidi:
Bugbot run

cmd/entire/cli/checkpoint/committed.go
```diff
@@ -629,45 +630,43 @@ func aggregateTokenUsage(a, b *agent.TokenUsage) *agent.TokenUsage {
...
-	}
-
-	// Redact secrets before chunking so content hash reflects redacted content
-	redactStart := time.Now()
-	redactCtx, redactTranscriptSpan := perf.Start(ctx, "redact_transcript")
-	transcript, err := redact.JSONLBytes(transcript)
-	if err != nil {
-		redactTranscriptSpan.RecordError(err)
-		redactTranscriptSpan.End()
-		return fmt.Errorf("failed to redact transcript secrets: %w", err)
```
pfleidi, line 656: The code above was moved to a new helper which does instrumentation etc. ☝️

computermode (dismissed):
I like the explicitness of this and the cleanup work... thank you for looking into this! It was a tricky one 🥲

pfleidi:
Bugbot run

computermode (approved):

Issue #938, opened 2026-04-13
Factory Droid Mission mode: PreToolUse and PostToolUse hooks fail with tool_use_id and schema mismatch


blackgirlbytes:
### Discord community member reported the following:

Original Conversation found [here](<https://discord.com/channels/1468014954689855716/1468014955533176977/1493211000055529552>)

They tried the Mission mode in Droid. The PreToolUse and PostToolUse hooks fail when tasks are delegated to Worker (subagent) inside the session, failed to capture pre-task state and to parse hook event

```
PreToolUse: failed to capture pre-task state: tool_use_id is required
PostToolUse: failed to parse hook event: failed to parse hook input: json: cannot unmarshal string into Go struct field postToolHookInputRaw.tool_response of type struct { AgentID string "json:\"agentId\"" }
```

### Steps to reproduce

1. Use Droid 0.99.0 in Mission mode
2. Delegate a task to a Worker (subagent)
3. Observe hook failures

### Environment

* Droid 0.99.0
* Entire CLI 0.5.4 (746a74cd)
* Go version: go1.26.2
* Fedora 43 via Homebrew
* Entire hook installed Mar 24

## Related

[ENCLI-174](https://linear.app/entirehq/issue/ENCLI-174/factory-droid-factoryai-droid-hooks-fail-with-schema-mismatch-on) and [ENCLI-175](https://linear.app/entirehq/issue/ENCLI-175/drod-cli-failed-to-capture-pre-task-state-tool-use-id-is-required)

<img src="https://github.com/user-attachments/assets/c63ca24a-380e-49f1-8fe2-d3674d81a401 " alt="Image" width="1074" data-linear-height="765" />

<img src="https://github.com/user-attachments/assets/bfa0182d-7b0a-4491-96c2-e282cfaf3f64 " alt="Image" width="749" data-linear-height="436" />

blackgirlbytes:
I believe this is fixed by #942

Issue #930, opened 2026-04-12
DROD CLI : failed to capture pre-task state: tool_use_id is required


SvenMeyer:
### What happened?

I am getting quite some error with factory.ai DROID CLI now.
```
$ entire --version
Entire CLI 0.5.4 (746a74cd)
Go version: go1.26.2
OS/Arch: linux/amd64
```

### Entire CLI version

0.5.4

### OS and architecture

ArchLinux / Manjaro KDE

### Agent

DOID CLI

### Terminal

KDE Konsole

### Logs / debug output

```bash
HOOKS  PreToolUse → Task
 └─ Script: entire hooks factoryai-droid pre-tool-use
 └─ Result : Exit code: 1
   stderr:
   failed to capture pre-task state: tool_use_id is required

 HOOKS  PreToolUse → Task
 └─ Script: entire hooks factoryai-droid pre-tool-use
 └─ Result : Exit code: 1
   stderr:
   failed to capture pre-task state: tool_use_id is required

   can you write the plan to `docs-dev/tasks/v0.2.2-openfort-implementation-with-onboarding.md` and    
   mark `docs-dev/tasks/v0.2.1-openfort-implementation-part-2.md` as superseeded                       
                                                                                                       
   then commit and push it all                                                                         

 HOOKS  PostToolUse → Task
 └─ Script: entire hooks factoryai-droid post-tool-use
 └─ Result : Exit code: 1
   stderr:
   failed to parse hook event: failed to parse hook input: json: cannot unmarshal string into Go struct
    field postToolHookInputRaw.tool_response of type struct { AgentID string "json:\"agentId\"" }
```

### Additional context

I had no problems with v0.5.3 afaik

rohitg00:
I'd like to pick this up if no one is on it yet.

Quick diagnosis from reading the v0.5.3...v0.5.4 diff in `cmd/entire/cli/agent/factoryaidroid/`:

**Bug 1 — `tool_use_id is required` on PreToolUse**

`parseSubagentStart` in `lifecycle.go` does fall back via `registerFallbackToolUseID` when `tool_use_id` is empty in the hook payload, so `event.ToolUseID` should be populated before reaching `CapturePreTaskState`. The fact that the error still fires suggests either:

- DROID is sending a `tool_use_id` key as a non-string type (e.g. `null` or numeric) so it decodes to `""` and the fallback path runs but emits empty, or
- The PreToolUse handler is being invoked for non-Task tools where the fallback is wired only on the Task path.

[...]

PR #942, opened 2026-04-13
Fix Factory fallback tool use IDs

### LINKED ISSUE #917: Factory Droid (factoryai-droid) hooks fail with schema mismatch on PreToolUse and PostToolUse — severely impacts Mission workflow
## What happened?

Entire's `factoryai-droid` hooks for `PreToolUse` and `PostToolUse` fail with exit code 1 on **every** invocation. This produces noisy error output in the Droid CLI on every tool call while Entire receives no data.

**PreToolUse error:**
```
HOOKS  PreToolUse → Task
 └─ Script: entire hooks factoryai-droid pre-tool-use
 └─ Result : Exit code: 1
   stderr:
   failed to capture pre-task state: tool_use_id is required
```

**PostToolUse error:**
```
HOOKS  PostToolUse → Task
 └─ Script: entire hooks factoryai-droid post-tool-use
 └─ Result : Exit code: 1
   stderr:
   failed to parse hook event: failed to parse hook input: json: cannot unmarshal string into Go struct field postToolHookInputRaw.tool_response of type struct { AgentID string "json:\"agentId\"" }
```

## Impact on Droid Missions

This is especially painful during **Droid Missions** — Factory's orchestrated multi-subagent workflow feature. Missions spawn multiple parallel `Task` (subagent) tool calls, and **each one triggers the failing PreToolUse hook**. With a mission running 5-10+ subagents, this means:

- 10-20+ failing hook executions per mission cycle
- Each failing hook adds latency to the already-parallelized workflow
- The cumulative slowdown dramatically degrades the automated/agentic experience
- The constant error output makes it hard to follow what the mission is actually doing

For context, a typical mission cycle looks like:

```
HOOKS  Stop
 └─ Script: entire hooks factoryai-droid stop
 └─ Result : Exit code: 0

HOOKS  PreToolUse → Task
 └─ Script: entire hooks factoryai-droid pre-tool-use
 └─ Result : Exit code: 1    ← FAILS

HOOKS  PreToolUse → Task
 └─ Script: entire hooks factoryai-droid pre-tool-use
 └─ Result : Exit code: 1    ← FAILS

HOOKS  PreToolUse → Task
 └─ Script: entire hooks factoryai-droid pre-tool-use
 └─ Result : Exit code: 1    ← FAILS
... (repeats for every subagent)
```

## Root cause

Two schema mismatches between what Entire expects and what Factory Droid provides in hook input JSON:

1. **Missing `tool_use_id`**: Entire's `pre-tool-use` handler requires a `tool_use_id` field, but Droid's PreToolUse hook input does not include it. This is the same issue documented for Claude Code at anthropics/claude-code#13241 — `tool_use_id` was never reliably sent by the agent platform.

2. **`tool_response` type mismatch**: Entire expects `tool_response` to be a structured object with an `agentId` field (`struct { AgentID string }`), but Droid sends `tool_response` as a plain string (the raw tool output text). This causes the Go JSON unmarshal error.

## Droid's actual hook input format (v0.98.0)

Per Factory docs (https://docs.factory.ai/reference/hooks-reference), Droid sends:

**PreToolUse:**
```json
{
  "session_id": "...",
  "transcript_path": "...",
  "cwd": "...",
  "permission_mode": "...",
  "hook_event_name": "PreToolUse",
  "tool_name": "Task",
  "tool_input": { ... }
}
```
No `tool_use_id` field.

**PostToolUse:**
```json
{
  "session_id": "...",
  "transcript_path": "...",
  "cwd": "...",
  "permission_mode": "...",
  "hook_event_name": "PostToolUse",
  "tool_name": "Task",
  "tool_input": { ... },
  "tool_response": "<raw string output, not a structured object>"
}
```
`tool_response` is a string, not an object with `agentId`.

## Steps to reproduce

1. Install Entire CLI v0.5.3
2. In a git repo, run `entire enable --agent factoryai-droid`
3. Start Droid in that repo: `droid`
4. Ask Droid to perform any tool call (e.g. read a file)
5. Observe PreToolUse and PostToolUse hook errors in Droid output
6. Run a Droid Mission (multi-subagent workflow) to see the amplified impact

## Entire CLI version

0.5.3

## OS and architecture

macOS ARM64 (darwin/arm64)

## Agent

Factory Droid (factoryai-droid)

## Suggested fix

1. Make `tool_use_id` optional in the PreToolUse parser — generate a fallback ID or skip the field when absent
2. Handle `tool_response` as either a string or a structured object (union type / `json.RawMessage`) rather than requiring the `agentId` struct
3. Alternatively, correlate tool calls via `tool_name` + `session_id` + timestamp instead of requiring `tool_use_id`
4. Consider graceful degradation: if a hook fails to parse, continue with partial data rather than exiting with code 1 on every invocation

## Additional context

- Related: #237 (hooks failing with Claude Code for similar schema reasons)
- Related: #750 (cannot unmarshal Go struct — same class of JSON parsing error)
- Related: #859 (Droid agent shows as 'Unknown' — may be downstream of hooks failing to capture data)
- Anthropic-side root cause: anthropics/claude-code#13241 (`tool_use_id` not sent by agent platforms)
- Droid hook docs: https://docs.factory.ai/reference/hooks-reference
- Workaround: Remove all Entire hooks from `.factory/settings.json` project settings (which is what I have had to do to make Missions usable)

gtrrz-victor:
## Summary
Fix Factory AI Droid fallback `tool_use_id` handling so repeated Worker runs with identical inputs do not collide while pre/post task hooks still correlate correctly.

## What Changed
- replaced the deterministic fallback-only path with a per-invocation fallback `tool_use_id` generated during `PreToolUse`
- added a small Factory-only pending fallback state file under `.entire/tmp` so `PostToolUse` can recover and consume the exact ID for the matching Worker invocation when Factory omits `tool_use_id`
- threaded hook parsing context through the Factory lifecycle parser so the fallback state can be resolved relative to the repo tmp dir
- added focused lifecycle coverage proving repeated identical missing-`tool_use_id` invocations stay unique, correlate in LIFO order, and clean up temporary state

## Verification
- `go test ./cmd/entire/cli/agent/factoryaidroid`
- `mise run test:e2e --agent factoryai-droid TestFactoryTaskHooksDoNotFail`

fixes: https://github.com/entireio/cli/issues/917, https://github.com/entireio/cli/issues/930, https://github.com/entireio/cli/issues/938

### DIFF
--- a/cmd/entire/cli/agent/factoryaidroid/lifecycle.go
+++ b/cmd/entire/cli/agent/factoryaidroid/lifecycle.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/entireio/cli/cmd/entire/cli/agent"
@@ -231,11 +232,15 @@ func (f *FactoryAIDroidAgent) parseSubagentStart(stdin io.Reader) (*agent.Event,
 	if err != nil {
 		return nil, err
 	}
+	toolUseID := raw.ToolUseID
+	if toolUseID == "" {
+		toolUseID = fallbackToolUseID(raw.SessionID, raw.ToolName, raw.ToolInput)
+	}
 	return &agent.Event{
 		Type:       agent.SubagentStart,
 		SessionID:  raw.SessionID,
 		SessionRef: raw.TranscriptPath,
-		ToolUseID:  raw.ToolUseID,
+		ToolUseID:  toolUseID,
 		ToolInput:  raw.ToolInput,
 		Timestamp:  time.Now(),
 	}, nil
@@ -246,16 +251,20 @@ func (f *FactoryAIDroidAgent) parseSubagentEnd(stdin io.Reader) (*agent.Event, e
 	if err != nil {
 		return nil, err
 	}
+	toolUseID := raw.ToolUseID
+	if toolUseID == "" {
+		toolUseID = fallbackToolUseID(raw.SessionID, raw.ToolName, raw.ToolInput)
+	}
 	event := &agent.Event{
 		Type:       agent.SubagentEnd,
 		SessionID:  raw.SessionID,
 		SessionRef: raw.TranscriptPath,
-		ToolUseID:  raw.ToolUseID,
+		ToolUseID:  toolUseID,
 		ToolInput:  raw.ToolInput,
 		Timestamp:  time.Now(),
 	}
-	if raw.ToolResponse.AgentID != "" {
-		event.SubagentID = raw.ToolResponse.AgentID
+	if agentID := parseHookToolResponseAgentID(raw.ToolResponse); agentID != "" {
+		event.SubagentID = agentID
 	}
 	return event, nil
 }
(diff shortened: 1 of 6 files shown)


gtrrz-victor:
**Follow-up: fix lint and architecture test violations**
- Fixed `gosec` (G304) and `wrapcheck` lint errors in `loadFallbackToolUseState`
- Replaced forbidden `cli/testutil` import in `lifecycle_test.go` with inline `git.PlainInit` to satisfy architecture boundary test

gtrrz-victor:
**Follow-up: bytes-based null check in `parseHookToolResponseAgentID`**
- Replaced `string(raw) == "null"` with `bytes.TrimSpace` + `bytes.Equal` to avoid allocation and handle whitespace-padded JSON

Soph (approved):

PR #876, opened 2026-04-08
Add timeout to explain summary generation

peyton-alt:
## Summary
Adds a timeout to checkpoint summary generation for `entire explain --generate` so the command cannot hang indefinitely when the Claude CLI stalls.

## What Changed
- scopes the 30 second timeout to `explain --generate` in `explain.go`
- keeps `summarize.GenerateFromTranscript` generic for other call sites
- preserves shorter parent deadlines and clamps longer parent deadlines to the 30 second safety timeout
- improves Claude CLI cancellation and deadline handling so context cancellation surfaces cleanly
- prints `Generating checkpoint summary...` while `explain --generate` is waiting
- adds unit coverage for default timeout, shorter parent deadline, longer parent deadline, and cancellation behavior

## Why
A stalled Claude summary generation path currently leaves users waiting without a bounded failure mode. This change keeps the fix narrowly scoped to `explain --generate` instead of changing shared summarize behavior for other callers.

Current state before this fix:
`entire explain --generate`
`^Cfailed to generate summary: failed to generate summary: claude CLI failed (exit -1):`

## Verification
- `go test ./cmd/entire/cli -run 'TestGenerateCheckpointAISummary|TestExplainCmd_RejectsPositionalArgs'`
- `go test ./cmd/entire/cli/summarize`
- `mise run lint`
- `mise run test`

PR #964, opened 2026-04-16
feat(explain): live progress and observable-state timeout diagnostic

### LINKED ISSUE #1198: summary_timeout_seconds is ignored by checkpoint explain --generate
## Summary

`summary_timeout_seconds` is accepted by settings, but `entire checkpoint explain --generate` still uses the hardcoded 30s timeout.

This makes the setting look configurable even though it has no effect for the current summary generation path.

## Environment

- Entire CLI: `0.6.1 (774ee21f)`
- OS/Arch: `darwin/arm64`
- Go: `go1.26.2`

I also checked current `main` / nightly `v0.6.2-nightly.202605130712.1622abc5`, and the generate path still appears to use the hardcoded `checkpointSummaryTimeout`.

## Reproduction

With a large checkpoint transcript:

```bash
entire checkpoint explain --checkpoint ebd73f222193 --generate
```

Output:

```text
Using Claude Code for summary generation.
Generating checkpoint summary...
✗ Summary generation timed out after 30s
```

Then configure a longer timeout:

```json
{
  "enabled": true,
  "summary_timeout_seconds": 3600
}
```

Run the same command again:

```bash
entire checkpoint explain --checkpoint ebd73f222193 --generate
```

It still times out after 30s.

## Expected Behavior

`summary_timeout_seconds` should either:

1. be honored by `checkpoint explain --generate`, or
2. be removed/hidden until it is supported.

Ideally:

- unset or non-positive: keep current default behavior
- positive value: use that value as the summary generation deadline

## Actual Behavior

The setting round-trips through config, but `explain --generate` still uses the package-level default:

```go
const defaultCheckpointSummaryTimeout = 30 * time.Second
var checkpointSummaryTimeout = defaultCheckpointSummaryTimeout
```

The settings field exists:

```go
SummaryTimeoutSeconds int `json:"summary_timeout_seconds,omitempty"`
```

and has a helper:

```go
func (s *EntireSettings) SummaryTimeoutValue() time.Duration
```

but that helper does not appear to be consumed by the generate path.

## Related Context

There is an open PR that appears related:

- https://github.com/entireio/cli/pull/964

That PR seems to move toward stream-driven progress and a different deadline model, but it is still open. In the meantime, the released CLI exposes a setting that appears configurable but does not affect the behavior.

## Why This Matters

Large real-world checkpoints can exceed 30 seconds to summarize, especially when the transcript is long. In that case the only workaround is to avoid `--generate` or patch/build the CLI locally.

A small fix could wire `SummaryTimeoutValue()` into the summary deadline selection while preserving the current 30s default.

Happy to open a narrow PR if maintainers prefer that over waiting for #964.

peyton-alt:
https://entire.io/gh/entireio/cli/trails/164

## Summary

Two paired user-facing improvements to `entire explain --generate`, building on PR #1204 (which made the 30s timeout configurable but kept a 5-minute default and didn't change the silent-wait UX):

1. **Live progress on TTY.** Real-time phase events (sending request → first token / TTFT → writing tokens → done) instead of silent wait. In-place line updates on TTY; one line per event in CI/piped; ASCII glyphs under `ACCESSIBLE=1`.
2. **Observable-state timeout diagnostic.** When `--summary-timeout-seconds` fires, the error message reflects what the subprocess was actually doing. Streaming providers (Claude) get phase-specific attribution; non-streaming providers (Codex/Cursor/Copilot/Gemini) surface captured stderr and distinguish "produced no output" from "was generating output when killed". The three-cause shrug from PR #876 is removed.

Plus one policy change that enables both:

3. **Automatic deadline removed.** `--generate` no longer has a 5-minute default cap. Timeouts are opt-in only via `--summary-timeout-seconds` flag or `summary_timeout_seconds` setting (both surfaces from #1204 preserved). With no automatic cap, every diagnostic users see is for a timeout they consciously requested.

## What users see now

**TTY, normal run:**
```
Generating checkpoint summary... (transcript: 40.7 KB, provider: claude-code)
→ Sending request to provider...
→ Provider responded (TTFT 1.6s, 13k cached input tokens) -- generating...
→ Writing summary... (~Nk tokens)             ← updates in place
✓ Summary generated (7.6s, 420 output tokens)
✓ Summary generated for <checkpoint-id>
```

**TTY, network down + `--summary-timeout-seconds 10`:**
```
Generating checkpoint summary... (transcript: 40.7 KB, provider: claude-code)
→ Sending request to provider...
✗ Timed out after 10s: provider sent request but received no response
  cause    network/firewall, provider API degraded, or auth check stuck
  try      check connectivity to the provider, then retry
```

**TTY, transcript too large + `--summary-timeout-seconds 3`:**
```
→ Provider responded (TTFT 1.6s, 20.2k cached input tokens) -- generating...
→ Writing summary... (~24 tokens)
✗ Timed out after 3s: model responded but did not finish
  cause    transcript may be too large for the chosen cap, or model is slow
  try      raise --summary-timeout-seconds or pick a faster model
```

**Non-TTY (`| cat`):** same lines, one per event, no `\r` in-place updates, no ANSI styling (lipgloss auto-strips).

**`ACCESSIBLE=1`:** `→` becomes `->`, no ANSI escapes.

## Architecture

- **`agent.StreamingTextGenerator`** optional capability interface with `AsStreamingTextGenerator` detector. Defined alongside the existing `TextGenerator` interface. Claude Code is the first (currently only) implementer.
- **`(*ClaudeCodeAgent).GenerateTextStreaming`** invokes `claude --print --output-format stream-json --include-partial-messages --verbose`, pipes stdout through a 4 MiB-buffered NDJSON scanner, and dispatches `ProgressPhase` callbacks (`Connecting → FirstToken → Generating(×N) → Done`). Falls back to plain `GenerateText` when stderr indicates the CLI rejected streaming flags (older Claude CLI builds).
- **`summarize.TextGeneratorAdapter`** (the production summary-generator path used by all providers via `resolveCheckpointSummaryProvider`) gained streaming-capability detection and dispatches to `GenerateTextStreaming` when available. `summarize.ClaudeGenerator` got the same treatment for legacy callers.
- **`summaryProgressWriter`** renders progress events using existing `statusStyles` (lipgloss). In-place `\r` updates on TTY; one line per event otherwise; ASCII glyphs in accessible mode. Throttled to 500ms-OR-25%-jump in non-TTY mode to avoid spam.
- **`summaryAttempt`** value type captures observable state during a run (phases reached, captured stderr, stdout byte count) so the diagnostic path can attribute timeouts without depending on the UI struct.
- **`*agent.TextGenerationError`** wraps `RunIsolatedTextGeneratorCLI` failures with captured stderr + stdout byte count, so the diagnostic can surface real subprocess output for non-streaming providers.
- **`providerCLIName`** maps `types.AgentName` to the user-facing CLI binary in diagnostic `try:` rows (`run \`claude\` directly to confirm it works`).

## What changed from the original draft design (in case you remember the old PR)

The original draft (commits before this rebase) had a TTY-gated deadline strategy with an idle-based 5min watchdog + 30min hard wall-clock cap, plus an opt-in `summary_timeout_seconds` override. This redesign drops the automatic deadline entirely — timeouts are pure opt-in via the flag/setting from #1204. Rationale: the bug behind issue #1198 was that the auto-default truncated legitimate generations. Making the deadline opt-in fixes the root cause and removes a layer of mechanism (no idle watchdog, no separate wall cap, no goroutine).

Also intentionally NOT included: a stderr substring classifier for non-streaming providers (would have been "chunk 8" in the original plan). Decision: ship verbatim stderr passthrough rather than speculative phrase-matching without real CLI fixtures. The raw stderr is always surfaced in the diagnostic rows.

## Test plan

- [x] `mise run check` (fmt + lint + unit + integration + E2E Vogon canary) passes locally
- [x] Manual: TTY happy path — progress phases render with in-place updates, final ✓ with duration + token count
- [x] Manual: `--summary-timeout-seconds 3` on a real Claude generation → "model responded but did not finish" with cause + try rows
- [x] Manual: Network offline + `--summary-timeout-seconds 10` → "provider sent request but received no response" (phase-distinguished from "never sent its request")
- [x] Manual: `| cat` → one line per event, no `\r`, no ANSI
- [x] Manual: `ACCESSIBLE=1` → `->` instead of `→`
- [x] Manual: `entire explain --help | grep summary-timeout` → "no automatic deadline applies"

Closes #1198 (issue was already closed at #1204 time; this delivers the diagnostic improvement that #1204 deferred).

---

> [!NOTE]
> **Medium Risk**
> Touches `explain --generate` execution/timeout behavior and adds a new streaming Claude CLI path; failures could affect summary generation UX and error reporting across providers.
> 
> **Overview**
> `entire explain --generate` now reports *live, phase-based progress* during summary generation (TTY in-place updates, non-TTY line output, ASCII in `ACCESSIBLE` mode) and uses an observable-state `summaryAttempt` to produce more actionable timeout messages.
> 
> Adds optional `agent.StreamingTextGenerator` support (capability + detector) and implements Claude Code streaming via `claude --output-format stream-json` with NDJSON parsing, progress callbacks (connecting/TTFT/generating/done), and fallback to non-streaming on older CLIs.
> 
> Removes the implicit 5-minute default deadline for `--generate` (timeouts are now opt-in via flag/setting), and enhances non-streaming provider diagnostics by wrapping CLI errors with captured stderr and stdout byte counts (`TextGenerationError`), plus a shared provider→binary lookup (`SummaryCLIBinaryName`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 252e320a537d4c816b105158e835d5b8e59f7c65. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


peyton-alt:
@bugbot review

peyton-alt:
@bugbot review

gtrrz-victor (approved):

PR #990, opened 2026-04-20
Accept checkpoint ID or commit SHA as positional arg to explain

peyton-alt:
## Summary

- `entire explain <id-or-sha>` now takes a checkpoint ID *or* a commit SHA/ref as a positional argument, so `--generate`/`--raw-transcript` no longer require the user to pick between `-c/--checkpoint` and `--commit`.
- Resolution is checkpoint-first (preserves short-prefix matching for IDs and the existing temp-checkpoint fallback), then falls back to git commit resolution via the `Entire-Checkpoint` trailer.
- `--checkpoint`/`-c` and `--commit` are kept as explicit overrides for backward compatibility and for disambiguating short hex prefixes that could resolve to either.
- Commits without an `Entire-Checkpoint` trailer still print the existing friendly "No associated Entire checkpoint" message and exit cleanly — no behavior change on that path.

## Test plan

- [x] `mise run check` (fmt + lint + unit + integration + E2E canary) passes locally
- [x] Manual simulation against a real repo covering 16 scenarios (success paths, backward-compat flag paths, conflict errors, missing-trailer, gibberish input, `--generate` routing for both forms)
- [x] New `TestExplainCmd_PositionalArgConflictsWithFlags` covers positional + `--checkpoint` / `-c` / `--commit` / `--session` rejection
- [x] Existing `runExplain` call sites updated for the new `target` parameter
- [ ] Reviewer sanity-check that the 7-char-prefix collision tradeoff (checkpoint wins, `--commit` available as override) matches expectations

---

> [!NOTE]
> **Medium Risk**
> Changes the `entire explain` routing and error semantics for checkpoint/commit resolution, including side-effect modes (`--generate`, `--raw-transcript`) that write or output checkpoint data. Risk is moderate due to new ambiguity handling and fallback logic that could affect which checkpoint is targeted.
> 
> **Overview**
> `entire explain` now accepts an optional positional `checkpoint-id | commit-sha` argument and auto-resolves it by trying checkpoint IDs (including temp checkpoints) first, then falling back to resolving a git commit and following its `Entire-Checkpoint` trailer.
> 
> Side-effect modes are tightened: `--generate`/`--raw-transcript` are allowed with positional targets and `--commit`, refuse ambiguous short targets when `--generate` is set, and error (instead of exiting 0) when a resolved commit has no trailer. Checkpoint-not-found now returns a typed `checkpoint.ErrCheckpointNotFound` wrapper to enable reliable fallback detection, and tests are expanded to cover the new routing/validation cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 733fce210feca07a93269093fb0c225ffbefdb2b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/explain.go
+++ b/cmd/entire/cli/explain.go
@@ -89,30 +89,31 @@ func newExplainCmd() *cobra.Command {
 	var searchAllFlag bool
 
 	cmd := &cobra.Command{
-		Use:   "explain",
+		Use:   "explain [checkpoint-id | commit-sha]",
 		Short: "Explain a session, commit, or checkpoint",
 		Long: `Explain provides human-readable context about sessions, commits, and checkpoints.
 
 Use this command to understand what happened during agent-driven development,
 either for self-review or to understand a teammate's work.
 
-By default, shows checkpoints on the current branch. Use flags to filter or
-explain specific items.
+By default, shows checkpoints on the current branch. Pass a checkpoint ID or
+commit SHA as a positional argument to explain a specific item, or use flags.
+
+Viewing specific items:
+  entire explain <id-or-sha>           Auto-detects checkpoint ID or commit SHA
+  entire explain --checkpoint <id>     Force interpretation as checkpoint ID
+  entire explain --commit <ref>        Force interpretation as commit ref
 
 Filtering the list view:
   --session      Filter checkpoints by session ID (or prefix)
 
-Viewing specific items:
-  --commit       Explain a specific commit (shows its associated checkpoint)
-  --checkpoint   Explain a specific checkpoint by ID
-
-Output verbosity levels (for --checkpoint):
+Output verbosity levels (when explaining a specific item):
   Default:         Detailed view with scoped prompts (ID, session, tokens, intent, prompts, files)
   --short          Summary only (ID, session, timestamp, tokens, intent)
   --full           Parsed full transcript (all prompts/responses from entire session)
   --raw-transcript Raw transcript file (JSONL format)
 
-Summary generation (for --checkpoint):
+Summary generation:
   --generate    Generate an AI summary for the checkpoint
   --force       Regenerate even if a summary already exists (requires --generate)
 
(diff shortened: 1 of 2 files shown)


peyton-alt:
@bugbot review

peyton-alt:
@bugbot review

cmd/entire/cli/explain.go
```diff
@@ -212,6 +252,102 @@ func runExplain(ctx context.Context, w, errW io.Writer, sessionID, commitRef, ch
...
+		slog.String("target", target),
+		slog.String("checkpoint_error", checkpointErr.Error()))
+
+	repo, repoErr := openRepository(ctx)
+	if repoErr != nil {
+		// Composed message beats errors.Join here — the latter renders
+		// two lines (one per error) and users act on the first/stale one.
+		return fmt.Errorf("no checkpoint matched %q, and commit fallback failed: %w", target, repoErr)
+	}
+	hash, resolveErr := repo.ResolveRevision(plumbing.Revision(target))
```
pjbgf: If `target` resolves to multiple hashes, first one wins which could be the wrong commit. Instead, we could run `repo.Storer.(filesystem.ObjectStorage).HashesWithPrefix()` to check all objects with that prefix, and then check which ones are commits. If more than one we error asking for a longer abbreviated commit ID.

pjbgf (dismissed):

Soph (dismissed):
Did a review and added some follow up fixes directly: 

- Reworked the auto-resolution path to reuse a single repo/checkpoint lookup, which removes the duplicate repo open and duplicate committed-checkpoint listing on the `--generate` path.
- Added `logging.Warn` when the `--generate` ambiguity guard has to degrade because repo/checkpoint lookup fails, instead of failing silently.
- Added a regression test to lock the sentinel contract for temporary checkpoints: if a temp checkpoint matches with `--generate`, we now return `errCannotGenerateTemporaryCheckpoint` and do not fall back to commit resolution.
- Added the missing `Use` string assertion and cleaned up the small error-format/readability nits called out in review.

pjbgf (approved):

PR #1002, opened 2026-04-21
Fix SSH signing for 1Password and bare public key setups

pfleidi:
## Summary

- PR #960 introduced checkpoint commit signing via go-git's `auto` signer library. This broke setups where `gpg.ssh.program` is configured (e.g. 1Password's `op-ssh-sign`), because go-git signs natively via the SSH agent protocol while the external program uses a separate mechanism. The signing key typically isn't in the SSH agent, causing warnings on every push.
- Detects `gpg.ssh.program` in the raw git config and skips native signing silently with a debug log. Checkpoint commits will be unsigned in this setup, which is acceptable since signing is best-effort.

## Test plan

- [x] Unit tests for `hasSSHSignProgram` (4 cases: nil, empty, program set, gpg without ssh.program)
- [x] `mise run check` passes (fmt, lint, unit, integration, E2E canary)
- [x] Manual verification: `git push` no longer emits signing warnings with 1Password SSH setup


cmd/entire/cli/objectsigner.go
```diff
@@ -74,6 +85,16 @@ var scopeName = map[config.Scope]string{
...
 }
 
+// hasSSHSignProgram checks whether gpg.ssh.program is set in the raw config.
+// go-git's Config struct does not expose this field, so we read it directly.
+func hasSSHSignProgram(raw *format.Config) bool {
+	if raw == nil {
+		return false
+	}
+
+	return raw.Section("gpg").Subsection("ssh").Option("program") != ""
```
pjbgf: As official per [docs](https://git-scm.com/docs/git-config#Documentation/git-config.txt-gpgformatprogram):
> The default value for ... `gpg.ssh.program` is `"ssh-keygen"`.

```suggestion
    program := raw.Section("gpg").Subsection("ssh").Option("program")
	return (program != "" && program != "ssh-keygen")
```

cmd/entire/cli/objectsigner.go
```diff
@@ -74,6 +85,16 @@ var scopeName = map[config.Scope]string{
 	config.SystemScope: "system",
 }
 
+// hasSSHSignProgram checks whether gpg.ssh.program is set in the raw config.
+// go-git's Config struct does not expose this field, so we read it directly.
+func hasSSHSignProgram(raw *format.Config) bool {
```
pjbgf: Maybe?

```suggestion
func hasCustomSSHSignProgramSet(raw *format.Config) bool {
```

computermode (approved):

pjbgf (approved):

PR #963, opened 2026-04-16
feat: classify Claude CLI errors into actionable user messages

peyton-alt:
## Summary

Makes `entire explain --generate` surface actionable error messages when the
Claude CLI fails, instead of generic "failed to parse summary JSON" output
or silent empty errors.

## Problem

The Claude CLI returns most operational errors (auth, rate limit, invalid
model) at **exit 0** with `is_error: true` in its JSON envelope. The prior
implementation didn't check `is_error`, tried to JSON-parse the error text
as a `Summary`, and exposed users to messages like:

- `failed to parse summary JSON: ... (response: There's an issue with the selected model...)`
- `failed to generate summary: failed to generate summary: claude CLI failed (exit -1):` (double-wrapped)

For CLI crashes with no output (OOM, SIGKILL) users saw literally `Claude failed to generate the summary:` — nothing after the colon.

## What changed

**Typed error classification.** Introduces `*claudecode.ClaudeError` with a `Kind` discriminator (`auth` / `rate_limit` / `config` / `cli_missing` / `unknown`) populated from HTTP status or stderr fallback:

- 401/403 → auth, 429 → rate limit, other 4xx → config
- Envelope with auth phrasing but no status → also classified as auth (handles older CLI builds that omit `api_error_status`)
- Envelope with `is_error:true` + `result:null` → preserved with its HTTP status instead of discarded as a parse error
- Binary not on PATH → `cli_missing`
- Structured errors survive `errors.As` all the way from `claudecode.GenerateText` → `summarize.Generate` → explain

**User-facing message mapping.** `formatCheckpointSummaryError` turns each Kind into an actionable message with remediation:

| Kind | User sees |
|---|---|
| Auth | `Claude authentication failed: <detail>` + `` Run `claude login` and retry `` |
| Rate limit | `Claude rejected the summary request due to rate limits or quota: <detail>` + `Wait and retry` |
| Config | `Claude rejected the summary request: <detail>` + `Check your Claude CLI config and selected model` |
| CLI missing | `Claude CLI is not installed or not on PATH` |
| Timeout | 3-cause diagnostic (transcript size / network / API status) |
| Unknown with no message | Falls back to HTTP status or exit code so output is never empty |

**Settings groundwork.** Adds `summary_timeout_seconds` field and `SummaryTimeoutValue()` accessor. The field is not yet consumed by the generate path; the stacked follow-up PR wires it into TTY-gated deadline selection.

## Scope

All behavior changes are contained to:

[...]

PR #1005, opened 2026-04-22
feat(agent): one typed error for every summary provider (closes #963)

peyton-alt:
https://entire.io/gh/entireio/cli/trails/31103d7bc0c5

## Summary

Replaces per-agent error handling on the summary path with one shared typed
error, so `explain --generate` gives the same actionable message whichever
provider failed.

`agent.TextGenError{Kind, Provider, Message, APIStatus, ExitCode, Cause}`
replaces `claudecode.ClaudeError`. All six providers (Claude, Codex, Gemini,
Cursor, Copilot CLI, Pi) plus external plugins return it; `renderTextGenError`
maps (Kind × Provider) to the user-facing label and remediation row.

**Scope grew twice after this PR was opened — see History below.** It is now
larger than any recently merged PR in this repo, so the reviewing notes at the
bottom flag where the real risk sits.

## What the user sees

| Scenario | Before | After |
|---|---|---|
| Stale key, Claude, streaming (the default path) | `claude stream failed: Invalid API key · Please run /login: exit status 2` | `Claude authentication failed` / `try: run \`claude login\` and retry` |
| Auth failure, non-Claude | raw stderr, no label | `Codex authentication failed` + stderr verbatim |
| Binary absent | `exec: "agent": executable file not found in $PATH` | `Cursor CLI is not installed or not on PATH` |
| Quota exhausted (402/429) | unclassified | `rejected the summary request due to rate limits or quota` / `try: wait and retry` |

## Design

- **Two complementary error types, composed not competing.** `TextGenError`
  classifies (drives the label); `agent.TextGenerationError` carries evidence
  (`Stderr`, `StdoutBytes`) and is the only signal on the ctx path, where
  classification is meaningless. Failures wrap the former in the latter so
  `errors.As` finds either and `errors.Is` finds the ctx sentinel.
- **One status→Kind mapping** (`KindForHTTPStatus`), shared by the stderr
  classifier and Claude's envelope classifier. They drifted twice when
  separate; now pinned by `TestEnvelopeAndStderrClassifiersAgree`.
- **Classification requires an HTTP-status context**, not just isolated digits
  — `:` and `.` are non-word chars, so `\b4\d{2}\b` matched stack frames
  (`foo.js:401:12`) and decimals (`2.403 seconds`).
- **Stdout is classified with the machine-shaped patterns only.** Stdout-primary
  CLIs (codex/cursor/copilot) put real diagnostics there, but stdout on a
  summary call is often the model's prose — and "the user hit a 401
  Unauthorized" is a description, not a diagnosis.
- Precedence is by specificity (Auth > RateLimit > Config), not document order.

## History — how the scope changed

1. **Original scope:** extend #963's Claude-only typed errors to the other
   providers.
2. **#964 merged 2026-07-23** and rewrote the same surface with its own error
   type, inverting this PR's queue assumption. `51c8b2670` reconciles them by
   composing the two types rather than picking one; that merge also preserves
   the `apiKeyHelper` auth injection #964 added, which a naive resolution would
   have dropped.
3. **Finish-or-close decision:** rather than defer the remaining gaps to
   follow-up PRs, `6e5c08941` completed the migration — streaming path,
   external plugins, Pi (which gained a `CommandRunner` so it could be tested
   at all), and a registry-completeness guard.

## Reviewing notes

- **~600 lines are production logic**; ~1360 are tests and ~375 comments across
  34 files. The logic surface is much smaller than the diff suggests.
- **The riskiest commit is `51c8b2670`**, the reconciliation merge. All three
  serious bugs found during review originated at that seam. Tests now guard it
  (five are mutation-verified — each was confirmed to fail when the code was
  deliberately broken) but it is the part most worth a human read.
- **Known limitation:** external plugins have dynamic names, so they cannot
  appear in the compile-time display/remediation tables and fall back to the
  registry key.

## Test plan

- [x] `mise run test` — 8540 unit tests
- [x] `mise run test:integration` — 448 tests
- [x] `golangci-lint run` on a cleaned cache — 0 issues
- [x] Five guards mutation-verified: composition contract, apiKeyHelper argv,
      evidence wire, prose-vs-diagnostic split, cross-classifier agreement
- [ ] Real-agent E2E not run (typed-error surface change; the canary and
      integration suites cover the subprocess boundary)


peyton-alt:
Brought up to date with current main via a **merge commit** (`4f4810da3`) — all 23 original commits and their checkpoint linkage are preserved. What the merge integrates:

- **explain's structured failure blocks**: `formatCheckpointSummaryError` keeps main's `(label, rows, error)` contract; this PR's `renderTextGenError` was re-ported to return the structured triple, with remediation rendered as lowercase `try` rows. Claude wording preserves 963's baseline; non-Claude providers show their CLI stderr verbatim (no synthesized remediation on top).
- **Pi support**: main added a Pi agent after this branch forked, calling the helper this PR removes — Pi's `GenerateText`/`ListModels` are migrated to the same `RunIsolatedTextGeneratorCLIRaw` + `HandleTextGenResult` pattern (no phrase rules; HTTP-status baseline + verbatim stderr, consistent with the classifier feedback on this PR), and Pi is wired into the provider display-name/remediation maps.
- `explain_summary_error_test.go` re-pinned to the structured contract — every `(label, rows, error)` triple asserted exactly.

Verified: full `mise run test:ci` (unit + integration + canary) green on exactly this tree; lint clean. GitHub should report it mergeable now.

Note for the queue: #964/#1230 stack on the same error surface — landing this first lets them adopt `TextGenError` instead of their interim raw carrier (see the triage notes).

peyton-alt:
Closing as not worth the review cost. Recording the reasoning so this isn't re-litigated from scratch.

**The valuable part already shipped.** The one case where a user got nothing
useful was Claude exiting 0 with its error inside the JSON envelope — stderr
empty, the real message in a field — which without parsing produced a garbage
summary or an "unsupported JSON" error. #963 fixed that and it is on `main`.

**What was left was cosmetic.** Everything this PR added converts an
already-actionable message into a prettier already-actionable message. The
clearest example, a stale API key on the path Claude actually uses:

```
before:  claude stream failed: Invalid API key · Please run /login: exit status 2
after:   Claude authentication failed
           message: Invalid API key · Please run /login
           try:     run `claude login` and retry
```

The "before" is ugly, but the provider had already told the user to run
`/login`. Nobody is blocked by it.

**The design paid too much for that.** ~220 lines of classification machinery
(Kind taxonomy, a 3-alternative HTTP-status regex, two scanners, specificity
precedence, per-provider remediation tables) existed to derive a *label* from
text that is then displayed directly underneath it. That intricacy had a cost:
the classifier was the source of most of the defects found during review —
prose being misread as diagnostics, a 402 divergence between two copies of the
status mapping, stack frames matching as auth failures.

**Size.** 2368 additions across 34 files, against a recent maximum of 1881 for
anything merged in this repo. Roughly 600 lines of that are production logic;
the rest is tests and comments.

## Preserved

Branch `feat/all-agents-summary-error-handling` is pushed at `c7f4f8447` and
left in place. It is green: 8540 unit tests, 448 integration tests, lint clean,
up to date with `main`, no open findings.

## If this need resurfaces

Worth reviving, in order:

1. **Streaming failures return raw Go error strings.**
   `claudecode/generate_streaming.go` had 9 of 10 failure returns as bare
   `fmt.Errorf`. Since `TextGeneratorAdapter` prefers streaming, that is the
   path `explain --generate` actually takes for Claude. Tidying the message
   shape there is a handful of lines — see commit `6e5c08941`.
2. **`agent.TruncateStderr`** — trim, 500-byte cap, valid-UTF-8 on provider text
   before it reaches the terminal. Small and correct wherever third-party
   stderr is displayed.
3. **Do not rebuild the Kind taxonomy.** If a label is wanted, derive it at the
   single formatting site from (provider, err) without a typed error: the
   caller already knows which provider it selected, so `Provider` does not need
   to live in an error type, and the three cases that genuinely need code are
   binary-missing, Claude's exit-0 envelope, and ctx cancellation.

PR #1008, opened 2026-04-22
Consolidate TTY detection and narrow ACCESSIBLE to output format

gtrrz-victor:
https://entire.io/gh/entireio/cli/trails/56bb41c9cd0b

## Summary

Make prompt behavior explicit for CLI flows so subprocesses do not need fake TTY overrides and required prompts do not silently succeed in headless runs.

## What Changed

- Added `ENTIRE_PROMPTS=auto|never|plain` in `interactive` so prompt policy is explicit instead of inferred solely from TTY detection.
- Kept the hook-specific `ENTIRE_TEST_TTY` override for subprocess-based hook tests, while moving regular CLI subprocess control to `ENTIRE_PROMPTS`.
- Changed `cli.NewForm().Run()` to return `ErrPromptUnavailable` when a prompt is required but no interactive path exists.
- Added `RunOptional()` for the cases where skipping a prompt in headless mode is intentional.
- Added `entire login --no-browser` so non-interactive login flows can print the approval URL directly instead of relying on fake TTY behavior.
- Updated `scripts/local-device-auth-smoke.sh` to use `ENTIRE_PROMPTS=never` and `--no-browser`.
- Documented `ENTIRE_PROMPTS` in root help and added focused unit tests around prompt mode parsing, required prompt behavior, and the no-browser login path.

## Verification

- `go test ./cmd/entire/cli ./cmd/entire/cli/interactive -run 'TestRunLogin_NoBrowser_PrintsApprovalURL|TestFormRun_NoPromptAvailable_ReturnsErrPromptUnavailable|TestFormRunOptional_NoPromptAvailable_ReturnsNil|TestForcePlainOutput_UsesPromptModePlain|TestPromptMode_'`
- `mise run lint`
  - Fails in this environment because the installed `golangci-lint` is `v1.64.8` while the repo config expects golangci-lint v2.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -619,25 +619,19 @@ Trailers:
 - Handle all errors explicitly—don't leave them unchecked.
 - Reference `.golangci.yml` for enabled linters before writing Go code.
 
-## Accessibility
+## Interactive prompts and accessibility
 
-The CLI supports an accessibility mode for users who rely on screen readers. This mode uses simpler text prompts instead of interactive TUI elements.
+The CLI separates two concerns:
 
-### Environment Variable
+1. **Can we prompt at all?** — handled by `interactive.CanPromptInteractively()`. Returns false when no `/dev/tty` is available, or when running inside a known AI-agent subprocess (`GEMINI_CLI`, `COPILOT_CLI`, `PI_CODING_AGENT`, `GIT_TERMINAL_PROMPT=0`). Single source of truth.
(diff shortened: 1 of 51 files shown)


gtrrz-victor:
Followup: split TTY detection into two named entry points.

- `CanPromptInteractively` now uses `term.IsTerminal(os.Stdin.Fd())` —
  standard Go idiom, covers every cobra command.
- `CanPromptFromHook` keeps the `/dev/tty` probe because git redirects
  hook stdin; this is the only entry point that consults
  `ENTIRE_TEST_TTY`.
- `OverrideForTest` swaps both at once, so existing
  `forceInteractive` / `forceNonInteractive` helpers keep working.

Net: env-var escape hatch scope shrinks to the git-hook code path only.
Normal CLI paths are free of any test-only env var.

PR #1014, opened 2026-04-23
Guard `entire attach` against overwriting checkpoints from other machines

Soph:
https://entire.io/gh/entireio/cli/trails/cf1eb592e195

## Summary

A production report surfaced a scenario where `entire attach` could silently clobber a colleague's session data. If Alice creates checkpoint `abc123` on her machine and pushes the code commit (with `Entire-Checkpoint: abc123` trailer) without also pushing `entire/checkpoints/v1`, then Bob pulls the commit and runs `entire attach`, Bob's attach reuses ID `abc123`, writes a fresh session 0 locally, and on push overwrites Alice's session data.

This branch adds two layers of defence and one diagnostic tripwire.

### `entire attach` refuses to overwrite
When HEAD already carries an `Entire-Checkpoint:` trailer, attach now verifies the checkpoint is present on the local `entire/checkpoints/v1` branch before writing:

1. **Fast path**: read locally first — no network if we already have it.
2. **Fallback**: if missing, run the same metadata fetch chain `entire resume` uses (`checkpoint_remote` → treeless origin fetch → full origin fetch → remote-tracking).
3. **Refuse**: if it's still missing after the refresh, error out with an actionable hint. The suggested `git fetch` command targets the configured `checkpoint_remote` when present, otherwise `origin`.

### Tripwire in `writeStandardCheckpointEntries`
If a session-0 write is about to replace metadata from a *different* session ID (a tree-corruption / stale-summary shape, not a routine overwrite path), emit a loud `logging.Warn` with both session IDs so future regressions leave a trace.

### Tests
- `TestAttach_AppendsAsAdditionalSessionWhenIDDiffers` — positive path: two attaches on the same commit land as session 0 and session 1 under the same checkpoint.
- `TestAttach_RefusesWhenCheckpointMissingFromLocalBranch` — negative path: commit bears a trailer for a nonexistent checkpoint; attach errors with the actionable hint and creates nothing.
- `TestWriteStandardCheckpointEntries_WarnsOnUnexpectedSessionZeroOverwrite` — verifies the tripwire fires and captures both session IDs.

## Behaviour notes

- Offline repeat-attach on an already-local checkpoint: zero network calls.
- Offline attach against a missing checkpoint: best-effort fetch logs a warning, then the refuse error prints.
- Commits without an `Entire-Checkpoint:` trailer are unaffected — new attaches create fresh IDs as before.
- The refuse guard protects both the v1 write and the v2 dual-write, since it runs before either.

## Test plan
- [ ] `mise run check`
- [ ] Manual: create a checkpoint locally, delete `entire/checkpoints/v1` locally, run `entire attach` against the commit → verify refuse + hint
- [ ] Manual: run `entire attach` twice on the same session-free commit → verify second attach appends as session 1 without a network fetch

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes `entire attach` behavior around checkpoint reuse and git metadata fetching, which can affect users’ ability to attach sessions and may surface new failure modes in repos with unusual ref/remote setups.
> 
> **Overview**
> Prevents `entire attach` from accidentally overwriting an existing checkpoint ID created on another machine by **verifying the referenced checkpoint exists locally before writing**; if missing, it attempts the same metadata refresh chain as `resume` and then **refuses with an actionable `git fetch` hint**.
> 
> Adds a diagnostic tripwire in `checkpoint` writes that warns when a session-0 write would replace metadata for a different session ID (stale/corrupt tree shape), plus new tests covering append-as-new-session behavior, refusal on missing checkpoint metadata, and the warning emission.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a5be08cc568afd5f532592a08de866adb20e1de5. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/attach.go
+++ b/cmd/entire/cli/attach.go
@@ -136,15 +136,16 @@ func runAttach(ctx context.Context, w io.Writer, sessionID string, agentName typ
 	// Determine checkpoint ID: reuse from HEAD if one exists, otherwise generate new.
 	checkpointID, isExistingCheckpoint := resolveCheckpointID(headCommit)
 
-	// Refresh the metadata branch if HEAD already references a checkpoint.
-	// Returns a possibly-freshly-opened repo handle; see refreshForExistingCheckpoint.
-	repo = refreshForExistingCheckpoint(ctx, logCtx, repo, isExistingCheckpoint)
+	// If HEAD references an existing checkpoint, make sure we have it locally
+	// before writing — otherwise we'd create a fresh session 0 under the same
+	// ID and overwrite the original on push.
+	repo, err = ensureCheckpointAvailable(ctx, logCtx, repo, checkpointID, isExistingCheckpoint)
+	if err != nil {
+		return err
+	}
 
 	// Write directly to entire/checkpoints/v1.
 	store := cpkg.NewGitStore(repo)
-	if err := verifyCheckpointLocallyAvailable(ctx, store, checkpointID, isExistingCheckpoint); err != nil {
-		return err
-	}
 
 	author, err := GetGitAuthor(ctx)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


cmd/entire/cli/checkpoint/committed.go
```diff
@@ -341,6 +355,17 @@ func (s *GitStore) writeStandardCheckpointEntries(ctx context.Context, opts Writ
...
+	// Tripwire: if we're writing session 0 and there was already session-0
+	// metadata for a DIFFERENT session ID, emit a loud warning. This is a
+	// tree-corruption / stale-summary shape, not a routine overwrite path.
+	if existingSessionZeroMeta != nil && existingSessionZeroMeta.SessionID != opts.SessionID {
+		logging.Warn(ctx, "checkpoint write overwrites session 0 with a different sessionID — potential overwrite regression",
+			slog.String("checkpoint_id", opts.CheckpointID.String()),
+			slog.String("existing_session_id", existingSessionZeroMeta.SessionID),
+			slog.String("write_session_id", opts.SessionID),
+			slog.Bool("existing_summary_nil", existingSummary == nil))
+	}
```
computermode: Is it expected for this to log and then continue to write anyway?

cmd/entire/cli/attach.go
```diff
@@ -244,6 +255,107 @@ func getHeadCommit(repo *git.Repository) (*object.Commit, error) {
...
+	if err != nil {
+		return false, err //nolint:wrapcheck // Caller wraps with checkpoint ID context
+	}
+	return summary != nil, nil
+}
+
+// suggestCheckpointFetchCommand returns a git fetch command string the user
+// can paste, aware of whether checkpoint_remote is configured.
+func suggestCheckpointFetchCommand(ctx context.Context) string {
+	const ref = "entire/checkpoints/v1:entire/checkpoints/v1"
```
computermode: Is this only an issue for v1? Do we need to check if it's v2 and have a different ref here, I mean?

computermode:
```
For v2-only mode, it's the v2 /main ref, which
// has no remote-tracking analog and is therefore already local-only by
// construction.
```

Seems like it but just double-checking

computermode (dismissed):
Added some questions for my understanding but looks good 🙌🏻

cmd/entire/cli/attach.go
Soph: no it was, fixed it, thanks!

cmd/entire/cli/checkpoint/committed.go
Soph: yeah this was initially more investigation / debugging, but you are right, now it fails!

computermode (approved):
Thank you!

pfleidi (approved):

PR #1022, opened 2026-04-24
strategy: guard against empty-session metadata stubs

Soph:
https://entire.io/gh/entireio/cli/trails/a7fd3b96a21d

Two defensive guards in CondenseSession to prevent writing metadata.json checkpoints with no transcript and no real session attribution behind them:

1. Post-redaction skip gate (skipIfPostRedactionEmpty). The existing pre-redaction gate at line 165 catches sessions with neither a transcript nor tracked files. It cannot catch the case where the transcript was non-empty pre-redaction but redaction silently dropped it (malformed JSONL from an external agent), AND the session's tracked files don't overlap with the committed file set. The new gate runs after redaction and after filterFilesTouched so it sees the post-filter view; if both transcript and FilesTouched are now empty, return Skipped instead of writing a stub.

2. Narrow the filterFilesTouched fallback (sessionHasEvidenceOfWork). The old fallback assigned every committed file (minus .entire/) to any session with empty FilesTouched. That was designed for mid-turn commits before SaveStep ran, but it also fired for sessions registered at SessionStart that never produced anything — e.g., a Codex companion session whose hooks ran with a null transcript_path and never reached SaveStep. The fallback now requires evidence the session is a real participant: either a non-empty transcript was extracted, or a prior SaveStep recorded a checkpoint (StepCount > 0).

Also extracts redactOrDrop helper so CondenseSession stays under the maintidx threshold after the new skip gate is added.

Tests cover the post-redaction skip in the no-overlap + redaction-fail case, plus four unit tests for the narrowed filterFilesTouched contract: applies fallback with StepCount evidence, applies fallback with transcript evidence, skips fallback when neither is present, and intersects normally when FilesTouched is already populated.

---

> [!NOTE]
> **Medium Risk**
> Changes condensation skip/attribution behavior and could affect which sessions get persisted or credited for a commit, especially around mid-turn commits and redaction failures.
> 
> **Overview**
> **Prevents empty-session checkpoint stubs during manual-commit condensation.** `CondenseSession` now performs an additional *post-redaction* skip check, so if redaction drops the transcript and `FilesTouched` filters to empty, the session is skipped instead of persisting a metadata-only checkpoint.
> 
> **Tightens file attribution fallback.** `filterFilesTouched` only falls back to “all committed files (excluding `.entire/`)” when the session has evidence of work (non-empty transcript or `StepCount > 0`), avoiding false attribution to sessions that never produced output.
> 
> Adds focused regression/unit tests covering the new post-redaction skip and the updated `filterFilesTouched` fallback/intersection behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 323d8fb6365f9aa64bfafc971d4628b74822d258. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/strategy/condense_skip_test.go
+++ b/cmd/entire/cli/strategy/condense_skip_test.go
@@ -2,6 +2,7 @@ package strategy
 
 import (
 	"context"
+	"errors"
 	"os"
 	"path/filepath"
 	"testing"
@@ -9,6 +10,7 @@ import (
 
 	"github.com/entireio/cli/cmd/entire/cli/checkpoint/id"
 	"github.com/entireio/cli/cmd/entire/cli/session"
+	"github.com/entireio/cli/redact"
 
 	"github.com/go-git/go-git/v6"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


dipree (approved):
Looks good. The gating order (pre-redaction gate → filterFilesTouched → redactOrDrop → post-redaction gate) is correct, and sessionHasEvidenceOfWork cleanly separates legitimate mid-turn commits from empty companion sessions. Tests cover both paths — the redaction-failure test is serial (t.Chdir) so the global redactSessionJSONLBytes override doesn't race with the parallel filterFilesTouched unit tests. Lint and full strategy suite pass locally.

PR #1035, opened 2026-04-25
Fix v1-to-v2 migration dropping attribution metadata

pfleidi:
https://entire.io/gh/entireio/cli/trails/5558d19b2501

## What

The v1-to-v2 checkpoint migration silently dropped two types of attribution metadata:

1. **Session-level `prompt_attributions`** (71 sessions affected in validation repo) — `buildMigrateWriteOpts` omitted `PromptAttributionsJSON` from the write options, so the diagnostic payload was never passed to the v2 writer.

2. **Root-level `combined_attribution`** (2 checkpoints affected) — `writeCheckpointSummary` only preserved `CombinedAttribution` from an already-existing target summary. During fresh migration the target doesn't exist yet, so the v1 value was lost.

All functional checkpoint data (transcripts, prompts, content hashes) migrated correctly — these are metadata-only losses.

## How

- Added `PromptAttributionsJSON: m.PromptAttributions` to `buildMigrateWriteOpts`.
- Added `CombinedAttribution *InitialAttribution` field to `WriteCommittedOptions`.
- Updated `writeCheckpointSummary` to prefer `opts.CombinedAttribution` when non-nil, falling back to the existing summary value. This is backward-compatible: normal condensation never sets this field (combined attribution is computed post-commit via `UpdateCheckpointSummary`).
- Plumbed `summary.CombinedAttribution` from the v1 root summary through `buildMigrateWriteOpts` in `migrateOneCheckpoint`.
- Added two regression tests covering both bugs.

## Note

The broader gap where `updateCombinedAttributionForCheckpoint` in hooks only writes to the v1 store (not v2) is a separate issue affecting new v2-only checkpoints, not migration. Tracked separately.

---

> [!NOTE]
> **Low Risk**
> Low risk: changes only plumb and persist additional attribution metadata during migration and summary writing, with regression tests covering the previously dropped fields.
> 
> **Overview**
> Fixes v1→v2 checkpoint migration to **preserve attribution metadata** that was previously dropped.
> 
> `buildMigrateWriteOpts` now carries session-level `prompt_attributions` into v2 writes, and `WriteCommittedOptions` gains `CombinedAttribution` so `writeCheckpointSummary` can persist the v1 root-level `combined_attribution` even on first-time migrations (falling back to any existing summary when not provided).
> 
> Adds regression tests ensuring both `prompt_attributions` and `combined_attribution` survive migration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c5175de5db20aac39a7fde74ab2580470fc12a6e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/checkpoint/checkpoint.go
+++ b/cmd/entire/cli/checkpoint/checkpoint.go
@@ -296,6 +296,11 @@ type WriteCommittedOptions struct {
 	// Uses json.RawMessage to avoid importing session package.
 	PromptAttributionsJSON json.RawMessage
 
+	// CombinedAttribution is holistic attribution across all sessions.
+	// Used during migration to preserve v1 root summary attribution.
+	// During normal condensation this is nil (computed post-commit via UpdateCheckpointSummary).
+	CombinedAttribution *InitialAttribution
+
 	// Summary is an optional AI-generated summary for this checkpoint.
 	// This field may be nil when:
 	//   - summarization is disabled in settings
--- a/cmd/entire/cli/checkpoint/committed.go
+++ b/cmd/entire/cli/checkpoint/committed.go
@@ -471,12 +471,14 @@ func (s *GitStore) writeCheckpointSummary(opts WriteCommittedOptions, basePath s
 		return fmt.Errorf("failed to aggregate session stats: %w", err)
 	}
 
-	var combinedAttribution *InitialAttribution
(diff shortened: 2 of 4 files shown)


pfleidi:
Bugbot run

computermode (approved):

PR #1044, opened 2026-04-27
Support external agents to generate summary in entire explain

peyton-alt:
> [!NOTE]
> **Medium Risk**
> Medium risk because it changes provider discovery/validation and executes external agent binaries (including optional transcript compaction) in the summary generation path, which could affect reliability and output handling.
> 
> **Overview**
> **Summary generation now supports external providers.** `entire explain --generate` and the post-commit auto-summary path can use discovered `entire-agent-*` plugins that advertise `capabilities.text_generator`.
> 
> Provider resolution/validation is updated to trigger external discovery when needed, treat external providers as “available” without PATH checks for built-in CLIs, and auto-enable `external_agents` when configuring an external summary provider.
> 
> When an external agent’s transcript format can’t be summarized, the CLI can optionally call the agent’s `compact_transcript` capability, *redact the compacted output*, and use it for summary generation; condensed transcript parsing also gains a fallback to compact-transcript JSONL. Tests and docs are expanded to cover these behaviors, and lint config is updated for new interface-return cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ca3a01483aac016b85e97715f55a562baddbfeba. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


peyton-alt:
@bugbot review

peyton-alt:
@bugbot review

cmd/entire/cli/setup.go
```diff
@@ -176,6 +176,9 @@ func updateSummaryGenerationSettings(ctx context.Context, w io.Writer, provider,
 		if err := validateSummaryProvider(provider); err != nil {
 			return err
 		}
+		if ag, getErr := getSummaryAgent(types.AgentName(provider)); getErr == nil && external.IsExternal(ag) {
+			s.ExternalAgents = true
```
Soph: I wonder if we should log / surface this to the customer that this is done? But not sure. Like it's pretty clear you need to enable external agents for this to work, but it also means they work for other things?

Soph (dismissed):

cmd/entire/cli/setup.go
peyton-alt: Added a one-line notice that surfaces the side effect when
 (and only when) the flag actually flips false→true:

> Note: external agents are now enabled for the rest of Entire too — not just summaries.

  It fires from any path where picking an external summary provider
  implicitly turns on external_agents:
  - entire enable --summarize-provider <external> on an already-configured
   repo
  - The entire explain summary picker (auto-resolve or interactive TUI),
  which now also uses DiscoverAndRegisterAlways so installed external
  plugins show up as candidates regardless of the flag's current state —
  install is the user's opt-in, picking it is when the flag flips.

  Skipped for: built-in providers, flag-already-on (idempotent), the
  post-commit auto-summary hot path (no perf cost), and the entire enable
  agent picker (different flow where the user is explicitly choosing an
  agent — not a side effect).

Soph (approved):

PR #1045, opened 2026-04-27
Improve database credential redaction after betterleaks

peyton-alt:
## Summary

Redact for database credential gaps that remain after #1043 switched redaction to Betterleaks.

- Redacts low-entropy database connection strings that Betterleaks/default entropy detection miss, including Postgres keyword DSNs, SQL Server strings, ODBC strings, and JDBC/DB URL query-password forms.
- Redacts bounded DB credential values such as `DB_PASSWORD=...`, `PGPASSWORD=...`, and `REDIS_PASSWORD=...` while preserving placeholders and already-redacted values.
- Redacts structured JSON database password fields like `{"db":{"password":"...","host":"...","user":"..."}}` without redacting unrelated fields that happen to share the same value.
- Keeps over-redaction guardrails for non-credential URLs, SSH remotes, prose, paths, IDs, placeholders, and already-redacted values.
- Updates security/privacy docs to describe the split between Betterleaks, existing credentialed URI detection, and the additional DB-specific redaction layer.

## Scope After #1043

- Private key block redaction is covered by Betterleaks default rules, so this PR does not add custom private-key block logic.
- Generic `scheme://user:password@host` URI userinfo redaction is covered by the existing `credentialedURIPattern` in `redact/redact.go`, so the DB-specific detector no longer owns userinfo URLs.
- This PR focuses on DB DSNs, DB URL query passwords, bounded DB credential values, and structured JSON DB password fields.

## Also Included

Two unrelated concurrency fixes that surfaced during this work:

- `cmd/entire/cli/logging/logger.go` — holds the read lock across `l.Log` so `Init`/`Close` cannot close the log writer mid-write. Adds `TestLogging_ConcurrentInitAndLog` as a regression guard (now exercises `Close` too).
- `cmd/entire/cli/dispatch_tui_test.go` — drops `t.Parallel()` on two tests that mutate the package-level `newDispatchProgram` factory; adds a doc comment on the var so future contributors don't reintroduce the race.

Bundled here rather than split into separate PRs because both are small.

## Testing

- `GOCACHE=/tmp/entire-go-cache go test ./redact -count=1`
- `mise run check`
- `mise run lint`

## Notes

`mise run check` and `mise run lint` were run outside the sandbox because golangci-lint fetches its schema, Go uses the user cache, and integration tests bind localhost `httptest` servers.

---

> [!NOTE]
> **Medium Risk**
[...]

PR #1068, opened 2026-04-28
tighten redaction coverage

Soph:
https://entire.io/gh/entireio/cli/trails/254

## Follow-up to #1045: tighten redaction coverage

Addresses the review findings on the merged redact PR — gaps that don't break correctness today but leak common DB credential shapes.

### Changes

- **fix(redact): match `*_ROOT_PASSWORD` style env vars** — Allow internal segments and multi-char separators between vendor prefix and `password`/`passwd`/`pwd`. Covers `MYSQL_ROOT_PASSWORD`, `MARIADB_ROOT_PASSWORD`, `MONGO_INITDB_ROOT_PASSWORD`, `MSSQL_SA_PASSWORD`, `DB__PASSWORD`. Applied to both env-var and JSON-key regexes.
- **fix(redact): redact ADO.NET quoted password values** — `Password="se;cret;here"` (semicolons inside the quoted value) was leaking. Added `"…"`/`'…'` to the semicolon-DSN value class.
- **fix(redact): expand placeholder set and normalize dotted JSON keys** — `<password>`, `your_password`, `*****`, mask-char runs, `placeholder`, `secret_here` etc. now recognized as docs placeholders (no false redaction). `db.password` / `mysql.root.password` from flattened-config exporters normalize to `_`-form so they hit the credential key regex.
- **test(redact): pin shell shorthand, idempotency, and cross-context collision** — Pins three previously-unasserted behaviors: `mysql --password=…` is *not* redacted (known gap), `String` is idempotent (`f(f(x)) == f(x)`), and the keyed-JSON scanner is keyed by `(key, value)` not `(path, value)`.
- **refactor(redact): collapse duplicated shape regex and split state** — Vendor alternation extracted to a single `dbPasswordKeyShape` const composed into both regexes. Placeholder values consolidated into one map.

### Test plan

- [x] `mise run check` (fmt + lint + unit + integration + canary) passes
- [x] Positive tests for each broadened shape
- [x] Negative tests pin no over-redaction (`DB_PASSWORD_HASH`, `MYSQL_USER_ID`, all mask placeholders)

---

> [!NOTE]
> **Medium Risk**
> Changes core redaction regexes and placeholder heuristics, which could cause new false positives/negatives in log/transcript output despite added tests.
> 
> **Overview**
> **Tightens secret redaction coverage for database credentials** by unifying vendor/password key matching into a shared `dbPasswordKeyShape`, expanding it to catch common env/JSON keys like `*_ROOT_PASSWORD` (and multi-separator variants), and treating dotted keys (e.g. `db.password`) as credential keys during JSON normalization.
> 
> Improves connection-string parsing to redact ADO.NET/semicolon DSNs where `Password` values are quoted and contain embedded semicolons, and expands “placeholder” detection (e.g. `<password>`, `your_password`, mask runs) to avoid redacting obvious documentation defaults.
> 
> Adds tests to cover the new credential shapes, placeholder edge cases, idempotent redaction (`String(String(x)) == String(x)`), dotted JSON keys, cross-context value collisions, and explicitly pins the known non-redaction gap for `--password=...` CLI flags.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dfe3760750a232417df7f0089c6f876f98af4bbc. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redact/redact.go
+++ b/redact/redact.go
@@ -383,15 +383,45 @@ func isPlaceholderSecretValue(value string) bool {
 	if strings.HasPrefix(normalized, "${") && strings.HasSuffix(normalized, "}") {
 		return true
 	}
+	// Documentation placeholders like <password> or <your-db-password>.
+	if strings.HasPrefix(normalized, "<") && strings.HasSuffix(normalized, ">") && len(normalized) > 2 {
+		return true
+	}
 	if _, ok := redactedPlaceholderForms[normalized]; ok {
 		return true
 	}
+	if isRepeatedCharPlaceholder(normalized) {
+		return true
+	}
 	switch normalized {
-	case "xxx", "xxxx", "changeme", "example":
+	case "changeme", "example", "placeholder",
+		"your_password", "your_db_password", "your_secret",
+		"secret_here":
 		return true
 	default:
 		return false
 	}
+}
+
+// isRepeatedCharPlaceholder reports whether s is a non-empty run of a single
+// masking character commonly used to redact values in docs and screenshots,
+// e.g. "***", "xxxx", "....", "----".
+func isRepeatedCharPlaceholder(s string) bool {
+	if s == "" {
+		return false
+	}
+	first := s[0]
+	switch first {
+	case '*', 'x', '.', '-':
+	default:
+		return false
+	}
+	for i := 1; i < len(s); i++ {
+		if s[i] != first {
+			return false
+		}
+	}
+	return true
 }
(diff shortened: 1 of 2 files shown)


peyton-alt:
@bugbot review

peyton-alt (approved):

gtrrz-victor (approved):

PR #1082, opened 2026-04-30
Fix Cursor session mis-identified as Claude Code

ashtom:
https://entire.io/gh/entireio/cli/trails/269

## Summary

When Cursor Agent forwards a single user prompt to both `.cursor/hooks.json` and `.claude/settings.json`, two `entire` hook processes can claim the same `session_id` with different `agent_type`s. The loser's wrong `agent_type` can win the last write — `entire status` then shows e.g. `Claude Code (composer-2)` for a Cursor session.

Two layered defenses + repair on the next turn:

- **SessionStart agent hint** (first-writer-wins): the first agent to fire `SessionStart` writes `<sessionID>.agent` with `O_CREATE|O_EXCL`. Subsequent hooks for the same session see the existing file and no-op. `InitializeSession` reads this hint at TurnStart so the firing hook can't relabel a session that another agent already claimed.
- **Transcript-path override** (strongest signal): at `TurnStart`, if the transcript path lives inside another agent's `GetSessionDir(repo)`, that agent wins regardless of which hook is firing or what the hint says. This same check (`correctSessionAgentType`) also runs each turn against existing sessions, so a session that was mis-attributed on the racing turn is corrected the moment a later turn's transcript points to the true owner.
- **Forwarded-event filter**: once `state.AgentType` is recorded, lifecycle events fired by other agents are skipped (`SessionStart` and `TurnStart` are exempt — they need to reach the setup paths to perform repair).

Concurrent `InitializeSession` calls for the same session_id are accepted (no per-session lock — see [`e983e08a`](https://github.com/entireio/cli/commit/e983e08a) for the rationale: in field logs Cursor IDE only forwards `SessionStart`/`Stop`, not `UserPromptSubmit`, so the race doesn't actually fire in practice; the transcript-path repair handles any case where it does).

The SessionStart banner is gated by a separate `<sessionID>.banner` marker so a non-`HookResponseWriter` agent (Cursor) winning the ownership race can't suppress the banner from a banner-capable agent firing SessionStart for the same session ([`e38f1c60`](https://github.com/entireio/cli/commit/e38f1c60)).

## Test plan

- [x] `mise run fmt` clean
- [x] `mise run lint` clean
- [x] All unit tests pass — added `TestAgentForTranscriptPath`, `TestPathHasDirPrefix_CaseSensitivity` (Windows-aware), `TestStoreAgentTypeHint_*`, `TestClaimSessionStartBanner_*`, `TestResolveSessionAgentType_*`, `TestCorrectSessionAgentType_*`, `TestInitializeSession_HintWinsRaceAtTurnStart`, `TestInitializeSession_TranscriptPathRepairsExistingState`, `TestInitializeSession_ConcurrentCallsConvergeToTranscriptOwner` (also under `-race`), `TestHandleLifecycleSessionStart_NonWriterClaimDoesNotSuppressBanner`, `TestHandleLifecycleSessionStart_BannerClaimedOnce`
- [x] `mise run test:integration` passes
- [x] `mise run test:e2e:canary` passes (4/4)
- [x] `GOOS=windows go build ./...` and `go vet` clean
- [x] Verified live in `entire-helpers` repo: existing wrong session repaired itself on the next prompt via transcript-path override
- [ ] Verify on Linux/Windows (CI)


Soph:
@bugbot review

dipree (approved):

PR #1093, opened 2026-05-01
Copy login device code to clipboard

blackgirlbytes:
## Summary
- copy the device auth user code to the clipboard during `entire login`
- keep login usable when clipboard access fails by printing a warning and preserving the manual code path
- add focused tests for clipboard copy success and failure

## Tests
- `mise run check`
- `mise run lint`

https://github.com/user-attachments/assets/4c056d2f-b7fe-4b84-96cc-3c6e91ae3e89


Soph (commented):
@blackgirlbytes could we do the clipboard add when you hit enter? That way you know more what is happening and are a bit more in control? Like "Press Enter to copy the code into your clipboard and open ... in your browser..."

blackgirlbytes:
> @blackgirlbytes could we do the clipboard add when you hit enter? That way you know more what is happening and are a bit more in control? Like "Press Enter to copy the code into your clipboard and open ... in your browser..."

totally can do it! @Soph YAY!

blackgirlbytes:
@Soph updated for pressing Enter

Soph (approved):

Issue #411, opened 2026-02-18
prepare-commit-msg hook not executed after session restart (ended → idle)


hyunjae-labs:
Hi Entire team,

Thank you for creating such a useful tool for tracking AI agent sessions. I've encountered what appears to be a bug in version 0.4.4 and wanted to report it with detailed reproduction steps and evidence.

## Summary

The `prepare-commit-msg` hook is not executed when a session restarts from "ended" state, causing checkpoint creation to fail.

## Environment

- **Entire CLI Version**: 0.4.4 (2f0ad9ab)
- **OS**: Linux (WSL2)
- **Claude Code**: 2.1.44
- **Strategy**: auto-commit

### Expected

When a Claude Code session is active and a git commit is made:
1. `prepare-commit-msg` hook executes
2. `entire hooks git prepare-commit-msg` is called
3. `Entire-Checkpoint` trailer is added to commit message
4. Event is logged to `.entire/logs/entire.log`

### Actual

After a session transitions from "ended" → "idle":
1. `prepare-commit-msg` hook executes (Git side)
2. But `entire hooks git prepare-commit-msg` produces no output
3. No `Entire-Checkpoint` trailer added
4. **No event logged** to `.entire/logs/entire.log`

### Step 1: Setup

```bash
cd /home/devuser/projects/atlas
entire status
# Output shows: Active Session 2c4172e4
```

### Step 2: Session State Transitions (from entire.log)

```
2026-02-18 00:32:09 - SessionStop (idle → ended)
2026-02-18 00:32:20 - SessionStart (ended → idle)  ← Session restarted!
```

### Step 3: Git Commit After Restart

```bash
# Time: 00:33:00 (10 seconds after session restart)
git add . && git commit -m "feat: some feature"
```

### Step 4: Check Results

```bash
# Check commit
git log -1 --format=fuller
# ❌ No Entire-Checkpoint trailer

# Check entire.log
grep "prepare-commit-msg" .entire/logs/entire.log | tail -1
# Last event: 23:45:53 (BEFORE session restart)
# No event at 00:33:00
```

### Timeline Analysis

```
Atlas Project (.entire/logs/entire.log):

23:45:53 - prepare-commit-msg executed ✅
          (last event before session restart)

00:32:09 - SessionStop (idle → ended)
00:32:20 - SessionStart (ended → idle)

00:44:25 - Git commit
00:45:39 - Git commit
00:46:32 - Git commit
00:48:30 - Git commit
00:58:27 - Git commit
[...]

PR #1102, opened 2026-05-04
fix: entire attach preserves BaseCommit and active phase

rkfir-dn:
Related to https://github.com/entireio/cli/issues/411

When entire attach was called on a session with an empty BaseCommit (e.g., after a hook initialization failure, or on a session created before entire was enabled), saveAttachSessionState wrote PhaseEnded and left BaseCommit empty. This prevented the prepare-commit-msg hook from recognizing the session, so subsequent commits never received Entire-Checkpoint trailers.

Two changes in saveAttachSessionState:

1. Populate BaseCommit (and AttributionBaseCommit) from repo.Head() when the state's BaseCommit is empty. This activates the session so the prepare-commit-msg hook attaches trailers to future commits in the same session — eliminating the need to call entire attach again after a hook failure.

2. Do not unconditionally set Phase to PhaseEnded. If the session is already active (Phase.IsActive()), preserve that phase. Previously, calling entire attach on a running session would set PhaseEnded and break all subsequent commits in that session.

### DIFF
--- a/cmd/entire/cli/attach.go
+++ b/cmd/entire/cli/attach.go
@@ -207,7 +207,7 @@ func runAttach(ctx context.Context, w io.Writer, sessionID string, agentName typ
 	}
 
 	// Create or update session state.
-	if err := saveAttachSessionState(logCtx, existingState, sessionID, ag.Type(), transcriptPath, checkpointID, meta, tokenUsage); err != nil {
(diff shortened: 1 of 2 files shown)


Soph (approved):
@rkfir-dn thanks for your contribution, I added tests validating it but your changes look good.

alishakawaguchi (approved):

PR #1114, opened 2026-05-05
Checkpoints v2: Fix migration rerun logic (speedups, avoid duplicated migration efforts)

computermode:
https://entire.io/gh/entireio/cli/trails/292

## Summary

Fixes a few issues that arose when rerunning `entire migrate --checkpoints v2:
* rerunning the migration could duplicate `/full/<n>` archives on supposed no-ops (one of our repos grew from gen 46 -> 86), 
* interrupting a migration could result in a partially completed state that re-running may not fix properly
* rerunning the migration resulted in an unconditional transcript walk across all the trees every time, even if there was nothing else to be done.

For repos with thousands of checkpoints, the issues above meant that migrations would take 20+ minutes to run even if there was nothing to be done.

### Changes

* skip when every session has `/full` artifacts, pack only the missing sessions when some are absent, fresh-migrate when the checkpoint isn't in `/main` yet, `--force` to redo from scratch
* generation-metadata repair only runs when the loop packed something, and just-packed archives are excluded since the packer writes correct `generation.json` in-memory. A new `BuildFullSessionArtifactsIndex` turns per-session presence checks into O(1) map lookups, and `ReadSessionMetadata` reads only `metadata.json`
* no-op migration reruns now take seconds to run

---

> [!NOTE]
> **Medium Risk**
> Touches checkpoint migration and generation repair flows, changing when archives are created and when metadata repair runs; mistakes could skip needed packing or leave inconsistent `/full/*` state. Scope is contained to migration/repair tooling, but impacts large repos and historical data layout.
> 
> **Overview**
> Improves `entire migrate --checkpoints v2` rerun behavior so already-migrated checkpoints are **skipped** when `/full/*` artifacts are present, while partially migrated checkpoints (written to `/main` but missing `/full/*`) **resume** by packing only missing sessions instead of re-migrating everything.
> 
> Speeds up the migration hot path by prebuilding a `BuildFullSessionArtifactsIndex` for O(1) per-session `/full/*` presence checks, adding a lightweight `ReadSessionMetadata` (metadata-only) read, and computing `generation.json` timestamps from already-loaded transcripts via `AggregateTranscriptTimestamps`.
> 
> Reduces post-migration overhead by running `RepairV2GenerationMetadata` **only when** new `/full/<n>` refs were written, and adds `ExcludeRefs` support so freshly packed generations are not re-walked during repair. Removes legacy v1 transcript cleanup and compact-transcript backfill/fast-path logic from the migration rerun flow, with tests updated to lock in the new idempotent/resume semantics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4b333b08659f6a120c222e194b3d46f388d71eae. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmd/entire/cli/checkpoint/v2_committed.go
```diff
@@ -135,6 +136,111 @@ func (s *V2GitStore) findFullSessionArtifacts(checkpointID id.CheckpointID, sess
...
+			if errors.Is(refErr, plumbing.ErrReferenceNotFound) {
+				continue
+			}
+			return nil, fmt.Errorf("read %s: %w", refName, refErr)
+		}
+		rootTree, treeErr := s.repo.TreeObject(rootTreeHash)
+		if treeErr != nil {
+			return nil, fmt.Errorf("read %s root tree: %w", refName, treeErr)
+		}
+		if err := s.indexFullSessionsInTree(rootTree, index); err != nil {
```
pfleidi: This might be a purely stylistic choice so feel free to ignore this comment.

The role of `indexFullSessionsInTree` within `BuildFullSessionArtifactsIndex` is a bit odd to me: We're creating `index` in this function only to have it mutated by `indexFullSessionsInTree`. That isn't what I would have expected.

I'm wondering if it would be easier to read/understand to have a function `listFullSessionsInTree()` return a list of sessions per ref and merge them into the tree in `BuildFullSessionArtifactsIndex()` to avoid passing an object around and mutating it in multiple places which imo makes following the data flow harder than it needs to be.

cmd/entire/cli/checkpoint/v2_committed.go
```diff
@@ -135,6 +136,111 @@ func (s *V2GitStore) findFullSessionArtifacts(checkpointID id.CheckpointID, sess
 	return fullSessionArtifacts{}, nil
 }
 
+// FullSessionArtifactsIndex answers "does this session have complete /full/*
+// artifacts?" with an O(1) map lookup. Build it once via
+// BuildFullSessionArtifactsIndex.
+type FullSessionArtifactsIndex map[string]struct{}
```
pfleidi, line 142: Given that this acts like a Set, could this be a `map[string]bool`?

cmd/entire/cli/migrate.go
```diff
@@ -185,14 +191,24 @@ type migratedFullSession struct {
...
+// freshly-packed refs list that production code uses to gate the repair pass.
 func migrateCheckpointsV2(ctx context.Context, repo *git.Repository, v1Store *checkpoint.GitStore, v2Store *checkpoint.V2GitStore, progressOut io.Writer, force bool) (*migrateResult, error) {
+	result, _, err := migrateCheckpointsV2Internal(ctx, repo, v1Store, v2Store, progressOut, force)
+	return result, err
+}
+
+// migrateCheckpointsV2Internal returns the /full/<n> refs the packer wrote so
+// the caller can pass them as exclusions to the generation-metadata repair
+// pass.
+func migrateCheckpointsV2Internal(ctx context.Context, repo *git.Repository, v1Store *checkpoint.GitStore, v2Store *checkpoint.V2GitStore, progressOut io.Writer, force bool) (*migrateResult, []plumbing.ReferenceName, error) {
```
pfleidi: Part of me is wondering whether the split between `migrateCheckpointsV2` and `migrateCheckpointsV2Internal` is necessary. Could we inline the code changes here and update the signature of `migrateCheckpointsV2()` instead?

pfleidi (dismissed):

cmd/entire/cli/checkpoint/v2_committed.go
computermode: > map[string]struct{} is the idiomatic Go set type with zero-byte values vs map[string]bool's 1-byte values. Kept as-is.

let me know how you feel about this explanation, happy to update it

cmd/entire/cli/checkpoint/v2_committed.go
computermode: Totally agree with this yeah - updated in the latest commit. Thank you!

cmd/entire/cli/checkpoint/v2_committed.go
pfleidi: If it works and it's the idiomatic way to do this in Go, who am I to argue? 👍

pfleidi (approved):

PR #1127, opened 2026-05-06
Support headless auth token storage

### LINKED ISSUE #1036: Feature Request: Support for non-keyring secret storage (Headless/CI  environments)
### Problem or use case

The Entire CLI currently relies on the system keyring (via the Secret Service D-Bus
     interface) to store authentication tokens. On headless Linux environments (SSH, WSL, Docker,
     or CI/CD runners), this often fails because no unlocked "default" or "login" collection is
     available, preventing users from logging in.
Many users operate in environments where `gnome-keyring-daemon` or `dbus` are not running
      or are difficult to configure. Providing a standard way to bypass the system keyring would
      make Entire much more accessible for remote development and automation.

### Desired behavior

```shell
When running in a headless or non-GUI environment, entire should provide an alternative to the
  system keyring for persistent authentication.
```

### Proposed solution

To support headless environments more gracefully, I propose adding a fallback or
      configurable secret store backend:
    
     Allow users to specify a local file for secret storage via an
      environment variable (e.g., `ENTIRE_SECRETS_PATH=~/.entire/credentials`) or a flag.
     **Environment Variable Fallback:** Allow providing the authentication token directly via
      an environment variable (e.g., `ENTIRE_AUTH_TOKEN`) to support non-interactive CI/CD
      workflows.
     **Plain-text Fallback:** A `--no-keyring` flag that stores the token in plain text
      within the `.entire` directory (with a warning about security implications).

### Alternatives or workarounds

Currently, I have to manually launch `dbus-launch` and pipe a password into
      `gnome-keyring-daemon --unlock`, which is not ideal for many workflows.

peyton-alt:
https://entire.io/gh/entireio/cli/trails/307

## Summary

Adds explicit non-keyring CLI auth paths for headless, container, and CI environments. Implements all three options from issue #1036.

- **`ENTIRE_AUTH_TOKEN`** — read-only env-var bearer for non-interactive workflows. Scoped to the production API origin (`https://entire.io`) so a stray `ENTIRE_API_BASE_URL` override can't leak a prod token to a staging/dev endpoint.
- **`ENTIRE_SECRETS_PATH`** — opt-in JSON file-backed store. Plaintext, `0600` perms enforced on read, atomic temp+rename writes, refuses relative paths and group-readable files.
- **`--no-keyring`** — fail-closed login flag. Preflights `ENTIRE_SECRETS_PATH` before device auth and refuses to fall back to the OS keyring if no non-keyring backend is configured.

Keeps OS keyring as the default backend. No automatic plaintext fallback.

### Source-aware command behavior

`auth status`, `auth list`, `auth revoke`, and `logout` now look up the token via `GetTokenInfo` and behave correctly per source:

- `auth status` reports provenance (e.g. "supplied by ENTIRE_AUTH_TOKEN" vs "stored in file /path" vs "stored in OS keychain") and emits the right re-auth guidance per source.
- `logout` with an env token doesn't pretend to delete the env var — it tells the user to unset it.
- `auth revoke --current` with an env token treats 401 as idempotent success, but **fails loud (non-zero, no "Revoked" claim)** on any other server error — env tokens have no local copy to fall back on, so a 5xx must not be silently swallowed.

Fixes #1036.

## Validation

- `mise run check` (fmt + lint + unit + integration + canary E2E) on every commit
- 16-writer concurrency test on the file backend (passed before the file-locking commit was reverted; atomic temp+rename writes still guarantee torn-free reads)
- Live binary replication of both reported user scenarios:
  - **kustrun** (remote container, locked keyring): `entire login` with `ENTIRE_SECRETS_PATH` set writes the token to a `0600` JSON file; `auth status` reads it back source-tagged. The original "failed to unlock correct collection" error is captured verbatim in `TestRunLogin_SaveFailureIncludesHeadlessHint`.
  - **digiSal** (CI/headless): `ENTIRE_AUTH_TOKEN=fake entire auth status` triggers a source-aware "unset ENTIRE_AUTH_TOKEN to re-authenticate" message instead of telling a CI runner to "run entire login".

## Reviews addressed

- **Copilot inline review** (PR comment on `auth.go`): idempotent `revoke --current` for env tokens — `423551486` (with one re-iteration after Codex adversarial review).
- **Codex adversarial review**:
  - [high] Env-token revoke false success on non-401 → `423551486`.
  - [medium] Env token bypassed baseURL scoping → `deab46b2e`.
  - [medium] Dropped `--no-keyring` removed fail-closed login → `b203b2308`.
  - [medium] Unlocked file-store read-modify-write → addressed in `19b26afc5`, then **reverted** (`d63ed2685`) for scope. The race requires two concurrent CLI invocations sharing one credentials file, which is exotic in practice; deferred to a follow-up if observed. Atomic temp+rename writes still guarantee readers never see torn content.

## Notes

`ENTIRE_CHECKPOINT_TOKEN` remains separate and only applies to checkpoint Git fetch and push operations.


peyton-alt:
@bugbot review

peyton-alt:
@bugbot review

blackgirlbytes:
eek sorry for mentioning your PR in my forked repo @peyton-alt . I was showing codex this as an example..and it mentioned it in a PR

diwu-sf:
@peyton-alt is this PR abandoned? We still need headless support on dev VMs that dont run with desktop GUI.

PR #1147, opened 2026-05-08
V2 checkpoints migration performance improvements

computermode:
https://entire.io/gh/entireio/cli/trails/321

### Summary

Tested with a repo with ~2500 checkpoints. Reduced migration time to ~3.5 minutes (`migrate_checkpoints  216869ms`), down from ~5.5 minutes (`migrate_checkpoints  337738ms`).

Validated with a fresh migration of the repo locally and compared with the version pushed up to GH, ensured there were no regressions in the data pushed up. See the comparison report here: https://gist.github.com/computermode/1c06c434317fb8fe7df18b4598913ab8

Script to compare repos: https://gist.github.com/computermode/599fea82d7ae0147716997de8f19576a

### Biggest Redundancies

| Area | Before | After |
|---|---|---|
| Raw transcript packing | Re-chunked and re-stored v1 transcript blobs into v2 | Reuses existing v1 blob hashes |
| Generation metadata | Rescanned raw transcripts for timestamps | Uses already-loaded checkpoint `CreatedAt` first |
| Compact offsets | Compacted full transcript, then compacted scoped suffix again | Caches raw-position -> compact-line offset |
| `/main` flush | Repeated tree surgery per checkpoint inside each batch | Builds subtrees, applies root changes once |
| V2 existence checks | Read v2 summary per checkpoint | Builds one v2 presence index |

### Where It Showed Up In Trace

- `pack_full_generation_total`: duplicate raw transcript packing
- `pack_full.generation_timestamps_total`: raw transcript timestamp rescans
- `migrate_one.compact_transcript_total`: duplicate compact passes
- `flush_main_total`: repeated `/main` tree rewrites
- per-checkpoint migration overhead: repeated v2 summary lookups

### Key Point

The batch size was already `100`; the slowdown was inside each batch from repeated blob work, transcript parsing, and Git tree rewrites.

---

> [!NOTE]
> **Medium Risk**
> Touches checkpoint migration and v2 git ref-writing logic; while covered by new invariants/tests, mistakes could lead to missing or mis-indexed checkpoint data during migration.
> 
> **Overview**
> Speeds up v1→v2 checkpoint migration by **batching v2 `/main` writes**: migration now buffers per-session `WriteCommittedOptions` and flushes them via a new `WriteCommittedMainBatch` that updates `/main` with a single commit/CAS per batch.
> 
> Avoids redundant work during `/full/*` packing by **reusing existing v1 transcript blob hashes** (surfaced via `SessionContent.TranscriptBlobHashes`) and by preferring checkpoint `CreatedAt` when generating `generation.json`, falling back to transcript timestamp scans only when needed.
> 
> Adds a compact-transcript offset cache and cumulative perf annotation (`perf.Annotate`) to reduce repeated offset computation and make phase totals visible, alongside new tests validating batch-vs-sequential tree equality, single-commit behavior, and raw-blob reuse.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 961c1959ab13159cc81456509659e97ecd1e879f. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmd/entire/cli/checkpoint/v2_committed.go
```diff
@@ -59,6 +59,285 @@ func (s *V2GitStore) WriteCommittedWithSessionIndex(ctx context.Context, opts Wr
...
+	sessions := make([]SessionFilePaths, len(ordinals))
+
+	for sessionIndex, ordinal := range ordinals {
+		opts := batch[ordinal]
+		sessionPath := fmt.Sprintf("%s%d/", basePath, sessionIndex)
+		filePaths, err := s.writeMainSessionToSubdirectory(opts, sessionPath, entries)
+		if err != nil {
+			return plumbing.ZeroHash, err
+		}
+		sessions[sessionIndex] = filePaths
```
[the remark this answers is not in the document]
pfleidi: I think we might be able to ignore this.

cmd/entire/cli/checkpoint/v2_committed.go
```diff
@@ -59,6 +59,285 @@ func (s *V2GitStore) WriteCommittedWithSessionIndex(ctx context.Context, opts Wr
...
+	groups := []*cpGroup{}
+	byID := map[id.CheckpointID]*cpGroup{}
+	for i, opts := range batch {
+		g, ok := byID[opts.CheckpointID]
+		if !ok {
+			g = &cpGroup{cpID: opts.CheckpointID}
+			byID[opts.CheckpointID] = g
+			groups = append(groups, g)
+		}
+		g.ordinals = append(g.ordinals, i)
```
pfleidi, line 102: I might be misunderstanding this, but it looks like we're creating a lookup table for a checkpoint ID to return on which position in the batch a checkpoint ID can be found.

- If we're already using `id.CheckpointID` as key, do we need `cpGroup` even? Could `byID` be something like `map[id.CheckpointID][]int`?
- While passing ordinals around works, I find it hard to follow. 

I wonder if a lookup table like ` map[id.CheckpointID][]WriteCommittedOptions` would make this easier to understand 🤔

pfleidi (approved):

cmd/entire/cli/checkpoint/v2_committed.go
computermode: We chatted about this offline - I will get this addressed in a followup. Thank you for reviewing and raising this point!

computermode: I have to resolve this comment to merge but I will still address it in the followup 🥲

PR #1149, opened 2026-05-08
fix(settings): activate hooks for entire enable --local

### LINKED ISSUE #1123: Git hooks silently no-op when only .entire/settings.local.json exists (entire enable --local)
## Description

Running `entire enable --local` installs git hooks but creates only `.entire/settings.local.json`. The hooks then silently do nothing because their activation check requires `.entire/settings.json` to exist.

## Steps to reproduce

1. Fresh repo with no prior Entire setup.
2. Run `entire enable --local`.
3. Make changes and commit — hooks fire but do nothing (no checkpoints saved).
4. Run `entire enable` (which creates `.entire/settings.json`) — hooks now work.

## Root cause

The hooks' `PersistentPreRunE` calls `settings.IsSetUpAndEnabled()`, which internally calls `IsSetUp()`. That function checks only for `.entire/settings.json`:

```go
// settings.go
func IsSetUp(ctx context.Context) bool {
    _, err = os.Stat(settingsFileAbs)  // only checks settings.json
    return err == nil
}

func IsSetUpAndEnabled(ctx context.Context) bool {
    if !IsSetUp(ctx) {  // returns false when only settings.local.json exists
        return false
    }
    ...
}
```

`IsSetUpAny()` already exists and correctly checks either file, but `IsSetUpAndEnabled` doesn't use it.

## Fix

Change `IsSetUpAndEnabled` to use `IsSetUpAny` instead of `IsSetUp`. `Load()` already merges both files and defaults `Enabled: true` when `settings.json` is absent, so the loaded value is correct — the only broken piece is the presence check.

## Workaround

Run `entire enable` (without `--local`) after `entire enable --local` to create `.entire/settings.json`.

ivasuy:
## Summary

Fixes #1123.

This PR switches `IsSetUpAndEnabled` to `IsSetUpAny`, which already exists and correctly checks for either file. `Load()` already merges both files and defaults `Enabled: true` when only the local file is present, so the loaded value was always correct. The only broken piece was the presence check.

## Changes

- `cmd/entire/cli/settings/settings.go`
  - One-line fix: `IsSetUpAndEnabled` now calls `IsSetUpAny(ctx)` instead of `IsSetUp(ctx)`.
  - Updated the doc comment to explain the local-only setup case.

- `cmd/entire/cli/settings/settings_test.go`
  - Added regression coverage:
    - `TestIsSetUpAndEnabled_LocalOnly`
    - `TestIsSetUpAndEnabled_LocalOnlyDisabled`
    - `TestIsSetUpAndEnabled_NoFiles`
    - `TestIsSetUpAndEnabled_BaseOnlyEnabled`
  - Added a new `setupSettingsRepo` helper using `testutil.InitRepo` so `paths.WorktreeRoot()` resolves correctly through `AbsPath`.
  - Kept the existing `setupSettingsDir` for fake `.git` directory cases, but avoided using it for tests that need real path resolution.

## Verification

```bash
mise run check
```

Result:

```text
fmt + lint + test:ci — all green
```

```bash
go test ./cmd/entire/cli/settings/ -run TestIsSetUpAndEnabled -v
```

Result:

```text
passed
```

I also confirmed the new `TestIsSetUpAndEnabled_LocalOnly` regression test fails on `main` without the one-line fix and passes with it. That confirms the test locks in the actual fix instead of asserting tautologically.

End-to-end smoke test:

- Created a scratch repo.
- Ran:

```bash
entire enable --local
```

- Made a commit.
- Confirmed a checkpoint was saved on `entire/checkpoints/v1`.

Before the fix, the same flow produced no shadow branch and no checkpoint.

### DIFF
--- a/cmd/entire/cli/settings/settings.go
+++ b/cmd/entire/cli/settings/settings.go
@@ -631,10 +631,13 @@ func IsSetUpAny(ctx context.Context) bool {
 }
 
 // IsSetUpAndEnabled returns true if Entire is both set up and enabled.
-// This checks if .entire/settings.json exists AND has enabled: true.
+// This checks if either .entire/settings.json or .entire/settings.local.json
+// exists AND the merged settings have enabled: true. Using IsSetUpAny here
+// ensures hooks fire for `entire enable --local` setups, where only the
+// local settings file exists.
 // Use this for hooks that should be no-ops when Entire is not active.
 func IsSetUpAndEnabled(ctx context.Context) bool {
-	if !IsSetUp(ctx) {
+	if !IsSetUpAny(ctx) {
 		return false
 	}
 	s, err := Load(ctx)
--- a/cmd/entire/cli/settings/settings_test.go
+++ b/cmd/entire/cli/settings/settings_test.go
@@ -8,6 +8,8 @@ import (
 	"strings"
 	"testing"
 	"time"
+
+	"github.com/entireio/cli/cmd/entire/cli/testutil"
 )
 
 const (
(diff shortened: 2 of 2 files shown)


Soph:
Thanks for opening this PR. The issue #1123 was closed by #1713 and hardened by #2056. Sorry that your change didn't made it.

PR #1182, opened 2026-05-12
Speed up entire clean --all and add progress + make checkpoints v2 commands respect ctrl+c

computermode:
https://entire.io/gh/entireio/cli/trails/357

## Summary

`entire clean --all` previously hung silently for many seconds (or indefinitely on repos with lots of archived v2 generations) before any output appeared, and Ctrl-C wouldn't reliably escape because go-git's tree iteration ignores context cancellation.

After incorporating review feedback, this PR makes the following changes:

1. Trust `generation.json` over raw transcripts **(TRADE-OFF, please see below).** This is the primary fix; on a real repo with 18 archived generations the command now completes in ~1s instead of hanging for almost a minute.
2. **Concurrent v2 enumeration.** v2 archived-generation listing runs in a goroutine concurrent with the fast local enumeration, with a derived context so a failure in the local path cancels the v2 path immediately rather than blocking on it.
3. **Reliable Ctrl-C, scoped to this command.** Inside `runCleanAll` only, an `installForceExitOnSecondSignal` helper arms a second-Ctrl-C → `os.Exit(130)` escape hatch in case the rare slow tree-walk fallback is hit and go-git refuses to honor cancellation. `main.go`'s top-level signal handling is unchanged. A `ctx.Err()` check between candidates lets the first Ctrl-C exit cleanly between generations, and `context.Canceled` is mapped to `NewSilentError` so cancellation exits quietly without printing `Error: ...`.

## Trade-off, please read

The hot path was walking every archived checkpoint's full-tree via go-git and parsing each transcript JSONL to derive timestamps. That's where the multi-second-per-generation hang came from. Archived `/full/` refs intentionally don't carry per-session `metadata.json` (the split is: `/main` has metadata + prompts, `/full` has transcripts only), so `generation.json` at the tree root was the only cheap source — and the prior code consulted it only as a fallback when the tree walk found no timestamps.

This PR **inverts the preference**: read `generation.json` first; fall back to the tree walk only when `generation.json` is missing or incomplete. `generation.json` is written at archival time and actively maintained by `generation_repair.go`, so in practice it is authoritative.

**Behavioral implications:**

| `generation.json` state | Outcome |
|---|---|
| Matches reality (normal case) | No observable difference; cleanup is orders of magnitude faster. |
| Stale/wrong, says "newer than actual" | Generation is skipped from cleanup even though eligible. **Safe direction** — data stays on disk longer than retention. |
| Stale/wrong, says "older than actual" | Generation is cleaned up before its real newest timestamp passes retention. **Unsafe direction** — data deleted prematurely. Mitigation: `generation_repair.go` keeps `generation.json` correct. |

The test `TestCleanCmd_All_UsesRawTranscriptTimeForV2GenerationRetention` previously pinned the opposite preference. It's renamed to `TestCleanCmd_All_PrefersGenerationJSONOverRawTranscript`, its assertion is flipped, and the inline comment documents the new contract.

If a user reports a generation deleted prematurely due to a wrong `generation.json`, the cheapest fix is to run the repair pass; the next-cheapest is to revert the preference inversion in `cleanup.go` (~6 lines).

## Test plan

- [x] `mise run fmt && mise run lint && mise run test` all clean (one pre-existing unrelated flake in `TestResolveAgentForRewind/dynamically_registered_agent_resolves_by_type` exists on `main` too)
- [x] Smoke test on a real repo with 18 archived v2 generations: previously hung indefinitely, now completes in ~1 second with no chatter on the output
- [x] Verified two-Ctrl-C escape hatch kills the process within 1 second if a go-git tree walk is in flight (rare slow-fallback path)
- [ ] Reviewer: confirm the trade-off documented above is acceptable for your usage


cmd/entire/main.go
```diff
@@ -28,6 +28,11 @@ func main() {
 	go func() {
 		<-sigChan
 		cancel()
+		// Second signal forces exit. Some long-running operations (e.g.
+		// go-git tree walks) don't honor context cancellation, so cancel()
+		// alone may not stop the process. Give the user an escape hatch.
+		<-sigChan
+		os.Exit(130)
```
computermode: hmm I'm not sure we want to have this exit at the top level here - I only intend to add it for `entire clean --all` for now... not sure what potential consequences there are for allowing it at a top level like this / if it was an intentional design choice to not respect ctrl+c for other commands.

computermode: Done — moved the second-SIGINT force-exit into a scoped helper (`installForceExitOnSecondSignal`) that only runs inside `runCleanAll`. `main.go` is back to the original single-signal cancel(); other commands are unaffected. Pushed in 74da9d509.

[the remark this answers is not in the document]
computermode: Fixed in 74da9d509 — the scoped signal channel in `installForceExitOnSecondSignal` uses buffer 2 so a fast second press isn't dropped.

cmd/entire/cli/clean.go
```diff
@@ -278,30 +278,48 @@ func runCleanAll(ctx context.Context, cmd *cobra.Command, force, dryRun bool) er
...
+		}()
+	} else {
+		close(v2Done)
+	}
+
 	// List all items (sessions, shadow branches) — not just orphaned ones
 	items, err := strategy.ListAllItems(ctx)
 	if err != nil {
+		<-v2Done
 		return fmt.Errorf("failed to list items: %w", err)
```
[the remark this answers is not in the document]
computermode: Fixed in 74da9d509 — `runCleanAll` now derives a context for the v2 goroutine via `context.WithCancel` and calls `cancelV2()` on the `ListAllItems` error path before waiting. The v2 goroutine's `ctx.Err()` check between iterations picks up the cancel on its next pass, so a failure in local enumeration no longer blocks on slow v2 work.

cmd/entire/cli/strategy/cleanup.go
```diff
@@ -363,22 +368,35 @@ func ListEligibleV2Generations(ctx context.Context, s *settings.EntireSettings)
 	cutoff := time.Now().AddDate(0, 0, -s.GetFullTranscriptGenerationRetentionDays())
 	cleanupItems := make([]CleanupItem, 0, len(candidates))
 
-	for _, candidate := range candidates {
+	for i, candidate := range candidates {
+		if err := ctx.Err(); err != nil {
+			return nil, warnings, fmt.Errorf("listing v2 generations cancelled: %w", err)
```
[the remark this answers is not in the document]
computermode: Fixed in 74da9d509 — the loop now returns `ctx.Err()` unwrapped (with a `//nolint:wrapcheck` justification), and `runCleanAll` maps `context.Canceled` to `NewSilentError` so Ctrl-C exits quietly with no "Error: ..." message.

cmd/entire/cli/strategy/cleanup.go
```diff
@@ -364,21 +364,31 @@ func ListEligibleV2Generations(ctx context.Context, s *settings.EntireSettings)
...
-		if !foundCheckpointTimes {
-			var genErr error
-			gen, genErr = store.ReadGeneration(treeHash)
-			if genErr != nil {
+		// Prefer the precomputed timestamps in generation.json — walking every
+		// checkpoint tree via go-git is expensive and generation.json (written
+		// at archive time and repaired by generation_repair.go) is the
+		// authoritative timestamp source in practice. The tree walk remains as
+		// a fallback for older generations whose generation.json is missing or
+		// incomplete.
```
pfleidi: That's a very good point! Especially given how costly the repair step during the migration is 😓 👍

cmd/entire/cli/clean.go
```diff
@@ -278,33 +279,87 @@ func runCleanAll(ctx context.Context, cmd *cobra.Command, force, dryRun bool) er
...
+
 	return runCleanAllWithItems(ctx, cmd, force, dryRun, items, tempFiles)
 }
 
+// installForceExitOnSecondSignal arms a goroutine that exits the process with
+// code 130 if the user presses Ctrl-C twice while this scope is active. The
+// first signal continues to be handled by the top-level cancel() in main.go.
+// Buffer size 2 so a fast second press isn't dropped by signal.Notify.
+// Returns a stop function that the caller MUST defer.
+func installForceExitOnSecondSignal(ctx context.Context) func() {
```
pfleidi: Are we sure this is necessary? I'd personally prefer if one `ctrl + c` would stop the clean command as soon as possible. I think we should ensure that the context object from `main.go` is threaded through everywhere appropriately and checked like we're doing it here: https://github.com/entireio/cli/pull/1182/changes#diff-07318303542b6b6131c6919c8589b5880d3010571721b57d34c1b66fb081b92bR367-R369

pfleidi (dismissed):

pfleidi (approved):

PR #1194, opened 2026-05-12
Remove --force flag from `entire migrate --checkpoints v2`

computermode:
https://entire.io/gh/entireio/cli/trails/366

The flag was hidden and only used internally during early versions of the migration checkpoints command during testing and iteration. Its per-checkpoint prune was O(N × G) over archived generations (each checkpoint triggered tree rebuilds + commits on /main, /full/current, and every /full/<n>), turning real-world reruns into hour-long jobs vs. minutes for non-force.

Non-force semantics — skip v1 checkpoints already in v2, copy over the rest — already match the desired behavior, and the resume path still backfills missing /full/* artifacts and compact transcripts. If a future bug ships bad v2 data, recourse is a targeted repair, not a wholesale re-migration. Reintroducing force later remains an option.

Drops the flag, the force plumbing through runMigrateCheckpointsV2 / migrateCheckpointsV2 / migrateOneCheckpoint, and the prune helpers (pruneV2CheckpointForForce, pruneV2CheckpointRef,
pruneV2ArchivedCheckpointRef, addRecomputedGenerationJSON, pruneCheckpointFromRoot). Updates the post-push migration hint to stop suggesting --force.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes checkpoint migration behavior by removing the ability to forcibly re-migrate/overwrite existing v2 data, which could affect operational recovery workflows (though the command is hidden and rerun/resume behavior remains).
> 
> **Overview**
> `entire migrate --checkpoints v2` no longer supports `--force`; the flag, its plumbing, and the expensive per-checkpoint v2 prune/rewrite helpers have been removed.
> 
> Migration now always treats existing v2 checkpoints as *already migrated*, only performing **resume repairs** (pack missing `/full/*` artifacts and backfill missing compact transcripts) and skipping otherwise; tests and the post-push migration hint were updated accordingly to stop referencing `--force`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c07c4705ece3b92066b6e656d0df1014f648bc7. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/migrate.go
+++ b/cmd/entire/cli/migrate.go
@@ -38,7 +38,6 @@ import (
 
 func newMigrateCmd() *cobra.Command {
 	var checkpointsFlag string
-	var forceFlag bool
 	var dryRunFlag bool
 
 	cmd := &cobra.Command{
@@ -52,9 +51,6 @@ func newMigrateCmd() *cobra.Command {
 			}
 			if checkpointsFlag != "v2" {
 				return fmt.Errorf("unsupported checkpoints version: %q (only \"v2\" is supported)", checkpointsFlag)
-			}
-			if dryRunFlag && forceFlag {
-				return errors.New("--dry-run and --force cannot be combined")
 			}
 
 			ctx := cmd.Context()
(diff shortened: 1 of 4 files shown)


computermode:
I'm going to just remove it from the user-facing messaging for now and then we can see about removing it later just in case we'll need it for the generation ref logic changes coming up.

Issue #1195, opened 2026-05-12
Search returns 0 results for Claude Code checkpoints despite successful capture (CLI 0.6.1 + 0.6.2-nightly)


WrangleMeThis:
### Summary

In a real production monorepo with Entire installed via Path B (Entire wrapping Lefthook), capture works end-to-end for Claude Code sessions — trailers are attached, checkpoints land in `entire/checkpoints/v1`, the branch pushes to a separate private checkpoint remote, `entire activity` / `entire dispatch` / `entire checkpoint explain` all see the data — but `entire search` **and** `entire checkpoint search` **return 0 results for every query**, including queries that match exact content from the captured transcripts.

This is the Claude Code variant of [#1171](<https://github.com/entireio/cli/issues/1171>) (Codex variant, same author, brand-new repo repro). Filing separately because:

* Different agent (Claude Code, not Codex)
* Different repo profile (real multi-package monorepo, not a clean canary)
* Same end-state regardless: search is empty while every other surface sees the data

### Environment

* Entire CLI: pre-flight done on `0.6.2-nightly.202605090651.30e58ea3`, production rollout on stable `0.6.1`
* macOS arm64
* Agent: Claude Code ([claude.ai](<http://claude.ai>) / Claude Code CLI)
* Lefthook 1.13.6 (project) / 2.1.6 (system PATH)
* Hook architecture: Path B — Entire wraps Lefthook via `entire configure --force` in `prepare` script in `package.json`
* Separate private checkpoint remote: `agiterra/tankloop-checkpoints-private`
* GitHub App: Entire installed on `agiterra` org with `repository_selection=all`

### What works

* `entire status` — Enabled, Claude Code + Codex agents active
* Capture: every commit gets an `Entire-Checkpoint: <id>` trailer
* Push: `entire/checkpoints/v1` syncs to the checkpoint remote on every `git push`
* `entire checkpoint list` — shows local checkpoints (2 on `main`)
* `entire checkpoint explain <id>` — shows the full transcript, prompts, file list, tokens
* `entire activity` — shows 11 checkpoints across 3 repos, including all of today's `agiterra/tankloop` work, with proper agent attribution (Claude Code 64% / Codex 36%)
* `entire dispatch` — generates digests citing real repo activity
* PII redaction working in captured transcripts (verified: real emails replaced with `REDACTED_EMAIL`)

### What fails

All search variants return empty:

```
$ entire search "lefthook" --json
[...]

Issue #1171, opened 2026-05-09
Search stays empty after successful Codex & Claude checkpoint capture in clean repo repro


WrangleMeThis:
### Summary

In a brand-new clean GitHub repo, Entire successfully captures a Codex checkpoint, injects an `Entire-Checkpoint` trailer, pushes `entire/checkpoints/v1`, and shows the checkpoint in both `entire activity` and `entire dispatch`, but `entire search` / `entire checkpoint search` still return 0 results for an exact unique token from the committed file.

This repro was done after normalizing git identity and using a strict order of operations to avoid mixed authors, preexisting branches, or GitHub-generated starter commits.

### Clean repro

Environment:
- Entire CLI 0.6.1
- macOS arm64
- Agent: Codex CLI

Order of operations:
1. Set git identity to `Brian Sweet <brian@agiterra.io>`.
2. Create a brand-new local repo with `git init -b main`.
3. Make the initial commit locally: `chore: initial empty repo`.
4. Create the GitHub repo from that local history and push it.
5. Run `entire enable --agent codex --force --telemetry=false`.
6. Start Codex inside the repo.
7. Have Codex create a plain text file with a unique token, commit it, keep the hook-added trailer, push, and exit cleanly.

Controlled repo:
- `BrianSweet/entire-orderops-local-canary-20260509-130023`

Controlled canary commit:
- commit: `df6c7e091ac2493aea605da737cc6780ffca6173`
- trailer: `Entire-Checkpoint: bee65c2cc740`
- committed token: `zz-entire-orderops-local-canary-20260509-f6`

### What works

- `git show -s --format=%B HEAD` shows the trailer
- `git ls-remote --heads origin main entire/checkpoints/v1` shows both branches on origin
- `entire checkpoint list` shows checkpoint `bee65c2cc740`
- `entire checkpoint explain --commit df6c7e091ac2493aea605da737cc6780ffca6173 --short --no-pager` succeeds
- `entire dispatch --since 1d --repos BrianSweet/entire-orderops-local-canary-20260509-130023` sees the canary test
- `entire activity` shows the repo and the Codex checkpoint/commit

### What fails

Both of these return 0 results for the exact token from the committed file:

```bash
entire search "zz-entire-orderops-local-canary-20260509-f6" --json

entire checkpoint search "zz-entire-orderops-local-canary-20260509-f6" \
  --json --repo BrianSweet/entire-orderops-local-canary-20260509-130023
```

Observed result from both:

```json
{
[...]

PR #1210, opened 2026-05-14
feat(search): add --local fallback that greps the local checkpoint branch

timon0305:
## Summary

Adds `--local` to `entire search` / `entire checkpoint search`. Skips the remote service and does case-insensitive substring matching over the local `entire/checkpoints/v1` branch (prompts, transcripts, file paths, branch name).

Refs #1195, #1171. The remote indexing bug isn't fixed here — but users with confirmed-captured checkpoints (visible in `activity` / `dispatch` / `explain`) get an immediate working path while it's diagnosed.

- Honors `--branch`, `--json`, piped/accessibility output.
- Reuses the existing JSON shape so scripted consumers don't branch on source.
- Skips auth entirely (works offline).
- Out of scope: TUI, `--repo *`, author/date filters.

## Test plan

- [x] 6 new unit tests pass (transcript token, files-touched, branch filter, empty query, no match, snippet windowing)
- [x] `go build ./...`, `go vet`, `gofmt` clean
- [x] E2E against the built binary on a seeded fixture repo: JSON output, branch filter (`main` and `feature`), empty query, no-match (`total: 0`), case-insensitive match, top-level `entire search --local` alias

### DIFF
--- a/cmd/entire/cli/search_cmd.go
+++ b/cmd/entire/cli/search_cmd.go
@@ -169,6 +169,7 @@ prompts, transcripts, file paths, and the checkpoint branch name.`,
 
 			// JSON output: explicit flag or piped/redirected stdout
 			if jsonOutput || !isTerminal {
+				maybePrintLocalFallbackHint(ctx, cmd.ErrOrStderr(), resp, query)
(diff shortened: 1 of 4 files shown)


timon0305:
@claude review

PR #1214, opened 2026-05-14
feat(redact): OpenAI Privacy Filter with pre-push architecture

peyton-alt:
https://entire.io/gh/entireio/cli/trails/401

## Summary

Adds an opt-in 8th redaction layer — the [OpenAI Privacy Filter](https://github.com/openai/privacy-filter) (`opf`) binary, a 1.5B-parameter token-classification model that catches names, emails, phone numbers, addresses, dates, URLs, account numbers, and secret-shaped values the 7 always-on regex layers miss. Runs **only at push time**, gated by an interactive Yes/No/Always prompt; commits stay on the fast 7-layer pipeline so per-turn latency is unchanged.

## What's in this PR

**Core OPF integration:**
- `opf` binary shellout via a generic `opfRuntime` interface (production: `\x1e`-joined batched stdin; tests inject fakes)
- Settings schema under `redaction.openai_privacy_filter.{enabled,categories,command,timeout_seconds,prompt_default}`
- Eight categories: `private_person`, `private_email`, `private_phone`, `private_address`, `private_url`, `private_date`, `account_number`, `secret`
- Process-scoped circuit breaker: first runtime failure disables OPF for the rest of the invocation
- Unknown category names rejected at settings load time
- JSON-aware leaf extraction with within-blob batching (one shellout per blob, all prose-shaped leaves)

**Pre-push architecture:**
- Post-commit emits 7-layer-only blobs to local `entire/checkpoints/v1`; pre-push rewrites those blobs with OPF, builds new commits carrying `Entire-OPF-Applied: true` trailer, CAS-updates the local v1 ref, then pushes
- Idempotent re-parenting: commits already tagged are not re-redacted on retry
- Shard scoping: only redacts files inside the commit's own shard (`<id[:2]>/<id[2:]>/*`) — files outside the shard are byte-copied since they belong to other commits' shards
- Fail-closed extension policy: every regular file inside the shard (except `content_hash.txt`) flows through redaction; future blob types (`.md`, no-extension) are covered by default rather than slipping through

**Interactive prompt + env override:**
- `huh` Yes/No/Always prompt (Dracula theme, matches existing CLI styling) before pre-push rewrite
- `ENTIRE_OPF=yes|no` env var overrides per-push
- `prompt_default: "always"` persists to `.entire/settings.local.json` when the user picks Always
- Non-TTY contexts (CI, agent-driven pushes) auto-run with a stderr progress notice

**Sentinel errors (fail-closed):**
- `V1DivergedError` — local v1 has commits not on remote (refuses to rebase)
- `BootstrapTooLargeError` — first push with too many unpushed commits (configurable via `ENTIRE_OPF_BOOTSTRAP_LIMIT`)
- `V1RefMovedError` — another worktree advanced the ref mid-rewrite (CAS conflict)
- `OPFRuntimeFailedError` — OPF binary failed mid-rewrite; push aborts before tagging anything

**Shadow-branch cleanup:**
- After successful push, vestigial shadow branches (ended sessions, no pending turn checkpoints) are auto-deleted
- Conservative safety predicate: active sessions or pending finalize protect their shadow

**Privacy contract:**
- Short-return from the OPF runtime trips the circuit breaker + aborts the rewrite (defense against silent under-redaction)
- Fail-closed everywhere: any runtime failure aborts the push rather than tagging 7-layer content as 8-layer-applied

## Why this shape

OPF inference is ~6s of CPU model-load per call. Earlier iterations ran OPF at commit time, which meant every commit in an agent loop paid that cost — terrible UX for sessions that produce 10+ commits before pushing. Moving OPF to pre-push aligns the cost with the user-visible "content leaves the machine" moment: one prompt per push, one wait, opt-out per push via env var.

## Test plan

- [x] `mise run check` (fmt + lint + test:ci + Vogon canary + roger-roger external-agent canary) — green
- [x] Real-agent smoke test on a fresh repo with `private_person` + `private_email` enabled
- [x] End-to-end pre-push rewrite verified against a real `opf` binary
- [x] All four sentinel error paths exercised by integration tests
- [x] Fail-closed contract tests:
  - Short-return runtime trips breaker + aborts
  - Mid-rewrite runtime failure aborts before CAS
  - Diverged remote refuses to proceed
  - Bootstrap cap respects `ENTIRE_OPF_BOOTSTRAP_LIMIT`
- [x] Privacy contract tests:
  - `.md` and no-extension blobs inside a shard get redacted (slip-through prevention)
  - Idempotent re-run preserves trailer + tree


peyton-alt:
@bugbot review

cmd/entire/cli/strategy/manual_commit_opf_rewrite.go
```diff
@@ -0,0 +1,568 @@
...
+		tree, err := repo.TreeObject(oldCommit.TreeHash)
+		if err != nil {
+			return plumbing.ZeroHash, fmt.Errorf("load tree: %w", err)
+		}
+		// Parse the shard path from the commit subject. Falls back to
+		// "" (walk everything) for bootstrap commits and unrecognized
+		// subjects — the conservative default still produces correct
+		// output, just slower.
+		shardPath := parseShardPathFromCommitMessage(oldCommit.Message)
+		newTree, err = rebuildTreeWithOPF(ctx, repo, tree, "", shardPath)
```
Soph: **🔴 [Critical] Shard-scoping under-redacts every checkpoint except the newest in a multi-commit push**

`rebuildV1Commit` rebuilds each commit's tree from `oldCommit.TreeHash` (the original 7-layer tree), redacting only that commit's own shard and byte-copying the rest. The comment assumes out-of-shard files are "already redacted (prior OPF-applied push)" — but that only holds when pushing one checkpoint at a time.

For a push of 2+ new checkpoints A,B: `WriteCommitted` builds B on A's *original* (7-layer) tree. The rewrite redacts shard_A in A′ and shard_B in B′, but B′ byte-copies shard_A from B's original tree = **7-layer**. The pushed tip B′ references the 7-layer shard_A blob, which `git push` uploads. The OPF'd shard_A blob lives only inside commit A′, not at the tip.

In the designed "many commits then one push" workflow (this PR's own motivation), only the single most-recent checkpoint per push is OPF-redacted at the tip; all others ship at 7-layer. Bootstrap / opt-in-after-history is the same defect at its worst. A fully-functioning runtime still tags these commits `Entire-OPF-Applied: true`.

**Fix:** rebuild each commit on its *rewritten parent's* tree (carry forward already-OPF'd shards, redact only the new shard), or redact all shards. Add a multi-commit test asserting the tip tree has every shard OPF-applied (current tests cover only single commits + error counts).

cmd/entire/cli/strategy/cleanup.go
```diff
@@ -118,6 +118,77 @@ func ListShadowBranches(ctx context.Context) ([]string, error) {
...
+	states, err := ListSessionStates(ctx)
+	if err != nil {
+		return 0, fmt.Errorf("list session states: %w", err)
+	}
+
+	// Build a set of shadow branch names that must be preserved
+	// because at least one session still depends on them.
+	protected := map[string]bool{}
+	for _, s := range states {
+		if s.EndedAt != nil && len(s.TurnCheckpointIDs) == 0 {
```
Soph: **🔴 [Critical] CleanupPushedShadowBranches can permanently delete un-condensed checkpoints**

The predicate `EndedAt != nil && len(TurnCheckpointIDs) == 0` is treated as "safe to delete", but `TurnCheckpointIDs` is populated only *after* condensation (PostCommit) and cleared at turn end (session/state.go:155-162). It is empty for a session that did work (checkpoints on its shadow branch) but was never `git commit`-ted — i.e. `EndedAt!=nil`, `TurnCheckpointIDs==[]`, `FullyCondensed==false`, with its only checkpoint copy on the shadow branch.

Sequence: agent works → agent stops (EndedAt set) → user pushes **any** branch before committing → cleanup (runs on every successful push) deletes the shadow branch → checkpoints lost.

The proven-safe predicate already exists: `manual_commit_hooks.go:981,996` gate on `FullyCondensed && Phase == PhaseEnded`. This substitutes a weaker proxy, and `cleanup_pushed_shadow_test.go` only models ended/finalized fixtures, so it encodes the buggy contract.

**Fix:** gate on `FullyCondensed && Phase == PhaseEnded`; add a fixture for the ended-but-not-condensed state.

redact/opf.go
```diff
@@ -0,0 +1,463 @@
...
+//     span boundaries)
+const opfBatchSeparator = "\x1e"
+
+// Redact runs OPF on a single text input.
+func (s *shellOut) Redact(ctx context.Context, text string, categories []string) ([]Span, error) {
+	if len(categories) == 0 {
+		return nil, nil
+	}
+	batch, err := s.RedactBatch(ctx, []string{text}, categories)
+	if err != nil || len(batch) == 0 {
```
Soph: **🟠 [High] Fail-closed short-return guard missing on the single-input OPF path**

The "runtime returned fewer spans than inputs → trip breaker" guard exists only on the batched JSONL path (`redact/redact.go:674-678`). The single-text path (`BytesWithPrivacyFilter` → `StringWithPrivacyFilter` → `detectOPF`) trips the breaker only on an outright *error*. A runtime returning `(emptySpans, nil)` yields 7-layer output, `OPFBreakerTripped()` stays false, the commit is tagged `Entire-OPF-Applied: true`, and future pushes skip it.

Reachable in-tree: the fail-closed extension policy (`checkpoint/committed.go:1989-1992`) routes every non-`.json/.jsonl` blob in the shard (and JSON-parse-failure fall-through) through this path. Today's production `shellOut` returns a parsed set or an error so it doesn't hit this, but it's a latent fail-open hole gated only on the `opfRuntime` contract. `Redact()` also folds `len(batch)==0` into the `err!=nil` branch here, hiding it.

**Fix:** mirror the completeness check into `detectOPF`/`Redact`, or route all in-shard blobs through the batched path; document the `opfRuntime` contract.

redact/opf.go
```diff
@@ -0,0 +1,463 @@
...
+	defer cancel()
+
+	var buf strings.Builder
+	starts := make([]int, len(inputs))
+	for i, in := range inputs {
+		if i > 0 {
+			buf.WriteString(opfBatchSeparator)
+		}
+		starts[i] = buf.Len()
+		buf.WriteString(strings.ReplaceAll(in, "\n", " "))
```
Soph: **🟠 [High] Record-separator (0x1e) inside a leaf desyncs the OPF batch protocol**

Inputs are joined with 0x1e after replacing only newlines (this line). The comment claims 0x1e "doesn't appear in real text", but leaves are arbitrary agent-transcript content (tool output, pasted files, terminal dumps with control bytes). An embedded 0x1e injects a spurious record boundary, so that leaf is mis-classified or its spans dropped at the boundary — no error, no breaker trip. Other leaves stay correct (offsets are byte-accurate), so it's a per-leaf fail-open.

**Fix:** strip/replace 0x1e (ideally all C0 control bytes) symmetric with the newline flattening; add a test.

cmd/entire/cli/strategy/cleanup.go
```diff
@@ -118,6 +118,77 @@ func ListShadowBranches(ctx context.Context) ([]string, error) {
...
+func CleanupPushedShadowBranches(ctx context.Context) (int, error) {
+	branches, err := ListShadowBranches(ctx)
+	if err != nil {
+		return 0, fmt.Errorf("list shadow branches: %w", err)
+	}
+	if len(branches) == 0 {
+		return 0, nil
+	}
+
+	states, err := ListSessionStates(ctx)
```
Soph, line 163: **🟠 [High] Cleanup read-modify-delete TOCTOU vs concurrent session start**

`cleanup.go` snapshots `ListShadowBranches`/`ListSessionStates` (this read) then runs `DeleteShadowBranches` against the stale snapshot. The CLI supports concurrent sessions per worktree; a session starting in that window (new shadow branch / state file) is not in `protected` and can be deleted.

**Fix:** re-validate per-branch immediately before deleting, gate under the session-state lock, or document the limitation.

cmd/entire/cli/trailers/trailers.go
```diff
@@ -270,3 +283,31 @@ func AppendCheckpointTrailer(message, checkpointID string) string {
...
 }
+
+// HasOPFApplied reports whether the commit message carries an
+// `Entire-OPF-Applied: true` trailer. Any other value (or absence) is
+// treated as "OPF not applied" so the pre-push rewrite considers the
+// commit a candidate for OPF redaction. Pinning the value to literal
+// "true" — rather than just trailer presence — prevents a future
+// "Entire-OPF-Applied: false" or "skipped" from accidentally meaning
+// "yes, applied."
+func HasOPFApplied(commitMessage string) bool {
```
Soph, line 293: **🟡 [Medium] Unanchored OPF trailer regex enables a privacy-filter bypass**

`opfAppliedTrailerRegex` matches `Entire-OPF-Applied:\s*(\S+)` anywhere in the message (not just the trailer block). `entire/checkpoints/v1` commit messages carry condensed session data; content containing the literal `Entire-OPF-Applied: true` makes `HasOPFApplied` (this func) return true, skipping OPF while marking the commit applied. The `false`-injection direction is safe (re-triggers OPF); only `true`-injection bites.

**Fix:** anchor with `(?m)^` or parse only the trailer block.

cmd/entire/cli/settings/settings.go
```diff
@@ -896,9 +960,82 @@ func mergeRedaction(dst *RedactionSettings, data json.RawMessage) error {
...
+			return err
+		}
+	}
 	return nil
 }
 
+// validateOPFSettings rejects unknown category names so typos surface at
+// parse time. Silent zero-detection of a privacy category is effectively
+// a correctness bug — the user thinks they're protected but they're not.
+func validateOPFSettings(opf *OPFSettings) error {
```
Soph, line 983: **🟡 [Medium] `timeout_seconds` is not validated at the settings layer**

`validateOPFSettings` validates categories and `prompt_default` but not `timeout_seconds`. Negative/zero/garbage values are silently coerced to 30s in `redact.ConfigurePrivacyFilter`, contrary to the "surface typos at parse time" philosophy stated in this function. No test covers timeout values.

**Fix:** reject `< 0` (or at least log) for parity with the category-validation rationale.

cmd/entire/cli/strategy/manual_commit_opf_rewrite.go
```diff
@@ -0,0 +1,568 @@
...
+// entire history as a "bootstrap" since URL-based remotes have no
+// tracking refs locally.
+//
+// Returns ZeroHash with no error when the remote has no v1 yet (genuine
+// bootstrap case). Fetch failures fall back to ZeroHash + a warning
+// log; the rewrite then treats the push as bootstrap rather than
+// blocking the user on a transient network issue.
+func resolveRemoteV1Tip(ctx context.Context, repo *git.Repository, target string) (plumbing.Hash, error) {
+	if !remote.IsURL(target) {
+		return readV1Tip(repo, plumbing.NewRemoteReferenceName(target, paths.MetadataBranchName))
```
Soph: **🟡 [Medium] Stale remote-tracking ref weakens v1 divergence detection**

`resolveRemoteV1Tip` reads `refs/remotes/<target>/entire/checkpoints/v1` (this line) without fetching. A stale tracking ref means the `V1DivergedError` check compares against an outdated remote tip. The server's non-fast-forward rejection is a backstop, but the local divergence guarantee is weaker than the comment implies.

cmd/entire/cli/strategy/manual_commit_opf_prompt.go
```diff
@@ -0,0 +1,186 @@
...
+	if err := writeSubObject(raw, "redaction", redactionRaw); err != nil {
+		return err
+	}
+	if err := settings.SaveLocalRaw(path, raw); err != nil {
+		return fmt.Errorf("save local settings: %w", err)
+	}
+	return nil
+}
+
+func readSubObject(parent map[string]json.RawMessage, key string) map[string]json.RawMessage {
```
Soph: **🟢 [Low] persistOPFPromptDefaultAlways can clobber sibling redaction config**

`readSubObject` (this func): a malformed `redaction` value in `settings.local.json` causes it to return an empty map; the subsequent write overwrites the entire `redaction` slot, destroying sibling config (`custom_redactions`, PII settings). Narrow trigger (already-malformed file), warnings-only failure path. Log before overwriting so it's diagnosable.

cmd/entire/cli/checkpoint/committed.go
```diff
@@ -1882,7 +1883,13 @@ func (s *GitStore) copyMetadataDir(metadataDir, basePath string, entries map[str
...
 		}
 
-		// Create blob from file with secrets redaction
+		// Create blob from file with 7-layer secrets redaction.
+		// Post-commit emits 7-layer-only blobs; the pre-push rewrite
+		// (strategy/manual_commit_opf_rewrite.go) walks the resulting
+		// tree, re-redacts these blobs with OPF when enabled, and
+		// rewrites entire/checkpoints/v1 into 8-layer commits before
+		// they leave the local machine.
+		_ = ctx // ctx not needed by the 7-layer path; kept on caller signature for future use
```
Soph: **🟢 [Low] Dead ctx plumbing in copyMetadataDir / createRedactedBlobFromFile**

`copyMetadataDir` threads `ctx` then discards it (this line), while `createRedactedBlobFromFile` hardcodes `context.Background()` into `RedactBlobBytes`. Harmless today (this path uses `usePrivacyFilter=false`) but misleading. Either thread the real ctx through to `RedactBlobBytes` or drop the unused parameter.

redact/opf.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+	batched := buf.String()
+
+	cmd := s.commandRunner(callCtx, s.command,
+		"--device", "cpu",
+		"--output-mode", "typed",
+		"--format", "json",
+		"--no-print-color-coded-text",
+	)
+	cmd.Stdin = strings.NewReader(batched)
```
Soph, line 380: **🟢 [Low] Unbounded OPF stdin/stdout buffers and single whole-batch timeout**

The full batched string is fed to `cmd.Stdin` (this line) and Stdout/Stderr are captured into `bytes.Buffer` with no size cap; a pathological transcript or a misbehaving `opf` can OOM the CLI. The timeout is also a single budget for the entire batch, so a large-but-valid transcript can hard-block pushes (fail-closed → push aborts) with no partial progress. Availability concern; reasonable as a follow-up.

Soph (commented):
## Review — OpenAI Privacy Filter (pre-push architecture)

Posting as inline comments (this PR's branch `feat/openai-privacy-filter` has no Entire trail of its own; the linked trail 401 is the merged `feat/opf-prepush`). The architecture is solid — clean `opfRuntime` abstraction, no shell injection, well-designed sentinel errors, strong happy-path tests. But there are **two privacy-correctness defects that silently defeat the feature's promise** plus one **data-loss bug**, none caught by existing tests (the tests encode the buggy contracts).

**Verdict: 🔴 Request changes.** Recommend blocking on the two Critical findings + the High short-return guard; the rest can be follow-ups. See inline comments for: C1 multi-commit shard-scoping under-redaction, C2 cleanup data loss, H1 short-return guard, H2 0x1e desync, H3 cleanup TOCTOU, M1 trailer-regex bypass, M2 timeout validation, M3 stale tracking ref, L1–L3.

**🟡 [Medium] (not inlineable — pre-existing code) Recovery path `cherryPickOnto` preserves redaction+trailer but is untested.** `metadata_reconcile.go:316` / `createCherryPickCommit:411` copy the commit message verbatim and reference already-redacted OIDs, so the mechanism is sound — but there's no OPF-then-recovery integration test pinning the redaction/trailer-preservation invariant across the reconcile/recovery path. Add one.

**Done well:** no shell injection (`exec.CommandContext(name, fixedArgs...)`); `HasOPFApplied` pins to the literal `"true"`; settings reject unknown categories at parse time; breaker is `atomic.Bool` + CAS; `RedactBlobBytes` never returns plaintext on failure. The fail-closed *intent* is right throughout — the gaps are specific paths that bypass it.

Soph (approved):

PR #1229, opened 2026-05-19
feat(attach): add --target-repo and --dry-run flags for skill orchestration

alishakawaguchi:
## Summary

Adds two flags to `entire session attach` to enable agent sessions running outside an Entire-enabled repo (or across multiple repos) to be linked into the affected Entire-enabled repos without `cd`'ing or guessing what will be amended.

- `--target-repo <path>`: makes attach operate on the given repo instead of cwd. All downstream cwd-coupled lookups (`openRepository`, `git rev-parse --git-common-dir`, session state store path) resolve against the target. The session group's `PersistentPreRunE` skips its cwd-based worktree-root check when this flag is set; the chdir + restore happens inside attach's `RunE` via a deferred restore.
- `--dry-run [--json]`: emits the planned action, checkpoint id, transcript size, target repo, and agent without writing checkpoint metadata, saving session state, or amending HEAD. The short-circuit fires **before** the transcript is read into memory (uses `os.Stat` for size) and **before** `logging.Init` opens a log file — so a multi-repo orchestrator preview stays fast and side-effect-free. JSON goes to stdout for clean piping; warnings stay on stderr.

## What this solves

Today the user runs an agent from a higher-level folder (e.g. `~/Projects/devenv`) and the agent makes changes across multiple child repos. None of the child repos' commits get linked to that agent session, because Entire is repo-scoped and the parent-level launch leaves no session state in the children.

The skill consuming these flags is in entireio/skills#26 (`session-crosslink`) — it does the discovery and orchestration: find the running agent's session id and transcript, infer affected child repos from `files_touched` plus optional explicit input, call `attach --dry-run --json` per repo to render a preview, then on confirmation re-run without `--dry-run` to amend each child repo's HEAD with the checkpoint trailer.

Investigation that motivated this: from a non-Entire-enabled parent, `entire session list --json` returns `[]` because session state is stored under each repo's git common dir; the parent has no store. Once a child repo has the session id, `attach` already works — it resolves the transcript via the agent runtime's filesystem layout, not via any prior session store entry. The missing piece was a way to call attach against a specific repo (not cwd) and to preview before amending.

## Action values

`--dry-run --json` emits one of:

- `would_add_trailer` — HEAD has no checkpoint trailer; a new checkpoint will be created and the commit amended.
- `would_link_existing_in_head` — HEAD already has an `Entire-Checkpoint:` trailer; the session is added to that existing checkpoint.
- `would_skip_existing_in_state` — the session is already linked to a checkpoint in this repo's session store; no-op.

## Test plan

- [x] New integration test: `--target-repo` amends the target repo's HEAD when CLI is run from an unrelated cwd, and session state lands in the target (not cwd).
- [x] New integration test: `--dry-run --json` returns valid JSON on stdout, leaves HEAD hash + message unchanged, and writes no session state file.
- [x] New integration test: `--json` without `--dry-run` returns an explanatory error.
- [x] Existing `TestAttach_*`, `TestSession_*`, `TestReview_*` integration suites pass.
- [x] `cmd/entire/cli` unit tests pass.
- [x] `golangci-lint run --new-from-rev=main` reports 0 issues.

## Out of scope

- A separate `entire transcript locate` command — the skill currently hardcodes per-runtime transcript paths.
- Linking *all* commits in a child repo to the same session — attach links to HEAD (or `--commit`, which is a fast-follow if needed).
- Cross-store session id lookup (`--source-repo`) — not needed; attach already resolves transcripts via the agent runtime's filesystem, not via any source repo's session store.

### DIFF
--- a/cmd/entire/cli/attach.go
+++ b/cmd/entire/cli/attach.go
@@ -8,6 +8,7 @@ import (
 	"log/slog"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -19,6 +20,7 @@ import (
 	"github.com/entireio/cli/cmd/entire/cli/checkpoint/id"
 	"github.com/entireio/cli/cmd/entire/cli/checkpoint/remote"
 	"github.com/entireio/cli/cmd/entire/cli/interactive"
+	"github.com/entireio/cli/cmd/entire/cli/jsonutil"
 	"github.com/entireio/cli/cmd/entire/cli/logging"
 	"github.com/entireio/cli/cmd/entire/cli/paths"
 	"github.com/entireio/cli/cmd/entire/cli/session"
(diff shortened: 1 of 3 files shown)


alishakawaguchi:
Closing — agent-only approach is sufficient. The skill that motivated this PR (entireio/skills#26, \`session-crosslink\`) now uses the shipped CLI (\`cd <repo> && entire session attach <id> --force\`) and computes the preview locally by inspecting \`git log -1\`'s commit message and \`.git/entire-sessions/<id>.json\` — same logic attach uses internally, no flag dependency.

Tradeoffs the agent-only path accepts:
- A few extra bash calls per candidate to inspect git/filesystem state instead of one \`--dry-run --json\` round-trip.
- Drift risk: if attach's action-resolution logic changes server-side, the skill's preview can disagree until the skill is updated.

Both are acceptable for the one-shot use case. Re-opening this PR (or a variant) only makes sense if multi-repo skill performance becomes a concern, or if CI/scripting consumers want a structured \`--dry-run --json\` API.

PR #1233, opened 2026-05-20
docs: add Pi to agent listings, hook table, and architecture guides

godswillumukoro:
Adds the Pi coding agent to documentation where it was previously absent, matching the existing in-tree implementation at cmd/entire/cli/agent/pi/.

Changes:
- README.md: Add Pi to intro text, agent hook config table, and agent-specific limitations section (documenting tree-shaped sessions and active-branch resolution)
- docs/security-and-privacy.md: Add Pi to supported agents list
- docs/architecture/agent-integration-checklist.md: Add Pi to canonical export and file-based agent examples
- docs/architecture/agent-guide.md: Add Pi JSONL transcript format section (tree-shaped entries, parentId chains, pijsonl package) and Plugin File Pattern section (TypeScript extension, hook lifecycle, session-ID caching, transcript capture)


Soph (approved):
Thanks!

I really think we should word some places more generic so we don't have to always add the agent names, but until then 👍

godswillumukoro:
> Thanks!
> 
> I really think we should word some places more generic so we don't have to always add the agent names, but until then 👍

Indeed. A more generic approach would definitely scale better long-term.

Happy to help however I can. Let me know anytime 🙂

PR #1242, opened 2026-05-20
Checkpoints v1.1: store v1 metadata at a custom ref

computermode:
https://entire.io/gh/entireio/cli/trails/406

## Summary

- Introduces `strategy_options.checkpoints_version: 1.1` (number `1.1` or string `"1.1"`). When set, checkpoint metadata is stored at the custom ref `refs/entire/checkpoints/v1.1` instead of the branch `refs/heads/entire/checkpoints/v1` — invisible to `git branch -a` and GitHub's branch UI, and not pulled by a default `git clone`. Same on-disk format, sharded paths, and trailers as v1; only the ref location differs. The `v1.1` suffix (vs. plain `v1`) keeps the custom ref distinct from the legacy branch so a single repo can hold both refs side-by-side while history is being migrated.
- Activation is **manual opt-in only**: a user edits `.entire/settings.json` to set `strategy_options.checkpoints_version` to `1.1`. `entire enable` does not auto-enroll.
- On existing v1 repos with prior history, the custom ref starts **empty** — past checkpoints on the legacy branch are NOT automatically reachable under 1.1. A future migration command can handle that; today, a user who wants their old history reachable can run, once:

  ```
  git update-ref refs/entire/checkpoints/v1.1 refs/heads/entire/checkpoints/v1
  ```

- Push hook refactored: `pushBranchIfNeeded` → `pushRefIfNeeded(localRef, trackingRef)`. The existing rebase-on-non-FF recovery flow now works at any ref namespace. `runEnable` idempotently installs `+refs/entire/checkpoints/v1.1:refs/entire/remotes/origin/checkpoints/v1.1` on origin when settings say 1.1, so `git fetch` updates the tracking ref.

**This is a starting point: v1.1 is the same as v1 right now, except for the custom ref location. Repos should not switch over to this version yet unless they are for testing purposes only.**

## Test plan

- [x] `mise run check` — unit + integration + e2e canary all pass; `golangci-lint` clean.
- [x] New unit tests:
  - `parseCheckpointsVersion` accepts `1.1` (float and string).
  - `EntireSettings.UsesCustomMetadataRef` true iff `checkpoints_version` is 1.1; `CheckpointsVersion()` still returns 1 (major) and `IsCheckpointsV2Enabled()` is false.
  - `checkpoint.MetadataRef` / `MetadataTrackingRef` resolve correctly per settings.
  - `installMetadataRefspec` is idempotent and no-ops without origin.
  - `runEnable` installs the refspec on a 1.1 repo, leaves it alone on a v1 repo.
- [x] New integration tests (`integration_test/v1_metadata_ref_workflow_test.go`):
  - Fresh repo with 1.1: custom ref created, legacy branch absent after first commit.
  - Legacy v1 regression: legacy branch is created, custom ref absent.
  - Push under 1.1: custom ref lands on the bare remote; legacy branch does not.
- [ ] Manual smoke: enable 1.1 on a real repo, run the one-time `git update-ref` above, verify `entire status` shows past checkpoints and a new commit lands at the custom ref.

## Notes for reviewers

- The custom ref is `refs/entire/checkpoints/v1.1` (the `v1.1` suffix distinguishes it from the legacy `refs/heads/entire/checkpoints/v1` branch so both can coexist during migration).
- **Legacy-branch history preservation was intentionally dropped from this PR.** An earlier iteration auto-copied the legacy branch tip onto the custom ref on first access (`checkpoint.PreserveV1History`); that was reverted. The custom ref now starts empty on existing repos, and bridging old history is a deliberate one-time `git update-ref` (see Summary) or a future migration command.
- The push refactor (`strategy/push_common.go`) is the largest mechanical change — every internal helper now takes `plumbing.ReferenceName` instead of `branchName string`. Behavior for legacy v1 branches is preserved by passing `NewBranchReferenceName(branch)` + `NewRemoteReferenceName("origin", branch)`.
- `EnsureMetadataBranch` and the two `GetMetadata*BranchTree` helpers gained a `context.Context` parameter so they can call the resolver. Callers updated mechanically.
- `paths.MetadataBranchName` is kept as-is and is still the source of truth for the legacy branch name; new ref names live next to it as `paths.MetadataRefName` and `paths.MetadataTrackingRefName`.
- `entire migrate v1 → 1.1` UX subcommand and `entire migrate cleanup-v1` are out of scope; today the path is "edit settings.json, then run anything." A future PR can wrap this in a friendlier flow.


computermode:
bugbot review

cmd/entire/cli/paths/paths.go
```diff
@@ -42,6 +42,31 @@ const (
 // MetadataBranchName is the orphan branch used by manual-commit strategy to store metadata
 const MetadataBranchName = "entire/checkpoints/v1"
 
+// MetadataRefName is the custom ref used when checkpoints_version is "1.1".
+// Lives under refs/entire/ for the same reasons V2*RefName does: invisible
+// in GitHub's branch UI and not fetched by default by `git clone`.
+const MetadataRefName = "refs/entire/checkpoints/v1"
```
pfleidi: I'm wondering if we should use `v1.1` here since v1 could be confusing when also dealing with `refs/heads/entire/checkpoints/v1`. 🤔 


```suggestion
const MetadataRefName = "refs/entire/checkpoints/v1.1"
```

computermode: 100% agree, thanks for documenting it as a comment here since I've already had weird conflicts with "two" v1 refs locally lol

cmd/entire/cli/settings/settings.go
```diff
@@ -1137,6 +1136,48 @@ func (s *EntireSettings) IsCheckpointsV2Enabled() bool {
 	return ok && val
 }
 
+// Recognized `strategy_options.checkpoints_version` literals. Accepted as
+// `int`, `float64`, or `string` per the parsing rules in
+// parseCheckpointsVersion.
+const (
+	checkpointsVersion1FloatLit   = 1.0
```
pfleidi, line 1143: Do we need both versions here? Also: Should we skip the v2 constants since they're not necessary anymore?

computermode: Yes, good catch - I initially started work on this before v2 support was fully removed.

computermode:
Closing this since we decided on another approach

PR #1244, opened 2026-05-21
pre-push: speed up checkpoint sync

prichodko:
https://entire.io/gh/entireio/cli/trails/408

## Why
`git push` runs Entire's `pre-push` hook, which also publishes `entire/checkpoints/v1`. When that metadata branch is non-fast-forward, the hook fetches the remote branch and reconciles local checkpoint commits before retrying the push.

In the slow case I hit, local and remote checkpoint metadata had both advanced: 131 local-only commits and 145 remote-only commits. The old recovery path replayed each local commit onto the remote tip. For every replayed commit it flattened the commit tree, parent tree, and current tip tree through go-git, then rebuilt the full tree. On a packed repo this meant repeated pack reads/decompression and the hook was still running after 5+ minutes.

## Fix
Instead of replaying commit history one commit at a time, the sync path now treats checkpoint metadata as a tree merge:

- Fetch the remote checkpoint ref as before.
- Find the merge-base between local and fetched remote metadata.
- Compute the aggregate file-level delta from merge-base to local with `git diff-tree --raw`.
- Apply that sparse delta once onto the fetched remote tree with `checkpoint.ApplyTreeChanges`.
- Create one merge commit with remote tip as first parent and local tip as second parent, then retry the push.

This keeps the remote side authoritative for remote-only changes and deletions. The local side contributes only its net changes since the true merge-base, so old local merge commits do not reintroduce remote-only checkpoint shards that were deleted later on the remote.

## Risk / behavior change
The resulting checkpoint metadata history is a merge commit instead of a linear cherry-pick chain. That is intentional: this branch is storage metadata, and preserving every replayed local commit is less important than fast, correct convergence. Existing fast-forward and disconnected-branch recovery paths are unchanged.

## Validation
- `mise run check`
- `mise exec -- go test ./cmd/entire/cli/strategy -run 'TestFetchAndRebase|TestHasUnpushedSessionsCommon|TestPushBranchIfNeeded|TestDoPushBranch'`
- `mise run lint:gofmt`
- `mise run lint:go`
- Manual pre-push sync on the reproduced diverged metadata branch: stuck after 5+ minutes before, 13.9s after, 0.047s no-op rerun.

---

> [!NOTE]
> **Medium Risk**
> Changes checkpoint metadata branch sync semantics from cherry-pick rebases to a tree-diff-based merge commit, which could affect history shape and edge cases around deletions/conflicts. The logic shells out to `git diff-tree`/`merge-base`, so correctness depends on parsing and on git CLI output stability.
> 
> **Overview**
> Checkpoint metadata sync during pre-push now **merges** diverged local/remote metadata branches by computing file-level changes since the merge-base (via `git diff-tree --raw`) and applying them as a single tree update onto the fetched remote tip, producing a merge commit rather than a linearized history.
> 
> Adds parsing for raw `diff-tree` output (including deletions and file modes) and updates sync tests to assert merge-parent ordering and preserve safeguards that prevent resurrecting remote-deleted checkpoint shards.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 394a12a1c21e3200e738e9afed42dd388af1f641. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/strategy/push_common.go
+++ b/cmd/entire/cli/strategy/push_common.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/go-git/go-git/v6"
 	"github.com/go-git/go-git/v6/plumbing"
+	"github.com/go-git/go-git/v6/plumbing/filemode"
 	"github.com/go-git/go-git/v6/plumbing/object"
 )
 
@@ -299,9 +300,8 @@ func printProtectedRefBlock(w io.Writer, ref, target string) {
 	fmt.Fprintln(w, banner)
 }
 
-// fetchAndRebaseSessionsCommon fetches remote sessions and rebases local commits
-// on top of the remote tip. Since checkpoint shards use unique paths, rebases
-// always apply cleanly.
+// fetchAndRebaseSessionsCommon fetches remote sessions and merges local changes
+// on top of the remote tip.
 // The target can be a remote name or a URL.
 func fetchAndRebaseSessionsCommon(ctx context.Context, target, branchName string) error {
 	ctx, cancel := context.WithTimeout(ctx, 2*time.Minute)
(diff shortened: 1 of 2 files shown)


pjbgf:
Closing this due to #1248 being merged - additional performance improvements are still incoming.

PR #1256, opened 2026-05-22
Preserve diverged local refs in SafelyAdvanceLocalRef

Soph:
https://entire.io/gh/entireio/cli/trails/419

## Summary

- The previous `SafelyAdvanceLocalRef` contract overwrote any local ref the target couldn't reach, which silently discarded unpushed local commits whenever a fetch landed sibling commits — e.g. checkpoint metadata produced on another machine sharing the same orphan-style refs (`entire/checkpoints/v1`, the V2 main ref). The objects survived in the loose-objects pool until git gc, but the branch ref no longer pointed at them.
- Tighten the helper so it only sets the ref when the move is non-destructive:
  - missing → create
  - local at/ahead of target → no-op (existing protection)
  - local strictly behind target → fast-forward
  - diverged or unrelated history → no-op, logged at debug level
- All three production callers (`promoteRemoteTrackingMetadataBranch`, `FetchMetadataBranch`, `PromoteTmpRefSafely`) sync orphan branches where "skip on divergence" is the correct semantic. In the rare divergent case, resume falls through to remote-tracking-tree reads (v1) or the V1 fallback (DualCheckpointReader → v1).
- Update `TestFetchV2MainFromURL_UpdatesExistingRef` to use the existing `advanceV2MainOnTop` helper. The prior setup advanced the remote via a second call to `createV2MainRef` (which always produces an unrelated orphan commit). That setup was only "updating" the local ref via the now-removed diverged-overwrite path; a descendant commit on top of the previous tip is what the real condensation flow produces.

## Companion PR

Stacks naturally with #1252 (which fixes the immediate user-visible "session log not available" bug). This PR is the broader hardening that came out of reviewing that fix — protects unpushed local checkpoint commits against destructive overwrite across all fetch paths.

## Test plan

- [x] `mise run check` passes (fmt, lint, unit, integration, canary)
- [x] New tests in `strategy/safely_advance_local_ref_test.go` cover all six branches:
  - `LocalMissing_SetsToTarget`
  - `LocalEqualsTarget_NoOp`
  - `LocalAhead_NoOp` (existing protection)
  - `LocalBehind_FastForwards`
  - `Diverged_PreservesLocal` (new behavior — failed on `main`)
  - `UnrelatedHistory_PreservesLocal` (new behavior — failed on `main`)
- [x] Existing `TestFetchMetadataBranch_DoesNotRewindLocalAhead` still passes
- [x] Existing `TestFetchV2MainFromURL_DoesNotRewindLocalAhead` still passes
- [ ] Manual: simulate a diverged metadata branch (orphan commits on two machines), confirm fetch no longer rewinds local

---

> [!NOTE]
> **Medium Risk**
> Changes git ref promotion logic used by metadata/v2 ref fetch paths; incorrect ancestor detection could cause refs to stop updating (or to remain diverged) and impact resume behavior, though it reduces risk of losing unpushed local commits.
> 
> **Overview**
> Prevents fetch/promotion code from **overwriting orphan-style local refs** when the fetched target has *diverged or unrelated history*, preserving unpushed local checkpoint commits. `SafelyAdvanceLocalRef` now only creates, no-ops, or fast-forwards refs, and logs a debug message when it detects divergence instead of updating.
> 
> Adds a dedicated unit test suite covering missing/equal/ahead/behind/diverged/unrelated scenarios, and updates `TestFetchV2MainFromURL_UpdatesExistingRef` to advance the remote v2 ref via a true descendant commit so the test exercises the intended fast-forward path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ca40dd00d14a3262156c4ae061f6912f7b97eab7. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/strategy/checkpoint_remote_test.go
+++ b/cmd/entire/cli/strategy/checkpoint_remote_test.go
@@ -918,10 +918,12 @@ func TestFetchV2MainFromURL_UpdatesExistingRef(t *testing.T) {
 	require.NoError(t, err)
 	hash1 := strings.TrimSpace(string(hash1Out))
 
-	// Add a second commit on the remote's v2 ref
-	createV2MainRef(ctx, t, remoteDir) // Creates a new orphan commit, updating the ref
+	// Advance the remote's v2 /main with a child commit so the new tip is a
+	// strict descendant of hash1 (the realistic CLI flow — condensation
+	// appends a new checkpoint commit on top of the previous tip).
+	advanceV2MainOnTop(ctx, t, remoteDir, hash1)
 
-	// Fetch again — should update
+	// Fetch again — should fast-forward
 	require.NoError(t, FetchV2MainFromURL(ctx, remoteDir))
 
 	hashCmd = exec.CommandContext(ctx, "git", "rev-parse", paths.V2MainRefName)
--- a/cmd/entire/cli/strategy/common.go
+++ b/cmd/entire/cli/strategy/common.go
@@ -126,13 +126,24 @@ func PromoteTmpRefSafely(ctx context.Context, tmpRefName, destRefName plumbing.R
 	return nil
 }
 
-// SafelyAdvanceLocalRef updates localRefName to point at targetHash, except
-// when the existing local ref is already at or ahead of targetHash. In that
-// case it leaves the local ref unchanged to avoid rewinding locally-ahead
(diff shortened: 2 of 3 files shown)


Soph:
Folding into #1252 since the changes are tightly coupled — the global helper tightening makes the resume fix safer, but each fix is small on its own. Branch will be deleted.

PR #1263, opened 2026-05-26
Fix Cursor hook misattribution and add token usage support

### LINKED ISSUE #1262: Cursor inadvertently invokes Claude Code hooks when `.cursor/hooks.json` is not installed
### What happened?

When using Cursor as the primary agent **without** `.cursor/hooks.json` installed (e.g. only Claude Code hooks exist in `.claude/settings.json`), Cursor sessions are incorrectly attributed to Claude Code.
 
The root cause is the first-writer-wins ownership model in `DispatchLifecycleEvent`. When `.cursor/hooks.json` is present, Cursor fires `entire hooks cursor session-start` first, calling `StoreAgentTypeHint` to claim the session. Claude Code's hooks later fire for the same session ID but the ownership check (`state.AgentType != ag.Type()`) detects the mismatch and skips them — everything works correctly.
 
When `.cursor/hooks.json` is **absent**, no Cursor hook fires to claim the session. Claude Code (running as a Cursor subagent) reads its own `.claude/settings.json` hooks and fires `entire hooks claude-code session-start`, incorrectly claiming ownership of a session that belongs to Cursor. All subsequent lifecycle events are then processed under the wrong agent, producing incorrect checkpoints, session metadata, and step counts.
 
In short: **the fix for this is present in the codebase (the ownership check), but it only works when Cursor hooks are also installed.** If only Claude Code hooks are installed, there is nothing to claim the session on Cursor's behalf first.

### Steps to reproduce

1. Have Claude Code hooks installed in `.claude/settings.json` (`entire agent add claude-code` or `entire enable`).
2. Do **not** install Cursor hooks — `.cursor/hooks.json` should not exist.
3. Open a Cursor session in the repository. Cursor will invoke Claude Code as a subagent.
4. Observe that `entire hooks claude-code session-start` fires and claims the session.
5. Check `.entire/logs/` — the session is attributed to `claude-code`, not `cursor`.
6. Now install Cursor hooks (`entire agent add cursor`) so `.cursor/hooks.json` exists, and repeat. The session is now correctly owned by `cursor` and the Claude Code hooks are skipped.

### Entire CLI version

0.6.2

### OS and architecture

macos 26 arm64

### Agent

cursor

### Terminal

_No response_

### Logs / debug output

```bash

```

### Additional context

- A PR is in progress to fix this.
- Relevant code: the ownership check lives in `DispatchLifecycleEvent` (`cmd/entire/cli/lifecycle.go` ~line 55). Session ownership is set via `StoreAgentTypeHint` in `handleLifecycleSessionStart`.
- When working correctly, logs contain: `"skipping forwarded hook for non-owning agent"` with `owning_agent=cursor` and `firing_agent=claude-code`.
- The fix likely involves ensuring that when Cursor is detected as the IDE (e.g. via an env var or by the presence of Cursor-specific hook payload fields), the session is pre-claimed for Cursor even without `.cursor/hooks.json` — or that `entire enable` / `entire agent add` warns the user to install Cursor hooks when using Cursor.

### LINKED ISSUE #1264: Cursor sessions report zero token usage (stop hook token fields not parsed)
### What happened?

Cursor's `stop` hook payload includes per-turn token accounting fields (`input_tokens`, `output_tokens`, `cache_read_tokens`, `cache_write_tokens`), but the CLI does not parse them. As a result, `entire status` always shows 0 tokens for Cursor sessions, and checkpoint metadata has no token data.
Unlike Claude Code (which embeds `usage` in the JSONL transcript) or Codex (which has `token_count` events in the rollout), Cursor's JSONL transcript contains **no usage fields at all**. The stop hook payload is the only authoritative source.

### Steps to reproduce

`entire status` should display token counts for Cursor sessions, and checkpoint metadata should include token usage.

### Entire CLI version

0.6.2

### OS and architecture

_No response_

### Agent

cursor

### Terminal

_No response_

### Logs / debug output

```bash

```

## Root cause
- `stopHookInputRaw` in `cmd/entire/cli/agent/cursor/types.go` does not define `input_tokens` / `output_tokens` / `cache_read_tokens` / `cache_write_tokens`
- `parseTurnEnd` in `cmd/entire/cli/agent/cursor/lifecycle.go` does not populate `event.TokenUsage`
- `handleLifecycleTurnEnd` in `cmd/entire/cli/lifecycle.go` falls through to `agent.CalculateTokenUsage()` which returns nil for Cursor (no `TokenCalculator` implemented)
## Token derivation
Cursor reports `input_tokens` as the **total** (cache_read + cache_write + fresh). The correct derivation:

SnowingFox:
close https://github.com/entireio/cli/issues/1262
close https://github.com/entireio/cli/issues/1264 
image
# Dispatch — feat/cursor-hook-guard-and-tokens
> 2026-05-26, 1 commit, 1 checkpoint (2a89e658287e)
## Summary
This branch fixes Cursor session misattribution when only Claude Code hooks are installed (#1262), and adds token usage reporting for Cursor sessions by parsing the `stop` hook payload — the only authoritative source, since Cursor's JSONL transcript contains no usage fields.
### Bug Fix: Cross-agent hook forwarding guard (#1262)
- Added `shouldSkipForwardedHook()` in `hook_guard.go` — checks if `event.SessionRef` (transcript path) belongs to a different registered agent via `AgentForTranscriptPath`
- Wired the guard into `executeAgentHook()` in `hook_registry.go`, before `DispatchLifecycleEvent` — silently skips with a debug log when a forwarded hook is detected
- Covers the scenario: user has `.claude/settings.json` but no `.cursor/hooks.json`; Cursor IDE fires Claude Code hooks; transcript path proves it's a Cursor session
### Feature: Cursor token usage from stop hook
- Added `TokenUsage *TokenUsage` field to `agent.Event` — allows hook parsers to inject authoritative token data that the framework prefers over transcript-based computation
- Extended `stopHookInputRaw` in `cursor/types.go` with `input_tokens`, `output_tokens`, `cache_read_tokens`, `cache_write_tokens` fields
- Implemented `tokenUsageFromStop()` in `cursor/lifecycle.go` — derives fresh input tokens as `max(0, total_input - cache_read - cache_write)`, matching the story/apps/cli approach
- Modified `handleLifecycleTurnEnd` in `lifecycle.go` to prefer `event.TokenUsage` over `agent.CalculateTokenUsage()` fallback
## Tests
- `hook_guard_test.go`: transcript belongs to other agent (skip), transcript belongs to self (pass through), full `executeAgentHook` integration test confirming no state file created
- `lifecycle_test.go`: `TestHandleLifecycleTurnEnd_PrefersEventTokenUsage` — verifies hook-provided tokens flow into session state
- `cursor/lifecycle_test.go`: token extraction from stop hook, nil when omitted, negative clamp to zero


Soph:
@SnowingFox thanks for your contribution! I had a look and I don't think this works yet. The issue is that the current code would overwrite the tokens again at the end since the same logic kicks in as for every other agent and tries to calculate the token usage from the script - where there isn't any for cursor. I let claude make two tests on top of your branch that proof the token calculation would be overridden but also highlights that there is more work to make sure the we add the right token count to each checkpoint. Since we would always get the whole token count in the hook of the session until then. 

Let me know if you want to address this otherwise I'm also happy to take over your work. 

Here is the draft PR with the two tests: https://github.com/entireio/cli/pull/1388

SnowingFox:
> @SnowingFox thanks for your contribution! I had a look and I don't think this works yet. The issue is that the current code would overwrite the tokens again at the end since the same logic kicks in as for every other agent and tries to calculate the token usage from the script - where there isn't any for cursor. I let claude make two tests on top of your branch that proof the token calculation would be overridden but also highlights that there is more work to make sure the we add the right token count to each checkpoint. Since we would always get the whole token count in the hook of the session until then.
> 
> Let me know if you want to address this otherwise I'm also happy to take over your work.
> 
> Here is the draft PR with the two tests: #1388

Thanks for the detailed feedback and the tests! That makes sense. I'll dig into #1388 and work on fixing the override issue + per-checkpoint delta calculation. Will push updates to this branch.

SnowingFox:
@Soph thanks again for the detailed review and for putting together #1388.

I pushed an update to this PR. The branch now includes the two acceptance tests from #1388:

- `TestCursorTokenUsage_SurvivesCondensation`
- `TestCursorTokenUsage_PerCheckpointScoping`

The fix keeps Cursor hook-provided token usage as a pending checkpoint-scoped delta, uses it during condensation only when transcript-derived usage is absent, and clears it after successful condensation so the next checkpoint does not inherit cumulative totals. I also merged current `main` and kept both `Event.SkillEvents` and `Event.TokenUsage` through the conflict.

Verified locally:

- `go test -tags integration ./cmd/entire/cli/integration_test -run TestCursorTokenUsage -count=1`
- `mise run test:integration`
- `mise run test`
- `mise run lint`

Could you please take another look when you have a chance?

SnowingFox:
Hi @Soph , just a gentle follow-up on this PR.

I’ve pushed the update addressing your previous feedback: the branch now includes the two acceptance tests from #1388, fixes the Cursor token usage override/condensation issue, and keeps the token usage scoped per checkpoint.

I also rebased/merged the latest main and verified it locally with the integration tests, full test suite, and lint.

Could you take another look when you get a chance? No rush, I’d mainly like to know if there are any remaining blockers or changes you’d like me to make.

computermode (approved):

Soph:
@SnowingFox thanks for your contribution!

PR #1307, opened 2026-06-01
auth: keyring deadline + post-review polish for control-plane client

pjbgf:
https://entire.io/gh/entireio/cli/trails/461

## Summary

Bundles the OS-keyring deadline with a small set of review-pass fixes
for the freshly-merged control-plane client (#1299):

- **auth: cap OS-keyring calls with a context deadline.** A hung
  keychain helper (locked Keychain, paused libsecret prompt) would
  block `entire login`/`logout` indefinitely; wrap every keyring read
  and write with a 3s deadline and surface the timeout as a clear
  error rather than a frozen process.
- **coreapi: skip sessionAuth so requests carry no stray `Cookie`
  header.** `bearerSource.SessionAuth` previously returned an empty
  `SessionAuth{}` and the generated `securitySessionAuth` would
  unconditionally add `Cookie: entire_session=` to every outbound
  request. Return `ogenerrors.ErrSkipClientSecurity` instead; the
  bearer alone satisfies the OR-requirement and no cookie is sent.
  Locked in with a wire-level unit test.
- **grant org add: validate `--role` at the CLI boundary.** Mirrors
  `parseProjectOwnerType` with a `parseOrgRole` mapping
  owner/admin/member onto the generated enum, so typos fail fast
  with a clear local message instead of an opaque server 422.
  Project and repo `--role` stay free-form because the spec for those
  doesn't enumerate values.
- **mirror create/remove: reject dot-only owner/repo segments.** The
  tightened owner charset (from #1299) already excludes dots, but
  `gitHubRepoPat` admits `.` and `..` — both would embed a literal
  `..` in the `/gh/<owner>/<repo>` slug and the token-exchange
  audience. Reject at the boundary; two negative tests cover `.` and
  `..`.
- **mirror clone probe: drop `http.DefaultClient`, cap the body
  read.** `DefaultClient` is process-global, has no `Timeout`, and is
  poisonable; replace with a purpose-built `probeClient` (own
  Transport, 15s timeout, 10s TLS handshake, no redirect-follow), and
  wrap the body in `io.LimitReader(8 MiB)` to match the JSON path in
  `api/client.go`.
- **mirror clone probe: floor token re-mints so STS hiccups don't
  amplify.** Re-mint on 401 was unconditional, so a flapping STS
  during the 30-minute wait could mint ~900 tokens at the 2s probe
  cadence. Track `lastMint` and skip re-mint inside a 30s window;
  probes keep firing, auth load is bounded.

---

> [!NOTE]
> **Medium Risk**
> Changes touch credential storage timeouts, control-plane HTTP auth headers, STS token re-mint behavior, and URL validation used for token audiences—mostly defensive but in security-sensitive paths.
> 
> **Overview**
> Adds **bounded OS keyring I/O** so login/logout cannot hang forever: every keyring save/get/delete runs through a **5s** deadline (overridable via `ENTIRE_KEYRING_TIMEOUT`) with platform-specific timeout errors.
> 
> Hardens **control-plane and mirror paths**: the Core API client skips `sessionAuth` via `ogenerrors.ErrSkipClientSecurity` so outbound requests do not send an empty `entire_session` cookie; `entire grant org add` validates `--role` locally; GitHub URL parsing rejects **dot-only** repo segments; the mirror clone probe uses a dedicated HTTP client (timeouts, no redirects), caps `info/refs` reads at **8 MiB**, and **rate-limits** repo-scoped token re-mints after 401s.
> 
> Pins **go-licenses** to **v2.0.1** in the license lint task for reproducible CI.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 614d8450ccd8e94c7250fbbe3ce35ad70c854a1a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/auth/keyring_timeout.go
+++ b/cmd/entire/cli/auth/keyring_timeout.go
@@ -1,0 +1,73 @@
+package auth
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"time"
+)
+
+// defaultKeyringTimeout caps how long every OS keyring call may take.
+// On Linux the keyring talks D-Bus to the Secret Service; in headless
+// environments (SSH without a graphical session, containers, WSL) no
+// provider is running and the call blocks indefinitely, freezing the
+// CLI. 5s is comfortably longer than any healthy round-trip while still
+// surfacing the hang to the user quickly.
+const defaultKeyringTimeout = 5 * time.Second
+
+// keyringTimeoutEnvVar overrides defaultKeyringTimeout. Accepts any
+// time.ParseDuration string; invalid or non-positive values fall back
+// to the default. Useful on slow keyrings or to extend the wait on a
+// system where the secret service is just sluggish.
+const keyringTimeoutEnvVar = "ENTIRE_KEYRING_TIMEOUT"
+
+func keyringTimeout() time.Duration {
+	v := os.Getenv(keyringTimeoutEnvVar)
+	if v == "" {
+		return defaultKeyringTimeout
+	}
+	d, err := time.ParseDuration(v)
+	if err != nil || d <= 0 {
+		return defaultKeyringTimeout
+	}
+	return d
+}
+
+// newKeyringContext returns a context that cancels after the configured
+// keyring timeout. Callers must invoke the returned cancel to release
+// the timer regardless of which select branch wins.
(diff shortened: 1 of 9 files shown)


pjbgf:
@cursor review

Soph (approved):

PR #1309, opened 2026-06-01
Harden cluster trust: ephemeral binding + HostInCluster specificity floor

Soph:
Follow-up security hardening to the remote-helper credential-scoping work (merged via #1306). Two independent fixes to the `entire://` clone trust model.

## 1. Stop silent persistent cluster auto-binding (`clusterdiscovery`)

A successful `/.well-known/entire-cluster.json` discovery match used to persist `cluster_contexts[host] = <context>` to `contexts.json`. That binding is **durable and silent**: after a single drive-by clone of an attacker URL (e.g. a malicious submodule whose `.well-known` names the victim's real core), every future op against that host skipped discovery and minted identity-bearing JWTs against the bound context — a persistent identity-leak channel with no UX to notice or revoke it.

- `ResolveContextForCluster` no longer persists a binding. It resolves the match **for the current invocation only** and re-evaluates the live `.well-known` each time, so the trust decision stays fresh and a one-off clone leaves no durable state. The sole caller (the non-interactive git-remote-entire helper) can't prompt, so ephemeral resolution is the right scoping. Explicit bindings (`entire-core context bind`, teammate tooling) are still honored.
- Bindings are now **auditable and revocable**: `entire auth contexts` lists any `cluster_contexts` entries with an unbind hint, and new `entire auth unbind <host>` revokes one without touching the underlying login context.

## 2. Floor `HostInCluster` wildcard at a registrable domain (`discovery`)

`HostInCluster` treated any `*.cluster` host as same-cluster via a bare suffix match. With a broad cluster host that makes an entire public suffix in-cluster (cluster `io` → every `*.io` inherits the Authorization-carry / replica-trust boundary), so a relinquished or misconfigured subdomain anywhere under that suffix becomes a credential receiver.

- The subdomain (wildcard) match is now gated on the cluster host being strictly more specific than its own public suffix (via `x/net/publicsuffix`). `*.entire.io` still matches `entire.io`; `*.io` and `*.co.uk` no longer wildcard-match. Exact matches and IP literals are unaffected.

## Scope note

With #1 in place an attacker can no longer durably pin `*.example.com` trust via a one-off clone. The `example.com` (eTLD+1, attacker-registrable) case still passes the public-suffix floor — fully closing that needs a signed/endorsed cluster manifest, which is intentionally out of scope here.

## Testing

- New tests: ephemeral resolution (no binding persisted, re-discovers each call), `HostInCluster` specificity-floor cases, binding surfacing in `auth contexts`, and `auth unbind` (idempotent, preserves the context).
- `mise run fmt && mise run lint` clean; `go test -tags=integration,authfilestore -race ./...` and the e2e canary pass.

---

> [!NOTE]
> **High Risk**
> Changes authentication trust boundaries for entire:// git operations and redirect/replica credential scoping; mistakes could break clones or leak tokens.
> 
> **Overview**
> Hardens the `entire://` clone trust model in two ways.
> 
> **Cluster auth resolution** no longer persists `cluster_contexts` after a successful `/.well-known` discovery match—each git operation re-reads discovery so a one-off malicious clone cannot leave a silent, durable auto-auth channel. Explicit bindings (e.g. deliberate `context bind`) still apply. **`entire auth contexts`** now lists those bindings with an unbind hint, and **`entire auth unbind <host>`** removes a binding without deleting the login context.
> 
> **`HostInCluster`** subdomain matching now requires the cluster host to be a registrable domain (via `golang.org/x/net/publicsuffix`), so bare suffixes like `io` or `co.uk` cannot treat an entire TLD as in-cluster for credential-carrying redirects/replicas.
> 
> Tests cover ephemeral discovery, binding audit/unbind, and the hostname specificity floor.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d8bc4b94e925ae24bc84e7377d14c0ee069b59ba. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/auth.go
+++ b/cmd/entire/cli/auth.go
@@ -154,6 +154,7 @@ func newAuthCmd() *cobra.Command {
 	cmd.AddCommand(newAuthRevokeCmd())
 	cmd.AddCommand(newAuthContextsCmd())
 	cmd.AddCommand(newAuthUseCmd())
+	cmd.AddCommand(newAuthUnbindCmd())
 	return cmd
 }
 
--- a/cmd/entire/cli/auth/context_store.go
+++ b/cmd/entire/cli/auth/context_store.go
@@ -116,6 +116,36 @@ func Contexts() ([]*contexts.Context, string, error) {
 	return f.Contexts, f.CurrentContext, nil
 }
 
+// ClusterBindings returns the cluster_host → context_name map from
+// contexts.json. Surfaced by `entire auth contexts` so users can audit
+// which cluster hosts auto-authenticate git operations with a stored
+// context — and spot any they don't recognise.
+func ClusterBindings() (map[string]string, error) {
+	f, err := contexts.Load(contexts.DefaultConfigDir())
+	if err != nil {
+		return nil, fmt.Errorf("load contexts: %w", err)
+	}
+	return f.ClusterContexts, nil
+}
+
+// UnbindCluster removes the cluster→context binding for host, reporting
+// whether one existed. Used by `entire auth unbind` to revoke a binding
+// without touching the underlying login context.
+func UnbindCluster(host string) (bool, error) {
+	var existed bool
+	if err := contexts.Modify(contexts.DefaultConfigDir(), func(f *contexts.File) (bool, error) {
+		if _, ok := f.ClusterContexts[host]; !ok {
+			return false, nil
+		}
+		delete(f.ClusterContexts, host)
+		existed = true
+		return true, nil
+	}); err != nil {
+		return false, fmt.Errorf("unbind cluster: %w", err)
+	}
+	return existed, nil
+}
+
 // ContextStore wraps the legacy keyring Store so token *reads* prefer the
 // active contexts.json context, falling back to the legacy
 // entire-cli/<authBaseURL> entry. Writes are inherited from Store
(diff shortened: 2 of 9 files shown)


Soph:
Thanks bots — addressed in 61b0c1027:

- **IP literals wildcard-matching (Copilot, real bug):** Fixed. `clusterAllowsSubdomains` now returns `false` for IP literals, so an IP cluster only matches the exact host — `evil.127.0.0.1` no longer suffix-matches `127.0.0.1`. Added a regression test and updated the doc comment to state IP clusters have no subdomain semantics.
- **Bindings hidden when no login contexts (Cursor, Medium):** Fixed. `runAuthContexts` no longer early-returns; it prints the empty-state hint and still falls through to the bindings section, so an orphaned binding (manual edit / deleted context) is surfaced. Added a test for that case.
- **Single-label cluster regression, e.g. `foo.localhost` (Cursor, Low):** Working as intended. A single-label name is its own public suffix, so it's rejected from the wildcard path on purpose — treating `*.localhost` (or any internal single-label TLD) as same-cluster is exactly the over-broad trust the floor exists to prevent. Exact-match still covers `localhost`/IP replicas, which is what the local-dev and test setups actually use. Documented in the `clusterAllowsSubdomains` comment.

alishakawaguchi (approved):

pjbgf (approved):

PR #1311, opened 2026-06-01
Add checkpoints v1.1 support to entire explain

computermode:
https://entire.io/gh/entireio/cli/trails/465

## Summary

Adds a v1.1 committed-checkpoint read path so `entire explain` reads from `refs/entire/checkpoints/v1.1` when `checkpoints_version` is `1.1`. When v1.1 is enabled, reads bind to the v1.1 custom ref and **never fall back to v1** — the v1.1 path is genuinely exercised rather than masked by a v1 fallback.

## Details

- `GitStore` gains a configurable `committedReadRef` for **reads only**. Writes and read-modify-write paths always target the v1 branch (the durable source of truth) and are mirrored to v1.1 separately by the write-time mirror.
- `NewCommittedReadStore` always binds reads to the v1.1 ref when v1.1 is enabled. Before returning, it best-effort **sync-then-reads**: advances the local-only v1.1 ref up to the v1 tip (seed when missing, advance when an ancestor, no-op when equal, leave a diverged ref as-is). Sync failures are logged, never gating the store choice — there is no v1 store fallback.
- The v1 tip for the sync is taken from the local v1 branch, or from `origin/entire/checkpoints/v1` when no local branch exists, so `entire explain` still works on a fresh clone whose v1 metadata is remote-only — without reading through a v1 store.
- `entire explain --generate` writes summaries through the v1 store (source of truth), then rebuilds the read store so v1.1 mode reloads the newly written summary after the next sync.
- The v1.1 custom ref keeps its local-only invariant: it is never pushed or fetched.

## Validation

- `go test ./cmd/entire/cli/checkpoint`
- `go test ./cmd/entire/cli -run Explain`
- `mise run lint`

Regression coverage: sync rules (seed, seed-from-origin, advance, equal, diverged, write-failure, no-v1-tip) and read-store behavior (ref selection, read equivalence with v1, remote-only metadata, and no-v1-fallback when the custom ref is unavailable or diverged).

### DIFF
--- a/cmd/entire/cli/checkpoint/committed_read_store.go
+++ b/cmd/entire/cli/checkpoint/committed_read_store.go
@@ -15,11 +15,13 @@ import (
 // NewCommittedReadStore returns a GitStore for reading committed checkpoints,
 // honoring the checkpoints_version setting.
 //
-// When v1.1 is enabled it first brings the local-only v1.1 custom ref up to the
-// v1 branch tip (sync-then-read). If the custom ref cannot be synced, or if it
-// has diverged from v1, reads fall back to the default v1 store so v1 remains
-// the source of truth and the existing remote-tracking fallback still applies.
-// When v1.1 is not enabled, this is the default v1 read store.
+// When v1.1 is enabled, committed reads resolve against the local-only v1.1
+// custom ref and NEVER fall back to v1: the store always binds to the v1.1 ref.
+// Before returning, it best-effort syncs that ref up to the v1 tip
+// (sync-then-read) so v1.1 carries v1's full history. v1 stays the source of
+// truth via the write-time mirror; reads stay on v1.1 so the v1.1 path is
+// genuinely exercised rather than masked by a v1 fallback. When v1.1 is not
+// enabled, this is the default v1 read store.
 //
 // Currently only the `entire explain` command reads through this factory. Other
 // read-only commands stay on NewGitStore (v1) for now and will be migrated in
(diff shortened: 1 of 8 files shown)


cmd/entire/cli/checkpoint/committed_read_store.go
```diff
@@ -0,0 +1,114 @@
...
+
+// NewCommittedReadStore returns a GitStore for reading committed checkpoints.
+// With checkpoints_version 1.1 it binds reads to the local-only v1.1 custom ref
+// (never falling back to v1) after best-effort syncing it to the v1 tip. Writes
+// still use NewGitStore (v1); only `entire explain` reads through this today.
+func NewCommittedReadStore(ctx context.Context, repo *git.Repository) *GitStore {
+	if !settings.MirrorsToV1CustomRef(ctx) {
+		return NewGitStore(repo)
+	}
+	syncV1CustomRefForRead(ctx, repo)
```
pfleidi: Is there a reason we're syncing the v1 branch with the custom ref when creating a new store. That seems like the constructor is a doing a bit too much work here.

cmd/entire/cli/checkpoint/store.go
```diff
@@ -12,11 +15,33 @@ var _ Store = (*GitStore)(nil)
...
 	blobFetcher BlobFetchFunc
+	// committedReadRef is the ref that committed-checkpoint *reads* resolve
+	// against. Defaults to the v1 branch; v1.1 read stores bind it to the
+	// local-only custom ref (paths.MetadataRefName).
+	//
+	// Writes intentionally do NOT use this ref: committed writes always target
+	// the v1 branch (the durable source of truth) and are mirrored to the v1.1
+	// custom ref separately by the strategy's write-time mirror. Pointing writes
+	// here would let a v1.1 read store write ahead of v1 and diverge from it.
+	committedReadRef plumbing.ReferenceName
```
pfleidi, line 26: Just as a general thought, not necessarily related to this particular change:

I've been wondering if it would make sense to have a general `committedRef` or something similar for rollouts swapping around the write order.

What I could imagine is having an iteration where we're defaulting to the v1.1 ref rather than the v1 branch going forward at some point. This doesn't need to happen in this PR, though.

pfleidi (dismissed):
I had two comments but only one question about the concrete changes to this PR. Looks great, otherwise!

cmd/entire/cli/checkpoint/store.go
computermode: Yeah that makes sense, I think keeping it out of scope for this PR makes sense for now and we can add it in a future one.

cmd/entire/cli/checkpoint/committed_read_store.go
computermode: Since the v1.1 ref isn't fetched/pushed at the moment, `git pull` will only pull in the v1 changes.. so this re-syncs v1.1 before we read from it. It's shared here because `explain` has multiple places we can read. I can pull it out separately as an explicit step if that's makes more sense.

pfleidi: Would it be possible doing that in a different code path outside of the constructor? I understand the reasoning behind doing it here but it feels like an unexpected side effect of calling the constructor. 🤔

pfleidi (approved):

PR #1316, opened 2026-06-01
checkpoint: committed-ref topology seam

computermode:
https://entire.io/gh/entireio/cli/trails/472

## Summary

Addresses @pfleidi's review comment on #1311:
> I've been wondering if it would make sense to have a general `committedRef`
> or something similar for rollouts swapping around the write order [...]
> defaulting to the v1.1 ref rather than the v1 branch.

The committed-metadata ref topology — *"the v1 branch is the source of truth
(written and pushed); the v1.1 custom ref is a local-only mirror that reads use
when `checkpoints_version: "1.1"` is opted in, and that is never pushed"* — was
re-derived inline from `settings.MirrorsToV1CustomRef` + the `paths` constants
at five sites. A future rollout phase that swaps the write order (v1.1 ref as
primary) would have meant editing all five in lockstep.

This PR names that topology once and routes every site through it.

## The seam

New `checkpoint.CommittedRefs` value object + resolver:

| Field   | Meaning |
|---------|---------|
| `Primary` | source of truth — written first, pushed/fetched |
| `Read`    | ref committed reads resolve against |
| `Mirror`  | advanced to `Primary` after writes, kept current before reads; empty when there is no mirror (local-only, never pushed) |

**Invariant (holds in both rollout phases):** `Read == Mirror` when a mirror
exists, else `Read == Primary`.

Two entry points: `ResolveCommittedRefs(ctx)` (disk settings) and
`ResolveCommittedRefsFromSettings(s)` (an injected `EntireSettings`, the path
`attach` uses; nil → v1-branch-only).

Sites routed through the resolver:

- `NewCommittedReadStore` → `ResolveCommittedRefs(ctx).Read`
- read-time sync (`syncV1CustomRefForRead` → `syncMirrorForRead(refs)`), gating on `HasMirror()`
- `strategy.mirrorMetadataToV1CustomRef` (hook side, log-only)
- `cli.mirrorToV1CustomRef` (error-returning), now taking the resolved refs
- `attach` / `explain` call sites

After the change, `MirrorsToV1CustomRef` and the two metadata-ref `paths`
constants are referenced for committed-ref topology **only** inside
`committed_refs.go`.

## Behavior is unchanged

Behavior-identical refactor: today `Primary` is always the v1 branch and
`Read`/`Mirror` are the v1.1 custom ref only when `checkpoints_version: "1.1"`
is opted in. Existing read-store, mirror, and sync tests pass unmodified
(beyond a renamed internal helper).

The future **write-order flip** (making the v1.1 ref `Primary`) becomes a
one-site change in `ResolveCommittedRefs`. The flip itself and its push/fetch
semantics are explicitly **out of scope**.

## Design doc

`docs/superpowers/specs/2026-06-01-committed-ref-topology-seam-design.md`
(local-only; that dir is gitignored in this repo).

## Validation

- `mise run check` (fmt + lint + unit/integration + Vogon canary) — green
- New: `TestResolveCommittedRefs`, `TestResolveCommittedRefsFromSettings`

### DIFF
--- a/cmd/entire/cli/attach.go
+++ b/cmd/entire/cli/attach.go
@@ -60,11 +60,13 @@ type attachOptions struct {
 	entireSettings *settings.EntireSettings
 }
 
-func (opts attachOptions) mirrorsToV1CustomRef(ctx context.Context) bool {
+// committedRefs resolves the committed-ref topology for this attach, honoring
+// an injected EntireSettings when present and otherwise reading disk settings.
+func (opts attachOptions) committedRefs(ctx context.Context) cpkg.CommittedRefs {
 	if opts.entireSettings != nil {
-		return opts.entireSettings.MirrorsToV1CustomRef()
+		return cpkg.ResolveCommittedRefsFromSettings(opts.entireSettings)
 	}
-	return settings.MirrorsToV1CustomRef(ctx)
+	return cpkg.ResolveCommittedRefs(ctx)
 }
 
 func newAttachCmd() *cobra.Command {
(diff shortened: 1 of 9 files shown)


pfleidi (approved):
Thanks for doing this!

PR #1319, opened 2026-06-02
Auth contexts: finish binding-removal cleanup, use "login server" wording

Soph:
Stacked on top of #1318 (`20260602-auth-contexts`).

#1318 removes the cluster→context binding model, but a few user-facing strings and package docs still describe bindings that no longer exist, and the user-facing copy used the internal term **core** for the JWT issuer. This cleans that up.

## Changes (docs/help text only — no logic)

- **`logout --help` / `--all` flag** — no longer claims to remove "cluster bindings" (logout removes logins; there are no bindings anymore).
- **`auth use` — doc comment, `--help` `Long`, and the cross-core runtime warning** — reframed: the active context is the preferred identity that authenticates any cluster its **login server** fronts, and the switch takes effect on the next operation. Drops the obsolete "bound / not yet bound" narrative.
- **`git-remote-entire` package doc** — describes cores-cache-or-`/.well-known` resolution instead of "an explicit cluster binding, else discovery".
- **`contexts` package doc** — drops the removed `cluster_contexts` field; reframes `current_context`.
- **`RenderLoginHint` + warning copy** — "core(s)" → "login server(s)" so user-facing text doesn't leak the internal issuer term.

## Verification

- `go build ./...` clean
- `mise run fmt` applied
- affected `cli` + `clusterdiscovery` tests pass
- `golangci-lint` 0 issues on all changed packages

### DIFF
--- a/cmd/entire/cli/auth_context.go
+++ b/cmd/entire/cli/auth_context.go
@@ -11,25 +11,25 @@ import (
 
 // newAuthUseCmd switches the active login context.
 //
-// For `git clone entire://…` the active context is only a fallback: a
-// cluster already bound to a context (cluster_contexts) keeps using that
-// binding, and the active context is consulted only the first time a
-// not-yet-bound cluster is resolved. Control-plane commands (auth status/
-// list/revoke, org/project/repo/grant) currently target the configured auth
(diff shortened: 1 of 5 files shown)


toothbrush (approved):
Yes this seems reasonable!  Thanks for cleaning up those spots i missed.

PR #1347, opened 2026-06-03
Count checkpoint "steps" by prompts, not file-modifying turns

dipree:
## Problem

The "steps" count shown for each session in the UI was often wrong — and frequently **0**. It was sourced from the CLI's `StepCount`, which only counted turns that *modified files* and reset to 0 after every condensation. Roughly 1 in 5 recent checkpoints displayed "0 steps".

Three issues:

1. **Bug — written before a step was recorded.** Checkpoints created mid-turn or by a plain `git commit` (commit-only sessions) condense before any step is recorded, so the count was 0.
2. **Bug — `entire attach` ignored it.** The attach path never set the count at all, so every attached checkpoint wrote 0.
3. **Wrong thing counted.** Even when non-zero, it counted *file-modifying turns*, not prompts — so conversational/planning turns didn't count.

E.g.

CleanShot 2026-06-03 at 18 26 53@2x

But session has clearly two prompts before the checkpoint https://entire.io/gh/entireio/cli/session/eaa96109-502f-4be2-9357-cf4fde90deef#timeline-46304ab84156

## Solution

One concept everywhere: **a checkpoint's "steps" = the number of user prompts in that checkpoint's window** (prompts since the previous checkpoint, with a deferred reset). Derived from the existing `SessionTurnCount` (counts every turn, and is safe for exec-mode agents that report turns via hooks):

```
steps = SessionTurnCount − PromptWindowBase   (floored at 1)
```

- **Deferred reset:** writing a checkpoint doesn't zero the window — it restarts only when the *next* prompt arrives. So two checkpoints with no prompt between them report the **same** count, not 0. The reset is gated on the turn count *actually advancing*, so a repeated/stale hook can't re-anchor early.
- **Floored at 1**, so we never display "0 steps" again.

### Scenarios

| Sequence | Result |
|---|---|
| 3 prompts → checkpoint A, then 2 prompts → checkpoint B | A = 3, B = 2 |
| 3 prompts → checkpoint A, then checkpoint B with no prompt between | A = 3, **B = 3** |
| checkpoint with no prompts counted yet (mid-turn) | 1 |

### `entire attach`

Attach is the same counting concept applied post-hoc. The hooks never fired, so instead of live turn events it counts the user prompts parsed from the transcript (`meta.TurnCount`), floored at 1. Attach produces a single checkpoint covering the whole session, so its window = the whole transcript = the total prompt count. (Earlier this counted `len(prompts)`, which only ever held the first prompt — so every attached session reported 1 regardless of length.)

### Related

The combined-attribution gate previously keyed off `checkpoints_count == 0`. Since that's now floored to ≥1, the gate moves to a new honest `SaveStepCount` metadata field ("did a real SaveStep run?").

Forward-only — existing rows keep their stored values (no backfill); the API/frontend pass the value through unchanged.


Soph (dismissed):

dipree:
@copilot resolve the merge conflicts in this pull request

evisdren (approved):

PR #1356, opened 2026-06-04
feat(antigravity): native token counts via title-slot tee

peyton-alt:
https://entire.io/gh/entireio/cli/trails/507

## Summary

Follow-up to #1287 (which intentionally deferred token counts). Adds native per-checkpoint token usage for the **Antigravity (agy)** agent, surfaced in committed checkpoint metadata exactly like every other agent — so the entire.io UI displays agy tokens with **zero web-side changes**.

**Why this approach:** agy exposes token usage *only* via the JSON payload piped to its title/statusline script on every agent state change (`context_window.{total_input_tokens, total_output_tokens, current_usage{...}}`, cumulative per conversation). It is absent from transcripts, hook payloads, logs, and the protobuf session state. The statusline schema is the single source — verified against real agy transcripts and the published docs.

**How it works (end-to-end):**
1. **`title-tee` shim** — `entire hooks antigravity title-tee [--wrap '<original>']` is installed into agy's global `~/.gemini/antigravity-cli/settings.json` `title` slot. On every state change it tees the JSON to a per-conversation append-only JSONL store (`<cache>/entire/antigravity/status/<conversation_id>.jsonl`) and pipes through to any pre-existing user title command. We use the `title` slot (not `statusLine`) — same payload, lower blast radius.
2. **`OutOfBandTokenSource` capability** — new interface for agents whose token data lives outside the transcript. Built-in only (external agents excluded).
3. **Delta computation** — baseline snapshot captured at TurnStart (stored in `PrePromptState`), delta = latest cumulative totals − baseline at TurnEnd (the Codex cumulative-minus-baseline pattern, sourced from the tee file). Input/output exact; cache/api-call counts best-effort sampled.
4. **Additive lifecycle fallback** — only fires when transcript-based calc returns nil; strict no-op for claude-code/gemini/codex/etc.
5. **Condensation fallback** — accumulated `SessionState.TokenUsage` copied into committed metadata for non-`TokenCalculator` agents.
6. **Doctor check** — warns when agy hooks are installed but the title-tee isn't configured.

**Install/uninstall semantics:** install wraps + preserves any existing user title command (self-describing via `--wrap`); uninstall (agent-removal only, never per-repo `disable` since the slot is global) restores the original or clears a bare tee, and leaves foreign/user-wrapped commands untouched.

## Notable properties

- **Safe in agy's render path:** the shim never exits non-zero and never writes noise to stdout (stdout is the window title). Measured ~36ms warm/invocation (≤50ms budget); ~55µs/append on a 500-line file.
- **Privacy:** snapshot store holds only token *counts*, never prompts/file contents. All new logging is metadata-only.
- **Degrades to nil, never errors** when agy lacks the title hook or the user removes the config.

## Known v1 limitations (documented in code)

- Resumed conversations (`agy --continue`) whose title-tee hasn't fired before the first TurnStart over-count that one turn by the prior cumulative total.
- Cache-creation/cache-read/api-call counts are best-effort (snapshot sampling); input/output totals are exact.

## Test Plan

- [x] Unit: snapshot store (dedup, garbage tolerance, prune), title installer (wrap/unwrap/idempotent/round-trip incl. embedded single quotes), delta math (baseline, nil-baseline, clamp-on-backwards, unparseable-baseline degradation), capability gating.
- [x] Integration: `TestAntigravity_TokenUsageInCheckpointMetadata` — seeds snapshots, drives the real hook lifecycle, commits, asserts `token_usage` 3500/300 in **both** per-session `metadata.json` and the `CheckpointSummary` aggregate on `entire/checkpoints/v1`.
- [x] Doctor: configured/not-configured/not-installed cases.
- [x] `mise run check` green (unit `-race` + integration + Vogon canary 58/58). Vogon unaffected (degrades to nil).
- [x] Shim latency re-measured against final binary (~36ms warm).

---

> [!NOTE]
> **Medium Risk**
> Modifies global Antigravity user config and checkpoint token metadata paths; failures degrade to missing counts rather than blocking hooks, but misconfigured title-tee affects all agy workspaces on the machine.
> 
> **Overview**
> Adds **native per-checkpoint token usage for Antigravity**, which only exposes `context_window` totals via the global **title** command JSON—not transcripts or lifecycle hooks.
> 
> A **`title-tee` shim** (`entire hooks antigravity title-tee`) is installed into `~/.gemini/antigravity-cli/settings.json`, tees state JSON into a per-conversation JSONL snapshot cache, and optionally chains the user’s original title command via `--wrap`. Hook install always attempts title-tee setup (even when repo hooks are already idempotent) so upgrades can repair a missing global slot.
> 
> New **`OutOfBandTokenSource`** captures a baseline at turn start (`PrePromptState.token_baseline`) and computes turn deltas at turn end when transcript token calc is nil; condensation copies accumulated session token usage into committed metadata for OOB agents only. **`entire doctor`** warns when repo Antigravity hooks exist but title-tee is not configured; title-tee is removed on agent removal, not per-repo disable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit acad86fb0c04986aa4608a1816485945375d4869. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/agent/agent.go
+++ b/cmd/entire/cli/agent/agent.go
@@ -5,6 +5,7 @@ package agent
 
 import (
 	"context"
+	"encoding/json"
 	"io"
 	"os/exec"
 
@@ -186,6 +187,31 @@ type TokenCalculator interface {
 
 	// CalculateTokenUsage computes token usage from the transcript starting at the given offset.
 	CalculateTokenUsage(transcriptData []byte, fromOffset int) (*TokenUsage, error)
+}
+
+// OutOfBandTokenSource provides token usage from a source other than the
+// transcript. Antigravity is the only agent that needs this: agy never writes
+// token data into its transcript or hook payloads — its title/statusline pipe
+// is the only surface, captured to disk by `entire hooks antigravity
+// title-tee` (see agent/antigravity/statusline.go).
+//
+// Flow: the lifecycle calls SnapshotTokenBaseline at TurnStart and stores the
+// opaque baseline in PrePromptState; at TurnEnd (when transcript-based
+// calculation yields nothing) it calls CalculateTokenUsageSince to get the
+// checkpoint-scoped delta — the same cumulative-totals-minus-baseline pattern
+// Codex uses, sourced out-of-band.
+type OutOfBandTokenSource interface {
+	Agent
+
+	// SnapshotTokenBaseline returns an opaque, agent-defined marker of the
+	// current cumulative token position for the session. A nil baseline with
+	// nil error means "no usage observed yet" (delta will count from zero).
+	SnapshotTokenBaseline(ctx context.Context, sessionID string) (json.RawMessage, error)
+
+	// CalculateTokenUsageSince computes usage between the baseline and now.
+	// A nil result with nil error means no data is available (degrade to no
+	// token counts, never to an error).
+	CalculateTokenUsageSince(ctx context.Context, sessionID string, baseline json.RawMessage) (*TokenUsage, error)
 }
 
 // TextGenerator is an optional interface for agents whose CLI supports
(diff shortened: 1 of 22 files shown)


peyton-alt:
@bugbot review

PR #1382, opened 2026-06-05
fix: bootstrap metadata branch from checkpoint_remote during enable

### LINKED ISSUE #1374: entire enable` creates empty orphan `entire/checkpoints/v1` instead of fetching from configured `checkpoint_remote`
### What happened?

When `entire enable` is run on a second device in a repo that already has a `checkpoint_remote` configured and an existing `entire/checkpoints/v1` branch in that remote, it creates a fresh empty orphan branch locally rather than fetching the existing one. This makes `entire checkpoint list` return no results, and a subsequent `git fetch` from the checkpoint remote is rejected as non-fast-forward because the local orphan has a different history.

### Steps to reproduce

1. Enable Entire on device A with a `checkpoint_remote` configured in `.entire/settings.json`
2. Commit and push — checkpoint data is written to the checkpoint remote's `entire/checkpoints/v1` branch
3. Clone or pull the repo on device B
4. Run `entire enable -y --agent claude-code` on device B
5. Run `entire checkpoint list` — shows 0 checkpoints
6. Attempt to fetch: `git fetch https://github.com/<org>/checkpoints entire/checkpoints/v1:entire/checkpoints/v1` — rejected as non-fast-forward

## Expected behaviour

`entire enable` should detect that a `checkpoint_remote` is configured and that `entire/checkpoints/v1` already exists there, and fetch it locally instead of creating an empty orphan branch.

## Actual behaviour

A fresh empty orphan `entire/checkpoints/v1` is created locally, conflicting with the real branch in the checkpoint remote. The workaround is:
```bash
git fetch https://github.com/<org>/checkpoints entire/checkpoints/v1:entire/checkpoints/v1 --force
```

### Entire CLI version

Entire CLI 0.6.3 (refs/tags/v0.6.3) Go version: go1.26.3 OS/Arch: darwin/arm64

### OS and architecture

Darwin 25.5.0 arm64 Darwin / Linux 6.18.26 x86_64 GNU/Linux

### Agent

Claude Code

### Terminal

cmux

### Logs / debug output

```bash

```

### Additional context

I setup `entire` in my NixOS flake here, which contains `.entire/settings.json` and other configuration: https://github.com/stackptr/rc/pull/502

alishakawaguchi:
Fixes #1374

## Problem

When `entire enable` runs on a second device in a repo with a configured `checkpoint_remote` and an existing `entire/checkpoints/v1` branch on that remote, it created a fresh empty orphan branch locally instead of fetching the existing one. Consequences:

- `entire checkpoint list` returned 0 results (local ref exists but points at an empty tree, so the origin fallback never triggers)
- `git fetch <checkpoint-remote> entire/checkpoints/v1:entire/checkpoints/v1` was rejected as non-fast-forward (the orphan shares no ancestry with the real branch)

Root cause: `EnsureMetadataBranch` only consults the local ref and `origin/entire/checkpoints/v1` before minting an empty orphan — it never consults the configured `checkpoint_remote`, even though `entire resume` already fetches from it first.

## Fix

`EnsureSetup` now calls a new `bootstrapMetadataFromCheckpointRemote` before opening the repo handle: `remote.Configured` → `remote.FetchURL` → `fetchMetadataBranchIfMissing`. Notes on placement and behavior:

- **In `EnsureSetup`, not `EnsureMetadataBranch`** — covers both enable paths *and* the lifecycle hook path (`lifecycle.go`), keeps `EnsureMetadataBranch` a pure repo-handle function (existing unit tests call it with CWD = real repo, which itself has a `checkpoint_remote` configured — an in-function fetch would mutate the real repo during tests), and the repo handle opened afterward sees the freshly fetched packfiles
- **Before the handle is opened** — avoids go-git's stale packfile cache after a CLI fetch
- **Best-effort and one-time** — only fetches when the local branch is missing; failure (e.g. remote branch doesn't exist before the first push from any device) falls back to orphan creation as before, so device A's first enable is unchanged
- `fetchMetadataBranchIfMissing` now returns `(fetched bool, err error)` so enable can print `✓ Created local branch 'entire/checkpoints/v1' from checkpoint remote`

## Testing

- New regression test `TestHTTPS_EnableBootstrapsMetadataFromCheckpointRemote` (watched fail before the fix): device A pushes checkpoints to a separate HTTPS checkpoint remote; device B clones origin and runs `entire enable` via the real binary; asserts the local metadata branch tip equals the checkpoint remote tip and contains device A's checkpoint metadata
- Strengthened existing `fetchMetadataBranchIfMissing` tests with the new `fetched` return assertions
- `mise run test:integration`: 395/395 pass; E2E canary: all pass; `mise run fmt && mise run lint`: clean
- Pre-existing unrelated failures in `cmd/entire/cli` unit tests (`TestRunAuthStatus_RendersSessionsTable`, `TestExplainCmd_*`) fail identically on the clean tree — environment-sensitive, not from this change

---

> [!NOTE]
> **Medium Risk**
> Changes shared setup/enable metadata initialization for multi-repo checkpoint routing; failures still fall back to orphan creation, but wrong fetch behavior could hide checkpoints or cause sync conflicts.
> 
> **Overview**
> Fixes **#1374**: on a fresh clone with a separate **`checkpoint_remote`**, `entire enable` no longer creates an unrelated empty **`entire/checkpoints/v1`** orphan when checkpoints only exist on that remote (not on **origin**).
> 
> **`EnsureSetup`** now runs **`bootstrapMetadataFromCheckpointRemote`** *before* opening the go-git repo handle: if a checkpoint remote is configured, it resolves the fetch URL and calls **`fetchMetadataBranchIfMissing`** so **`EnsureMetadataBranch`** sees real history instead of minting an empty branch. Bootstrap is best-effort (warnings + orphan fallback when the remote branch does not exist yet). When a fetch actually creates the local branch, enable prints **`✓ Created local branch 'entire/checkpoints/v1' from checkpoint remote`**.
> 
> **`fetchMetadataBranchIfMissing`** returns **`(fetched bool, err error)`**; callers (including push settings resolution) ignore the bool except for that message.
> 
> **Tests:** new HTTPS integration **`TestHTTPS_EnableBootstrapsMetadataFromCheckpointRemote`** (device A pushes to checkpoint remote; device B clones origin and runs enable; local tip and checkpoint metadata must match the remote). Unit tests assert the new **`fetched`** return value.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aa5169fad9b12405553972ccf8ed259d03a8e9b1. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Soph:
#1384 handles this, closing.

PR #1384, opened 2026-06-07
fix(strategy): fetch metadata branch from checkpoint_remote on enable

Soph:
https://entire.io/gh/entireio/cli/trails/534

## Problem

Relates to #1374.

`EnsureMetadataBranch` only consulted the local branch and origin's remote-tracking ref when setting up `entire/checkpoints/v1`. When a separate `checkpoint_remote` is configured, the metadata history lives in *that* remote rather than origin, so on a second device neither source existed and `entire enable` fell straight through to creating an **empty orphan** branch. That orphan left `entire checkpoint list` empty and was rejected as non-fast-forward on the next fetch from the real remote.

## What changed

`EnsureMetadataBranch` now treats a configured `checkpoint_remote` as the authoritative store and consults it **before** origin:

- **Local branch missing** → bootstrap from the checkpoint remote (direct ref set, no replay).
- **Local branch holds no checkpoint data** → force-replace the un-initialized orphan with the real branch from the checkpoint remote (force-set rather than a safe advance, so the disjoint orphan commit is discarded instead of replayed onto the real tip).
- It **never seeds from origin** while a `checkpoint_remote` is configured, so a stale `origin/entire/checkpoints/v1` tracking ref can no longer shadow the real checkpoints or get replayed back into the checkpoint remote.

The "un-initialized" check now ignores orphan-init artifacts (`vercel.json`) rather than requiring a literal empty tree — vercel-enabled repos write `vercel.json` into the orphan, which previously made the branch look non-empty and skip healing. Ignoring only the known init files (instead of treating "no checkpoint shards" as empty) avoids clobbering a branch that carries other real content.

Settings are resolved against the repository being modified (its worktree root) rather than the ambient CWD, keeping the decision consistent with the repo handle. In normal use CWD is the repo root, so this is a no-op.

All paths are non-fatal: a missing/unreachable `checkpoint_remote` falls back to keeping the local branch or creating a fresh orphan.

## Tests

- `TestEnsureMetadataBranch_CheckpointRemoteTakesPrecedenceOverOrigin` — a stale origin ref is ignored in favor of the checkpoint remote.
- `TestHealEmptyOrphanMetadataFromCheckpointRemote_HealsVercelOnlyOrphan` — a `vercel.json`-only orphan heals.
- `TestBootstrapMetadataFromURL_*`, `TestHealEmptyOrphanFromURL_*` — bootstrap/heal mechanics against a local remote.

`mise run fmt && mise run lint` clean; full unit suite (6138) and integration suite (394) green.

---

> [!NOTE]
> **Medium Risk**
> Changes metadata branch creation during enable/setup and can force-reset local refs when healing empty orphans; wrong remote URL or fetch behavior could leave users with missing or incorrect checkpoint history, though failures fall back without hard errors.
> 
> **Overview**
> Fixes **#1374** by making `EnsureMetadataBranch` treat a configured **`checkpoint_remote`** as the source of truth during `entire enable` / setup, instead of only local + `origin/entire/checkpoints/v1`.
> 
> When **`checkpoint_remote` is set**, setup now **bootstraps** a missing local metadata branch from that remote (ref set to the fetched tip, no replay) or **heals** a local branch that has no real checkpoint data by **force-replacing** the orphan with the remote tip (avoids replaying a disjoint empty commit). It **does not seed from `origin`** in that mode, so stale `origin/entire/checkpoints/v1` cannot shadow the real store.
> 
> **`isEmptyMetadataBranch`** is replaced by **`metadataBranchHasData`**, which treats tips with only orphan-init files (e.g. **`vercel.json`**) as uninitialized so vercel-enabled second devices still heal. Settings resolution uses the **repo worktree root** via `settings.WithWorktreeRoot` for consistent `checkpoint_remote` detection.
> 
> New helpers in **`checkpoint_remote.go`** (`BootstrapMetadataFromCheckpointRemote`, `HealEmptyOrphanMetadataFromCheckpointRemote`, test seam `resolveCheckpointRemoteURLFn`) plus broad integration tests cover bootstrap, heal, origin precedence, and vercel-only orphans. Unreachable/missing checkpoint remote remains **non-fatal** (keep local or create orphan).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bcab96d8ba07bf60e6f2223e02c4e3ef6147741c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/strategy/checkpoint_remote.go
+++ b/cmd/entire/cli/strategy/checkpoint_remote.go
@@ -4,14 +4,17 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"os"
 	"strings"
 	"time"
 
 	"github.com/entireio/cli/cmd/entire/cli/checkpoint"
 	"github.com/entireio/cli/cmd/entire/cli/checkpoint/remote"
 	"github.com/entireio/cli/cmd/entire/cli/logging"
+	"github.com/entireio/cli/cmd/entire/cli/paths"
 	"github.com/entireio/cli/cmd/entire/cli/settings"
 
+	"github.com/go-git/go-git/v6"
 	"github.com/go-git/go-git/v6/plumbing"
 )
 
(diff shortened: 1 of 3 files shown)


peyton-alt:
Closing in favor of #1719, which fixes the same #1374 bug against the current codebase (this branch predates the ResolveCommittedRefs→ResolveRefs / PersistentRefs refactors and no longer applies) and adds the piece this PR was missing: the checkpoint-remote fetch is gated to explicit enable flows via WithCheckpointRemoteBootstrap, so EnsureSetup on the per-turn hook hot path never blocks on a network fetch (here, the heal ran a fetch on every TurnStart while the local branch had no checkpoint data yet — e.g. device A between enable and the first condensation).

Three design points from this PR are better than #1719's current state and have been filed as findings on trail 825 to be folded in there:
- init-file-aware emptiness (metadataBranchHasData + vercel.json allowlist) so vercel-only orphans heal — finding 019f6744-8a9
- never seed from origin while a checkpoint_remote is configured (stale origin tracking-ref shadowing) — finding 019f6744-d87
- urlTargetsCheckpointRepo guard against remote.FetchURL's origin fallback — finding 019f6744-ea1

Thanks — the authority-rule design here survives, just landing via the other branch.

PR #1386, opened 2026-06-08
Fix: Keep Entire pre-push alive after Lefthook reinstalls hooks

### LINKED ISSUE #1349: entire and lefthook fight over the pre-push hook; session push gets silently disabled
### What happened?

In a repo that also uses [lefthook](https://github.com/evilmartians/lefthook) to manage git hooks, entire and lefthook fight over `.git/hooks/pre-push`. After any change to `lefthook.yml`, lefthook silently reclaims the hook and entire's session-log-on-push stops working — with no error — until someone manually re-wraps it.

**How the conflict arises**

1. lefthook installs and manages `pre-push`. Its installed hooks self-sync on every run, gated by a checksum of `lefthook.yml` stored in `.git/info/lefthook.checksum`.
2. `entire enable` / `entire configure --force` wraps the existing hook: it renames the current `pre-push` to `pre-push.pre-entire` and installs its own wrapper that runs the entire handler then chains down:
   ```sh
   # Entire CLI hooks
   if command -v entire >/dev/null 2>&1; then entire hooks git pre-push "$1" || true; else :; fi
   # Chain: run pre-existing hook
   "$(dirname "$0")/pre-push.pre-entire" "$@"
   ```
   At this point both run on push — entire pushes session logs, then chains to lefthook. 👍
3. **But** whenever `lefthook.yml` changes, lefthook's checksum goes stale. Its next self-sync (at the top of any lefthook-managed hook, e.g. the next commit's `pre-commit`) sees that `pre-push` is no longer *its* hook, backs entire's wrapper up to `pre-push.old`, and reinstalls its own bare hook. **entire's session push is now silently gone.**
4. On the next `lefthook.yml` change after that, lefthook also emits a recurring warning, because its earlier backup collides:
   ```
   sync hooks: ❌
   could not replace the hook: can't rename pre-push to pre-push.old - file already exists
   ```

**Why there is no clean user-side fix** (verified on 0.7.3)

The natural fix — make lefthook the sole owner of `pre-push` and run entire's push as a lefthook job/script — is blocked:

- `entire configure --force` re-installs entire's pre-push wrapper **even with `push_sessions: false`** → you cannot stop entire owning the hook.
- `entire hooks git pre-push <remote>` is itself **gated by `push_sessions`** (no-ops when off) → you cannot disable the wrapper and still call the handler from a lefthook script.
- `entire configure --force` **re-wraps whatever is at `pre-push` with no detection** — if lefthook reclaimed it, entire silently wraps lefthook's bare hook again, with no idempotency check or "another manager owns this" warning.

The only local mitigation is a self-heal step that re-wraps after each re-sync, which still can't cover a push that happens with no intervening commit.

**Proposed fixes (any one resolves it)**

1. **Idempotent / manager-aware install** — if `pre-push` already chains to `pre-push.pre-entire` (or contains entire's wrapper), don't blindly re-wrap; detect known managers (lefthook, husky, pre-commit, `core.hooksPath`).
2. **Decouple handler from install** — an option to install the handler *behavior* without owning the `pre-push` file, so managers like lefthook can invoke `entire hooks git pre-push` as a job/script. (Today `--skip-push-sessions` disables the handler too, defeating this.)
3. **Register via the host manager** — when lefthook is detected, register entire's push as a lefthook script (lefthook scripts receive the hook's `$@`, including the remote) instead of taking the file.
4. **Self-repair marker** — let entire re-assert/repair its wrapper automatically so a reclaim heals without manual `entire configure --force`.

### Steps to reproduce

1. A repo using lefthook with a `pre-push:` section; run `lefthook install`.
2. `entire enable` (wraps `pre-push`, chaining to lefthook).
3. Edit `lefthook.yml` (any change) and make a commit or push.
4. lefthook's self-sync reclaims `pre-push`; entire's session push silently stops (no error). Confirm with `head .git/hooks/pre-push` — the entire wrapper is gone.

### Entire CLI version

Entire CLI 0.7.3

### OS and architecture

macOS 26.5, arm64 (Darwin 25.5.0)

### Agent

claude code

### Terminal

ghostty

### Logs / debug output

```bash

```

### Additional context

- lefthook 2.0.12; `.entire/settings.json` committed (team-wide setup), so this affects every macOS dev on the team using entire.
- The 0.7.3 pre-push wrapper added a `command -v entire` guard (no-ops cleanly when entire is off-PATH) — good, but it does not address this ownership conflict.
- Also reproduces across git worktrees, which share a single `.git/hooks` dir.
- No log output to attach: the failure is silent (entire's hook is simply removed; nothing is logged).

stale2000:
## Summary

https://entire.io/gh/stale2000/cli/session/019ea5bb-9549-7083-b95f-7ecb498cbc24 

Closes #1349.

This keeps Entire's existing Git hook ownership behavior unchanged while adding a Lefthook-local safety net for the failure mode reported in the issue.

When a repository has a YAML Lefthook config, `entire enable` still installs the normal Entire `.git/hooks/pre-push` wrapper. It now also ensures `lefthook-local.yml` contains an `entire-push-sessions` pre-push command that only runs if Lefthook has reclaimed `.git/hooks/pre-push` and removed the `Entire CLI hooks` marker.

Normal state:
- Git runs Entire's `pre-push` wrapper.
- Entire chains to Lefthook.
- The Lefthook-local fallback sees the Entire marker and does nothing.

Recovered state:
- Lefthook has rewritten `.git/hooks/pre-push`.
- Git runs Lefthook directly.
- The Lefthook-local fallback sees the Entire marker is missing and runs `entire hooks git pre-push {1} || true`.

## Details

- Adds and updates `lefthook-local.yml`, not team-owned `lefthook.yml`.
- Adds the local Lefthook config path to `.git/info/exclude`.
- Preserves unrelated existing local Lefthook commands.
- Removes only Entire's local Lefthook command from `entire disable`.
- Keeps non-Lefthook hook behavior unchanged.
- Handles `.config/lefthook.yml` detection in addition to the existing Lefthook config names.

## Verification

Passed:
- `gofmt -w cmd/entire/cli/strategy/hook_managers.go cmd/entire/cli/strategy/hook_managers_test.go cmd/entire/cli/strategy/hooks.go cmd/entire/cli/strategy/hooks_test.go`
- `/Users/stale2000/.local/share/mise/installs/golangci-lint/2.11.3/golangci-lint-2.11.3-darwin-arm64/golangci-lint run ./cmd/entire/cli/strategy/...`
- `go test -tags=integration ./cmd/entire/cli/integration_test -run 'Test(UserPromptSubmit_ReinstallsOverwrittenHooks|HookOverwrite_MidTurnWipe_NextPromptRecovers)'`
- `git diff --check`

Manual real-Lefthook checks:
- Verified with Lefthook `v1.13.6` that `lefthook-local.yml` is merged into `pre-push` runs.
- Verified `{1}` expands to the Git pre-push remote name (`origin`).
- Verified the generated safety-net command calls `entire hooks git pre-push origin` when `.git/hooks/pre-push` lacks the Entire marker.
- Verified the same command skips when `.git/hooks/pre-push` contains `Entire CLI hooks`.

Known verification gaps:
- `mise run check` and `mise run fmt` did not reach the repo tasks because mise/aqua hit GitHub 504s while downloading ShellCheck and tmux.
- `go test ./cmd/entire/cli` still has an unrelated date-sensitive failure in `TestRunAuthStatus_RendersSessionsTable` where the rendered local date is `2025-12-31` but the test expects `2026-01-01`.

## Entire Integration

This PR commit was attached to the active Codex transcript with Entire:

- Session: `019ea5bb-9549-7083-b95f-7ecb498cbc24`
- Checkpoint: `965634231c03`
- Commit includes `Entire-Checkpoint: 965634231c03`
- `entire session current` resolves the attached Codex session.
- `entire checkpoint list` shows checkpoint `965634231c03` on this branch.
- `git push` ran the Entire `pre-push` hook and pushed `entire/checkpoints/v1` to the fork.

### DIFF
--- a/cmd/entire/cli/strategy/hook_managers.go
+++ b/cmd/entire/cli/strategy/hook_managers.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/entireio/cli/cmd/entire/cli/paths"
+	"gopkg.in/yaml.v3"
 )
 
 // hookManager describes an external hook manager detected in a repository.
@@ -29,8 +30,8 @@ func detectHookManagers(repoRoot string) []hookManager {
 		{"Overcommit", ".overcommit.yml", false},
 	}
 
-	// Lefthook supports {.,}lefthook{,-local}.{yml,yaml,json,toml}
-	for _, prefix := range []string{"", "."} {
+	// Lefthook supports {.config/,}{.,}lefthook{,-local}.{yml,yaml,json,toml}
+	for _, prefix := range []string{"", ".", ".config/"} {
 		for _, variant := range []string{"", "-local"} {
 			for _, ext := range []string{"yml", "yaml", "json", "toml"} {
 				name := prefix + "lefthook" + variant + "." + ext
@@ -74,7 +75,8 @@ func hookManagerWarning(managers []hookManager, cmdPrefix string) string {
 	specs := buildHookSpecs(cmdPrefix)
 
 	for _, m := range managers {
-		if m.OverwritesHooks {
+		switch {
+		case m.OverwritesHooks:
 			fmt.Fprintf(&b, "Warning: %s detected (%s)\n", m.Name, m.ConfigPath)
 			fmt.Fprintf(&b, "\n")
 			fmt.Fprintf(&b, "  %s may overwrite hooks installed by Entire on npm install.\n", m.Name)
@@ -94,7 +96,16 @@ func hookManagerWarning(managers []hookManager, cmdPrefix string) string {
 				fmt.Fprintf(&b, "      %s\n", cmdLine)
 				fmt.Fprintf(&b, "\n")
 			}
-		} else {
+		case m.Name == "Lefthook":
+			fmt.Fprintf(&b, "Note: %s detected (%s)\n", m.Name, m.ConfigPath)
+			fmt.Fprintf(&b, "\n")
+			if _, ok := lefthookLocalConfigPath(m.ConfigPath); ok {
+				fmt.Fprintf(&b, "  Added a local Lefthook pre-push safety net so Entire session pushes still run if Lefthook reinstalls hooks.\n")
+			} else {
+				fmt.Fprintf(&b, "  If Lefthook reinstalls hooks, run 'entire enable' to restore Entire's hooks.\n")
+			}
+			fmt.Fprintf(&b, "\n")
+		default:
 			fmt.Fprintf(&b, "Note: %s detected (%s)\n", m.Name, m.ConfigPath)
 			fmt.Fprintf(&b, "\n")
 			fmt.Fprintf(&b, "  If %s reinstalls hooks, run 'entire enable' to restore Entire's hooks.\n", m.Name)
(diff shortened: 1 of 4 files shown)


2bTwist:
I ran this branch against #1349 because I had just reproduced that issue from the other side.
Two results, one in the fix's favour and one gap.

## The safety net works on a newer lefthook than you tested

You verified with lefthook `v1.13.6`. I built this branch (`fa3b7b9`) and drove it on lefthook
`2.1.10`, which is where I hit #1349:

- `entire enable` writes the `entire-push-sessions` command into `lefthook-local.yml`.
- After lefthook reclaims `.git/hooks/pre-push`, a real `git push` runs it, and `{1}` expands to
  the remote name.
- Observed by putting a shim `entire` on `PATH` that logs its argv, so this is the actual
  invocation and not a reading of the config:

```
┃  entire-push-sessions ❯
✔️ entire-push-sessions (0.17 seconds)
--- entire invocations captured during the push ---
  entire hooks git pre-push origin
```

So the `{1}` templating and local-config merge both still behave on lefthook 2.x.

## The gap: the net is keyed to one hook, the collision is keyed to the repo's config

`gitHookNames` installs five hooks. lefthook reclaims whichever of them its config declares, so
which hooks break is decided by the repo's `lefthook.yml`, not by the hook name `pre-push`.

With a `lefthook.yml` declaring `commit-msg` and `pre-push`, on this branch:

| Entire hook | after `entire enable` | after lefthook reclaims |
|---|---|---|
| `prepare-commit-msg` | Entire | Entire, untouched |
| `commit-msg` | Entire | **reclaimed, no net, unreachable** |
| `post-commit` | Entire | Entire, untouched |
| `post-rewrite` | Entire | Entire, untouched |
| `pre-push` | Entire | reclaimed, **recovered by the net** |

The three untouched rows are untouched only because that config does not declare them. A repo whose
`lefthook.yml` has a `post-commit` section loses `post-commit` the same way.

`pre-push` is the positive control here: the same probe detects the net where this PR adds one, in
the same run in which it finds none for `commit-msg`. So that empty cell is a measured absence, not
a failure to look.

The practical effect is that `Closes #1349` would close the issue with commit-time capture still
silently dead for any repo whose lefthook config declares one of the other four hooks. That is
narrower than the original report, which is only about `pre-push`, so it may be deliberate scoping.
Worth saying out loud either way, since the fix is one `for` loop over `gitHookNames` away from
covering all five.

## Small thing

The branch is based on a June nightly (`f0e7e5b5`, `0.7.6-nightly.202606080755`). It builds and runs
fine, but it will want a rebase before it can land.

Happy to share the probe script if it is useful, or to help extend the net to the other hooks.

PR #1387, opened 2026-06-08
Sign checkpoints at push time

pjbgf:
https://entire.io/gh/entireio/cli/trails/537

 ## Summary

  Defer GPG/SSH signing of checkpoint commits from creation time to push time. During an agent session, checkpoint commits accumulate
  unsigned on the local `entire/checkpoints/v1` ref — no GPG agent calls per step. At pre-push, every local-only commit is cherry-picked
  onto the remote tip with strict signing, with a visible per-commit progress line and an interactive retry/skip/abort prompt on failure.

  ## Why

  - **Performance:** active sessions create many checkpoint commits; the per-step `SignCommitBestEffort` round-trip to the GPG/SSH agent
  dominated commit latency.
  - **Visibility:** signing was silent — users couldn't tell when the agent was being invoked or which commit it was signing.
  - **Correctness:** local-only checkpoints don't need signatures; only what leaves the machine.

  ## What changes

  - `CreateCommit` and the orphan metadata-branch init no longer sign.
  - `createCherryPickCommit` is split into `buildCherryPickCommit` + `persistCherryPickCommit` so a new push-time loop can sign each
  cherry-picked clone between build and persist.
  - New strict `SignCommit(ctx, commit) error` alongside the existing `SignCommitBestEffort` (now a logging wrapper). `ErrSigningDisabled`
  sentinel.
  - `PrePush` fetches the remote tip into the appropriate ref (temp ref for URL targets), runs the sign-loop over local-only commits,
  advances the local ref to the new signed tip, then pushes. On any abort the local ref stays untouched.
  - Cross-clone safety: empty-tree orphans are filtered out, and tree construction uses diff-based application (mirroring `cherryPickOnto`)
   so per-clone files accumulate onto the remote tip rather than overwriting it.

  ## UX

  [entire] Fetching latest from remote...
  [entire] Signing commits:
          10/10: Finalize transcript for Checkpoint: 17d941007e3b
  [entire] Pushing entire/checkpoints/v1 to checkpoint remote... done

  - TTY: a static header and a single status line that rewrites in place — no scrollback noise.
  - Non-TTY (CI): one line per commit, preserved for the audit trail.
  - Signing failure: `Retry, skip, or abort? [r/s/a]:`. Non-TTY defaults to skip (with a `logging.Warn`).
  - Setting: `sign_checkpoint_commits` (default `true`) gates the whole loop.

---

> [!NOTE]
> **High Risk**
> Changes when and how metadata refs are rewritten before push (cherry-pick, ref advance, cross-clone filtering); signing failures can block push or leave unsigned commits on the chain.
> 
> **Overview**
> **Checkpoint commit signing moves from per-step creation to pre-push**, so local `entire/checkpoints/v1` history stays unsigned during sessions and only commits that will leave the machine are signed.
> 
> `CreateCommit` and orphan metadata ref init **no longer call the signer**. Signing is centralized in a new strict `SignCommit` (returns `ErrSigningDisabled` and propagates signer errors); `SignCommitBestEffort` is a thin logging wrapper. Cherry-pick helpers split into **`buildCherryPickCommit` / `persistCherryPickCommit`** so push can sign between build and store.
> 
> **`PrePush`** runs **`signLocalCommitsForPush`** before each checkpoint ref push: fetch remote tip, collect local-only commits (`collectCommitsSince` now supports `exclude == ZeroHash`), optionally drop empty-tree orphan dupes on cross-clone pushes, then **`signAndPersistCommits`** (diff-based cherry-pick + per-commit signing, TTY progress, retry/skip/abort on failure). Abort leaves the local ref unchanged; signing off via `sign_checkpoint_commits` skips the loop. Metadata reconcile still uses best-effort signing on its cherry-pick path.
> 
> Tests cover unsigned `CreateCommit`, strict `SignCommit`, and end-to-end pre-push signing behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b2d3e845e6aaaaade398400d047b9de0ba655d38. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/checkpoint/committed.go
+++ b/cmd/entire/cli/checkpoint/committed.go
@@ -1959,7 +1959,7 @@ func GetGitAuthorFromRepo(repo *git.Repository) (name, email string) {
 
 // CreateCommit creates a git commit object with the given tree, parent, message, and author.
 // If parentHash is ZeroHash, the commit is created without a parent (orphan commit).
-func CreateCommit(ctx context.Context, repo *git.Repository, treeHash, parentHash plumbing.Hash, message, authorName, authorEmail string) (plumbing.Hash, error) {
+func CreateCommit(_ context.Context, repo *git.Repository, treeHash, parentHash plumbing.Hash, message, authorName, authorEmail string) (plumbing.Hash, error) {
 	now := time.Now()
 	sig := object.Signature{
 		Name:  authorName,
@@ -1977,8 +1977,6 @@ func CreateCommit(ctx context.Context, repo *git.Repository, treeHash, parentHas
 	if parentHash != plumbing.ZeroHash {
 		commit.ParentHashes = []plumbing.Hash{parentHash}
 	}
-
-	SignCommitBestEffort(ctx, commit)
 
 	obj := repo.Storer.NewEncodedObject()
 	if err := commit.Encode(obj); err != nil {
--- a/cmd/entire/cli/checkpoint/committed.go
+++ b/cmd/entire/cli/checkpoint/committed.go
@@ -1993,44 +1993,49 @@ func CreateCommit(ctx context.Context, repo *git.Repository, treeHash, parentHas
 	return hash, nil
 }
 
-// SignCommitBestEffort signs the commit using an on-demand object signer.
-// If signing is disabled, no signer can be created, or signing fails, the commit
-// is left unsigned and the error is logged.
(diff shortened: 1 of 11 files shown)


pjbgf:
Closing as refs-based checkpoints removed some of the pain-points that drove this work.

PR #1403, opened 2026-06-10
login: harden the browser sign-in fallbacks

Soph:
https://entire.io/gh/entireio/cli/trails/550

## What

Hardening follow-ups to #1366 (browser sign-in default), from its code review. Stacked on `login-loopback-browser-flow` — review/merge that first.

- **Bound the loopback wait at 5 minutes.** The device flow is bounded by the AS's `expires_in`; the browser flow waited forever if the user closed the tab. The timeout error points at `entire login --device` as the escape hatch, and Ctrl-C still reports as cancellation, not a timeout.
- **SSH sessions fall back to the device flow.** Over `ssh -t` the browser flow was chosen (a TTY is present), but the loopback listener binds 127.0.0.1 on the *remote* host where the user's browser can't reach it. Detect SSH via the `SSH_CONNECTION`/`SSH_CLIENT`/`SSH_TTY` vars sshd sets — same as `gh`/`gcloud` — and explain the fallback in one line on stderr.
- **Listener-bind failure falls back to the device flow.** If `StartBrowserAuth` can't bind (sandboxing, firewall, exhausted ports), warn and continue with the device flow instead of erroring out.

## How

The flow choice moved out of the cobra `RunE` into `runLoginAuto`, which takes a `loginFlowFacts` struct (`useDevice`/`canPrompt`/`sshSession`, detected once at entry) plus a `startBrowser` func, so the selection logic and all three fallback paths are unit-tested with fakes. `startLoginProcess` in the integration tests now blanks the inherited `SSH_*` vars centrally, so browser-flow subprocess tests stay on the browser path even when the developer runs the suite over SSH.

## Testing

- New unit tests: timeout (incl. Ctrl-C-is-not-a-timeout), SSH/headless/listener-failure fallbacks and their stderr messages, `--device` produces no commentary, `isSSHSession`.
- Existing browser-flow integration test still passes; `mise run check` green.

---

> [!NOTE]
> **Medium Risk**
> Changes authentication flow selection and timeout behavior for `entire login`, which is security-sensitive but limited to CLI UX and fallbacks with broad unit test coverage.
> 
> **Overview**
> Hardens **`entire login`** browser sign-in with timeouts, SSH-aware routing, and automatic fallbacks to the device-code flow.
> 
> **Browser wait is capped at 5 minutes** (`browserLoginTimeout`). Loopback redirect wait uses a context timeout starting after the Enter prompt; deadline exceeded yields an error that suggests `entire login --device`, while parent cancellation still surfaces as `context.Canceled`, not a timeout.
> 
> **Flow selection moves into `runLoginAuto`** with a `loginFlowFacts` struct (`--device`, interactive TTY, SSH session). Browser flow is skipped when SSH env vars indicate a remote session (one-line stderr explanation), when headless, or when `--device` is set. If `StartBrowserAuth` fails to bind loopback, the CLI warns and falls back to device flow instead of failing outright.
> 
> Integration **`startLoginProcess`** clears inherited `SSH_*` vars so browser-flow subprocess tests stay stable when the suite runs over SSH. Unit tests cover SSH/headless/listener-failure paths, timeout vs cancel, and `--device` silence on stderr commentary.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 334df98857aeee8ac36d898fdf2c8136c698af53. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/integration_test/login_test.go
+++ b/cmd/entire/cli/integration_test/login_test.go
@@ -216,13 +216,7 @@ func TestLogin_BrowserFlow_SavesToken(t *testing.T) {
 	}))
 	defer server.Close()
 
-	// Blank the SSH_* vars too: startLoginProcess inherits os.Environ(), so a
-	// developer running tests over SSH would otherwise flip the subprocess'
-	// isSSHSession() detection and route it to the device flow.
-	proc := startLoginProcess(t, server.URL, []string{
-		"ENTIRE_TEST_TTY=1",
-		"SSH_CONNECTION=", "SSH_CLIENT=", "SSH_TTY=",
-	}, "login", "--insecure-http-auth")
+	proc := startLoginProcess(t, server.URL, []string{"ENTIRE_TEST_TTY=1"}, "login", "--insecure-http-auth")
 
 	authURL := waitForBrowserPrompt(t, proc.stdout)
 	u, err := url.Parse(authURL)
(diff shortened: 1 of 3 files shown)


toothbrush (approved):
Ah, yes, these are all very sensible improvements i think :+1:

PR #1407, opened 2026-06-10
[bug]Keep auth status dates stable across timezones

stale2000:
Entire logs: https://entire.io/gh/stale2000/cli/session/019eb399-2d0c-7a52-bd77-ee57e66eb3fe#timeline-c93f3cd1bb07 
## Summary

`entire auth status` rendered active-session dates after converting server RFC3339 timestamps into the user's local timezone. That is correct for a full local datetime, but this UI only shows a date, so users west of UTC could see session dates shifted one day earlier than the server-provided date.

## Example

Server session timestamp:

```text
2026-01-01T00:00:00Z
```

The old formatter parsed that timestamp, called `Local()`, then rendered only `YYYY-MM-DD`.

For a user in `America/Los_Angeles`, the same instant becomes:

```text
2025-12-31 16:00:00 PST
```

Because the table hides the time and timezone, the user only saw:

```text
2025-12-31
```

That made the `CREATED`, `LAST USED`, and `EXPIRES` columns appear one day early even though the server timestamp was correct.

## Fix

Format the parsed RFC3339 timestamp in its encoded timezone instead of converting it through `Local()` first. For server UTC timestamps, this preserves the server calendar date:

```text
2026-01-01T00:00:00Z -> 2026-01-01
```

This keeps the date-only auth status table stable across user timezones. If we later want true local-time semantics, the UI should show a full local datetime including time and timezone context.

## Tests

- Added a regression test that forces `time.Local` to PST and verifies `2026-01-01T00:00:00Z` still renders as `2026-01-01`.
- Verified the existing sessions table test under `TZ=America/Los_Angeles`.

```text
TZ=America/Los_Angeles go test ./cmd/entire/cli -run 'TestRunAuthStatus_RendersSessionsTable|TestFormatAuthDate_DoesNotShiftUTCDateToLocalTimezone' -count=1 -v
go test ./cmd/entire/cli -count=1
mise run check
mise run lint
```


Soph (approved):
Thanks! 🙇

PR #1408, opened 2026-06-10
[bug]Fix rewind of tracked checkpoint deletions

stale2000:
Entire Logs: https://entire.io/gh/stale2000/cli/session/019eb3a9-4eee-7440-ac23-ca4fbe5cdd38

COMMENTS:
--------------------------------------------
I'm having a difficult time wrapping my head around this bug, but the logs pretty clearly keep insisting that there is an issue.  It might be better to use this as a guideline to implement a fix because I don't really understand how the fix works.
----------------------------------------------

## What was broken

`entire rewind` is supposed to make the working tree match the selected checkpoint. Checkpoint trees represent deleted files the same way Git trees do: the deleted path is simply absent from the tree.

Before this change, rewind only deleted extra untracked files and then restored files that were present in the checkpoint tree. It did not remove tracked files that were present in the current branch `HEAD` but absent from the checkpoint. That could leave stale files on disk after rewind.

## Concrete example

Start with the current branch containing two tracked files:

```text
HEAD / working tree before rewind
├── keep.txt
└── delete-me.txt
```

The selected checkpoint represents this state:

```text
checkpoint tree
└── keep.txt
```

`delete-me.txt` is not in the checkpoint because it had been deleted when the checkpoint was created.

Old behavior after rewind:

```text
working tree after rewind, before this fix
├── keep.txt        restored correctly
└── delete-me.txt   BUG: stale tracked file remained
```

New behavior after rewind:

```text
working tree after rewind, with this fix
└── keep.txt
```

## Why deletion is safe here

This does not blindly delete arbitrary files. The new deletion pass is scoped to files that are:

- tracked by the current `HEAD` commit, meaning the current branch commit contains that path
- absent from the selected checkpoint tree
- outside protected paths such as `.git`, `.entire`, and agent config directories

Existing untracked-file protections still apply. Untracked files that existed at session start are preserved, and ignored directories are not walked.

## What changed

- Rewind now compares tracked `HEAD` paths with the selected checkpoint tree before restoring present files.
- Tracked files missing from the checkpoint are removed from the working tree.
- `PreviewRewind` reports those tracked deletions so interactive users can see the files that will be removed.
- Added a regression test for a checkpoint that deletes a tracked file.

## Verification

- `mise run lint`
- `go test ./cmd/entire/cli/strategy -run 'TestShadowStrategy_(PreviewRewind|Rewind_)' -count=1`
- `go test ./cmd/entire/cli/strategy ./cmd/entire/cli/checkpoint`

## Notes

Full `mise run check` was attempted, but currently fails in unrelated date-sensitive test `TestRunAuthStatus_RendersSessionsTable`: expected `2026-01-01`, output rendered `2026-01-31`. Strategy/checkpoint tests passed during that run.

### DIFF
--- a/cmd/entire/cli/strategy/manual_commit_rewind.go
+++ b/cmd/entire/cli/strategy/manual_commit_rewind.go
@@ -354,15 +354,10 @@ func (s *ManualCommitStrategy) Rewind(ctx context.Context, w, errW io.Writer, po
 	}
 
 	// Build set of files tracked in HEAD
-	trackedFiles := make(map[string]bool)
-	//nolint:errcheck // Error is not critical for rewind
-	_ = headTree.Files().ForEach(func(f *object.File) error {
-		if err := ctx.Err(); err != nil {
-			return err //nolint:wrapcheck // Propagating context cancellation
-		}
-		trackedFiles[f.Name] = true
-		return nil
-	})
+	trackedFiles, err := trackedFilesInTree(ctx, headTree)
+	if err != nil {
+		return fmt.Errorf("failed to list HEAD files: %w", err)
+	}
 
 	// Get repository root to walk from there
 	repoRoot, err := paths.WorktreeRoot(ctx)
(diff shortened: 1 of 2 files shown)


Soph:
sorry, this took us way to long. `entire rewind` is gone, so closing this

PR #1415, opened 2026-06-11
[bug]Prevent checkpoint symlink target disclosure

stale2000:
Entire logs:  https://entire.io/gh/stale2000/cli/session/019eb504-3cfe-7a92-92ab-e996c653480d

COMMENTARY:  

This one is kinda important.  Its a security bug fix.

## Summary

Checkpoint creation currently materializes changed worktree paths with `os.Stat` followed by `os.ReadFile`. Because both calls follow symlinks, a changed symlink in the repository can be snapshotted as the contents of its target instead of as a Git symlink. For example, an attacker-controlled branch or generated workspace change could introduce `leaked-key -> ~/.ssh/id_rsa`; if Entire checkpoints that changed path, the checkpoint shadow history can persist the victim user's local private key contents.

This is important because checkpoint refs and metadata can later be pushed, bundled, or shared. The attacker does not need arbitrary code execution: they only need to get a symlink into the set of changed/touched files that Entire snapshots during a session or task checkpoint.

## Fix

- Use `os.Lstat` when checking checkpoint file paths so the checkpoint writer inspects the worktree entry itself instead of the symlink target.
- Store symlinks with `filemode.Symlink` and blob content from `os.Readlink`, matching Git's normal symlink representation.
- Keep regular file and executable handling unchanged.
- Use `Lstat` for existence checks so broken symlinks are not silently treated as deleted.

## Tests

- Added regression coverage for first session checkpoints where the symlink is discovered through `git status`.
- Added regression coverage for subsequent session checkpoints where the symlink is supplied through `NewFiles`.
- Added regression coverage for task checkpoints where the symlink is supplied through `ModifiedFiles`.
- Each test verifies the checkpoint tree stores a symlink blob containing the link target, not the external target file contents.

## Verification

- `go test ./cmd/entire/cli/checkpoint`
- `mise run lint`

### DIFF
--- a/cmd/entire/cli/checkpoint/checkpoint_test.go
+++ b/cmd/entire/cli/checkpoint/checkpoint_test.go
@@ -2067,6 +2067,129 @@ func TestWriteTemporary_FirstCheckpoint_CapturesUntrackedFiles(t *testing.T) {
 	}
 }
 
+// TestWriteTemporary_PreservesSymlinkWithoutReadingTarget verifies that changed
+// worktree symlinks are snapshotted as git symlinks, not as the target contents.
+func TestWriteTemporary_PreservesSymlinkWithoutReadingTarget(t *testing.T) {
+	tests := []struct {
+		name              string
+		isFirstCheckpoint bool
+	}{
+		{
+			name:              "first checkpoint untracked symlink",
+			isFirstCheckpoint: true,
+		},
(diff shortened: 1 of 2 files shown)


pjbgf (approved):
@stale2000 thanks for fixing this issue. 🙇

PR #1421, opened 2026-06-11
Compact external agent transcripts for --full/--verbose display

blackgirlbytes:
https://entire.io/gh/entireio/cli/trails/561

## What

`entire checkpoint explain --full` / `--verbose` rendered "(failed to parse transcript)" for external agents (goose, amp): the transcript display path only tries built-in agent parsers and stored compact JSONL, and never asks the external agent binary to convert its native transcript — unlike the summary path, which already routes through `CompactTranscript`.

This runs stored external-agent transcripts through the same compaction before the transcript section renders, in both the committed-checkpoint path and the temporary-checkpoint path. `maybeCompactExternalTranscriptForSummary` is renamed to `maybeCompactExternalTranscript` to reflect the second caller. The helper is a no-op for transcripts the built-in parsers already understand, and returns input unchanged when no compactor is available — so built-in agents are unaffected.

## Why

External agents that declare `compact_transcript` (amp, and now goose via entireio/external-agents#39) store transcripts in their native format. Summaries work because the summary path compacts; the display path was left behind, so the documented `--full` ("Parsed full transcript — all prompts/responses from entire session") shows a parse failure instead of the conversation.

## Testing

- `go test ./cmd/entire/cli -run 'Compact|Transcript|Explain'` passes
- Manually verified against a real goose checkpoint: before — "(failed to parse transcript)"; after — full conversation renders ([User]/[Assistant] turns + [Tool] calls)

## Related

- entireio/external-agents#39 (entire-agent-goose)
- entirehq/entire.io#2354 (web rendering for goose)


Soph (commented):
Looks good. One ask: Can you add at least a test that feeds an external (fake/stub) agent's native transcript through the committed-checkpoint render path and asserts [User]/[Assistant] turns appear instead of the parse-failure string.

Soph (approved):

PR #1427, opened 2026-06-12
docs: Update Homebrew install for Tap Trust (Homebrew 6.0)

Beehtrix:
Use fully qualified cask for scoped trust under Homebrew 6.0 Tap Trust.

### Description

Homebrew 6.0.0 introduced a new "Tap Trust" security model that requires explicit trust for third-party taps before their code is evaluated. The existing instructions use `brew tap` followed by `brew install`, which now triggers trust prompts when tap trust is required.

### Changes

Updated the macOS Homebrew installation snippet to use the fully-qualified cask path:

**Stable**

```bash
brew install --cask entireio/tap/entire
```

**Nightly**

```bash
brew install --cask entireio/tap/entire@nightly
```

This matches Homebrew's Tap Trust documentation, which states that installing a fully-qualified formula or cask (`user/repo/name`) trusts only that item and allows it to be installed even when tap trust is enforced.

### Why this is needed

Using the fully-qualified cask (`entireio/tap/entire`) gives scoped trust to the `entire` cask without requiring users to trust the entire tap. This avoids new Tap Trust warning prompts and keeps the install command simple for new users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,8 +83,8 @@ Entire currently ships two release channels:
 
 How to use each channel:
 
-- Homebrew stable: `brew install --cask entire`
-- Homebrew nightly: `brew install --cask entire@nightly`
+- Homebrew stable: `brew install --cask entireio/tap/entire`
+- Homebrew nightly: `brew install --cask entireio/tap/entire@nightly`
 - `install.sh` stable: `curl -fsSL https://entire.io/install.sh | bash`
 - `install.sh` nightly: `curl -fsSL https://entire.io/install.sh | bash -s -- --channel nightly`
 - Scoop: currently supports `stable` only via `scoop install entire/cli`
--- a/README.md
+++ b/README.md
@@ -44,12 +44,10 @@ With Entire, you can:
 
 ```bash
 # Install stable via Homebrew
-brew tap entireio/tap
-brew install --cask entire
+brew install --cask entireio/tap/entire
 
 # Or install nightly via Homebrew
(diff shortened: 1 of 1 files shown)


Soph:
Thanks for the contribution. I updated other places in docs too. But when testing it I did run into the issue that because we use `conflicts_with` we need to trust both casks (`nightly` and `stable) first or installing one will throw an error about the other. This is not great and I'd like to take some more time to come up with a better solution.

Beehtrix:
> Thanks for the contribution. I updated other places in docs too. But when testing it I did run into the issue that because we use `conflicts_with` we need to trust both casks (`nightly` and `stable) first or installing one will throw an error about the other. This is not great and I'd like to take some more time to come up with a better solution.

Hey @Soph - thanks for getting back on this and appreciate you looking into it! I also proposed something here on the tap itself: https://github.com/entireio/homebrew-tap/pull/4 but yeah, for what it’s worth, I'm hoping the docs change here still seems like an improvement for the common “pick one channel and install it once” path under Tap Trust, since fully‑qualified casks work as documented there.

The `conflicts_with` behavior you’re seeing when both `stable` and `nightly` exist definitely sounds like a deeper design/UX question for the tap itself, not just docs. I’m happy to test any follow‑up you land on, whether that’s keeping `conflicts_with` and documenting a “trust both casks” step for people who switch between them, or reworking how nightly is shipped.

Let me know if I can help in any way! :)

PR #1473, opened 2026-06-18
fix(activity): truncate repo names rune-safely in repo chart

Mohit-Katyal:
## Overview

`renderRepoChart` (the `entire activity` repo chart) shortened long repo names
with a byte slice — `name[:maxNameLen-1] + "…"`. For repo names containing
multi-byte UTF-8 (accents, non-Latin scripts, emoji), that can split a rune
mid-byte and emit invalid UTF-8 to the terminal.

This reuses the file's existing rune-aware `padOrTruncate` helper — the same one
the commit-list path already relies on — which also removes the duplicated
truncate-and-pad logic.

## Test

Adds `TestRenderRepoChart_UnicodeNameSafeTruncation`, mirroring the existing
`TestRenderCommitList_UnicodeMessageSafeTruncation`: it renders a long,
multi-byte repo name and asserts the output is valid UTF-8 and still truncated.
The test fails on the previous byte-slicing code and passes with this change.

> **Low risk** — rendering-only change in the activity dashboard; no auth, API,
> or on-disk behavior touched.

### DIFF
--- a/cmd/entire/cli/activity_render.go
+++ b/cmd/entire/cli/activity_render.go
@@ -352,11 +352,9 @@ func renderRepoChart(w io.Writer, sty activityStyles, repos []repoContribution) 
 	}
 
 	for _, r := range display {
-		name := r.Repo
-		if len(name) > maxNameLen {
-			name = name[:maxNameLen-1] + "…"
-		}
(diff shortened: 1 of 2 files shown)


Soph (approved):

Soph:
Thanks! 🫶

PR #1488, opened 2026-06-21
Session liveness: finalize ACTIVE sessions whose agent has exited

Soph:
https://entire.io/gh/entireio/cli/trails/583

## Problem

A session is marked `PhaseActive` while an agent takes a turn, and the lifecycle relies on a `SessionStop` hook firing to leave ACTIVE. When the agent process goes away mid-turn — a clean `/exit`/Ctrl-D, a crash, a kill, a closed terminal, or a reboot — no hook fires, so the session is stuck ACTIVE forever. The only prior mitigation was `IsStuckActive()`, a coarse **1-hour timeout**, which is both too slow (a session whose agent exited 2 minutes ago still shows active for an hour) and imprecise (a genuinely long, still-running turn gets flagged after an hour).

## Approach

Record the owning agent process's identity at each turn start and detect immediately when that process is gone.

- **`cmd/entire/cli/proclive`** — new stdlib + `x/sys/unix` leaf package. Captures `{PID, start-time fingerprint, host, boot}` by walking up the process tree from the hook to the first non-shell, non-`entire` (and non-`go`) ancestor. `Check()` returns Alive / Dead / Unknown — Dead on a missing PID, start-time mismatch (PID reuse), or reboot; **Unknown** (fail-closed) when it can't confirm the host/boot or the platform can't introspect (Windows), so callers fall back to the timeout.
- **`session.State`** gains `Owner` plus `OwnerLiveness()`/`OwnerExited()`. `OwnerExited()` is true only for an ACTIVE session whose owner is gone.
- **Turn start** (`InitializeSession`) records the owner alongside the branch, cleared-then-set each turn so a failed resolve never leaves a stale owner and the fingerprint tracks agent restarts.
- **`entire status` and `entire doctor`** run a shared `finalizeExitedSessions` sweep up front, finalizing exited sessions on the spot by replaying the missing `SessionStop` (`PhaseEnded` + condense) via the extracted `endSessionNow` — the same path the clean-stop hook runs. The sweep re-checks `OwnerExited()` under the session-state lock to avoid racing a turn that revived the session. Both surfaces also carry an `exited` label (human, `--json`, and doctor's stuck-session reason) as a fallback.

Terminology is **"exited"**, not "crashed" — it covers a clean exit as much as an abnormal one.

## Commits

1. `proclive: add process-liveness package`
2. `session: record owning process and detect exited sessions`
3. `strategy: capture session owner at each turn start`
4. `status, doctor: finalize sessions whose agent has exited`

## Testing

- Unit tests for `proclive` (live-process Alive/Dead, PID-reuse, `/proc` parsing with parens, unsupported-platform Unknown), `OwnerExited`, owner capture at turn start, the finalize sweep, and its under-lock revalidation.
- `mise run lint` → 0 issues; `mise run test:ci` green (unit + integration + Vogon + external-agent canary); cross-compiles on linux/darwin/windows.
- Live smoke: `entire status` finalized a planted exited session (`phase: ended`, `fully_condensed: true`) immediately rather than after an hour.

Reviewed in two passes by Codex; all findings addressed.

## Out of scope

- Windows process liveness (degrades to the timeout).
- `entire session list` "exited" labeling (easy follow-up reusing `OwnerExited`).

---

> [!NOTE]
> **Medium Risk**
> Touches session lifecycle, condensation, and doctor/status auto-finalization; incorrect liveness could end live sessions, though guards and fail-closed Unknown reduce that risk.
> 
> **Overview**
> Adds **process-based session liveness** so ACTIVE sessions are not stuck until the 1-hour inactivity timeout when the agent exits without a `SessionStop` hook.
> 
> A new **`proclive`** package records the owning agent at each turn start (PID + start fingerprint, host/boot guards) and reports alive/dead/unknown. Session state gains **`Owner`** plus **`OwnerExited()`**; turn start clears then re-captures the owner via **`captureSessionOwner`**.
> 
> **`endSessionNow`** centralizes mark-ended + eager condense (shared by SessionStop and the new sweep). **`markSessionEnded`** takes an optional **guard** and returns whether the session actually ended.
> 
> **`finalizeExitedSessions`** runs at the start of **`entire status`** and **`entire doctor`**, ending and condensing exited sessions under lock (re-checking `OwnerExited` to avoid races). Doctor **`classifySession`** treats exited owners as stuck immediately; status shows an **exited** label when finalize could not complete.
> 
> Linux/darwin get real introspection; other platforms degrade to **Unknown** and keep the timeout fallback.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b7e2f3ae69fbb0174e3acad1a772518d796b467e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/doctor.go
+++ b/cmd/entire/cli/doctor.go
@@ -120,6 +120,14 @@ func runSessionsFix(cmd *cobra.Command, force bool) error {
 	}
 	defer repo.Close()
 
+	// Finalize any ACTIVE session whose agent process has exited (no SessionStop
+	// hook fired). A gone process is unambiguous, so these are condensed on the
+	// spot rather than left for the interactive prompt below; the sweep marks
+	// them ended in place so classifySession won't re-flag them.
+	if n := finalizeExitedSessions(ctx, states); n > 0 {
+		fmt.Fprintf(cmd.OutOrStdout(), "Finalized %d exited session(s) (agent process gone).\n\n", n)
+	}
+
 	// Identify stuck sessions
 	now := time.Now()
 	var stuck []stuckSession
@@ -210,14 +218,22 @@ func classifySession(state *strategy.SessionState, repo *git.Repository, now tim
 
 	switch {
 	case state.Phase.IsActive():
-		if !state.IsStuckActive() {
-			return nil
-		}
-
 		var reason string
-		if state.LastInteractionTime != nil {
+		switch {
+		case state.OwnerExited():
+			// Detected immediately (no timeout wait): the owning agent process
+			// is gone. Normally finalized up front in runSessionsFix; this
+			// branch covers a session that couldn't be finalized there.
+			pid := 0
+			if state.Owner != nil {
+				pid = state.Owner.PID
+			}
+			reason = fmt.Sprintf("agent process %d exited (no longer running)", pid)
+		case !state.IsStuckActive():
+			return nil
+		case state.LastInteractionTime != nil:
 			reason = fmt.Sprintf("active, last interaction %s ago", now.Sub(*state.LastInteractionTime).Truncate(time.Minute))
-		} else {
+		default:
 			reason = fmt.Sprintf("active, started %s ago with no recorded interaction", now.Sub(state.StartedAt).Truncate(time.Minute))
 		}
 
--- a/cmd/entire/cli/lifecycle.go
+++ b/cmd/entire/cli/lifecycle.go
@@ -933,26 +933,39 @@ func handleLifecycleSessionEnd(ctx context.Context, ag agent.Agent, event *agent
 	// the transcript to extract file changes. Cleanup is handled by
 	// `entire clean` or when the session state is fully removed.
 
-	if err := markSessionEnded(ctx, event, event.SessionID); err != nil {
+	if _, err := endSessionNow(ctx, event, event.SessionID, nil); err != nil {
 		logging.Warn(logCtx, "failed to mark session ended",
-			slog.String("error", err.Error()))
-		// Don't attempt eager condense if we couldn't even mark the session ended —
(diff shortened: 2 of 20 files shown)


Soph:
Pushed fixes for the CI lint failure and all review findings (force-pushed; folded into the relevant commits):

**CI lint** — `proc_linux.go` returned unwrapped `os.ReadFile`/`strconv.Atoi` errors (wrapcheck). Local lint runs on darwin so it never compiled the linux file; wrapped them and now lint both `GOOS` locally.

**Cursor Bugbot**
- *JSON status skips exited finalize (Medium)* — `runStatusJSON` now runs `finalizeExitedSessions` before listing, matching the human path, so `--json` no longer leaves orphaned ACTIVE sessions or reports them as `exited`.
- *FullyCondensed set optimistically (Low)* — the sweep no longer hand-sets `FullyCondensed`; it reloads the session from disk after finalizing, so the snapshot reflects the true post-finalize state (condense is fail-open). This also fixed a latent issue where stale in-memory `StepCount` could make doctor re-flag a just-finalized session.

**Copilot**
- *FullyCondensed optimism* — same fix as above.
- *ResolveOwner records empty Host on hostname failure* — now fails closed: it records no owner if `os.Hostname()` fails, so the host guard is always present.

**Trail #583 finding (Medium, macOS boottime/NTP)** — darwin no longer records a boot guard; `kern.boottime` drifts when the wall clock is stepped and could falsely declare a live session dead. darwin relies on its absolute `P_starttime` fingerprint (fixed at process creation, distinguishes a reused PID across reboots); Linux keeps the `boot_id` guard. Resolved on the trail.

All commits build on darwin/linux/windows; `mise run lint` clean on both `GOOS`; unit + integration + canary green.

pfleidi (approved):

Issue #1471, opened 2026-06-18
Proposal: Store Checkpoints In Per-Checkpoint Git Refs


pfleidi:
## Problem

Entire currently stores committed checkpoints in one append-only Git branch, `entire/checkpoints/v1`. That has a few problems:

- The Git repository keeps growing without clear cleanup boundaries.
- GitHub treats `entire/checkpoints/v1` like a normal branch and may show “open a pull request” prompts after checkpoint pushes.
- In new repositories, GitHub has sometimes selected `entire/checkpoints/v1` as the default branch because no other branch existed yet.
- Pushes are broader than necessary because one branch represents all checkpoint history.

The proposed direction is to store each checkpoint under its own Git ref.

## Related Work

This should build on the checkpoint-store refactor tracked in https://github.com/entireio/cli/issues/1433.

Ideally the pluggable store boundary lands before the ref-store work starts. In practice, this work may happen in parallel across 2-4 people. The ref-store design should follow the direction of that issue and avoid introducing a second abstraction that would need to be reconciled later.

The migration approach can reuse ideas from https://github.com/entireio/cli/pull/1397, but production checkpoint refs should point to commits, not raw tree objects.

## Storage Model

Each checkpoint gets one stable ref: `refs/entire/checkpoints/<shard>/<checkpoint-id>`.

The ref points to a checkpoint commit containing the checkpoint tree. The first commit for a migrated checkpoint can be an orphan commit that reuses the existing `entire/checkpoints/v1` subtree. Later updates can advance the same ref and preserve per-checkpoint history.

The ref should not point directly to a raw tree object. Commits work better with existing Git tooling and let us keep using Git commit signing.

For new CLI versions, the ref store is authoritative. If a checkpoint is written to refs, the CLI should read it from refs so ref-store issues are visible during rollout.

## Metadata Versioning

Root checkpoint metadata should include both `cli_version` and `checkpoints_version`.

`cli_version` records the CLI that wrote the checkpoint. `checkpoints_version` records the checkpoint storage format, for example `refs-1`. Session metadata does not need a separate checkpoint storage version.

## Checkpoint IDs

Existing checkpoint IDs cannot change because they are already referenced from Git history. Legacy IDs should keep the current prefix-shard layout. It would be possible to create a commit/checkpoint mapping using ref structure but that would add additional complexity.

[...]

PR #1494, opened 2026-06-22
Record checkpoint storage version metadata

pfleidi:
https://entire.io/gh/entireio/cli/trails/636

**Why**

Refs #1471.

Checkpoint metadata needs to state which storage format created the root
checkpoint summary. Branch-backed checkpoints are the first format, and
existing metadata without a version should continue to read without migration.

**What changed**

Root checkpoint summaries now carry `checkpoint_version: "branch-v1"`.
Readers default omitted versions to `branch-v1`, and updates preserve any
explicit version already on the summary.

**Implementation decisions**

The field lives at the checkpoint root because all sessions in a checkpoint
share one storage format. Session metadata stays unchanged, and this remains
metadata-only; it does not introduce the later custom-ref configuration path.

**Tradeoffs and alternatives considered**

A generic `v1` would be shorter, but it would make the later custom-ref format
reuse awkward. `branch-v1` names both the current storage family and its first
version while leaving room for a future `refs-v1`.

---

> [!NOTE]
> **Low Risk**
> Metadata-only JSON field on checkpoint root summaries; no auth, migration, or session blob layout changes.
> 
> **Overview**
> Root **CheckpointSummary** metadata now records which storage format produced the checkpoint via **`checkpoint_version`**, set to **`branch-v1`** for branch-backed checkpoints on `entire/checkpoints/v1`.
> 
> **Readers** treat a missing field as **`branch-v1`** through **`normalizeCheckpointSummary`**, so older summaries need no migration. **Writers** stamp new summaries with **`branch-v1`** and **reuse any explicit version** already on the root summary when rewriting (e.g. **`WriteCommitted`**, **`UpdateCheckpointSummary`**). Session-level metadata is unchanged.
> 
> Tests cover legacy backfill on read/update, persistence of a future version like **`refs-v1`**, and docs show the new JSON field.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 90b15155fa81f244dbbd6253d35d2757737bbb43. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/checkpoint/checkpoint.go
+++ b/cmd/entire/cli/checkpoint/checkpoint.go
@@ -28,6 +28,9 @@ var (
 	ErrNoTranscript = errors.New("no transcript found for checkpoint")
 )
 
+// CheckpointVersionBranchV1 identifies the branch-backed checkpoint metadata format.
+const CheckpointVersionBranchV1 = "branch-v1"
+
 // Checkpoint represents a save point within a session.
 type Checkpoint struct {
 	// ID is the unique checkpoint identifier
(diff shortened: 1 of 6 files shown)


pfleidi:
Bugbot run

computermode (approved):

Soph (approved):

PR #1495, opened 2026-06-22
checkpoint: split into persistent/ephemeral stores with generic Read/Write

Soph:
https://entire.io/gh/entireio/cli/trails/639

## What

The persistent/ephemeral rework of the checkpoint store (issue #1433 follow-up; addresses the naming/abstraction review feedback). Done as a series of small, individually reviewable commits, each validated with `codex exec review` before the next.

**Structural split** — `GitStore` no longer does both jobs:
- `GitStore` is now the **persistent** (committed) store.
- a new independent **`ephemeralStore`** owns the shadow-branch (temporary) surface; the two share only package-level helpers.

**Generic, symmetric surface** on both stores (the review's `sto.Persistent.Read()` / `sto.Ephemeral().Read()` shape): `Read` (differentiated by return type), `List`, and `Write` (a sealed request union) on each.

**Writes named by scope** — the union says *what* is written, not the verb (the method is already `Write`):
- Persistent: `Write(Session{…} | SessionTranscript{…} | SessionSummary{…} | CheckpointAttribution{…})` — three session-level requests and one checkpoint-level.
- Ephemeral: `Write(Step{…} | TaskStep{…})` — a session step and a subagent task step (both produce an ephemeral shadow-branch checkpoint; named to match the strategy's `SaveStep`/`SaveTaskStep`). Bare nouns like `checkpoint.Checkpoint` were rejected (stutter; both requests write the same artifact).

**Reads tiered by the same scope split:**
- `CheckpointReader` (`Read`, `List`) — checkpoint-level.
- `SessionReader` (`ReadSessionContent`, `ReadSessionMetadata`, `ReadSessionPrompts`, `ReadSessionMetadataAndPrompts`) — session-level.
- `PersistentStore` composes `CheckpointReader + SessionReader + Writer`; callers depend on the narrowest tier they need.

**Vocabulary** — "Committed"/"Temporary" → "Persistent"/"Ephemeral" throughout (types, methods, the `Type` enum, `CommittedRefs`→`PersistentRefs`, file names), plus `InitialAttribution`→`Attribution` (only one attribution type; "Initial" was misleading).

**Facade**: `Stores.Primary`→`Stores.Persistent`, `Stores.Temporary()`→`Stores.Ephemeral()`.

## Commits

Split & vocabulary:
1. drop redundant `createCommit` wrapper
2. **split** the temporary store out of `GitStore` → `ephemeralStore`
3. `InitialAttribution` → `Attribution`
[...]

Issue #1433, opened 2026-06-15
Pluggable checkpoint stores: abstraction refactor (tracking)


Soph:
## Goal

Make it cheap to add and experiment with new checkpoint storage backends. Adding a store should mean: implement read + write (+ push/fetch if it needs a remote round-trip), register it, and select it via settings — **without** touching the ~15 construction sites or reimplementing git-specific machinery.

This is the runtime half that [#1051](https://github.com/entireio/cli/pull/1051) (settings schema v2: `checkpoints.{primary,mirrors,git}`) was missing — that PR modeled the right config shape but nothing consumes it yet.

## Findings driving the design

1. **One fat interface, one implementation, scattered construction.** `Store` (`cmd/entire/cli/checkpoint/checkpoint.go:70`) is implemented only by `GitStore`, built via `NewGitStore(repo, ResolveCommittedRefs(ctx))` duplicated across ~15 sites in `cli` and `strategy`.
2. **The interface conflates two unrelated halves:**
   - *Temporary/shadow* (`WriteTemporary*`, `ListTemporaryCheckpoints`) — per-step working-tree snapshots on git shadow branches. **Inherently git-only.**
   - *Committed* (`ReadCommitted`, `WriteCommitted`, `UpdateCommitted`, `ListCommitted`, `ReadSessionContent*`) — the permanent record + reads. **The pluggable part.**
3. **Push/fetch — the "update endpoint" — isn't in the store at all.** It lives in `strategy` (`checkpoint_remote.go`, `manual_commit_push.go`, `push_common.go`, `FetchMetadataBranch`), hardcoded to git refs. A new store can't carry its own sync.
4. **No store-type seam exists.** The only variation point, `CommittedRefs` (v1 vs v1.1 mirror), is just different *ref names on the same git backend*.
5. **Rewind is deprecated, not removed (shipped in today's release); removal will take time.** Rewind has **two modes**, and only one is git/temporary-only:
   - *Working-tree rewind* — restores files from shadow-branch commits. **Temporary/git-only.**
   - *Logs-only rewind* — `GetRewindPoints` builds logs-only entries from committed checkpoint metadata, and `RestoreLogsOnly` reads the committed summary + session content from `entire/checkpoints/v1` (`manual_commit_rewind.go:158,646,688`). **This DOES touch committed storage** — i.e. the pluggable surface.

[...]

PR #1504, opened 2026-06-23
checkpoint: extract persistent contract to api/checkpoint

Soph:
https://entire.io/gh/entireio/cli/trails/647

> **Stacked on #1495** (the persistent/ephemeral split + scope-based read/write API). Review/merge that first; this diff is against that branch. Carries the same latest-`main` merge as #1495.

## What

The final step of the #1433 reorder: extract the **persistent-checkpoint contract** into a new `github.com/entireio/cli/api/checkpoint` package — born with the persistent vocabulary and scope-based names (the whole reason we did the rename before relocating).

**Moved to `api/checkpoint`:**
- Persisted document types: `Metadata`, `CheckpointSummary`, `Summary`/`LearningsSummary`/`CodeLearning`, `Attribution`, `SessionFilePaths`, `SessionMetrics`, `CheckpointInfo`, `SessionContent`.
- Option types: `WriteOptions`, `UpdateOptions`, `PrecomputedTranscriptBlobs`.
- Contract surface:
  - read tiers `CheckpointReader` (`Read`, `List`) + `SessionReader` (the four session reads), composed into `PersistentStore`;
  - `Writer` + the sealed `WriteRequest` union (`Session` / `SessionTranscript` / `SessionSummary` / `CheckpointAttribution`);
  - the reader helpers (`ReadCheckpoint`, `ReadLatestSessionContent`, `ReadRawSessionLogForCheckpoint`), the sentinel errors, and `CheckpointVersionBranchV1`.

The contract depends only on **leaf** packages (`agent/types`, `checkpoint/id`, `redact`, go-git `plumbing`) — a backend can implement it without the CLI's agent/TUI/git machinery.

**Stays in `cmd/entire/cli/checkpoint`:** `GitStore` + methods, `Open`, the `Stores` facade, `PersistentRefs`, `AuthorReader`, `normalizeCheckpointSummary`, and the entire git-only ephemeral surface.

## Zero call-site churn

`cmd/entire/cli/checkpoint` imports the contract and re-exports every moved symbol via aliases (`aliases.go`):

```go
type CheckpointSummary = apicheckpoint.CheckpointSummary
var  ErrCheckpointNotFound = apicheckpoint.ErrCheckpointNotFound
func ReadCheckpoint(...) (...) { return apicheckpoint.ReadCheckpoint(...) }
```

All existing importers compile unchanged; the backend imports `api/checkpoint` directly. (Helpers are thin wrapper funcs, not mutable vars — addressing the earlier review note.)

## Cleanup (post-review `/simplify` pass)
- Extracted a shared `getSessionTree` helper: the four `SessionReader` methods each repeated the same context-check → fetching-tree → checkpoint-tree → session-subtree navigation (with identical `ErrCheckpointNotFound` wrapping) before reading their files. Behavior unchanged.

## Notes
- `PrecomputedTranscriptBlobs.isUsable` → exported `IsUsable` (its one caller is now across the package boundary).
- The `Write` union is sealed to `api/checkpoint`, so the impl-package unknown-request test was dropped (no longer expressible); per-request dispatch coverage remains.
- Doc-comment terminology aligned to "persistent" (was "committed") on the moved option types.

## Verification
`go build ./...`, full `go test ./...`, and `mise run lint` (0 issues) green; validated with `codex exec review` (which confirmed `errors.Is` still works across the alias boundary) and a 4-agent `/simplify` review (reuse / simplification / efficiency / altitude).


api/checkpoint/doc.go
```diff
@@ -0,0 +1,13 @@
...
+//
+// It is the pluggable surface from issue #1433: a storage backend implements
+// these interfaces and operates on these types without depending on the CLI's
+// heavy agent runtime, TUI, or git-implementation packages. (It depends only on
+// leaf value packages — agent/types, checkpoint/id — redact, and go-git
+// plumbing.) The git-backed implementation (GitStore, Open, the facade, and the
+// ephemeral shadow-branch surface) lives in cmd/entire/cli/checkpoint, which
+// imports this package and re-exports these symbols as aliases so existing CLI
+// call sites are unaffected.
+package checkpoint
```
pfleidi, line 13: Creating a new file just to define the package and some docs is a bit odd but not-blocking for me.

pfleidi (dismissed):

api/checkpoint/doc.go
pfleidi: We talked on Slack and this seems to be a common pattern in go.

cmd/entire/cli/checkpoint/persistent.go
```diff
@@ -1073,7 +1073,10 @@ func (s *GitStore) Read(ctx context.Context, checkpointID id.CheckpointID) (*Che
 // ReadSessionMetadata reads only the metadata.json for a specific session within a checkpoint.
 // This is a lightweight read that avoids fetching transcript/prompt blobs.
 // sessionIndex is 0-based.
```
pjbgf: Do we still need this comment?

```suggestion
```

pjbgf (changes requested):

cmd/entire/cli/checkpoint/persistent.go
Soph: moved the comment to the right place, thanks!

pjbgf (approved):

PR #1507, opened 2026-06-23
Add checkpoint format compatibility checks

pfleidi:
https://entire.io/gh/entireio/cli/trails/649

**Why**

Checkpoints can now carry an explicit `checkpoint_version`. Older CLIs need a narrow compatibility gate before they decode checkpoint data, otherwise user-driven read commands may try to interpret a format they do not support or fall back to unrelated remote-recovery behavior.

**What changed**

Adds a small checkpoint compatibility package that parses checkpoint format identifiers and reports unsupported read versions. Explain, export, resume, and rewind read paths now validate the stored `checkpoint_version` immediately after loading the checkpoint summary and before decoding session content or transcripts.

`branch-v1` remains the only read-supported checkpoint format in this PR. Legacy checkpoints with no stored version continue to normalize to `branch-v1` through the existing checkpoint summary normalization. Newer known formats such as `refs-v1` produce a clear unsupported-version error.

Unknown format families are intentionally treated as invalid rather than unsupported. That keeps the `Family` model constrained to families this CLI actually knows about. The trade-off is that a future CLI using a brand-new family will produce an invalid-version error in this older CLI until that family is added here, instead of the nicer unsupported-version error used for known but unreadable formats. We are accepting that trade-off to keep the first compatibility gate simple and explicit.

**Usage examples**

Supported legacy and `branch-v1` checkpoints continue to work:

```sh
entire checkpoint explain abc123def456
entire session resume my-branch
```

When a checkpoint stores a newer known version such as `refs-v1`, user-driven read commands fail with a clear error like:

```text
checkpoint abc123def456 uses unsupported checkpoint_version "refs-v1": not read-supported by this Entire CLI
```

For squash-merge commits with multiple checkpoint trailers, `entire session resume <branch>` now fails if any listed checkpoint cannot be read, including unsupported checkpoint versions. If all listed checkpoints are readable, resume chooses the checkpoint with the newest `CreatedAt` timestamp and prints that it is resuming from the latest checkpoint.

**Resume behavior details**

Unsupported-version errors now propagate from all resume read paths instead of being treated as missing metadata:

- direct checkpoint resume through the picker-backed path
- branch resume after reading local checkpoint metadata
- checkpoint-remote and origin fallback reads in `checkRemoteMetadata`
- single-session log restore before transcript content is read
- squash-merge checkpoint selection in `resolveLatestCheckpoint`

This keeps the behavior conservative: a newer or unreadable checkpoint nudges the user to upgrade or fetch the missing metadata rather than silently resuming from a different checkpoint that happened to be readable.

**Decisions made during development**

This PR only handles per-checkpoint read compatibility. It intentionally does not add the repository policy ref, policy command, hook warnings, or policy enforcement; those are split into follow-up PRs to keep this review focused.

The read check lives at each command boundary after `CheckpointSummary` is loaded instead of inside low-level checkpoint storage. That keeps storage capable of preserving newer metadata for future clients while user-driven commands stop before decoding unsupported content.

The parser recognizes only known format families. Known unreadable formats become unsupported-version errors; unknown families remain invalid-version errors by design.

**Technical tradeoffs**

Failing multi-checkpoint resume on the first unreadable checkpoint is stricter than attempting to pick the newest readable checkpoint. The tradeoff is intentional: the confusing case should be rare, and a clear error is easier to reason about than resuming from a checkpoint that may not represent the newest session state.

Treating unknown families as invalid is also intentional. It gives up a slightly nicer forward-compatibility error for brand-new families in exchange for keeping the typed family model narrow and avoiding acceptance of arbitrary typo families as syntactically valid policy input.

The PR does not verify checkpoint commit signatures. Signature verification is related to trusting policy/config changes, but it is outside the scope of this read-compatibility split.

**Reviewer notes**

This is PR 1 of the checkpoint policy split. Follow-up stacked PRs will add the hidden `entire policy checkpoint` command and repository policy storage/enforcement.

---

> [!NOTE]
> **Medium Risk**
> Touches core user flows (resume, rewind, explain/export) with stricter multi-checkpoint and remote-fallback behavior; mistakes could block legitimate resumes or mis-handle version strings, but changes are read-path only with broad test coverage.
> 
> **Overview**
> Adds a **`checkpointpolicy`** package that parses `family-vN` checkpoint formats and allows reads only for **`branch-v1`** (empty version still normalizes to branch-v1). Known but unreadable versions such as **`refs-v1`** return a typed unsupported-version error via **`EnsureCanReadVersion`** and **`IsUnsupportedVersion`**.
> 
> **Explain**, **export** (JSON and transcript streaming), **resume**, **rewind**, and **manual-commit rewind** now call that check right after loading **`CheckpointSummary`** and before session content or transcripts are decoded. Unsupported versions fail with an explicit message instead of being treated as missing metadata or falling through to remote recovery.
> 
> **Resume** behavior is stricter for squash merges with multiple checkpoint trailers: **`resolveLatestCheckpoint`** fails on the first unreadable checkpoint (including unsupported version) rather than skipping bad IDs and picking the newest readable one; user messaging refers to resuming from the **latest checkpoint**. Remote metadata fallback surfaces unsupported-version errors when that is what was found on the remote tree.
> 
> Tests cover policy parsing, rejection paths across explain/export/resume, and e2e resume output expectations.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit efc3ca2b4fd92be9deb70abc668cbd6315e495b7. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pfleidi:
Bugbot run

cmd/entire/cli/resume.go
```diff
@@ -327,13 +336,16 @@ func resumeFromCurrentBranch(ctx context.Context, w, errW io.Writer, branchName
...
-func resolveLatestCheckpoint(ctx context.Context, store checkpointInfoReader, checkpointIDs []id.CheckpointID) (*strategy.CheckpointInfo, error) {
+// resolveLatestCheckpoint reads metadata for each checkpoint ID and returns the
+// readable checkpoint with the latest CreatedAt.
+func resolveLatestCheckpoint(ctx context.Context, store checkpointInfoReader, checkpointIDs []id.CheckpointID) (*strategy.CheckpointInfo, bool, error) {
 	infoMap := make(map[id.CheckpointID]strategy.CheckpointInfo, len(checkpointIDs))
 	for _, cpID := range checkpointIDs {
 		metadata, readErr := readCheckpointInfoFromStore(ctx, store, cpID)
 		if readErr != nil {
+			if checkpointpolicy.IsUnsupportedVersion(readErr) {
+				return nil, false, readErr
```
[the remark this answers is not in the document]
pfleidi: This is intentional to keep the logic simple and encourage users to upgrade.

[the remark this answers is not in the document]
pfleidi: This is intentional and will be updated in the PR description rather than in code.

pfleidi:
Bugbot run

computermode (approved):

cmd/entire/cli/checkpointpolicy/version.go
```diff
@@ -0,0 +1,46 @@
...
+	if version == "" {
+		version = checkpoint.CheckpointVersionBranchV1
+	}
+
+	format, err := ParseFormat(version)
+	if err != nil {
+		return fmt.Errorf("checkpoint %s has invalid checkpoint_version %q: %w", checkpointID, version, err)
+	}
+	if !CanRead(format) {
+		return &unsupportedVersionError{
```
pjbgf, line 25: Maybe use value-based error instead?

```suggestion
		return unsupportedVersionError{
```

cmd/entire/cli/checkpointpolicy/version.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"github.com/entireio/cli/cmd/entire/cli/checkpoint"
+)
+
+func IsUnsupportedVersion(err error) bool {
+	var unsupported *unsupportedVersionError
+	return errors.As(err, &unsupported)
```
pjbgf, line 12: Doing from memory, so maybe the syntax is slighlty off, but we are better off using `error.Is` in this case:

```suggestion
	return errors.Is(err, unsupportedVersionError)
```

cmd/entire/cli/strategy/manual_commit_rewind.go
```diff
@@ -651,6 +652,9 @@ func (s *ManualCommitStrategy) RestoreLogsOnly(ctx context.Context, w, errW io.W
 	if err != nil {
 		return nil, fmt.Errorf("failed to read checkpoint: %w", err)
 	}
+	if err := checkpointpolicy.EnsureCanReadVersion(point.CheckpointID.String(), summary.CheckpointVersion); err != nil {
+		return nil, fmt.Errorf("%w", err)
```
pjbgf, line 656:
```suggestion
		return nil, err
```

cmd/entire/cli/explain.go
```diff
@@ -754,6 +755,9 @@ func loadCheckpointForExplain(ctx context.Context, lookup *explainCheckpointLook
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to read checkpoint: %w", err)
 	}
+	if err := checkpointpolicy.EnsureCanReadVersion(cpID.String(), summary.CheckpointVersion); err != nil {
+		return nil, nil, fmt.Errorf("%w", err)
```
pjbgf, line 759:
```suggestion
		return nil, nil, err
```

cmd/entire/cli/explain_export.go
```diff
@@ -265,6 +266,9 @@ func runExplainStreamTranscript(ctx context.Context, w, errW io.Writer, opts exp
 	if err != nil {
 		return fmt.Errorf("failed to read checkpoint: %w", err)
 	}
+	if err := checkpointpolicy.EnsureCanReadVersion(cpID.String(), summary.CheckpointVersion); err != nil {
+		return fmt.Errorf("%w", err)
```
pjbgf, line 270:
```suggestion
		return err
```

cmd/entire/cli/explain_export.go
```diff
@@ -359,6 +363,9 @@ func runExplainCheckpointJSON(ctx context.Context, w, errW io.Writer, opts expla
...
```
pjbgf, line 367:
```suggestion
		return err
```

cmd/entire/cli/resume.go
```diff
@@ -359,6 +368,9 @@ func readCheckpointInfoFromStore(ctx context.Context, store checkpointInfoReader
 	if err != nil {
 		return nil, fmt.Errorf("read checkpoint: %w", err)
 	}
+	if err := checkpointpolicy.EnsureCanReadVersion(checkpointID.String(), summary.CheckpointVersion); err != nil {
+		return nil, fmt.Errorf("%w", err)
```
pjbgf, line 372:
```suggestion
		return nil, err
```

cmd/entire/cli/rewind.go
```diff
@@ -722,13 +729,19 @@ func restoreSessionTranscriptFromStrategy(ctx context.Context, cpID id.Checkpoin
 	if err != nil {
 		return "", fmt.Errorf("open checkpoint store: %w", err)
 	}
-	content, returnedSessionID, err := checkpoint.ReadRawSessionLogForCheckpoint(ctx, stores.Persistent, cpID)
+	summary, err := checkpoint.ReadCheckpoint(ctx, stores.Persistent, cpID)
+	if err != nil {
+		return "", fmt.Errorf("failed to read checkpoint: %w", err)
+	}
+	if err := checkpointpolicy.EnsureCanReadVersion(cpID.String(), summary.CheckpointVersion); err != nil {
+		return "", fmt.Errorf("%w", err)
```
pjbgf, line 737:
```suggestion
		return "", err
```

Issue #1336, opened 2026-06-02
Bulk Import existing Claude Code transcripts (~/.claude/projects/*/*.jsonl) into Entire to make cold start better


ishaan812:
### Problem or use case

Entire only captures Agent session data/ any session data when hooks fire during a live session. Users with months of pre-existing transcripts under ~/.claude/projects/<slug>/*.jsonl have no way to bring that history into Entire — entire session list, entire checkpoint explain, and token-usage aggregates all start from zero on adoption day. Current Issue only captures a probable solution for claude code though.

The main issue is the cold start of a repo onto entire and if I already have the transcripts available locally for my Repo, I should be able to add it to the context atleast, to use dispatch, recap and other entire features but not have first class features like rewind, checkpoint etc.

### Desired behavior

```shell
Add a new top-level command:
entire import claude-code [--path DIR] [--dry-run] [--since DATE] [--session ID]... [--force]
By default it auto-scopes to ~/.claude/projects/<SanitizePathForClaude(repoRoot)>/*.jsonl (the standard Claude Code layout for the current repo) and idempotently writes each transcript into Entire’s checkpoint metadata.
```

### Proposed solution

Imported checkpoints are read-only history. They preserve:

per-turn transcript slices (one checkpoint per user-prompt turn, matching live-session granularity)
token usage (input / output / cache create / cache read, with streaming dedup by message.id)
subagent token rollups (via the existing agent-<id>.jsonl lookup)
prompts, model name, timestamps, original cwd
They are not rewindable. Live sessions snapshot working-tree state at every step; historical transcripts don’t have that, and reconstructing it from Write/Edit tool_use blocks is fragile (misses out-of-band edits, conflicts with later commits). Faking a snapshot would be dishonest. entire checkpoint rewind <imported-id> should refuse with a clear message.

blackgirlbytes:
@pfleidi had a followup question for you @ishaan812 "Just to clarify: We'd essentially create new checkpoints but not assign them to any commits when doing import, correct?"

ishaan812:
Thats correct, thus making these imported sessions available for entire functionality (search, explain, using the UI for viewing sessions) other than checkpoint related things such as rewind.

ishaan812:
Hey @blackgirlbytes, I’d love to get your thoughts on whether this aligns with Entire’s long-term vision for the CLI and is something the team would want to support officially. I see a lot of value in it for my own workflows and would be happy to work on it, so just wanted to check product direction before i start working on it.

[...]

PR #1527, opened 2026-06-25
Add import tool for sessions prior to enabling Entire

computermode:
https://entire.io/gh/entireio/cli/trails/668

## What

Adds the ability to import a coding agent's pre-existing local transcripts into Entire as read-only checkpoints, so a repo adopting Entire isn't starting from zero (`checkpoint list`/`explain`, `search`, `tokens`, and server-side recap/UI work on day-one history). Supports #1336 (enable doesn't prompt to import yet - requires manual kickoff for validation first). Claude Code is the only agent implemented; the package is structured so others can be added.

Surface: a hidden command, discoverable via `entire labs`:

```
entire import claude-code [--path DIR] [--dry-run] [--session ID]...
```

There is intentionally **no** enable-time prompt this pass — importing is an explicit, opt-in command.

## How

- New `cmd/entire/cli/agentimport/` package. The orchestration (discovery loop, idempotent per-turn IDs, redaction, checkpoint write) is agent-agnostic; each agent plugs in an `Importer` (`Discover` + `SplitTurns`). A static list of supported importers backs `entire import <agent>` — adding an agent is a new `Importer`, no command wiring.
- Discovers `~/.claude/projects/<sanitized-repo-path>/*.jsonl` (fixed **one-month** lookback), splits each session into **per-user-prompt turns**, and writes one checkpoint per turn — reusing existing transcript parsing, per-turn token accounting (incl. subagent rollups), and redaction (`redact.JSONLBytes`).
- Imported checkpoints are stored on the normal `entire/checkpoints/v1` metadata branch, tagged `Kind:"imported"` with provenance (source, session/turn UUIDs, line range, content hash, import version). They're **commit-less** (no `Entire-Checkpoint` trailer) and surface in `list`/`explain` via an `Imported` flag, labeled `[imported]`.
- **Idempotent**: per-turn checkpoint IDs are `sha256(sessionID + "/" + turnUUID)`, so re-running skips already-imported turns.
- **Not rewindable**: `checkpoint rewind` refuses imported checkpoints with a clear read-only message.

## Tradeoffs / known limitations (do later)

- **Import is manual.** It is not offered during `entire enable` (an earlier iteration prompted at enable time; removed for now). Users run `entire import claude-code` when they want it.
- **Imported history is pushed** to the backend like normal checkpoints (it's what lights up server-side search/recap/UI). **Privacy note:** up to a month of old conversations (and provenance such as `transcript_path`) sync on the next push, unless checkpoint pushing is disabled. (An earlier iteration kept imports local-only on a separate `entire/imports/v1` ref; reverted — it caused explain-by-id and `[temporary]`-duplication bugs and was a cross-cutting overhaul. Single-ref is simpler; a local-only/opt-in-to-push mode could be revisited.)
- **Read-only** by design: imported checkpoints have no working-tree snapshot, so they are not rewindable.
- **Claude Code only** this pass (the package and `entire import <agent>` surface leave room for others).

## Reviewer notes

- `Kind:"imported"` + `Provenance` + an `Imported` umbrella flag were added to the checkpoint metadata contract (`api/checkpoint`).
- Verified locally: `mise run lint` (0 issues) and `mise run test:ci` (unit + integration + e2e canary) green; an end-to-end integration test covers dry-run, import, `[imported]` labeling (no `[temporary]` duplication), explain-by-id, idempotency, and rewind refusal.


api/checkpoint/metadata.go
```diff
@@ -136,6 +151,10 @@ type WriteOptions struct {
 	// Kind identifies the session purpose (e.g., "agent_review"). Empty for normal sessions.
 	Kind string
 
+	// Provenance records the source of an imported checkpoint. Only set when
+	// Kind == "imported".
+	Provenance *Provenance
```
pfleidi: I'm wondering if this would be interesting information to have for non-imported checkpoints as well. Are these import-specific or are we planning to extend this to regular checkpoints in the future?

api/checkpoint/metadata.go
```diff
@@ -262,6 +281,10 @@ type CheckpointInfo struct {
 	// Multi-session support
 	SessionCount int      // Number of sessions (1 if single session)
 	SessionIDs   []string // All session IDs that contributed
+
+	// Imported is true when this checkpoint was imported from pre-existing
+	// agent history (Kind == "imported"): read-only and commit-less.
+	Imported bool
```
pfleidi, line 268: On a quick glance, this looks somewhat redundant given that we could check `Kind == "imported"` and/or the existence of Provenance. Is there anything specific for which this boolean is needed that we can't do with the other values?

api/checkpoint/metadata.go
```diff
@@ -11,6 +11,21 @@ import (
 	"github.com/go-git/go-git/v6/plumbing"
 )
 
+// Provenance records where an imported checkpoint came from, so explain/search
+// can cite the original transcript honestly and re-import stays idempotent.
+// Only set for imported (Kind == "imported") checkpoints.
+type Provenance struct {
```
pfleidi: I couldn't find this information being read in this PR. Did I miss anything? It looks like we're writing it to `metadata.json` but I can't seem to find any cases where we're doing something with it. Am I missing something?

api/checkpoint/metadata.go
computermode: They're import specific for now yeah.. I'm not sure if we'll want to extend them but the option is there

api/checkpoint/metadata.go
computermode: We could use the other values, but Kind and Provenance live on the session level whereas this is on the checkpoint level, so I think we'd have to check each session without having this top-level bool here. We have a similar pattern on `CheckpointSummary` in `persistent.go` where we have bools like `HasReview` or `HasInvestigation` that serve as top-level bools for everything under that checkpoint.

pfleidi: Yeah, that makes sense. 👍

pfleidi (approved):

PR #1540, opened 2026-06-26
import: support cursor, pi, factory, codex, copilot, gemini

computermode:
https://entire.io/gh/entireio/cli/trails/676

## What

Extends `entire import` (PR #1527, Claude Code only) to the other supported agents: **Cursor, Pi, Factory AI Droid, Codex, Copilot CLI, and Gemini CLI**. Each `entire import <agent>` subcommand now surfaces automatically. Supports https://github.com/entireio/cli/issues/1336.

OpenCode is intentionally **out of scope**: it has no on-disk transcript Entire can read (transcripts only materialize via `opencode export`), and its sessions aren't repo-attributable.

## How

PR #1527 built the agent-agnostic orchestration (discovery loop, idempotent per-turn IDs, redaction, checkpoint write) behind an `Importer` seam. This adds one self-contained `Importer` per agent and registers them in the `importers` slice — no command wiring, since `import_cmd.go` iterates `agentimport.All()`.

Each importer reuses its agent's **existing** transcript methods rather than re-encoding parsing:

| Agent | Granularity | Tokens / Model |
|---|---|---|
| cursor | per-turn | none (nil tokens, empty model — records neither) |
| pi | per-turn | `CalculateTokenUsage` + `ExtractModel` (branch-aware) |
| factoryai-droid | per-turn | subagent-aware `CalculateTotalTokenUsage`; model from adjacent `.settings.json` |
| codex | per-turn | cumulative-usage delta; **global store**, filtered by `session_meta.cwd` |
| copilot-cli | per-turn | `CalculateTokenUsage`; **flat store**, filtered by `session.start` cwd/gitRoot |
| gemini | per-session | single JSON doc, message-index offsets → one checkpoint/session |

Codex and Copilot share a `repoMatches` helper (equal-or-descendant, symlink-normalized) so their global/flat stores only import the current repo's sessions.

Imported checkpoints keep the merged Claude behavior: commit-less, `Kind:"imported"`, `[imported]` label, not rewindable, idempotent per `sha256(sessionID + "/" + turnUUID)`.

No changes to `claude.go`, `Run`, the `Importer` interface, or `Turn`.

## Testing

- Per-agent unit tests: `Discover` (lookback / `--session` filter / repo-match) and `SplitTurns` (turn boundaries, prompt/model/tokens; Gemini single-turn).
- Registry test asserts all seven importers register with distinct names.
- End-to-end `Run` test through the Cursor importer (nil tokens / empty model) confirms the generic pipeline + idempotency.
- `mise run lint` (0 issues) and the import/agentimport test suites pass locally. Verified against sample fixtures: cursor 3 turns / pi 2 / factory 2 / codex 2 / copilot 2 / gemini 1.

---

> [!NOTE]
> **Medium Risk**
> Imports local transcripts into persistent checkpoints with format-specific parsing and repo attribution; mistakes could mis-attribute sessions or miscount tokens, but scope is additive with strong test coverage and no auth/data-path changes.
> 
> **Overview**
> Extends **`entire import`** beyond Claude Code by registering six new **`Importer`** implementations (Cursor, Pi, Factory AI Droid, Codex, Copilot CLI, Gemini CLI) in the static `importers` slice—existing command wiring via `agentimport.All()` picks them up automatically.
> 
> Each agent gets its own **`Discover`** (30-day lookback, optional session filter, agent-specific paths) and **`SplitTurns`** logic, mostly delegating tokens/model to existing agent helpers. **Codex** and **Copilot** filter global/flat session stores with shared **`repoMatches`** (cwd/gitRoot under repo, symlink-normalized). **Gemini** differs: one checkpoint per session with message-index offsets, not per user turn. **Cursor** imports with nil tokens and empty model.
> 
> Adds per-agent unit tests, a registry test for all seven names, shared test fixtures, and a **`Run`** end-to-end test through the Cursor importer for idempotency and checkpoint writes without token/model data.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2045376ffaa20acbe3cde17f741851dc37e9680e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/agentimport/copilot.go
+++ b/cmd/entire/cli/agentimport/copilot.go
@@ -92,53 +92,30 @@ func copilotSessionInRepo(path, repoRoot string) bool {
 // Token usage is delegated to the Copilot agent (per-turn slices sum
 // assistant.message outputTokens); the model is read once from the transcript.
 func (copilotImporter) SplitTurns(sf SessionFile, full []byte) ([]Turn, error) {
-	rawLines := splitRawLines(full)
-	var starts []int
-	for i, raw := range rawLines {
-		if _, ok := copilotPromptText(raw); ok {
-			starts = append(starts, i)
-		}
-	}
-
 	ag := &copilotcli.CopilotCLIAgent{}
 	model := copilotcli.ExtractModelFromTranscript(context.Background(), sf.Path)
-	turns := make([]Turn, 0, len(starts))
-	for k, start := range starts {
-		end := len(rawLines)
-		if k+1 < len(starts) {
-			end = starts[k+1]
-		}
-
-		truncated := joinLines(rawLines[:end])
-		tokens, err := ag.CalculateTokenUsage(truncated, start)
-		if err != nil {
-			return nil, fmt.Errorf("token usage for turn %d: %w", k, err)
-		}
-
-		var evt struct {
-			ID        string `json:"id"`
-			Timestamp string `json:"timestamp"`
-		}
-		if err := json.Unmarshal(rawLines[start], &evt); err != nil {
-			continue
-		}
-		ts, parseErr := time.Parse(time.RFC3339, evt.Timestamp)
-		if parseErr != nil {
-			ts = time.Time{}
-		}
-		prompt, _ := copilotPromptText(rawLines[start])
-
-		turns = append(turns, Turn{
-			LineStart: start,
-			LineEnd:   end,
-			UUID:      evt.ID,
-			Prompt:    prompt,
-			Model:     model,
-			CreatedAt: ts,
-			Tokens:    tokens,
+	return splitLineTurns(splitRawLines(full),
+		func(raw []byte) bool { _, ok := copilotPromptText(raw); return ok },
+		func(rawLines [][]byte, start, _ int, truncated []byte) (*Turn, error) {
+			tokens, err := ag.CalculateTokenUsage(truncated, start)
+			if err != nil {
+				return nil, fmt.Errorf("token usage: %w", err)
+			}
+			var evt struct {
+				ID        string `json:"id"`
+				Timestamp string `json:"timestamp"`
+			}
+			if err := json.Unmarshal(rawLines[start], &evt); err != nil {
+				//nolint:nilerr // skip defensively; the line already parsed in copilotPromptText
+				return nil, nil
+			}
+			ts, parseErr := time.Parse(time.RFC3339, evt.Timestamp)
+			if parseErr != nil {
+				ts = time.Time{}
+			}
+			prompt, _ := copilotPromptText(rawLines[start])
+			return &Turn{UUID: evt.ID, Prompt: prompt, Model: model, CreatedAt: ts, Tokens: tokens}, nil
 		})
-	}
-	return turns, nil
 }
 
 // copilotPromptText reports whether a raw events.jsonl line is a user.message
--- a/cmd/entire/cli/agentimport/copilot.go
+++ b/cmd/entire/cli/agentimport/copilot.go
@@ -1,6 +1,7 @@
 package agentimport
 
 import (
+	"bufio"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -73,11 +74,17 @@ func (copilotImporter) Discover(repoRoot, overridePath string, now time.Time, se
 // it in this repo (gitRoot or cwd is the repo root or a descendant). Sessions
 // whose location can't be determined are treated as not belonging to the repo.
 func copilotSessionInRepo(path, repoRoot string) bool {
-	data, err := os.ReadFile(path) //nolint:gosec // path discovered under the configured session dir
+	f, err := os.Open(path) //nolint:gosec // path discovered under the configured session dir
 	if err != nil {
 		return false
 	}
-	for _, raw := range splitRawLines(data) {
+	defer func() { _ = f.Close() }()
+
+	// Scan line-by-line and stop at the first session.start (normally line 0)
+	// rather than slurping the whole transcript, which can be large.
+	scanner := bufio.NewScanner(f)
+	scanner.Buffer(make([]byte, 0, bufio.MaxScanTokenSize), 10*1024*1024) // 10 MB max line
+	for scanner.Scan() {
 		var evt struct {
 			Type string `json:"type"`
 			Data struct {
@@ -87,7 +94,7 @@ func copilotSessionInRepo(path, repoRoot string) bool {
 				} `json:"context"`
 			} `json:"data"`
 		}
-		if err := json.Unmarshal(raw, &evt); err != nil || evt.Type != "session.start" {
+		if err := json.Unmarshal(scanner.Bytes(), &evt); err != nil || evt.Type != "session.start" {
 			continue
 		}
 		return repoMatches(evt.Data.Context.GitRoot, repoRoot) || repoMatches(evt.Data.Context.Cwd, repoRoot)
--- a/cmd/entire/cli/agentimport/copilot.go
+++ b/cmd/entire/cli/agentimport/copilot.go
@@ -16,6 +16,11 @@ import (
 	"github.com/entireio/cli/cmd/entire/cli/agent/types"
 )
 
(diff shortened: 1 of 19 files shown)


cmd/entire/cli/agentimport/pi.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+
+		// Bound token usage to [start, end): truncate to the first `end` lines,
+		// then let the agent helper slice from `start`.
+		truncated := joinLines(rawLines[:end])
+		tokens, err := ag.CalculateTokenUsage(truncated, start)
+		if err != nil {
+			return nil, fmt.Errorf("token usage for turn %d: %w", k, err)
+		}
+		model, mErr := ag.ExtractModel(joinLines(rawLines[start:end]))
```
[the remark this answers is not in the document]
computermode: Addressed in be123063a. `ExtractModel` now runs over the `truncated` (`rawLines[:end]`) buffer instead of `rawLines[start:end]`, so the file's beginning is retained and `parentId` chains reach the root. Truncating only the end is safe (parents are always earlier lines) and still scopes the model to this turn's end. Added a regression test (`TestPiSplitTurns_ModelInheritedOverPrefix`).

cmd/entire/cli/agentimport/pi.go
```diff
@@ -0,0 +1,160 @@
...
+		if mErr != nil {
+			model = ""
+		}
```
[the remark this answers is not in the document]
computermode: Addressed in be123063a — model extraction now uses the same `truncated` (0..end) buffer as the token calc, exactly as suggested, so active-branch resolution sees the full prefix back to the root.

cmd/entire/cli/agentimport/pi.go
```diff
@@ -0,0 +1,160 @@
...
+	for k, start := range starts {
+		end := len(rawLines)
+		if k+1 < len(starts) {
+			end = starts[k+1]
+		}
+
+		// Bound token usage to [start, end): truncate to the first `end` lines,
+		// then let the agent helper slice from `start`.
+		truncated := joinLines(rawLines[:end])
+		tokens, err := ag.CalculateTokenUsage(truncated, start)
```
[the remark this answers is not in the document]
computermode: The token calc already passes `truncated` = `rawLines[:end]` (the file from line 0, just the tail removed), so `ResolveActiveBranch` still walks `parentId` to the root — the documented hazard is removing the *beginning* (SkipLines), not the end. Passing the *full* transcript would instead make every turn include all later turns' tokens, breaking per-turn bounding. Clarified the rationale in a comment in be123063a; no behavior change to the token path.

cmd/entire/cli/agentimport/copilot.go
```diff
@@ -0,0 +1,167 @@
...
+					Cwd     string `json:"cwd"`
+					GitRoot string `json:"gitRoot"`
+				} `json:"context"`
+			} `json:"data"`
+		}
+		if err := json.Unmarshal(raw, &evt); err != nil || evt.Type != "session.start" {
+			continue
+		}
+		return repoMatches(evt.Data.Context.GitRoot, repoRoot) || repoMatches(evt.Data.Context.Cwd, repoRoot)
+	}
```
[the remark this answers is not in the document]
computermode: Addressed in be123063a. `copilotSessionInRepo` now scans the file line-by-line with a buffered scanner and returns at the first `session.start`, instead of reading the whole file and allocating via `splitRawLines`.

cmd/entire/cli/agentimport/factory.go
```diff
@@ -0,0 +1,110 @@
...
+
+		var env struct {
+			ID string `json:"id"`
+		}
+		if err := json.Unmarshal(rawLines[start], &env); err != nil {
+			continue
+		}
+		prompt, _ := factoryPromptText(rawLines[start])
+
+		turns = append(turns, Turn{
```
pfleidi: It looks like this isn't assigning the `CreatedAt` field. Is that not available in the original transcript?

pfleidi (dismissed):

Soph (dismissed):

cmd/entire/cli/agentimport/cursor.go
```diff
@@ -0,0 +1,78 @@
...
+				Timestamp string          `json:"timestamp"`
+			}
+			if err := json.Unmarshal(rawLines[start], &rec); err != nil {
+				//nolint:nilerr // skip defensively; the line already parsed in isUserPromptLine
+				return nil, nil
+			}
+			return &Turn{
+				UUID:      rec.UUID,
+				Prompt:    transcript.ExtractUserContent(rec.Message),
+				CreatedAt: parseTimestamp(rec.Timestamp),
```
Soph: **[Entire trail finding — HIGH] Cursor import silently drops every turn after the first**

Real Cursor transcript lines carry only `role` and `message` — there is no `uuid` or `timestamp` field (see `cursor/AGENT.md`: `{"role":"user","message":{...}}`). So `rec.UUID` is `""` for every turn, and `DeriveCheckpointID(sessionID, "")` produces the **same** checkpoint ID for all user prompts in a session. Every turn after the first is then skipped via the `existing` idempotency map → silent data loss. `CreatedAt` is also always zero.

The `cursor_test.go` fixtures inject synthetic `uuid`/`timestamp` fields, so the tests pass while real transcripts would import only one turn per session.

**Fix:** derive a stable per-turn key from the line index (like `codexImporter` uses `strconv.Itoa(start)`) or from `conversation_id`+offset, and fall back to the file modtime for `CreatedAt` (as `geminiImporter`/`factoryImporter` do). Please also update the fixtures to match the real on-disk format (no `uuid`/`timestamp`) so the regression is actually covered.

cmd/entire/cli/agentimport/copilot.go
```diff
@@ -0,0 +1,128 @@
...
+			if err != nil {
+				return nil, fmt.Errorf("token usage: %w", err)
+			}
+			var evt struct {
+				ID        string `json:"id"`
+				Timestamp string `json:"timestamp"`
+			}
+			if err := json.Unmarshal(rawLines[start], &evt); err != nil {
+				//nolint:nilerr // skip defensively; the line already parsed in copilotPromptText
+				return nil, nil
```
Soph: **[Entire trail finding — MEDIUM] Copilot numeric `timestamp` drops the whole session**

Copilot's `timestamp` can be a numeric epoch-millis value, not only an RFC3339 string — the agent's own `copilotcli/compat.go parseTimestamp` explicitly decodes both forms (tries `int64` first, then string). Here `timestamp` is decoded into a `string` field, so a numeric timestamp makes `json.Unmarshal` return an `UnmarshalTypeError`, which hits the `return nil, nil` defensive skip and silently drops the turn. A session whose events carry numeric timestamps therefore imports **zero** turns.

**Fix:** decode `timestamp` as `json.RawMessage` (or reuse the agent's dual-format `parseTimestamp`) instead of `string`, and add a fixture with a numeric `timestamp` so this path is covered.

cmd/entire/cli/agentimport/cursor.go
computermode: Addressed in b61250a1b. Cursor turns are now keyed by line index (`strconv.Itoa(start)`, the Codex pattern) instead of the always-empty `uuid`, so each prompt gets a distinct checkpoint ID and turns after the first are no longer skipped. `CreatedAt` falls back to the transcript file modtime (the Factory/Gemini pattern). Fixtures rewritten to the real on-disk format (only `role`+`message`) with assertions that turn keys are distinct and `CreatedAt` is non-zero.

cmd/entire/cli/agentimport/copilot.go
computermode: Addressed in b61250a1b. The `timestamp` field is now decoded as `json.RawMessage` and parsed via the agent's dual-format parser (exported as `copilotcli.ParseTimestamp`, which tries epoch-millis then RFC3339), so a numeric timestamp no longer triggers an `UnmarshalTypeError` and drops the session. Added `TestCopilotSplitTurns_NumericTimestamp` covering the numeric path.

Soph (approved):

PR #1581, opened 2026-06-30
Store full compacted transcript per checkpoint with a scope marker

Soph:
https://entire.io/gh/entireio/cli/trails/707

## What

`transcript.jsonl` (the compact transcript) now stores the **full compacted session** on every checkpoint, mirroring how `full.jsonl` already works, instead of only the per-checkpoint delta slice. Each checkpoint records a `compact_transcript_start` marker in its session metadata so consumers can segment a checkpoint's own slice as `compactLines[compact_transcript_start:]`.

## Why

Previously `transcript.jsonl` held only this checkpoint's delta. To reconstruct a session's compact view you had to stitch every checkpoint's slice together in order — so a single mid-history checkpoint lost, reverted, or dropped during a rebase left an unrecoverable gap. Storing the full compacted session per checkpoint makes each checkpoint **self-contained**: the session is reconstructable from any single surviving checkpoint, and a reader can open one file and segment by the marker rather than piecing deltas together.

This matches `full.jsonl`'s existing model and the project's "duplicate data on the checkpoints branch is acceptable" stance.

## How

- **`compact.FullWithBoundary`** compacts the whole transcript and returns the full bytes plus the boundary line index, derived as `lineCount(full) − lineCount(delta)`. This reuses each format's existing, independently-tested slicing behavior (line offsets / message-index / response-item index) rather than threading source indices through all seven emitters.
- **`CompactTranscriptStart *int`** in session metadata (`compact_transcript_start`), with a `GetCompactTranscriptStart()` accessor. A pointer so **nil = legacy delta-only file** (read from line 0) is distinguishable from **0 = full file, first checkpoint**.
- **`persistent.go`** threads the boundary into session metadata in both the normal write path and the pre-push OPF rewrite path. Extracted a shared `updateSessionMetadata` helper (also used by `replaceSkillEvents`).

### Boundary rounding (known, bounded limitation)

A single compact line can carry content from both sides of `StartLine` when its source straddles the boundary — most notably same-ID streaming assistant fragments, which compaction merges into one line that no integer offset can split. The boundary **rounds toward inclusion**: such a merged line stays in the slice, so `compactLines[marker:]` **never drops this checkpoint's content** but its head may repeat ≤1 merged line from the previous checkpoint. Pinned by a regression test; documented on the field and in the architecture doc.

## Tests / verification

- New: per-format boundary tests, straddle-rounding regression test, metadata nil-vs-0 round-trip, full end-to-end store tests (normal + OPF paths).
- Updated the two tests that encoded the old delta semantics to assert the new full-content + marker contract.
- `mise run lint` clean; full `mise run test:ci` green (unit + integration + Vogon canary).

## Follow-up (not in this PR)

The downstream **entire.io** reader must consume `compact_transcript_start`: slice at the marker, treat **absent** as legacy (read from line 0), and tolerate the ≤1-line head overlap. Docs (`metadata.go`, `sessions-and-checkpoints.md`, `CLAUDE.md`) now spell out the contract.

---

> [!NOTE]
> **Medium Risk**
> Changes checkpoint storage semantics and metadata on `entire/checkpoints/v1`; legacy checkpoints remain readable but external readers must adopt the new marker contract.
> 
> **Overview**
> **Compact transcripts now mirror `full.jsonl`:** each checkpoint writes the **entire** compacted session to `transcript.jsonl` instead of a per-checkpoint delta, so a session can be recovered from any surviving checkpoint without stitching prior slices.
> 
> **New `compact.FullWithBoundary`** compacts the whole transcript and returns a **boundary line index** (derived from full vs delta line counts) so readers slice this checkpoint as `compactLines[boundary:]`. Session metadata gains **`compact_transcript_start` (`*int`)** and **`GetCompactTranscriptStart()`** — absent means legacy delta-only files; explicit `0` means “full file, starts at line 0.”
> 
> **`GitStore` persistence** threads the boundary from `writeTranscript` / `writeCompactTranscript` into session metadata on normal writes and OPF finalize (`setCompactTranscriptStart` via shared **`updateSessionMetadata`**). Docs and tests are updated for the full-file contract, including bounded ≤1-line head overlap when streaming fragments straddle the boundary.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 799bb0ad7c97d02f236199dd88b15f6623877cc2. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Soph:
Thanks for the review. Addressed in 05e15705a:

**#3 — OPF rewrite could leave a stale, less-redacted `transcript.jsonl`** (also flagged by the branch's trail finding). Fixed: in `replaceTranscript`, when compact regeneration yields nothing (failure, empty, or oversized), the prior `transcript.jsonl` is now dropped and `CompactTranscriptStart` cleared, so no stale/under-redacted compact ships and the marker can't point at content that no longer matches the re-redacted `full.jsonl`. `setCompactTranscriptStart` now takes `*int` so it can clear the marker; regression test `TestUpdateCommitted_DropsStaleCompactWhenRegenerationProducesNone` added.

**#1 — full `transcript.jsonl` can exceed the 50MB cap and be dropped on long sessions.** Real but low-probability: the cap is on *compact* output (thinking/tool-noise stripped), so hitting 50MB needs an enormous session, and `full.jsonl` (chunked) stays authoritative with the compact regenerable from it. Chosen to **document as a known limitation** rather than add a chunked-read contract for `transcript.jsonl` now (the downstream entire.io reader isn't built yet). Documented in `sessions-and-checkpoints.md` and `CLAUDE.md`. Chunking the compact like `full.jsonl` remains a clean follow-up if it ever bites.

**#2 — `FullWithBoundary` compacts twice.** Intentional tradeoff: deriving the boundary as `lineCount(full) − lineCount(delta)` reuses every format's existing, independently-tested slicing instead of threading source indices through all 7 emitters. The common first-checkpoint case (`StartLine=0`) already skips the second pass. If checkpoint-write CPU ever shows up in profiles, a single-pass boundary for the line-offset formats is the targeted optimization.

computermode (dismissed):

computermode (approved):

PR #1586, opened 2026-06-30
Add telemetry for policy-blocked checkpoint hooks

pfleidi:
https://entire.io/gh/entireio/cli/trails/713

## Why

PR #1541 ("Gate checkpoint writes on policy support") made the CLI skip or reject Entire checkpoint work when a repository's checkpoint policy cannot be satisfied or cannot be read. When that happens inside a hook, checkpoint capture is lost — silently for git hooks, visibly for agent write hooks — and today we have no signal for how often policy gating actually blocks checkpoint creation in the field.

## What changed

Adds an opt-in PostHog event, `checkpoint_policy_blocked`, emitted at the git and agent hook policy gates whenever a hook is prevented from creating checkpoint data because the policy is unsupported or unreadable.

## Event schema

Emitted via the existing detached-subprocess telemetry path (`distinct_id` is the machine ID).

| Property | Values |
|---|---|
| `hook` | `prepare-commit-msg`, `commit-msg`, `post-commit`, `post-rewrite`, `session-start`, `turn-end`, `subagent-end`, `post-todo` |
| `hook_type` | `git` \| `agent` |
| `reason` | `policy_unsupported` \| `policy_unreadable` |
| `outcome` | `skipped` (git hooks + agent session-start) \| `blocked` (agent write hooks) |
| `agent` | agent name (omitted for git hooks) |
| `checkpoint_version`, `checkpoint_min_version` | admin-set policy versions (omitted when `policy_unreadable`) |
| `cli_version`, `os`, `arch` | standard, matching existing events |

All values are operational metadata — no prompts, file contents, commit messages, or paths.

- **Per-gate emission.** Each gated hook emits its own event, so one blocked `git commit` produces up to three events (`prepare-commit-msg`, `commit-msg`, `post-commit`), distinguished by `hook`. The three are separate processes with no shared in-memory state, so dedup is left to downstream analysis (`distinct_id` + timestamp + `hook_type`) rather than persisting a cross-process marker.
- **Rides the existing opt-in telemetry gate.** Emits only when `settings.Telemetry == true` and `ENTIRE_TELEMETRY_OPTOUT` is unset — identical to `cli_command_executed`. This undercounts (only opted-in users), but keeps the privacy model consistent.
- **Boundary.** The `telemetry` package stays settings-agnostic; the `settings.Telemetry` opt-in check lives in a thin CLI helper (`emitCheckpointPolicyBlocked`) to avoid an import cycle, mirroring how `TrackCommandDetached` leaves opt-in to its caller. The new builder/tracker mirror the established per-event `Build*`/`Track*` pattern rather than introducing a generic tracker.

## Technical tradeoffs

- **Over-counting vs. complexity:** per-gate emission over-counts a single user action, but cross-process dedup would add a persisted marker and new failure modes for a best-effort signal. Chose simplicity + downstream dedup.
- **Undercounting vs. privacy:** opt-in gating means the signal misses users who never enabled telemetry. Accepted for privacy consistency rather than adding a separate operational channel.

## Reviewer notes

Scope is intentionally limited to the hook gates. Strategy-internal condensation/finalization, the pre-push hook, and the command gates (`session attach`, `checkpoint import`, `explain --generate`) are not instrumented in this PR.

---

> [!NOTE]
> **Low Risk**
> Behavior of policy gates is unchanged; this only adds best-effort, opt-in telemetry with operational metadata and honors ENTIRE_TELEMETRY_OPTOUT.
> 
> **Overview**
> Adds **observability** for checkpoint policy gating introduced earlier: when hooks refuse checkpoint work because policy cannot be read or the CLI cannot satisfy it, the CLI can now emit a **`checkpoint_policy_blocked`** analytics event (same detached-subprocess path as other telemetry).
> 
> A new **`telemetry`** builder/tracker defines the event schema (`hook`, `hook_type`, `reason`, `outcome`, optional `agent` and policy versions, plus `cli_version`/`os`/`arch`). **`emitCheckpointPolicyBlocked`** in the CLI layer gates emission on **`settings.Telemetry`** and delegates tracking, keeping the telemetry package free of settings imports.
> 
> **Git hooks** call the emitter from `skipUnsupportedCheckpointPolicy` and a refactored **`skipUnreadableCheckpointPolicy`** helper (outcome **`skipped`**). **Agent hooks** emit on session-start skips and on write-path rejects (`turn-end`, `subagent-end`, `post-todo` via new **`agentWriteHookLabel`**), with outcomes **`skipped`** vs **`blocked`** and the agent name included. Payload tests cover supported vs unreadable cases; a small unit test covers hook label mapping.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d276ec2f162cbdd643bdcdbe5642d5c9d5212e34. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/checkpoint_policy_telemetry.go
+++ b/cmd/entire/cli/checkpoint_policy_telemetry.go
@@ -1,0 +1,19 @@
+package cli
+
+import (
+	"context"
+
+	"github.com/entireio/cli/cmd/entire/cli/telemetry"
+	"github.com/entireio/cli/cmd/entire/cli/versioninfo"
+)
+
+// emitCheckpointPolicyBlocked reports a checkpoint_policy_blocked telemetry
+// event when telemetry is opted in (settings.Telemetry == true). Best-effort
+// and non-blocking; failures to load settings simply suppress the event.
+func emitCheckpointPolicyBlocked(ctx context.Context, event telemetry.CheckpointPolicyBlockedEvent) {
+	s, err := LoadEntireSettings(ctx)
+	if err != nil || s.Telemetry == nil || !*s.Telemetry {
+		return
+	}
+	telemetry.TrackCheckpointPolicyBlocked(event, versioninfo.Version)
+}
--- a/cmd/entire/cli/hook_registry.go
+++ b/cmd/entire/cli/hook_registry.go
@@ -20,6 +20,7 @@ import (
 	"github.com/entireio/cli/cmd/entire/cli/logging"
 	"github.com/entireio/cli/cmd/entire/cli/paths"
 	"github.com/entireio/cli/cmd/entire/cli/strategy"
+	"github.com/entireio/cli/cmd/entire/cli/telemetry"
 	"github.com/entireio/cli/cmd/entire/cli/versioncheck"
 	"github.com/entireio/cli/cmd/entire/cli/versioninfo"
 	"github.com/entireio/cli/perf"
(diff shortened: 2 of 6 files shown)


pfleidi:
Bugbot run

computermode (approved):

PR #1595, opened 2026-07-01
feat(enable): offer to import agent history on first-time enable

computermode:
https://entire.io/gh/entireio/cli/trails/771

## What

First-time `entire enable` now proactively offers to import pre-existing agent history for the agents selected during setup, instead of leaving users to discover the hidden `entire import <agent>` command themselves.

## Behavior

- **First-run only** — gated on `!settings.IsSetUpAny(ctx)`, captured before any settings are written. Re-enable never re-offers.
- **Agent-level granularity**, scoped to the agents selected during enable, matched to importers by `AgentType`, and limited to agents with ≥1 discoverable session (30-day lookback, same as `entire import`).
- **Interactive**: a multi-select with everything unchecked; confirming with nothing selected imports nothing.
- **Non-interactive** (`--yes` or no TTY): auto-imports all eligible selected agents, no prompt.
- **Best-effort**: discovery/import failures are logged and surfaced as a notice but never fail `enable`.

## How

- New `cmd/entire/cli/setup_import.go`: `maybeOfferSessionImport` + helpers, reusing the same `strategy.EnsureRedactionConfigured()` + `agentimport.Run` flow as the standalone import command. Small test seams isolate orchestration from disk/TTY/checkpoint writes.
- Wired into **both** first-run enable paths after `strategy.EnsureSetup`: `runEnableInteractive` (bare / `--yes`) and `setupAgentHooksNonInteractive` (`--agent`). Both are reachable on re-enable, so the `firstRun` gate is load-bearing.

## Tests

- Unit (`setup_import_test.go`): first-run gate, non-interactive auto-import, no-eligible no-op, interactive selection, empty-selection skip, importer-by-type matching.
- Integration (`integration_test/enable_import_test.go`): first-run enable imports a discoverable fixture, no-history is silent, re-enable does not re-offer.

## Verification

- `mise run lint` — 0 issues
- `mise run test:ci` — unit + integration + e2e canary all pass

---

> [!NOTE]
> **Medium Risk**
> Writes imported checkpoints to the metadata branch on first enable; failures are non-fatal but mistaken auto-import in CI/no-TTY paths could add unexpected history.
> 
> **Overview**
> **First-time `entire enable` can now pull in existing agent session history** instead of requiring users to find `entire import` separately.
> 
> A new `maybeOfferSessionImport` flow runs after `strategy.EnsureSetup` on both interactive enable and `entire enable --agent <name>`. It only runs when `!settings.IsSetUpAny(ctx)` is captured **before** settings are written, so re-enable never re-offers or re-imports. For each selected agent with a registered importer and discoverable sessions (30-day lookback, same as standalone import), interactive users get an optional multi-select with nothing pre-checked; non-interactive (`--yes` or no TTY) auto-imports all eligible agents. Import uses the same `EnsureRedactionConfigured` + `agentimport.Run` path as `entire import` and is best-effort—failures are logged and surfaced as notes but do not fail enable.
> 
> Unit tests cover orchestration via injectable seams; integration tests cover first-run auto-import, silent no-op when no history, and no import on re-enable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 885859c030ab68875dc9879b0f126d916888828a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmd/entire/cli/setup_import.go
```diff
@@ -0,0 +1,189 @@
...
+			huh.NewMultiSelect[string]().
+				Title("Import existing sessions into Entire? (optional)").
+				Description("Space to select, enter to confirm. Select none to skip.").
+				Options(options...).
+				Value(&chosen),
+		),
+	)
+	if err := form.Run(); err != nil {
+		// Cancellation returns nil here => skip import; other errors propagate.
+		return nil, handleFormCancellation(w, "Import", err)
```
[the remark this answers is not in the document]
computermode: Addressed in 87360f3b4: `promptImportSelection` now takes `ctx` and runs the form with `form.RunWithContext(ctx)`, so parent cancellation stops the prompt and `handleFormCancellation` handles `context.Canceled`.

cmd/entire/cli/setup_import.go
```diff
@@ -0,0 +1,189 @@
...
+		return nil
+	}
+
+	selected := eligible
+	if !opts.Yes && interactive.CanPromptInteractively() {
+		selected, err = sessionImportPrompt(w, eligible)
+		if err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
computermode: Addressed in 87360f3b4: `maybeOfferSessionImport` no longer returns an error. On a prompt/UI failure it now logs a warning, prints a `Note:` to the user, and returns to skip import, so a terminal/UI error can never fail `entire enable`.

cmd/entire/cli/setup_import.go
```diff
@@ -0,0 +1,225 @@
...
+		return
+	}
+
+	eligible := sessionImportDiscover(ctx, agents, repoRoot)
+	if len(eligible) == 0 {
+		return
+	}
+
+	selected := eligible
+	if !opts.Yes && interactive.CanPromptInteractively() {
```
pfleidi: Right now, the [regular import command](https://github.com/entireio/cli/blob/8f5df76e3c5c17b762be7d488edd5875df86f59c/cmd/entire/cli/import_cmd.go#L59-L61) also validates a policy before running the import:

```golang
ensureCheckpointPolicyAllowsCheckpointData(ctx, repo)
```

I'm not sure whether this policy machinery will stay but for now, we should probably keep the behavior consistent in that regard.

cmd/entire/cli/setup_import.go
```diff
@@ -0,0 +1,225 @@
...
+		logging.Warn(ctx, "session import offer skipped: no worktree root", "error", err)
+		return
+	}
+
+	eligible := sessionImportDiscover(ctx, agents, repoRoot)
+	if len(eligible) == 0 {
+		return
+	}
+
+	selected := eligible
```
pfleidi, line 65: I think this means that the `enable` command will auto-import all eligible sessions if the command is run e.g. from an agent which doesn't have an interactive terminal (`interactive.CanPromptInteractively()` is false).

I think I'd be somewhat surprised by that behavior. Is that intended?

computermode: Do you mean if someone uses an agent to run `entire enable` on their behalf?

pfleidi: Yes, that's what I meant.

pfleidi: Or if they run enable as part of a script or something else that doesn't have an interactive shell.

computermode: Should be good to go now 👀

pfleidi (approved):

PR #1600, opened 2026-07-02
feat(mirror): warn when `repo mirror create` hits an admin-suspended mirror

toothbrush:
https://entire.io/gh/entireio/cli/trails/726

When `entire repo mirror create` re-registers an existing placement that an admin has suspended, it now prints a warning right after the placement message:

```
WARNING: this mirror has been suspended by an admin and won't be usable.
```

It skips the pointless clone-readiness poll (a suspended mirror never becomes ready) and treats the re-create as a non-fatal success (exit 0). The interactive wizard surfaces the same state as a per-mirror `suspended` status.

Consumes the `suspended` field on the `CreatedMirror` response (server side: entirehq/entiredb#2328, now deployed).

## Spec note
No spec refresh in this PR. The vendored `CreatedMirror.suspended` was hand-authored ahead of the entiredb deploy, and I verified it is **byte-identical** to what the deployed core now emits — so it's already authoritative. A full `curl` refresh of the live spec is *not* clean right now: the deployed core doesn't yet expose the `grantMirrorCollaborator`/`revokeMirrorCollaborator` endpoints the CLI depends on (also hand-authored ahead of their own deploy), so a wholesale overwrite would drop them and break the build. Deferring the full refresh until that surface is live.

---

> [!NOTE]
> **Low Risk**
> CLI-only UX and API client typing around mirror create; no auth or data-path changes, with behavior covered by new tests.
> 
> **Overview**
> When **`entire repo mirror create`** hits an **existing placement that an admin suspended**, the CLI now reads the new **`CreatedMirror.suspended`** field from the control plane and treats it as a distinct outcome instead of waiting for a clone that will never finish.
> 
> **One-shot create** still prints the placement (mirror id and URL), then emits a **stderr warning** that the mirror is suspended and unusable, **skips** the cloning spinner and **`GetMirror`** readiness poll, and **exits successfully** (no `SilentError`). **The interactive wizard** maps the same response to per-mirror status **`suspended`** (non-fatal, not counted as a batch failure).
> 
> Generated **coreapi** types/spec were updated so **`suspended`** is required on **`CreatedMirror`**. Tests lock in no polling and the warning / wizard status behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 35bd3240bea2f334e34c48736a1daf628c879c6b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


toothbrush:
Superseded by #1602, which combines this with the collaborator-verb cleanup into a single rebased PR.

PR #1619, opened 2026-07-02
cli/auth: unhide `entire auth token` and add `--jurisdiction`

georg:
https://entire.io/gh/entireio/cli/trails/743

## What

Two related changes to `entire auth token`:

1. **Unhide the command** so it appears in `entire auth --help` and `entire agent-help`, surfacing the supported way to authenticate ad-hoc requests.
2. **Add `--jurisdiction <slug>`** (e.g. `us`, `eu`) to mint a *jurisdictional identity token* for that jurisdiction's entire-api cells, plus a `Jurisdiction:` line in `entire auth status` so the slug is discoverable.

```
# control-plane bearer (default)
curl -H "Authorization: Bearer $(entire auth token)" "https://us.console.entire.io/api/v1/clusters"

# jurisdictional cell token
curl -H "Authorization: Bearer $(entire auth token --jurisdiction us)" "https://aws-us-east-2.api.entire.io/api/v1/me/activity"
```

## Why

`entire auth token` was a hidden, scripting-only escape hatch that printed the **control-plane** bearer. Two gaps:

- It wasn't discoverable — users and agents had no surfaced way to authenticate ad-hoc `curl` against the core API.
- The per-jurisdiction data-plane **cells** (e.g. `https://aws-us-east-2.api.entire.io/api/v1`) *reject* the control-plane bearer; they require a jurisdictional identity token (RFC 8693 exchange, `scope=openid`, `aud=<jurisdiction host>`, per COR-666). There was no way to obtain that token.

## Details

**Unhide** (`05f3feb`, `58f798b`, `ccc62a4`)
- Remove `Hidden: true` from `newAuthTokenCmd`; reword the doc comment to drop the "not part of the everyday surface" framing while keeping the "treat the output as a secret" caveat. `--insecure-http-auth` on the command stays hidden. `RunE` is otherwise unchanged; `agent-help` renders live from the Cobra tree so it picks up the visibility change automatically.

**`--jurisdiction`** (`2b08ff1`)
- New exported `auth.JurisdictionToken`, built by extracting a shared subject-resolution helper out of `NewEntireAPICellClient` (whose behavior is unchanged). **No new exchange/URL/security logic** — it reuses the audited `exchangeJurisdictionToken` / `jurisdictionAudience` / `jurisdictionCoreURL` helpers and returns the raw token instead of a client (skipping the cell-URL round-trip a client would need).
- **`ENTIRE_TOKEN` is honored** as the exchange subject (exclusive, fail-closed), with the environment (prod/staging/loopback) derived from the token's own `aud` — so CI works with only `ENTIRE_TOKEN` set, no `ENTIRE_API_BASE_URL`. Only the `--jurisdiction` path gains this; `NewEntireAPICellClient` / `entire api --to cell` are unchanged.
- The slug is a bare lowercase jurisdiction (normalized: `US`/`" us "` -> `us`); the one token works against any cell in that jurisdiction.
- `entire auth status` now shows the caller's home jurisdiction so the slug is discoverable non-interactively.

## Testing

- Unit tests for `JurisdictionToken` (stored-context + `ENTIRE_TOKEN` subject paths, asserting `subject_token`, `scope=openid`, and the jurisdiction audience).
- Command-level tests for `--jurisdiction` (env-token success + not-logged-in hint) and the new `auth status` slug line.
- `mise run lint` clean; the `cmd/entire/cli` and `cmd/entire/cli/auth` packages pass.

> Note: `mise run test:ci` fails locally only in unrelated `cmd/entire/cli/strategy` git-hook tests, caused by a global `~/.git_template` leaking hooks into test repos — those tests are untouched by this change and fail the same way on the base branch.

https://claude.ai/code/session_014ByabUEH7Egwn2rMt8o6bB

### DIFF
--- a/cmd/entire/cli/auth.go
+++ b/cmd/entire/cli/auth.go
@@ -132,19 +132,17 @@ func newAuthCmd() *cobra.Command {
 //
 //	curl -H "Authorization: Bearer $(entire auth token)" "$CORE/api/v1/clusters"
 //
-// Hidden: it emits a live credential, so it's a deliberate scripting escape
-// hatch, not part of the everyday surface. It resolves the same bearer the API
-// client would — ENTIRE_TOKEN verbatim when set, otherwise the active context's
-// login JWT, refreshed if it's near expiry — and prints nothing but the token
-// (errors and the not-logged-in hint go to stderr) so command substitution
-// stays clean.
+// It emits a live credential, so treat the output as a secret. It resolves the
+// same bearer the API client would — ENTIRE_TOKEN verbatim when set, otherwise
+// the active context's login JWT, refreshed if it's near expiry — and prints
+// nothing but the token (errors and the not-logged-in hint go to stderr) so
+// command substitution stays clean.
 func newAuthTokenCmd() *cobra.Command {
 	var insecureHTTPAuth bool
 	cmd := &cobra.Command{
-		Use:    "token",
-		Short:  "Print the active control-plane bearer token (for scripting)",
-		Hidden: true,
-		Args:   cobra.NoArgs,
+		Use:   "token",
+		Short: "Print the active control-plane bearer token (for scripting)",
+		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			// Refresh may exchange/refresh over the network; honor the
 			// plain-HTTP opt-in before resolving so local dev cores work.
--- a/cmd/entire/cli/auth.go
+++ b/cmd/entire/cli/auth.go
@@ -139,14 +139,13 @@ func newAuthTokenCmd() *cobra.Command {
 		Use:   "token",
 		Short: "Print the active control-plane bearer token — a live credential, treat as a secret",
 		Long: "Print the active control-plane bearer token to stdout so scripts and\n" +
-			"ad-hoc curl can authenticate against the core API without re-deriving the\n" +
-			"keychain slot.\n\n" +
+			"ad-hoc curl can authenticate against the control-plane API.\n\n" +
 			"The output is a live credential — treat it as a secret. It is the same\n" +
 			"bearer the API client uses: ENTIRE_TOKEN verbatim when set, otherwise the\n" +
 			"active context's login JWT (refreshed if it's near expiry). Only the token\n" +
 			"is printed to stdout; errors and the not-logged-in hint go to stderr so\n" +
 			"command substitution stays clean.",
-		Example: "  curl -H \"Authorization: Bearer $(entire auth token)\" \"$CORE/api/v1/clusters\"",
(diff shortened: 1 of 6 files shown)


cmd/entire/cli/auth.go
```diff
@@ -132,19 +132,17 @@ func newAuthCmd() *cobra.Command {
...
 func newAuthTokenCmd() *cobra.Command {
 	var insecureHTTPAuth bool
 	cmd := &cobra.Command{
-		Use:    "token",
-		Short:  "Print the active control-plane bearer token (for scripting)",
-		Hidden: true,
-		Args:   cobra.NoArgs,
+		Use:   "token",
+		Short: "Print the active control-plane bearer token (for scripting)",
+		Args:  cobra.NoArgs,
```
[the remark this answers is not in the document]
georg: Good call — fixed in 58f798b. The "treat as a secret" caveat now lives in the user-facing help, not just the Go doc comment:

- `Short` now reads "Print the active control-plane bearer token — a live credential, treat as a secret", so it shows in the `entire auth --help` subcommand list.
- Added a `Long` that spells out the caveat and resolution behavior, plus an `Example` with the curl one-liner (visible in `entire auth token --help`).
- Trimmed the doc comment to point at those instead of duplicating the prose.

toothbrush (dismissed):
good idea, thanks Georg!

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/),
 and this project adheres to [Semantic Versioning](https://semver.org/).
 
+## [Unreleased]
+
+### Added
+
+- `entire auth token --jurisdiction <slug>` mints a jurisdictional identity token (scope=openid, aud=jurisdiction host) for that jurisdiction's entire-api cells (e.g. `https://aws-us-east-2.api.entire.io/api/v1`), which reject the control-plane bearer; it exchanges `ENTIRE_TOKEN` when set, else the active login. `entire auth status` now shows the caller's home jurisdiction so the slug is discoverable.
+
```
Soph: please remove, we are not doing this consistently, maybe we should but then we somehow need to make sure every PR has it.

Soph (commented):
as written in slack, I'd rather want the agent to use: 

```
entire api api/v1/clusters
entire api --to cell "/api/v1/me/activity?timezone=UTC&timeframe=last-3-months"
```

Instead of juggling the token and using curl, or? Like fine making `auth token` visible but rather updated the agent-help to point it to `entire api`. I'll make a PR for that.

CHANGELOG.md
Soph:
```suggestion
```

Soph (approved):

PR #1620, opened 2026-07-03
Remove checkpoint_version from checkpoint metadata

pfleidi:
https://entire.io/gh/entireio/cli/trails/744

## Why

The `checkpoint_version` stamp in each checkpoint's root `metadata.json` was a pure label: which storage backend serves a checkpoint is decided by the settings topology, and nothing ever dispatched on the stored value. The `EnsureCanReadVersion` read gates at resume/explain/rewind/export were forward-compat tripwires that could only fire for checkpoint formats that never shipped. Both are dead complexity — every reader had to thread a version through code that had exactly one possible answer.

Existing checkpoints' stored `checkpoint_version` values are handled by an external migration, so no backward-compat handling for the stored attribute is needed here.

## What changed

Removed the read gates first (they read the field), then the field itself: `WriteOptions.CheckpointVersion`, `CheckpointSummary.CheckpointVersion`, the stamping plumbing through the write chain, the read-side backfill (`normalizeCheckpointSummary`), and the now-unused `CheckpointVersionRefsV1` constant. Two follow-up cleanups: the hidden `entire checkpoint policy` help text now describes `checkpoint_version` as a write guard rather than a format selector, and the branch↔refs seam test regained a primary-read discriminator.

- **The checkpoint policy subsystem is intentionally untouched.** Its `checkpoint_version` field remains a checkpoint-data write guard (`CanSatisfyPolicy`): a repo can still block writes from CLIs that don't support the configured version. It just no longer feeds a stamp into new summaries. `attach` switched from the deleted `checkpointVersionForNewCheckpoint` to the shared `ensureCheckpointPolicyAllowsCheckpointData` gate that explain and import already use.
- **`CheckpointVersionBranchV1` stays** — it anchors the policy defaults (`DefaultPolicy`, `Normalize`); only the refs-v1 constant became unused.
- **The seam test needed a new probe.** Dropping the version assert removed the only signal distinguishing a read served by the git-refs primary from one served by the git-branch mirror. The test now deletes the mirror's v1 branch and asserts the composed store still serves full reads — something only the primary can do.

## Reviewer notes

- Existing checkpoints and existing policy refs are left as-is; the stored-field cleanup ships as an external migration.
- Commits are ordered to be reviewable independently (gates → field → help text → seam test) and can be split into a stack if preferred.

---

> [!NOTE]
> **Medium Risk**
> Touches resume/rewind/explain/export and checkpoint persistence; behavior change is intentional but broad across user-facing flows. Stored-field cleanup for old checkpoints is assumed to happen outside this PR.
> 
> **Overview**
> Removes **`checkpoint_version`** from checkpoint root **`metadata.json`** and from the write path: **`WriteOptions`**, **`CheckpointSummary`**, stamping in branch/refs stores, and **`normalizeCheckpointSummary`** backfill on read. Drops **`CheckpointVersionRefsV1`** and **`checkpointpolicy.EnsureCanReadVersion`** / **`IsUnsupportedVersion`**, so **resume**, **rewind**, **explain**, and **export** no longer fail on per-checkpoint format labels.
> 
> **`entire session attach`** now uses **`ensureCheckpointPolicyAllowsCheckpointData`** instead of **`checkpointVersionForNewCheckpoint`** (policy still gates writes via **`CanSatisfyPolicy`**, but no longer chooses a version to stamp). Manual-commit condensation stops passing **`CheckpointVersion`** into writes.
> 
> Docs and **`entire checkpoint policy`** help describe **`checkpoint_version`** as a **write guard**, not a metadata format selector. The branch↔refs seam test deletes the v1 mirror branch and asserts reads still work via the git-refs primary.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dca230047ab5030b932f829cb67cda4491b80be5. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pfleidi:
Bugbot run

Soph (approved):

pjbgf (approved):

PR #1629, opened 2026-07-04
Emit ULID checkpoint IDs under the git-refs store

Soph:
https://entire.io/gh/entireio/cli/trails/753

New checkpoints written under the **git-refs** primary now get a 26-char **ULID** instead of a 12-hex id. This makes a checkpoint's id encode its storage backend — **ULID ⟹ ref**, legacy hex ⟹ branch — which is the missing piece for id-kind-aware read routing (a follow-up): reads can decide where a checkpoint lives from the id alone, no config, no guessing.

## What changed
- **`id.GenerateULID()`** — timestamp + crypto-random ULID via `oklog/ulid` (already a dependency, previously import-for-parsing-only). Canonical, passes `KindOf`/`Validate` as `KindULID`. `id.Generate()` (12-hex) is unchanged.
- **`checkpoint.GenerateCheckpointID(ctx)`** — the single place the format is decided: ULID when the primary is git-refs, else 12-hex. Fail-soft to hex on a missing/malformed checkpoints config. Layering stays clean (`checkpoint` already imports `id` + `settings`; `settings` doesn't import `checkpoint`).
- **Routed the checkpoint-id generation sites** through it: `attach`, both manual-commit hook paths, both condensation paths. **Turn ids and the investigate run id stay hex** — they're format-agnostic correlation tokens, not stored checkpoints.
- **`explain` guard fix** — its auto-ambiguity guard assumed a fixed 12-char id width; it now uses `id.MaxIDLength` (26) so a ULID target isn't wrongly treated as un-prefixable.

## Why no store changes
Read/write already handled both formats: `ShardFor` shards ULIDs on the last two chars, and `RefName`/`ParseRef`/`Validate` accept them. Only *emission* was missing (the deferred "Layer B" from the git-refs store PR).

## Testing
- `id.GenerateULID` unit test (valid, KindULID, 26 chars, unique); `GenerateCheckpointID` test (git-refs → ULID, default/git-branch → hex).
- Build clean, lint 0, checkpoint + strategy + cli suites pass, integration **390**.
- **git-refs canary passes with ULID emission** (58/58 +1 skip) — proves ULID checkpoints work end-to-end (minting → last-2-char shard ref names → reads → e2e assertions). git-branch canary unchanged (hex, 59/59).

## Follow-up
The id-kind read routing (git-refs primary reading old branch-only checkpoints, and vice-versa) builds on this — now that ULID ⟹ ref holds.

---

> [!NOTE]
> **Medium Risk**
> Changes how new checkpoint IDs are created across attach, hooks, and condensation; mis-routing format vs primary could affect metadata writes, though tests and fail-soft hex fallback limit blast radius.
> 
> **Overview**
> New committed checkpoints now get a **26-character ULID** when the checkpoints primary is **git-refs**, and keep the legacy **12-hex** ID for **git-branch** (and the default). The goal is to tie ID shape to storage (**ULID → ref**, hex → branch) so later read routing can pick the backend from the ID alone.
> 
> **`checkpoint.GenerateCheckpointID(ctx)`** is the single dispatcher: it reads checkpoints config and calls **`id.GenerateULID()`** or **`id.Generate()`**, failing soft to hex if config is missing or bad. **Attach**, **manual-commit** hook paths, and **condensation** (doctor / eager stop) now call this instead of **`id.Generate()`** directly; **`resolveCheckpointID`** takes **`context`** for that.
> 
> **`explain --generate`** ambiguity handling no longer assumes a 12-character checkpoint width; it uses **`id.MaxIDLength` (26)** so ULID targets are not skipped incorrectly.
> 
> Tests cover ULID generation and config-driven **`GenerateCheckpointID`** behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fbc294a037d783480d6afe2cecdc7a34d6909a6b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/attach.go
+++ b/cmd/entire/cli/attach.go
@@ -22,6 +22,7 @@ import (
 	"github.com/entireio/cli/cmd/entire/cli/logging"
 	cliReview "github.com/entireio/cli/cmd/entire/cli/review"
 	"github.com/entireio/cli/cmd/entire/cli/session"
+	"github.com/entireio/cli/cmd/entire/cli/settings"
 	"github.com/entireio/cli/cmd/entire/cli/strategy"
 	"github.com/entireio/cli/cmd/entire/cli/trailers"
 	"github.com/entireio/cli/cmd/entire/cli/validation"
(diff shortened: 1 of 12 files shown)


Soph:
@cursor review

nodo (approved):

PR #1313, opened 2026-06-01
Add Pi review-runner adapter

dipree:
https://entire.io/gh/entireio/cli/trails/467

## Summary

Adds **Pi** as a first-class `entire review` worker by implementing its `AgentReviewer` adapter. This is the concrete proof that the previous "only claude-code/codex/gemini" limit was just missing adapters, not a launch-capability wall.

Stacked on #1312 (review profiles). Review/merge that first.

## What changed

- **`cmd/entire/cli/agent/pi/reviewer.go`** — `NewReviewer()` returning a `ReviewerTemplate`:
  - `BuildCmd`: `pi --mode json --print [--model <model>] <prompt>` with `ENTIRE_REVIEW_*` env via `AppendReviewEnv`
  - `Parser`: maps Pi's `--mode json` `AgentSessionEvent` stream (`message_update` text deltas, `tool_execution_start`, `message_end`/`turn_end` usage, `agent_end`) into Entire's review `Event` stream. Falls back to `message_end` content text when no deltas streamed; uses `stopReason` for success.
- **`cmd/entire/cli/agent/pi/generate.go`** — `GenerateText` so Pi can also serve as a master / summary provider.
- **Wiring**: `launchableReviewerFor` returns `pi.NewReviewer()`; Pi added to summary provider binaries, default-profile agents, master-preference order, and config/help text.
- **Session matching**: Pi runs are matched to their lifecycle sessions by model too (same-agent/different-model workers resolve correctly); `agent_end` backfills the model via `ExtractModel`.

## Testing

- `reviewer_test.go`: name, argv/env shape with `--model`, JSON event-stream parsing, message-end fallback text
- manifest test for same-agent/different-model disambiguation
- `go vet ./cmd/entire/cli/agent/pi/...`
- `env -u PI_CODING_AGENT go test ./cmd/entire/cli/...`
- targeted review integration tests

---

> [!NOTE]
> **Medium Risk**
> Touches review orchestration, manifest session matching, and summary provider selection; behavior is well covered by unit tests but depends on Pi CLI JSON event stability.
> 
> **Overview**
> **Pi** is wired into `entire review` as a first-class worker alongside Claude Code, Codex, and Gemini, using the shared `ReviewerTemplate` pattern instead of marker fallback.
> 
[...]

PR #1651, opened 2026-07-06
feat(review): fan multi-skill workers out into parallel single-skill runs

peyton-alt:
https://entire.io/gh/entireio/cli/trails/772

Stacked on #1647 — review only the top commit; retarget to main after #1647 merges.

## Problem

A worker configured with N skills joined them into one child's prompt: the skills ran sequentially (or blended into one muddled pass), so selecting more skills grew the wait linearly. Measured live: a two-skill claude worker took ~9 minutes as a single child. The parallel machinery already existed one level up (worker slots fan out in `run_multi`); multi-skill selection just never used it.

## Changes

- **`explodeSkillWorkers`**: at plan time, each multi-skill worker becomes one worker per skill (keys like `claude-code:review`, deduped against existing worker names; agent/model/prompt preserved, explicit `Agent` set). Wait becomes max(skills) instead of sum(skills). Two exploded workers also engage the judge, extending crew+judge consolidation to single-agent multi-skill profiles.
- **`--agent` selects a filtered crew**: all of that agent's workers run (previously: ambiguity error). Single-agent path only when exactly one worker matches.
- **Session matching by skills**: same-agent same-model exploded workers defeat the agent+model matcher — assignments could cross, attributing tokens/transcripts to the wrong skill. `AgentRun` gains `Skills` (propagated through planned runs and both run paths), and the matcher requires skill-set agreement when both sides carry skills (`session.State.ReviewSkills` is recorded from `ENTIRE_REVIEW_SKILLS` by the lifecycle hook). Mirrors #1313's different-model disambiguation, with an adversarially-ordered test that fails without the discriminator.

## Testing

- TDD throughout (explosion, override semantics, cross-assignment, propagation — each watched fail first)
- `mise run test` (7,590), `test:integration` (390), canary — green
- Shim smoke: a two-skill profile spawns two one-skill children in parallel (~3s wall for both) plus one judge invocation

---

> [!NOTE]
> **Medium Risk**
> Touches review orchestration, `--agent` semantics, and manifest/session attribution; behavior is well covered by tests but wrong matching would mis-attribute findings.
> 
> **Overview**
> **Multi-skill workers now run in parallel** instead of one child that runs every skill together. At plan time, `explodeSkillWorkers` splits each worker with more than one skill into separate workers (keys like `worker:skill-slug`), so wall time tracks the slowest skill, not the sum. Model, prompt, and agent binding are preserved; keys are deduped so existing workers are not overwritten.
> 
> **`--agent` behavior changes** after explosion: `applyAgentOverride` / `selectProfileWorkers` return every worker for that agent. One match still uses the single-agent path; several exploded workers narrow the profile and run through the normal multi-reviewer flow (including judge when applicable), instead of failing on ambiguity.
> 
> **Manifest and run summaries carry skills** on `AgentRun` and planned runs (`ReviewerSkills` on configured reviewers). Session linking compares `run.Skills` to `session.State.ReviewSkills` when both sides have data, so same-agent, same-model parallel skill runs do not swap transcripts or tokens.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4c6f5847a9771c1b00db40906b31ab6af2b2a2cd. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


peyton-alt:
Note: this merged into `feat/review-child-context` (its base) via fast-forward — the combined stack now reviews as one PR: #1647.

PR #1664, opened 2026-07-07
feat(review): drop the default reviewer timeout — reviewers run until done

peyton-alt:
https://entire.io/gh/entireio/cli/trails/784

## Problem

Reviewers were killed at a wall-clock default (10m, later raised to 20m) that conflates "slow because working" with "hung". Observed live: a reviewer at ~90% completion killed with everything discarded. Plain agent invocation has no such cliff — the user is the liveness check.

Two findings shaped the fix:

- **No inactivity watchdog can replace the wall clock.** Legitimate review subagents run 12+ minutes while the reviewer's parent process emits zero output (measured on a real run), so a headless child offers no liveness signal that distinguishes a quiet working subagent from a hang. Any watchdog on available signals re-creates the kill-real-work bug.
- **The duration is the task, not a hang risk.** A controlled A/B on the same branch with the same two skills: plain `claude` session 9m57s vs `entire review` 9m08s. Reviews take as long as they take, wherever they run.

## Change

- `--timeout` default: `20m` → `0` (none). Reviewers run until they finish, exactly like a directly-invoked skill. The flag remains as an explicit opt-in hard cap for CI/cost-bounded runs.
- The judge keeps its own 5m default regardless — it's a single text-generation call; a new `judgeTimeoutArg` mapping prevents the reviewer's no-cap sentinel leaking through as "judge unbounded".
- A stuck reviewer in an interactive run is Ctrl+C (process-group kill already handles it, #1625-era fix); unattended liveness is the background-mode/durable-run design's job, not a kill timer's.

## Testing

- `TestReviewerTimeout_NoDefaultCap`, `TestJudgeTimeoutArg`, and the flag-resolution test updated to pin the new contract
- `mise run test` (7,604) green, lint clean

---

> [!NOTE]
> **Medium Risk**
> Default behavior change means long or stuck reviewer processes are no longer auto-cancelled unless users pass `--timeout`; multi-agent reviews may run much longer, though the judge consolidation step remains bounded.
> 
> **Overview**
> **`entire review` no longer applies a wall-clock cap to reviewers by default** (previously ~20m). Reviewers run until they finish, matching a skill invoked in a normal session; **`--timeout` stays an opt-in hard cap** for CI or cost limits, with updated flag help (default `0`).
> 
> **`reviewerTimeout`** drops the implicit default: only a positive `ReviewerTimeout` enables `context.WithTimeout`; unset, zero, or the no-cap sentinel means no reviewer deadline.
> 
> **The consolidating judge is handled separately** via **`judgeTimeoutArg`**: an explicit positive `--timeout` still bounds the judge; when reviewers are uncapped, the reviewer sentinel must not disable the judge—**the judge keeps its existing ~5m synthesis default**.
> 
> Tests pin the no-default-cap contract, judge mapping, and flag → resolver behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fbe6fe6090561768c2cc28b0f214cee29ec898ae. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/review/cmd.go
+++ b/cmd/entire/cli/review/cmd.go
@@ -250,7 +250,7 @@ To tag an already-finished session as a review, use
 	cmd.Flags().StringVar(&profileOverride, "profile", "", "review profile to run (default: review_default_profile or general)")
 	cmd.Flags().StringVar(&perRunPrompt, "prompt", "", "one-off instructions appended to this review run")
 	cmd.Flags().StringVar(&baseOverride, "base", "", "git ref to scope the review against (default: origin/HEAD → origin/main → origin/master → main → master)")
-	cmd.Flags().DurationVar(&reviewTimeout, "timeout", defaultReviewerTimeout, "max time each reviewer may run before it is cancelled and marked failed; also bounds the consolidating judge, whose timeout or error fails the review (0 disables both)")
+	cmd.Flags().DurationVar(&reviewTimeout, "timeout", 0, "optional hard cap per reviewer (default: none — reviewers run until they finish, like a skill invoked directly in a session). When set, it also bounds the consolidating judge; unset, the judge keeps its own 5m default")
 	// The listing modes and the action modes each select a distinct command
 	// behavior; combining them silently runs one and drops the rest, so reject
 	// the combination up front with a clear cobra error.
@@ -722,6 +722,19 @@ func resolveReviewerTimeoutArg(flagValue time.Duration) time.Duration {
 		return -1
 	}
 	return flagValue
+}
+
+// judgeTimeoutArg maps the resolved reviewer timeout to the judge's
+// ProviderTimeout three-state. The judge is a single text-generation call
+// with no event stream, so unlike reviewers it always keeps a bound: an
+// explicit --timeout governs it, otherwise the synthesis default (5m)
+// applies. The reviewer no-cap sentinel (negative) must not leak through —
+// it would disable the judge bound entirely.
+func judgeTimeoutArg(reviewerArg time.Duration) time.Duration {
+	if reviewerArg > 0 {
+		return reviewerArg
+	}
+	return 0
 }
 
 // runReview executes the main review flow.
@@ -1231,7 +1244,7 @@ func runMultiAgentPath(
 		profileName:       profileName,
 		task:              profile.Task,
 		masterName:        masterLabel,
-		judgeTimeout:      timeout,
+		judgeTimeout:      judgeTimeoutArg(timeout),
 		onSynthesisResult: func(result string) {
 			aggregateOutput = result
 		},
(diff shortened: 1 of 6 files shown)


Soph (dismissed):

cmd/entire/cli/review/cmd.go
```diff
@@ -1231,7 +1244,7 @@ func runMultiAgentPath(
 		profileName:       profileName,
 		task:              profile.Task,
 		masterName:        masterLabel,
-		judgeTimeout:      timeout,
+		judgeTimeout:      judgeTimeoutArg(timeout),
```
pfleidi, line 1233: I'm not sure I understand the value of this function: It seems to only ensure that the judge timeout is a positive number. Couldn't the same be accomplished by something like `math.Max(timeout, 0)`? Can this helper be removed?

cmd/entire/cli/review/run.go
```diff
@@ -34,29 +34,23 @@ func reviewerModelName(r reviewtypes.AgentReviewer) string {
...
+// reliable liveness signal exists for a headless child that would let a
+// watchdog distinguish "working via a quiet subagent" from "hung". A stuck
+// reviewer is Ctrl+C in interactive runs (process-group kill handles it);
+// unattended callers that need a bound pass --timeout explicitly.
+//   - positive: hard cap.
+//   - zero or negative: no cap.
 func reviewerTimeout(cfg reviewtypes.RunConfig) time.Duration {
-	switch {
-	case cfg.ReviewerTimeout > 0:
+	if cfg.ReviewerTimeout > 0 {
```
pfleidi: This looks like the same logic as for the judge timeout. Can this be a `math.Max()` call as well?

cmd/entire/cli/review/cmd.go
peyton-alt: Half of it turned out to be removable entirely: chasing this showed the `-1` sentinel from `resolveReviewerTimeoutArg` became vestigial once the default was dropped (every remaining reader just checks `> 0`), so that whole helper is deleted in 1be574238 and the flag flows through unmapped. `judgeTimeoutArg` I kept as a named function — the point is the contract (the judge must never be uncapped regardless of the reviewer setting) rather than the arithmetic — but its body is now `max(reviewerArg, 0)` as you suggested. (Builtin `max`; `math.Max` is float64-only.)

cmd/entire/cli/review/run.go
peyton-alt: Yes — `max(cfg.ReviewerTimeout, 0)` in 6232e22f5. Kept the named function since its doc comment carries the no-default-timeout rationale, but the body is exactly your suggestion.

Soph (approved):

PR #1665, opened 2026-07-07
repo mirror list: add --repo filter and --sort

gtrrz-victor:
https://entire.io/gh/entireio/cli/trails/786

## What

Makes `entire repo mirror list` easier to scan as mirror counts grow.

- **`--repo <substr>`** — client-side, case-insensitive substring filter on the
  repo name. The control plane already filters by `--owner`/`--provider`/`--cluster`
  server-side but not by repo name, so this fills that gap. Applies to the table,
  `--json`, and `--show-available`.
- **`--sort <column>`** — sort by a displayed column (header name, case-insensitive);
  a leading `-` sorts descending (e.g. `--sort -repo`).
- **Deterministic default order** — owner/repo ascending, breaking duplicate-repo
  ties (same repo mirrored on several clusters) by cluster host, instead of
  arbitrary server order.

Columns are unchanged (`REPO`, `CLONE URL`, `PRIVATE`).

## Why

The list printed in arbitrary order with no way to narrow it. Server-side repo
filtering would require an entiredb change; `--repo` is the in-repo win. Shared
`sortRows`/`filterByRepo` helpers keep it reusable across control-plane lists.

## Test plan

- Unit + command-level tests: `--repo` filtering, `--sort` (asc/desc, unknown
  column errors), default owner/repo ordering with cluster tiebreak, `--json`
  and `--show-available` paths.
- `mise run fmt` clean; golangci-lint (v2) 0 issues; full `cmd/entire/cli` package passes.

---

> [!NOTE]
> **Low Risk**
> Client-side list presentation only; no API or auth changes. Filtering/sorting runs after fetch and is covered by new tests.
> 
> **Overview**
> Adds **`--repo`** and **`--sort`** to `entire repo mirror list` so long mirror lists are easier to narrow and read.
> 
> **`--repo`** applies a case-insensitive substring filter on the repo name after the API returns (owner/provider/cluster stay server-side). It affects the table, **`--json`**, and **`--show-available`**, and empty matches still encode as `[]`.
> 
> **`--sort`** orders by a displayed column header (optional `-` for descending). With no **`--sort`**, existing mirrors use a new default: owner/repo ascending, then cluster host when the same repo appears on multiple clusters.
> 
> Shared helpers **`sortRows`**, **`filterByRepo`**, and **`sortMirrorsDefault`** in `corecmd` / `repo_mirror` back the behavior, with unit and command-level tests (including **`execMirrorList`** so **`--json`** is exercised with the parent control-plane flags).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 08812170360dbcb5a326b523391bdbfd8013322b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


gtrrz-victor:
**Follow-up: addressed sort review feedback**

Replaced the generic `sortRows[T]` / `sortMirrorsDefault` split with concrete sorters. Fixes two ordering bugs the review flagged:

- **Whitespace parity** — `desc` was read from the raw spec but the column name from the trimmed spec, and the two list paths passed raw vs trimmed. Now a shared `parseSortColumn` trims before reading direction, so `--sort " -repo"` behaves the same on both the mirror and `--show-available` lists.
- **Deterministic ties on every column** — only default / `--sort repo` had the `(owner/repo, clusterHost)` tiebreak; other columns fell back to a single-key sort with arbitrary server order. `sortMirrors`/`sortAvailable` now always apply the owner/repo(+cluster) tiebreak.

Also de-generalized per review — two concrete 3-column sorters instead of `sortRows[T any]`. Added unit tests (`TestSortMirrors`, `TestSortAvailable`) plus command-level coverage for `--sort private` tiebreak and `--sort " -repo"`.

_(Supersedes the body's "Shared `sortRows` helpers" note — `sortRows`/`sortMirrorsDefault` are gone.)_

gtrrz-victor:
**Follow-up: fail fast on bad `--sort`**

- `--sort` now validated in `PreRunE` — an unknown column errors before any control-plane fetch, instead of after walking every page.
- Corrected `--repo` help + `mirrorColumns` doc: it filters the repo name only; owner/provider/cluster stay server-side filters.
- Verified: mirror-list test suite green; golangci-lint v2 clean.

Soph (approved):

Issue #1717, opened 2026-07-10
entire enable in a non-repository directory creates and pushes a GitHub repository without confirmation


andrewluxem:
## Version

Entire CLI 0.8.42 on darwin/amd64.

## Summary

Running `entire enable` from a directory that was not already a Git repository caused Entire to:

1. Run `git init`
2. Stage the directory with behavior equivalent to `git add -A`
3. Create an initial commit
4. Create a new private GitHub repository named after the directory
5. Push the repository

No confirmation was requested before initializing Git, creating the GitHub repository, or pushing its contents.

## Environment

The command was accidentally run from a workspace parent directory containing:

- 82 normal Git repositories
- 17 linked Git worktrees
- 20 true loose top-level items
- Dependency-cache content
- Audit notes, drafts, QA files, reports, and other working material

## Observed impact

The newly created private repository contained approximately 1,084 entries:

```text
99 gitlinks
985 ordinary blobs
```

The 99 gitlinks reconciled exactly to:

```text
82 normal repositories
17 linked worktrees
```

The nested repository and worktree contents were not copied. They were represented as commit-SHA gitlink pointers.

The linked worktrees' uncommitted changes were also not captured because a gitlink references the checked-out commit, not working-tree modifications.

The 985 ordinary blobs came from the actual loose-file and cache content:

```text
772 dependency-cache blobs
213 real working files
```

The real loose files included material such as:

* Audit notes
* Drafts
* QA records
* Issue-import files
* Operational summaries
* Deployment and account notes

The repository was created and pushed within seconds, before the directory mistake was noticed.

A pattern-based review found no apparent real credential among the tested credential formats. The repository was private and was deleted after the incident was contained.

## Expected behavior

When `entire enable` is run outside an existing Git repository, Entire should stop and clearly explain that the current directory is not a repository.

If repository initialization is intentionally supported, it should require separate explicit confirmation before each externally visible or destructive step:

1. Running `git init`
2. Staging the directory
3. Creating a GitHub repository
4. Pushing any commit

## Suggested safeguards

At minimum:

* Require explicit confirmation before initializing Git in a non-repository directory.
* Require a second explicit confirmation before creating or pushing a remote repository.
* Display the absolute directory path before either action.
* Display a staged-file count and size summary before committing or pushing.
* Hard-stop or show a high-severity warning when the directory contains nested `.git` directories or `.git` worktree pointer files.

[...]

PR #1720, opened 2026-07-10
fix(enable): confirm before init/create/push in a non-repo directory

computermode:
https://entire.io/gh/entireio/cli/trails/828

## What

`entire enable` in a directory that isn't a git repo offers to `git init`, create a GitHub repo, and push. The interactive confirms defaulted to **Yes**, so reflexively pressing Enter could blow through the whole cascade and publish the directory's contents.

Addresses https://github.com/entireio/cli/issues/1717.

This change makes each outward-facing step an independent, **default-No** confirmation:

| Step | Prompt (interactive, default No) | If declined |
|------|----------------------------------|-------------|
| Init | `Warning: Not a git repository. Initialize a new one in <path>?` | Nothing happens — no `git init` |
| Create remote | `Create a GitHub repository for <path>?` | Local repo only; no remote |
| Push | `Push the initial commit to <owner/repo>?` | Repo created + origin set, **nothing published**; prints manual `git push` step |

- Opting into one step no longer implies the next.
- Prompts show the **absolute path** (visible in interactive and accessible modes) so a wrong-directory mistake is obvious.

## Unchanged

Non-interactive behavior is preserved: without `-y`/`--init-repo` it declines; `--yes`/`--repo-*` flags still bypass the prompts (documented happy path).

## Tests

Default-No guards for all three prompts; a finalize test that the push is skipped (repo still created) when the push decision is No. Shared `withInteractivePromptStdin` test helper, with the existing prompt test migrated onto it.

---

> [!NOTE]
> **Medium Risk**
> Changes default interactive behavior for repo init and remote publishing (safety fix); scripted flows without flags now require explicit confirmation unless `--yes` or repo flags are used.
> 
> **Overview**
> Interactive `entire enable` bootstrap in a non-git folder no longer treats Enter as consent for **git init**, **GitHub repo creation**, or **push**. Each step is a separate confirm that **defaults to No**, with titles that show the **absolute directory path** (and a **Warning:** on init) so accidental runs in the wrong folder are obvious.
> 
> **Push** is decoupled from create: `bootstrapState` gains `push`, init collects it via `confirmPushToRemote`, and finalize only runs `git push` when the user opted in—otherwise it still runs `gh repo create` / sets `origin` and prints manual `git push` guidance.
> 
> `--yes` and explicit `--repo-*` flags still skip these prompts (unchanged non-interactive happy path). Tests add `withInteractivePromptStdin`, default-No coverage for all three prompts, and a finalize test when `push` is false.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fd63242180cca49d8d45ddc18f7c29a491c85818. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/setup_github.go
+++ b/cmd/entire/cli/setup_github.go
@@ -345,7 +345,7 @@ func confirmCreateGitHubRepo(cwd string) (bool, error) {
 	form := NewAccessibleForm(
 		huh.NewGroup(
 			huh.NewConfirm().
-				Title(fmt.Sprintf("Create a GitHub repository for %s?", cwd)).
+				Title(fmt.Sprintf("Create a GitHub repository for %q?", cwd)).
 				Value(&confirmed),
 		),
 	)
@@ -371,7 +371,7 @@ func confirmPushToRemote(fullName string) (bool, error) {
 	form := NewAccessibleForm(
 		huh.NewGroup(
 			huh.NewConfirm().
-				Title(fmt.Sprintf("Push the initial commit to %s?", fullName)).
+				Title(fmt.Sprintf("Push the initial commit to %q?", fullName)).
 				Value(&confirmed),
 		),
 	)
@@ -409,7 +409,7 @@ func confirmInitRepo(_ io.Writer, cwd string, opts GitHubBootstrapOptions) (bool
 	form := NewAccessibleForm(
 		huh.NewGroup(
 			huh.NewConfirm().
-				Title(fmt.Sprintf("Warning: Not a git repository. Initialize a new one in %s?", cwd)).
+				Title(fmt.Sprintf("Warning: Not a git repository. Initialize a new one in %q?", cwd)).
 				Value(&confirmed),
 		),
 	)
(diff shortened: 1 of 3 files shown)


cmd/entire/cli/setup_github.go
```diff
@@ -219,6 +220,21 @@ func runGitHubBootstrapInitWith(ctx context.Context, w, errW io.Writer, opts Git
 		}
 	}
 
+	// Step 6: separate guard for the push. Pushing publishes the directory's
+	// contents to the remote — a distinct outward-facing action from creating
+	// the repo — so confirm it on its own and default to No. Only relevant
+	// when we'll create a GitHub repo and have a commit to push. Gated like
+	// the create confirm: explicit flags / --yes / non-interactive keep the
+	// documented happy path (push).
+	push := useGitHub && commit
```
pfleidi: I wonder if the safer path here would be defaulting to not push unless otherwise specified

alishakawaguchi (approved):

PR #1734, opened 2026-07-13
fix(auth): auth token --jurisdiction should follow the active context

jagregory:
## The bug

With multiple login contexts, `entire auth token --jurisdiction <slug>` ignores the **selected** context and always mints against `entire.io`.

`entire auth token` (no flag) resolves its bearer from the **active context**, so a selected partial.to context prints a partial.to token. But `--jurisdiction` goes through `JurisdictionToken` → `resolveCellSubject` → `resolveStoredCellSubject`, which resolves *which login to use* by **discovery against the data host** (`api.BaseURL()`, default `entire.io`). `selectContext` only lets the active context win if it is eligible for that host — a partial.to context isn't eligible for entire.io — so it falls back to the sole entire.io-eligible context. The whole exchange (subject, core, audience) then goes to entire.io regardless of what's selected.

## The change

`--jurisdiction`'s stored path now resolves its subject from the **active login context** (new `resolveActiveContextCellSubject`), consistent with plain `auth token`: the active context's refreshed login JWT is the exchange subject, and its own core drives the environment family and exchange target. So a selected partial.to context mints a partial.to jurisdiction token.

Scope is limited to this one scripting command:
- Only `JurisdictionToken`'s non-`ENTIRE_TOKEN` branch changes.
- `NewEntireAPICellClient` / the cell-client factory still use `resolveStoredCellSubject` (data-host discovery), which is correct — they dial the data plane.
- The `ENTIRE_TOKEN` path is unchanged (the token's own aud already drives its environment).

## Before / after (two contexts, partial.to selected)

- before: `--jurisdiction us` → `aud=https://us.entire.io`, exchange at `https://us.auth.entire.io`
- after: `aud=https://us.partial.to`, exchange at `https://us.auth.partial.to`

## Tests

- `TestJurisdictionToken_StoredContext` rewritten to seed an active context (no data-host discovery stub); asserts the mint follows that context's environment with no `ENTIRE_API_BASE_URL` set.
- `TestJurisdictionToken_StoredContextFollowsActiveContext`: two contexts (prod entire.io + staging partial.to), partial.to active → partial.to audience/subject/exchange-core.

### DIFF
--- a/cmd/entire/cli/auth/cell_data_api.go
+++ b/cmd/entire/cli/auth/cell_data_api.go
@@ -262,16 +262,67 @@ type cellSubject struct {
 	httpClient     *http.Client
 }
 
-// resolveCellSubject picks the jurisdiction-exchange subject: ENTIRE_TOKEN when
-// set (exclusive, fail-closed), otherwise the active stored login context. This
-// is the ENTIRE_TOKEN-aware dispatcher used by JurisdictionToken;
-// NewEntireAPICellClient calls resolveStoredCellSubject directly so its behavior
-// is unchanged.
+// resolveCellSubject picks the jurisdiction-exchange subject for
+// JurisdictionToken (the `entire auth token --jurisdiction` scripting helper):
+// ENTIRE_TOKEN when set (exclusive, fail-closed), otherwise the ACTIVE stored
+// login context.
+//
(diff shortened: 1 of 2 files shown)


toothbrush (dismissed):
Seems reasonable! :+1:

computermode (approved):

Issue #1505, opened 2026-06-23
Windows binaries are unsigned — blocked by Smart App Control / SmartScreen


genkey6:
## What happened?

On Windows 11 with **Smart App Control (SAC)** in Enforce mode, the unsigned `entire.exe` (~48 MB Go binary) is killed on launch with:

```
An Application Control policy has blocked this file
... did not meet the Enterprise signing level requirements
```

The blocking policy GUID is `{0283ac0f-fff1-49ae-ada1-8a933130cad6}` — the built-in Smart App Control policy. Because Git hooks spawn a fresh `entire` process on every relevant action, **every new entire process fails** once SAC enforces. (A long-running background process started before enforcement can keep working from memory until the next reboot, which is why it looks like "it worked yesterday, broken today.")

### Root cause

`entire.exe` is shipped **unsigned**. The release pipeline only signs macOS binaries:

- `.goreleaser.yaml` / `.github/workflows/release.yml` apply Developer ID signing + notarization for **darwin** builds only. There is no Authenticode signing step for the **windows** builds.
- GitHub Releases, the entire.io installer, and the Scoop bucket all distribute the **same unsigned exe**, so reinstalling / switching channels does not help.

Smart App Control only allows binaries that are **signed** or that have an **established Microsoft cloud reputation**. A new, unsigned, low-reputation Go binary evaluates to zero reputation and is blocked. Unlike AppLocker / WDAC, **SAC has no per-file allowlist** — the only way to run an unsigned binary is to turn SAC off entirely, which is **irreversible** (SAC cannot be re-enabled without reinstalling Windows) and disables OS-wide malware protection. So "allow just this exe" is not possible for affected users.

SAC auto-promotes itself from evaluation mode (`state=2`) to enforce mode (`state=1`) on its own ML-driven schedule with no user action and little notice, so this can hit any new Win11 machine at an arbitrary time.

## Steps to reproduce

1. Fresh Windows 11 install with Smart App Control left at the default (evaluation mode, which later auto-promotes to Enforce).
2. Install entire (Scoop / entire.io installer / GitHub Releases download).
3. Once SAC is in Enforce mode, run any command or trigger a Git hook that spawns `entire.exe`.
4. The process is blocked: `An Application Control policy has blocked this file`.

## Expected behavior

The official Windows `entire.exe` should run on a default Windows 11 machine without requiring the user to disable Smart App Control.

## Suggested fix

[...]

PR #1735, opened 2026-07-13
Fix Windows Cursor hooks + Scoop app-dir naming (#1424)

### LINKED ISSUE #1424: Cursor hooks not firing on Windows 11 (Scoop install)
### What happened?

         I'm on Windows 11 using Cursor and installed Entire CLI via Scoop. The package installed as cli instead of entire, which caused --absolute-git-hook-path to fail with "failed to resolve symlinks for C:\Users\admin\scoop\apps\cli\current\entire.exe: The system cannot find the path specified."

        I manually fixed the git hooks (post-commit, prepare-commit-msg) and .cursor/hooks.json to use the full absolute path /c/Users/admin/scoop/apps/cli/current/entire.exe. Everything looks correctly configured — entire status shows enabled, git hooks exist, hooks.json has correct entries with no BOM. However entire session list and entire checkpoint list always show 0 after Cursor agent sessions and commits. The Cursor agent hooks (sessionStart, stop, etc.) do not appear to be firing at all. Root cause seems to be Scoop naming the package cli instead of entire, breaking Entire's internal symlink resolution on Windows.

## Environment
- **OS:** Windows 11
- **Shell:** Git Bash (MINGW64)
- **Agent:** Cursor
- **Install method:** Scoop
- **Entire version:** 0.7.5

## Problem
Cursor agent hooks are not firing. `entire session list` and `entire checkpoint list` 
always show 0 after Cursor agent sessions and commits.

## Root Cause
Scoop installs the package as `cli` instead of `entire`, so the binary lives at:
`C:\Users\admin\scoop\apps\cli\current\entire.exe`

This breaks Entire's internal symlink resolution. Every time any command tries to 
resolve the path, it fails with:

> failed to resolve symlinks for C:\Users\admin\scoop\apps\cli\current\entire.exe:
> The system cannot find the path specified.

This affects:
- `entire configure --absolute-git-hook-path` — fails completely
- `entire enable --agent cursor --force` — fails to install git hooks
- `entire agent add cursor` — fails to install git hooks

## What I Tried
- Manually rewrote `.git/hooks/post-commit` and `.git/hooks/prepare-commit-msg` 
  to use the full absolute path
- Manually rewrote `.cursor/hooks.json` with full path and correct UTF-8 encoding 
  (no BOM)
- Set `ENTIRE_BIN` environment variable to the full path
- `entire status` shows **Enabled**, git hooks exist, hooks.json is correct

Despite all of this, no sessions or checkpoints are ever recorded.

## Expected Behavior
Entire should either:
1. Detect the correct binary path regardless of Scoop package name, **or**
2. Document a manual workaround for Windows + Scoop users

## Possible Fix
Allow users to manually set the binary path via an env variable like `ENTIRE_BIN` 
that Entire respects internally when resolving symlinks.

### Entire CLI version

Entire CLI 0.7.5

### OS and architecture

Windows-11 pro 64 bit

### Agent

cursor

### Terminal

WIndows Terminal 

### Logs / debug output

```bash

```

### Additional context

_No response_

karthik-rameshkumar:
[https://entire.io/gh/entireio/cli/trails/841](<https://entire.io/gh/entireio/cli/trails/841>)

Fixes entireio/cli#1424.

## Background

entireio/cli#1424 reports that on Windows 11 (Scoop install, Cursor), Entire records nothing — `session list` / `checkpoint list` stay at 0 even though `entire status` shows enabled and hooks look correctly configured. Investigation found **three independent root causes** stacked together.

### 1\. Scoop installs the app dir as `cli`, not `entire`

The published bucket manifest is literally `cli.json` (verified live), so the binary lands at `…\scoop\apps\cli\current\entire.exe`. GoReleaser defaults the scoop manifest name to the project name, which resolves to the repo name (`cli`). The `scoops:` block set no `name:`.

**Fix:** set `scoops[].name: entire`.

### 2\. `--absolute-git-hook-path` fails on Scoop's `current\` junction

`hookCmdPrefix` calls `filepath.EvalSymlinks(os.Executable())`, which can't resolve Scoop's `current\` **NTFS directory junction** and returns *"The system cannot find the path specified."* — the exact error in the report. This blocks `entire enable --agent cursor`, `agent add cursor`, and `configure --absolute-git-hook-path`.

**Fix:** on Windows, fall back to the unresolved `os.Executable()` path (extracted into `resolveHookExePath` so all branches are unit-tested cross-platform). That path is valid and, on Scoop, actually **more stable** than the resolved one — `current\` survives version updates that repoint the junction, whereas a resolved versioned path disappears on the next update.

### 3\. Cursor agent hooks silently never fire on Windows (the actual "nothing recorded")

`cursor.InstallHooks` hardcoded the POSIX `sh -c '…'` wrapper. Cursor spawns hook commands via the native OS shell, so on a Windows host with no runnable POSIX `sh` the wrapper fails to launch — and because it's the *silent* wrapper, the failure is swallowed (no hook, no error). The shared OS-aware path (`UseWindowsProductionHooks` + `WrapProductionSilentHookCommandForOS`) already exists and was built explicitly "so the other sh-based agents can adopt the Windows fallback" (commit `1cf8eadc2`) — but only codex had adopted it.

**Fix:** migrate Cursor to the shared path. It's **probe-gated**: a no-op on non-Windows and on Windows hosts where `sh` works; it only swaps in the native `cmd.exe` wrapper when no `sh` is present.

## Scope notes

* The other sh-based agents (gemini, claude-code, factory-droid) still hardcode the sh wrapper and have the same latent bug; **copilot-cli deliberately stays sh-based** because it runs hook commands via a `bash` field (not the native shell), so the cmd.exe wrapper would break it. A blanket migration is unsafe — auditing each agent's execution model is left to a follow-up.
* Windows **Authenticode signing** (Smart App Control) is a separate issue (entireio/cli#1505), not addressed here.
* Existing Scoop users on `cli` won't auto-migrate; they'd `scoop uninstall cli` + `scoop install entire`. The stale `cli.json` in the bucket can be removed separately.

## Tests

* `resolveHookExePath`: resolves normally / Windows falls back on failure / non-Windows surfaces the error (injected GOOS + resolver, runs on any host).
* Cursor: Windows probe-success keeps sh wrappers; probe-failure installs cmd.exe wrappers (via `SetWindowsHookProbeForTesting`).
* `go build ./...`, `golangci-lint` (changed pkgs), and `go test ./cmd/entire/cli/agent/... ./cmd/entire/cli/strategy/...` all pass locally.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -100,7 +100,14 @@ release:
   prerelease: auto
 
 scoops:
-  - repository:
+  # Name the manifest (and therefore the Scoop app directory) "entire". Without
+  # this, goreleaser defaults the manifest name to the project name, which
+  # resolves to the repo name ("cli"), so `scoop install` lands the binary in
+  # …\scoop\apps\cli\current\entire.exe. That mismatched app-dir name is
+  # surprising ("scoop install cli"?) and fed the Windows hook-path bug in
+  # https://github.com/entireio/cli/issues/1424.
+  - name: entire
+    repository:
       owner: entireio
       name: scoop-bucket
       token: "{{ .Env.TAP_GITHUB_TOKEN }}"
--- a/cmd/entire/cli/agent/cursor/hooks.go
+++ b/cmd/entire/cli/agent/cursor/hooks.go
@@ -130,13 +130,21 @@ func (c *CursorAgent) InstallHooks(ctx context.Context, localDev bool, force boo
 	subagentStartCmd := cmdPrefix + HookNameSubagentStart
 	subagentEndCmd := cmdPrefix + HookNameSubagentStop
 	if !localDev {
-		sessionStartCmd = agent.WrapProductionSilentHookCommand(sessionStartCmd)
-		sessionEndCmd = agent.WrapProductionSilentHookCommand(sessionEndCmd)
-		beforeSubmitPromptCmd = agent.WrapProductionSilentHookCommand(beforeSubmitPromptCmd)
-		stopCmd = agent.WrapProductionSilentHookCommand(stopCmd)
-		preCompactCmd = agent.WrapProductionSilentHookCommand(preCompactCmd)
-		subagentStartCmd = agent.WrapProductionSilentHookCommand(subagentStartCmd)
-		subagentEndCmd = agent.WrapProductionSilentHookCommand(subagentEndCmd)
+		// Cursor spawns hook commands through the native OS shell (cmd.exe on
+		// Windows), so a `sh -c '…'` wrapper silently fails to launch on a
+		// Windows host without a working POSIX sh — no hook fires and, because
+		// this is the *silent* wrapper, no error surfaces (issue #1424).
+		// UseWindowsProductionHooks probes for a runnable sh and only swaps in
+		// the native cmd.exe wrapper when one is absent, so this is a no-op on
+		// hosts (incl. all non-Windows) where the sh wrapper already works.
+		useWindowsHooks := agent.UseWindowsProductionHooks(ctx, localDev)
+		sessionStartCmd = agent.WrapProductionSilentHookCommandForOS(sessionStartCmd, useWindowsHooks)
+		sessionEndCmd = agent.WrapProductionSilentHookCommandForOS(sessionEndCmd, useWindowsHooks)
+		beforeSubmitPromptCmd = agent.WrapProductionSilentHookCommandForOS(beforeSubmitPromptCmd, useWindowsHooks)
+		stopCmd = agent.WrapProductionSilentHookCommandForOS(stopCmd, useWindowsHooks)
+		preCompactCmd = agent.WrapProductionSilentHookCommandForOS(preCompactCmd, useWindowsHooks)
+		subagentStartCmd = agent.WrapProductionSilentHookCommandForOS(subagentStartCmd, useWindowsHooks)
+		subagentEndCmd = agent.WrapProductionSilentHookCommandForOS(subagentEndCmd, useWindowsHooks)
 	}
 
 	count := 0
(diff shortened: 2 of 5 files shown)


karthik-rameshkumar:
## Self-review (addressed in 4dd458656)

Reviewed the diff adversarially and found one real defect, now fixed on the branch.

**Finding — medium — Cursor lacked replace-on-change migration.** `cursor.InstallHooks` appended a hook only when the exact command string was absent, and removed existing Entire hooks only under `--force`. Once this PR lets the wrapper form change between installs (`sh -c` ↔ `cmd.exe`, via the newly-adopted `UseWindowsProductionHooks` probe — whose non-memoized design is *meant* to drive self-correcting migration), a non-`force` reinstall after a probe-result flip would leave the stale `sh -c` entry **and** add the `cmd.exe` one. That dead duplicate could double-fire (duplicate `session-start`/`stop` → duplicate checkpoints) if `sh` later worked at hook runtime. Codex avoids this with `syncHookCommand`; Cursor didn't have the equivalent.

**Fix:** added `syncEntireHook` (mirrors codex): no-op if the desired command is already present, else remove any existing Entire hook of that type (any wrapper form) before appending — clean replacement without `--force`, non-Entire entries preserved. Added `TestInstallHooks_WindowsProbeFlipMigratesCleanly` asserting no `sh -c` wrapper survives an sh→cmd.exe reinstall and no duplicates remain.

**Verified good:** `scoops[].name` confirmed a real GoReleaser field (controls manifest/app name); `resolveHookExePath` branches tested cross-platform; existing Cursor/strategy assertions unaffected (Linux test host → sh form). `go build ./...`, `golangci-lint` (changed pkgs), and the full cursor + strategy suites pass locally. No other findings.

gtrrz-victor (approved):

PR #1738, opened 2026-07-13
Fix agent hooks hanging on stdin EOF on Windows/Git Bash (#1398)

### LINKED ISSUE #1398: Bug: Gemini CLI hooks session-start hangs indefinitely on Windows (Scoop install)
### What happened?

The `entire hooks gemini session-start` command hangs indefinitely when running on Windows (MINGW64/Git Bash, installed via Scoop). Command never returns, so Gemini's hooks likely time out and get killed, leading to no session ever being started. Killing it with Ctrl+C and checking `.entire/logs/*.log` shows no session start entry. This is a bug in Entire CLI on Windows.

### Steps to reproduce

1. Open Git Bash / MINGW64 on Windows
2. Install Entire CLI using Scoop (version 0.7.5)
3. Run: `entire hooks gemini session-start`
4. Observe that the command hangs indefinitely
5. Press Ctrl+C to stop it
6. Check log output: `cat .entire/logs/*.log | tail -20`

### Entire CLI version

Entire CLI 0.7.5

### OS and architecture

Windows, installed via Scoop (0.7.5)

### Agent

Gemini CLI

### Terminal

Git Bash / MINGW64

### Logs / debug output

```bash
See `.entire/logs/*.log` output after manually killing the session-start process.
```

### Additional context

Workaround: Use WSL (Windows Subsystem for Linux) instead of Git Bash — session-start works correctly there. Expected behavior: Command should return quickly and log session start. Please fix terminal compatibility on Windows.

karthik-rameshkumar:
[https://entire.io/gh/entireio/cli/trails/844](<https://entire.io/gh/entireio/cli/trails/844>)

Fixes entireio/cli#1398.

## Symptom

`entire hooks gemini session-start` hangs indefinitely on Windows (Git Bash / MINGW64, Scoop install). Gemini's hook times out and gets killed, so no session is ever recorded. WSL is unaffected. Reported for Gemini, but the cause is shared across all agents.

## Root cause

The blocking call is the shared helper `ReadAndParseHookInput` (`cmd/entire/cli/agent/event.go`):

```go
data, err := io.ReadAll(stdin)   // blocks until stdin reaches EOF
```

`io.ReadAll` returns only when stdin is **closed** (EOF). Agents deliver the hook payload by piping JSON to the hook process, but some keep the *write* end of that pipe open for the hook's lifetime rather than closing it right after the write. On Windows/Git Bash the full payload arrives but EOF never does, so `io.ReadAll` waits forever. WSL closes stdin normally, which is why it worked there.

Every agent's `ParseHookEvent` routes through this helper (cursor, claude-code, codex, factory-droid, gemini, opencode, copilot), so the hang was latent for all of them on Windows.

## Fix

Two changes in `ReadAndParseHookInput`:

1. **Stream-decode with** `json.Decoder` **instead of** `io.ReadAll`**.** A decoder returns as soon as one complete JSON value has been read — it does not wait for EOF. This is the core fix: once the payload arrives, the hook proceeds regardless of whether the parent ever closes stdin.
2. **Short-circuit an interactive-terminal stdin.** If stdin is a TTY there is no payload coming (the command was run by hand, or an agent left the console attached instead of wiring a pipe). Rather than block, return promptly as `empty hook input`. This also makes the reporter's manual `entire hooks gemini session-start` repro fail fast with a clear message instead of hanging.

Empty-input (`empty hook input`) and malformed-JSON (`failed to parse hook input`) error semantics are preserved.

### Scope note

If an agent on Windows genuinely never delivers the payload on stdin (an upstream bug rather than a held-open pipe), this converts an indefinite hang into a fast, clearly-logged failure rather than a working session — that boundary is the best the CLI can do without the payload. The streaming-decoder path fixes the case where the payload *does* arrive (the likely scenario, given WSL — same CLI code — works).

## Tests

`cmd/entire/cli/agent/event_test.go`:

* `ReturnsBeforeEOF` — writes a payload to an `io.Pipe` and **never closes the writer**, asserting the read returns within 3s. This test would hang (and fail) under the old `io.ReadAll`. Direct regression guard for entireio/cli#1398.
* `EmptyInputEOF`, `MalformedJSON`, `ValidPayload` — preserve error/parse semantics.

`go build ./...`, `golangci-lint` (agent pkg, 0 issues), and the full `./cmd/entire/cli/agent/...` tree pass locally.

### DIFF
--- a/cmd/entire/cli/agent/copilotcli/lifecycle.go
+++ b/cmd/entire/cli/agent/copilotcli/lifecycle.go
@@ -169,11 +169,13 @@ func (c *CopilotCLIAgent) buildSubagentStop(env *hookEnvelope) *agent.Event {
 }
 
 func (c *CopilotCLIAgent) readHookEnvelope(stdin io.Reader) (*hookEnvelope, error) {
-	data, err := io.ReadAll(stdin)
+	// Stream one JSON value rather than io.ReadAll so the hook never blocks
+	// waiting for stdin EOF that some agents don't send on Windows (issue #1398).
+	raw, err := agent.ReadHookInputRaw(stdin)
 	if err != nil {
-		return nil, fmt.Errorf("failed to read hook input: %w", err)
+		return nil, fmt.Errorf("read hook input: %w", err)
 	}
-	return parseHookEnvelope(data)
+	return parseHookEnvelope(raw)
 }
 
 // resolveTranscriptRef computes the transcript path from the session ID.
--- a/cmd/entire/cli/agent/event.go
+++ b/cmd/entire/cli/agent/event.go
@@ -189,6 +189,14 @@ func ReadAndParseHookInput[T any](stdin io.Reader) (*T, error) {
 // (e.g. key-name fallbacks, or forwarding the bytes to a subprocess) use this
 // directly, while the common case uses ReadAndParseHookInput.
 func ReadHookInputRaw(stdin io.Reader) (json.RawMessage, error) {
+	return ReadHookInputRawLimited(stdin, -1)
+}
+
+// ReadHookInputRawLimited is ReadHookInputRaw with a ceiling of limit bytes on
+// the JSON value (limit < 0 means unlimited). It is used at the external/plugin
+// boundary to bound an untrusted payload — without reintroducing the EOF-wait
+// hang, since the streaming decoder still returns on the first complete value.
+func ReadHookInputRawLimited(stdin io.Reader, limit int64) (json.RawMessage, error) {
 	// If stdin is an interactive terminal there is no payload coming at all: the
 	// command was run by hand, or the agent left the console attached instead of
 	// wiring up a pipe. Decoding would block waiting for input that never comes,
@@ -197,8 +205,12 @@ func ReadHookInputRaw(stdin io.Reader) (json.RawMessage, error) {
 		return nil, errors.New("empty hook input")
 	}
 
+	r := stdin
+	if limit >= 0 {
+		r = io.LimitReader(stdin, limit)
+	}
 	var raw json.RawMessage
-	if err := json.NewDecoder(stdin).Decode(&raw); err != nil {
+	if err := json.NewDecoder(r).Decode(&raw); err != nil {
 		if errors.Is(err, io.EOF) {
 			return nil, errors.New("empty hook input")
 		}
--- a/cmd/entire/cli/agent/event.go
+++ b/cmd/entire/cli/agent/event.go
@@ -5,7 +5,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"time"
+
+	"golang.org/x/term"
 )
 
 // EventType represents a normalized lifecycle event from any agent.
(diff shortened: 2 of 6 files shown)


karthik-rameshkumar:
## Self-review (addressed in 3054b0c50)

Reviewed the diff adversarially. One significant finding, now fixed on the branch.

**Finding — high — the fix's scope was incomplete; four hook readers still hung.** The first commit only fixed agents that route through `ReadAndParseHookInput`. These read the payload with their own EOF-dependent `io.ReadAll(stdin)` and retained the #1398 hang:

- `copilotcli` `readHookEnvelope` — **which the original PR body listed as fixed** (it wasn't)
- `pi` `ParseHookEvent`
- `external` `ParseHookEvent`
- claude-code **post-todo** (`parseSubagentCheckpointHookInput` in the `cli` package)

**Fix:** extracted the terminal guard + streaming decode into a shared primitive (`ReadHookInputRaw`) plus `StdinLooksInteractive`, and wired the bypassers to it:
- `pi` and claude-code post-todo → `ReadAndParseHookInput` (JSON-into-type).
- `copilot` → `ReadHookInputRaw` (its `parseHookEnvelope` needs raw bytes for `sessionId`/`session_id` key fallbacks).
- `external` forwards raw stdin bytes verbatim to its subprocess (may be empty/non-JSON), so it keeps `io.ReadAll` but adds the interactive-terminal guard. Its piped-held-open case is a documented residual, since fixing it properly needs a framing decision on the plugin contract.

So the corrected scope: **cursor, claude-code (incl. post-todo), codex, factory-droid, gemini, opencode, copilot, and pi are all fixed**; external gets the interactive-terminal guard with a noted residual.

**Verified good:** existing tests only assert error *presence* (not message text) for the moved parsers, so routing them through the shared primitive is safe; wrapcheck skips the generic `ReadAndParseHookInput` but flagged the non-generic `ReadHookInputRaw`, so copilot's call site wraps it. `go build ./...`, `golangci-lint` (agent + cli packages, 0 issues), and the full `./cmd/entire/cli/agent/...` + `cli` hooks tests pass locally.

karthik-rameshkumar:
## Addressed trail-review finding (commit 6d40e2549)

The review correctly caught that `external.ParseHookEvent` still used `io.ReadAll(io.LimitReader(stdin, maxParseHookBytes))` for the non-interactive (piped) case — the exact #1398 hang: an external/plugin agent that keeps the stdin pipe's write end open after delivering its payload never sends EOF, so that read would block forever. The `StdinLooksInteractive` guard only short-circuited the TTY/no-payload case.

**Fix:** `external` now streams a single JSON value like every other agent path. Added `ReadHookInputRawLimited` (the streaming raw reader plus a byte ceiling) so external keeps its 10 MB bound without waiting for EOF; the decoded raw bytes are still forwarded verbatim to the `parse-hook` subprocess, so the plugin payload contract is unchanged. This is safe because the external `parse-hook` stdin is the host's hook payload, which is JSON (confirmed by the agent's own test fixtures, e.g. `{"session_file": "..."}`).

**Result:** no hook-input path waits for stdin EOF anymore.

**New regression tests** (`event_test.go`): `ReadHookInputRawLimited` returns before stdin EOF (would hang under the old `io.ReadAll`), and rejects an over-limit payload instead of reading unbounded. `go build ./...`, `golangci-lint` (0 issues), and the full `./cmd/entire/cli/agent/...` tree pass locally.

gtrrz-victor (approved):

PR #1744, opened 2026-07-14
fix(strategy): defer checkpoint push until a normal remote branch exists (#1743)

### LINKED ISSUE #1743: Prevent checkpoint metadata from becoming the default branch on a repository’s first push
## Problem

When Entire is enabled before a repository has any normal branch on its remote, the first ordinary push can publish `entire/checkpoints/v1` before the user branch. GitHub may then select the metadata branch as the repository default branch.

`ghRepoCreate` already bypasses the pre-push hook for its initial push (`git push --no-verify ...`) to avoid this ordering. The generic/manual path is still exposed.

## Reproduction

1. Create a local repository and enable Entire.
2. Add an empty GitHub remote (or use bootstrap with the initial commit skipped).
3. Make normal commits and enough agent activity to create local checkpoint metadata.
4. Run the first normal push, for example `git push -u origin HEAD`.

The pre-push hook calls `ManualCommitStrategy.PrePush`, which independently pushes checkpoint refs before Git pushes the requested user branch. On an empty GitHub repository, `entire/checkpoints/v1` can become the first remote branch and therefore the default branch.

## Expected behavior

The first remote branch for the user repository must be the user’s pushed branch. Checkpoint metadata should be published only after a non-metadata branch exists on that remote.

A configured, distinct `checkpoint_remote` must keep its existing behavior; its metadata-only repository is intentionally separate.

## Proposed direction

Pass Git pre-push stdin through `hooks git pre-push` into the strategy and parse the ref-update lines. When pushing to the user remote and it has no normal branches yet, defer checkpoint publication for that invocation if the pending ref update creates a normal branch. The next push can publish the queued metadata.

Avoid a blanket "empty remote" skip because it would incorrectly suppress the intentional separate-checkpoint-remote case.

## Regression coverage

Add an integration test with an empty remote, a local checkpoint branch/ref, and a real `git push` through the hook. Assert the user branch is created first and the checkpoint metadata is not pushed until a later push.

karthik-rameshkumar:
[https://entire.io/gh/entireio/cli/trails/849](<https://entire.io/gh/entireio/cli/trails/849>)

Closes entireio/cli#1743.

## Problem

When Entire is enabled before a repository has any branch on its remote, the first ordinary push could publish `entire/checkpoints/v1` before the user's branch. On an empty GitHub repo, GitHub then selects the metadata branch as the repository **default branch**.

`ghRepoCreate` already bypasses the pre-push hook for its initial push; the generic/manual push path was still exposed.

## Fix

* Git's pre-push hook now routes through a new `PrePushFromGitHook` (vs. direct `PrePush` callers), which enables first-user-branch protection.
* `deferCheckpointPushUntilNormalBranch` inspects the **actual push target(s)** for the user remote — including a distinct `pushurl` resolved via `git remote get-url --push --all` (new `PushTargetsInDir` helper) — and defers automatic checkpoint publication until a non-metadata (`entire/checkpoints/v1`) branch exists there.
* A configured, separate `checkpoint_remote` is intentionally **exempt** (dedicated metadata store), matching the issue's requirement.
* The OPF rewrite still runs before deferral, so a user push that explicitly includes `v1` stays redacted.
* Deferral **fails closed**: any error resolving/inspecting the remote defers publication while the user's own push proceeds; a later push publishes the queued metadata.

## Tests

* `TestGitPushWithHooks_DefersCheckpointsUntilFirstUserBranchExists` — empty remote, real `git push` through the hook, asserts the user branch lands first and checkpoints are published only on a later push. Runs on both `git-branch` and `git-refs` backends.
* `TestGitPushWithHooks_DefersCheckpointsUntilPushURLHasUserBranch` — origin's fetch URL has a branch but pushes route to a different empty repo; guard must inspect the pushurl.
* `TestPrePushFromGitHook_DeferralStillRunsOPF` — asserts OPF advances the local `v1` ref before deferral.

Verified the tests fail without the wiring change (reproducing the exact bug) and pass with it; existing `SyncsCheckpointsToRemote` still passes (no regression on the normal path). `mise run fmt && mise run lint` clean; full unit suite (7,849 tests) passes.

### DIFF
--- a/cmd/entire/cli/checkpoint/remote/git.go
+++ b/cmd/entire/cli/checkpoint/remote/git.go
@@ -11,10 +11,17 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/entireio/cli/cmd/entire/cli/logging"
 	"github.com/entireio/cli/cmd/entire/cli/settings"
 )
+
+// stampConfigTimeout bounds the local git-config reads/writes that mark a newly
+// created checkpoint remote as skipped. They run detached from the fetch's
+// context (see stampNewlyCreatedRemote), so a bound guards against a stuck
+// config lock hanging the caller.
+const stampConfigTimeout = 10 * time.Second
 
 // CheckpointTokenEnvVar is the environment variable for providing an access token
 // used to authenticate git push/fetch operations for checkpoint branches.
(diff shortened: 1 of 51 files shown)


cmd/entire/cli/strategy/manual_commit_push.go
```diff
@@ -132,6 +159,71 @@ func (s *ManualCommitStrategy) PrePush(ctx context.Context, remote string) error
...
+			if branch != fields[1] && branch != paths.MetadataBranchName {
+				return false
+			}
+		}
+	}
+
+	logging.Info(ctx, "checkpoint push deferred until a normal remote branch exists",
+		slog.String("remote", ps.remote),
+	)
+	return true
```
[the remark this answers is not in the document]
karthik-rameshkumar: Good catch — and it pushed the fix in the right direction. This landed alongside a separate change that narrowed the guard, which happens to resolve exactly this.

The function (now `deferCheckpointPushOnEmptyRemote`) no longer returns "publish" on the *first* target with a branch. It inspects every push target (`git remote get-url --push --all`) and now:

- defers the moment **any** target has no `refs/heads/*` (truly empty), and
- publishes **only when every** target already has a branch.

So with multiple `pushurl` entries where one destination is still empty, we defer — `entire/checkpoints/v1` can't become the first/default branch on that empty destination. Once the user's branch lands on all of them, a later push publishes.

Added a regression test for precisely this — `TestGitPushWithHooks_DefersCheckpointsWhenAnyPushURLTargetIsEmpty` — which configures `origin` with two push destinations (one populated, one empty), asserts deferral on both while any target is empty, and asserts publication once all have a branch. Runs on both checkpoint backends.

karthik-rameshkumar:
## Review-response summary

For reviewers picking this up: the diff grew across several review rounds. Here's the arc so the commit history reads clearly, oldest → newest.

**Core fix**
- `d5d8ac6` — defer the pre-push-hook checkpoint metadata push when the user's push remote has no branches yet, so the user's first branch (not `entire/checkpoints/v1`) becomes the repo default on GitHub. Routes the hook through a new `PrePushFromGitHook`.

**Correctness follow-ups (each from a specific review point)**
- `a05a6a8` — narrowed the guard to fire only on a **truly empty** remote. The first cut deferred whenever there was no *non-metadata* branch, which broke checkpoint sync to a remote that already had `v1` but no user branch (the `TestAlternates_…_CheckpointSync` canary). Once any head exists, deferring can't prevent the default and only blocks legitimate syncs.
- `eca8e7f` — regression test for the multi-`pushurl` case (Copilot's note): defer if **any** push destination is empty, not just the first.
- `ba7e1a9` — bounded the probe with a 10s timeout. The hook ctx has no deadline and the probe runs synchronously before the user's push, so a stalled remote could have hung every `git push`.
- `30a6c53` — expire the memoization marker via a TTL. A permanently-trusted marker would let a force-emptied / recreated-empty remote skip the guard and re-hit the bug.

**Performance**
- `65e002a` — memoize the "remote has branches" observation so the guard costs no network round trip on steady-state pushes (only during the first-push window / once per TTL).

**Latest review pass (multi-agent review of the full diff)**
- `e907fa2` — (1) **redact** push-target URLs before logging (they can embed credentials); (2) `PushTargetsInDir` no longer uses `IsURL`'s bare-`@` heuristic, which misclassified a remote *name* containing `@` and fail-closed-deferred it forever — now follows git's own scp detection (colon-before-slash); (3) empty-target now errors instead of returning `[]string{""}`; (4) marker moved to the sibling-convention `.git/entire-push-bootstrap/`, routed through `os.Root`, read content+mtime in a single open.
- `b27efd1` — corrected the marker's teardown home: swept on `entire disable` (matching the `entire-session-locks` precedent for sibling non-session-data caches) rather than enrolled in `entire clean`'s session-data GC. Fixes a leak where the marker survived `disable`, and drops the exhaustive-switch maintenance surface.

**Design note (raised and dismissed):** parsing pre-push stdin ref-updates instead of probing the remote was considered (issue #1743's own suggestion) but doesn't work for the multi-`pushurl` case — stdin's `remote-sha` reflects git's fetch-URL view and can't tell whether a *pushurl* target is empty, which is exactly what this guard must detect.

Net behavior: first push to a fresh remote lands the user's branch first; checkpoint metadata publishes on the next push; steady-state pushes pay nothing; a separate `checkpoint_remote` is exempt throughout. Covered by integration tests on both checkpoint backends plus unit tests for the probe/marker/classification logic.

pjbgf:
This is only needed for `git-branch`, as the ref created is under `refs/heads/`. We are transitioning away from `entire/checkpoints/v1`, so it would be nice to gate these changes to only take place for when type is not `git-refs`:

```json
  "checkpoints": {
    "primary": {
      "type": "git-refs"
    }
  }
```

That way, if/when we ever remove the `git-branch` code this could go away with it.

cmd/entire/cli/strategy/manual_commit_push.go
```diff
@@ -132,6 +164,210 @@ func (s *ManualCommitStrategy) PrePush(ctx context.Context, remote string) error
...
+	f, err := root.Open(pushBootstrapMarkerFileName)
+	if err != nil {
+		return "", false
+	}
+	defer func() { _ = f.Close() }()
+	info, err := f.Stat()
+	if err != nil {
+		return "", false
+	}
+	data, err := io.ReadAll(f)
```
pjbgf: Let's add an `io.LimitReader` here to limit the max amount of memory this can consume.

cmd/entire/cli/checkpoint/remote/git.go
```diff
@@ -241,6 +242,59 @@ func LsRemoteInDir(ctx context.Context, dir, remote string, patterns ...string)
 	return lsRemote(ctx, dir, remote, patterns...)
 }
 
+// PushTargetsInDir resolves the endpoint or endpoints Git will use for a push.
+// A named remote may configure pushurl, which differs from its fetch URL and
+// can contain more than one destination. URLs and local paths are already
+// concrete push targets and are returned unchanged.
+func PushTargetsInDir(ctx context.Context, dir, target string) ([]string, error) {
```
pjbgf: I'd refrain from trying to get (or fetch) all remotes. The hook receives what remote is being used, let's just use that one instead.

cmd/entire/cli/strategy/manual_commit_push.go
```diff
@@ -132,6 +164,210 @@ func (s *ManualCommitStrategy) PrePush(ctx context.Context, remote string) error
...
+	// trusted for pushBootstrapTTL before the remote is re-probed. That bounds
+	// both the network cost (at most one probe per TTL) and the window in which a
+	// re-emptied remote could wrongly skip the guard.
+	fingerprint := pushTargetsFingerprint(targets)
+	if stored, fresh := readPushBootstrapMarker(ctx); fresh && stored == fingerprint {
+		return false
+	}
+
+	for _, target := range targets {
+		out, lsErr := checkpointremote.LsRemoteInDir(probeCtx, dir, target, "refs/heads/*")
```
pjbgf: We should refrain from fetching all user's remotes. Different targets may require different authentication methods and that could freak users out (i.e. unexpected Yubikey flashing).

pjbgf (changes requested):
@karthik-rameshkumar thanks for looking into this. On top of gating this behind a check on the checkpoint type, I think we could:

1. Check whether `.git/refs/remotes/<remote>/HEAD` exists
2. Fetch the specific remote passed to the hook, or default to `origin`

I probably would avoid the additional round-trip to the server (step 2), and just rely on the first for deciding when commits should be pushed first. But there could be some edge cases for that. 

The step 2 adds an additional physical presence check (for users with SSH+ssk enabled), so from an UX perspective users would require at least 3 (fetch, checkpoint push, commit push) confirmations which is not ideal.

cmd/entire/cli/checkpoint/remote/git.go
karthik-rameshkumar: Done. Dropped `PushTargetsInDir/get-url --push --all` entirely. The guard now uses only the remote git hands the hook.

cmd/entire/cli/strategy/manual_commit_push.go
karthik-rameshkumar: Fixed. No network on the pre-push path at all now. Replaced the per-target ls-remote with a local git for-each-ref `refs/remotes/<remote>/* `check, so nothing dials any remote

cmd/entire/cli/strategy/manual_commit_push.go
karthik-rameshkumar: The marker file this read from is gone (the local ref check needs no memoization), so there's no unbounded read left.

karthik-rameshkumar:
> @karthik-rameshkumar thanks for looking into this. On top of gating this behind a check on the checkpoint type, I think we could:
> 
> 1. Check whether `.git/refs/remotes/<remote>/HEAD` exists
> 2. Fetch the specific remote passed to the hook, or default to `origin`
> 
> I probably would avoid the additional round-trip to the server (step 2), and just rely on the first for deciding when commits should be pushed first. But there could be some edge cases for that.
> 
> The step 2 adds an additional physical presence check (for users with SSH+ssk enabled), so from an UX perspective users would require at least 3 (fetch, checkpoint push, commit push) confirmations which is not ideal.

Reworked per your direction. The guard is now git-branch-only (git-refs stores under `refs/entire/*`, which a forge can't make the default, so it publishes on first push, no guard). 

The decision is purely local: check for a `refs/remotes/<remote>/*` tracking ref, no `ls-remote`/fetch, so no SSH-sk touch prompts. 

One note: I went with any tracking ref rather than HEAD specifically — git push doesn't create `refs/remotes/<remote>/HEAD` (only clone/set-head do), so a HEAD check would defer forever in the init+push flow; any tracking ref appears after the first push and self-heals.

pjbgf (approved):

PR #1746, opened 2026-07-14
fix(checkpoint): stamp adhoc checkpoint remotes at creation, not retroactively

pjbgf:
https://entire.io/gh/entireio/cli/trails/852

## What

Reworks how the CLI keeps filtered checkpoint fetches from cluttering the
user's git config with phantom remotes.

When `filtered_fetches` is enabled, checkpoint fetches run
`git fetch --filter=blob:none <url>`. Git records a URL-keyed
`remote.<url>.promisor` section so it can lazy-fetch filtered-out objects
later — but that section also turns the URL into a phantom remote that
`git fetch --all` / `git remote update` keep dialing. We stamp
`skipFetchAll`/`skipDefaultUpdate` on it so bulk fetches skip it.

Previously (#1733) we re-scanned and stamped **any** URL-keyed promisor
section on **every** filtered fetch, rewriting config for remotes we
didn't create. This changes it to stamp **only the section this fetch
creates**, leaving pre-existing remotes untouched so we never rewrite the
user's own git config.

## Changes

- Drop the promisor-key read; detect "did we just create this remote" by
  comparing `remote.<url>.*` section existence **before and after** the
  fetch.
- Stamp only when the section newly appeared, so remotes the user (or an
  older CLI) already had are left alone.
- Stamp **regardless of fetch success**: git writes the promisor section
  eagerly during connection setup, so a filtered fetch that later fails
  (e.g. a missing ref during `entire resume`) still leaves a phantom
  remote. The earlier draft returned early on error before stamping, and
  its "only if new" gate then made the miss permanent — the entry looked
  pre-existing on every retry and lingered unstamped forever. The
  post-fetch existence recheck also avoids inventing a section when the
  fetch dies before git writes anything.

## Why

- Avoid mutating config for remotes we didn't create.
- Keep adhoc checkpoint URL remotes out of the user's bulk-fetch surface
  without accumulating unstamped phantoms.

## Testing

- `TestFetch_FilteredURLFetchMarksNewRemoteSkipped` — new URL remote is
  stamped and excluded from `git fetch --all`.
- `TestFetch_FailedFilteredFetchStillStampsNewRemote` — a filtered fetch
  that fails after git records the promisor still gets stamped
  (regression test; verified it fails without the fix).
- `TestFetch_ExistingURLRemoteNotReStamped` — a pre-existing remote is
  left untouched.
- `TestFetch_UnfilteredFetchDoesNotCreateConfigSection`,
  `TestMarkRemoteSkipped_SetsBothKeys`, `TestGitRemoteSectionExists`.
- `mise run fmt` / `mise run lint` clean; package + integration
  config-guard tests pass.

## Note

Promisor entries created by an **older CLI version** are not back-filled
— by design, the CLI only stamps sections it creates itself, so it never
rewrites config the user already has. Cleaning up legacy unstamped
entries would be a separate change.

---

> [!NOTE]
> **Medium Risk**
> Changes when and how the CLI mutates local git config during checkpoint fetches; behavior is narrower and best-effort, but mistakes could leave phantom remotes or skip stamping in edge cases.
> 
> **Overview**
> Reworks how filtered checkpoint URL fetches avoid **phantom remotes** in local `.git/config`. Instead of re-stamping **every** URL-keyed promisor section on each successful filtered fetch, the CLI now records whether `remote.<url>.*` existed **before** the fetch and only stamps when that section is **newly created** by this fetch—so pre-existing user (or legacy) remotes are not rewritten.
> 
> Stamping runs **after the fetch attempt regardless of success** (still gated on “was new”), fixing cases where git writes the promisor section during connection setup but the fetch then fails or times out—previously an early return on error could leave an unstamped phantom forever. Config writes use a **detached context with a 10s timeout** so a cancelled fetch deadline does not skip the stamp. The stamp is narrowed to **`skipFetchAll` only** (no longer sets `skipDefaultUpdate`), with existence detected via **`gitRemoteSectionExists`** (exact URL subsection match).
> 
> Tests cover failed fetches, existing remotes left alone, cancelled-context stamping, and integration **git config guard** regex updates for the slimmer stamped section.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 740b7994bedfec52834196357e178dd07c6d3a19. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/checkpoint/remote/git.go
+++ b/cmd/entire/cli/checkpoint/remote/git.go
@@ -11,10 +11,17 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/entireio/cli/cmd/entire/cli/logging"
 	"github.com/entireio/cli/cmd/entire/cli/settings"
 )
+
+// stampConfigTimeout bounds the local git-config reads/writes that mark a newly
+// created checkpoint remote as skipped. They run detached from the fetch's
+// context (see stampNewlyCreatedRemote), so a bound guards against a stuck
+// config lock hanging the caller.
+const stampConfigTimeout = 10 * time.Second
 
 // CheckpointTokenEnvVar is the environment variable for providing an access token
 // used to authenticate git push/fetch operations for checkpoint branches.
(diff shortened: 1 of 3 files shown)


pjbgf:
@cursor review

pjbgf:
@cursor review

toothbrush (approved):

PR #1752, opened 2026-07-15
feat(plugins): Lua plugin + hooks system

suhaanthayyil:
https://entire.io/gh/entireio/cli/trails/857

## Summary

Adds a pure-Go embedded Lua runtime (`github.com/yuin/gopher-lua`) so third
parties can author no-build-step plugins that subscribe to lifecycle/git hooks
and contribute commands, layered on top of the existing kubectl-style
`entire-<name>` binary plugins without breaking them.

Grown phase by phase; each phase is a separate conventional commit.

## Design constraints (all met)

- **Pure Go / `CGO_ENABLED=0`** everywhere — gopher-lua only (MIT, in
  `.allowed-licenses`). Verified `CGO_ENABLED=0 go build ./...` and the release
  binary builds clean.
- **Manifest is `plugin.json`** parsed strictly (`DisallowUnknownFields`), no TOML.
- **Repo-local plugins (`.entire/plugins/`) never auto-run** — require an explicit
  allow-list entry, mirroring the `external_agents` opt-in posture.
- **Capability model**: `http`/`exec`/`fs`/`net` plus the mutating-hook grants
  `commit_msg`/`pre_push` are denied unless granted per-plugin; ungranted calls
  fail loud.

## Decision gates

- **Latency**: empty observer-hook dispatch ≈ **1.5µs/op** (benchmark) — far under
  the "few ms" gate. The no-plugins-enabled fast path does zero filesystem work.
- **Binary size**: release binary ≈ 69 MB; gopher-lua (pure Go) adds ~1–2 MB — not
  a blocker.

## Trust model

Opt-in at every step: inert until allow-listed, repo-local never auto-runs,
privileged APIs capability-gated, `install` only places files, and
`ENTIRE_PLUGINS_DISABLED=1` is a process-wide kill switch. The allow-list +
capability grants are the trust boundary. See
`docs/architecture/plugins-lua.md`.

## Phases

- [x] **Phase 1 — foundation**: gopher-lua runtime, sandboxed LState (curated
  stdlib, escape hatches stripped, context timeout), `plugin.json` manifest,
  loader/registry (user + repo dirs), `settings.plugins` allow-list with
  capability grants, observer hook-bus POC, unit tests.
- [x] **Phase 2 — observer hook bus**: wired at `DispatchLifecycleEvent`,
  post-`SaveStep` (`checkpoint_saved`), `PostCommit`, and `PrePush`; event→Lua
  table payloads; `entire.on`/`entire.log`/`entire.kv`, read-only accessors;
  capability-gated stubs for `http`/`exec`/`fs`/`net`.
- [x] **Phase 3 — capabilities + hardening**: real, capability-gated `http`/`exec`/
  `fs` (repo-confined) with timeouts + size caps; kill switch; configurable hook
  timeout; documented trust boundary.
- [x] **Phase 4 — mutating hooks**: `prepare_commit_msg` trailer (after the
  built-in `Entire-Checkpoint` trailer) and `pre_push` veto → non-zero exit,
  both capability-gated, with documented ordering vs OPF and multi-plugin
  semantics.
- [x] **Phase 5 — plugin commands**: `entire.command{...}` + `entire.print`;
  resolution order built-in > Lua command > `entire-<name>` binary.
- [x] **Phase 6 — distribution**: `entire plugin install <git-url|path>`
  (`--ref` pin), `entire plugin update`, list/remove surfacing Lua + binary
  plugins.
- [x] **Phase 7 — docs + SDK + examples**: `docs/architecture/plugins-lua.md`,
  CLAUDE.md/AGENTS.md + security-and-privacy.md updates, `examples/plugins/entire.lua`
  LSP stub, and two example plugins.

## Test plan

- [x] `CGO_ENABLED=0 go test ./cmd/entire/cli/ ./cmd/entire/cli/settings/... ./cmd/entire/cli/plugins/...` (2677 pass) and `./cmd/entire/cli/strategy/...` (814 pass)
- [x] `CGO_ENABLED=0 go build ./...` and release binary build
- [x] `golangci-lint` clean; dependency license check passes
- [x] End-to-end: install an example Lua plugin, allow-list it, run its command

Never merge — long-lived branch.


pfleidi:
Thanks for putting this together. Could you add a bit more context on what’s driving the need for a new plugin system and what problem it’s meant to solve?

This feels like a bigger design decision than we should make in a PR alone, especially since it could affect more than just this implementation. I think it would be useful to discuss it with a wider group first, maybe through an RFD that lays out the use cases, requirements, and alternatives.

We also already have plugin mechanisms for [external agents](https://docs.entire.io/agents/external-agent-plugins/build-agent-plugin) and CLI commands, both using the git/kubectl-style approach of finding plugins through $PATH. Depending on the goal here, for example, responding to events, we might be able to build on the existing CLI plugin system by adding hooks [similar to the external agent ones](https://docs.entire.io/agents/external-agent-plugins/lifecycle#phase-3-agent-session-hooks-firing), rather than introducing a third mechanism.

My main concern is making sure we agree on the problem and the overall approach before getting too far into the implementation.

jdx:
my 2 cents (feel free to disregard—this isn't my project after all):

1. I think it would be a big mistake to use more than 1 type of plugin in a project. I think 1 plugin should be able to do anything. I didn't do this with mise and I regret it now—though in that case I wasn't really faced with this as an option, it just evolved this way.
2. I think distribution of plugins is a problem with the current model since it requires end-users to have a working go compiler. Probably not a _total_ dealbreaker but I think for something to be super mainstream it would be a perennial issue and I don't see a good way to ever solve that with the current setup aside from baking go into entire-cli (which is a terrible idea). I don't think this is as much of a problem for k8s where the end-user environment is more controlled.

suhaanthayyil:
Closing, not merging — per the PR description itself ("Never merge — long-lived branch") and a fresh mergeability check against current \`origin/main\`.

The core embedded Lua plugin runtime (\`cmd/entire/cli/plugins/\`: sandbox, capability model, hook bus, manifest, command dispatch — Phases 1–5) is still genuinely distinct: main has no \`gopher-lua\` dependency and no \`cmd/entire/cli/plugins/\` package, so that part is not superseded and merges cleanly (auto-merge only touched \`CLAUDE.md\`, \`cli/lifecycle.go\`, \`settings/settings.go\`, \`strategy/manual_commit_*.go\`, \`main.go\`, go.mod/go.sum with zero conflicts).

The one real conflict is \`cmd/entire/cli/plugin_group.go\` — and it's a design collision, not a mechanical one. Since this branch was opened, main independently shipped a full binary-plugin distribution system (\`9e40149dd\` "Add remote plugin install, git-synced index discovery, and dependency resolution", plus semver-aware upgrade and checksum verification follow-ups) that redefines \`entire plugin install/list/remove\` around a git-synced index, semver tag resolution, checksummed release downloads, and dependency graphs — none of which existed when this PR wrote its own (simpler) git-clone-based install path for Lua plugins. Both sides now want to own the same \`entire plugin install <arg>\` entrypoint for a bare git URL (Lua-plugin repo vs. binary-release repo), with incompatible flag sets (\`--ref\` here vs. \`--pin\`/\`--yes\`/\`--no-deps\`/\`--allow-unverified\` on main). Reconciling that needs a product decision (how to disambiguate/probe, whether Lua plugins register in the index too), not a patch.

Recommend splitting this work: (1) land \`cmd/entire/cli/plugins/\` + the hook-bus wiring as its own PR against current main — it has no command-surface conflicts today; (2) design \`entire plugin install\`'s Lua-vs-binary disambiguation as a deliberate follow-up that accounts for main's now-shipped index/remote-install system, rather than reviving this branch's Phase 6 distribution code.

PR #1753, opened 2026-07-15
cli: scaffold internal-gated `entire ci` command group

dvydra:
https://entire.io/gh/entireio/cli/trails/858

## Why

We want an `entire ci` command group for managing CI-provider integrations (Buildkite first), but the surface is internal-only and not ready to ship to customers. This PR lays the foundation: the group skeleton plus the machinery that keeps it out of the public release, so later PRs can stack verbs onto it without re-litigating the gating.

No Buildkite verbs yet — just the group, the build-tag gating, a goreleaser internal build variant, and tests.

## What

New package `cmd/entire/cli/ci/`, split across build tags:

- `doc.go` (no build tag) — the single package doc comment, so exactly one exists under either build configuration (keeps revive/staticcheck happy regardless of which register file compiles).
- `register_internal.go` (`//go:build internal`) — exports `func Register(root *cobra.Command)`, which mounts a hidden `ci` group carrying the shared control-plane persistent flags (mirroring `newRepoCmd` / `addControlPlaneFlags`) with a placeholder `buildkite` subgroup that prints "coming soon".
- `register_stub.go` (`//go:build !internal`) — a no-op `Register(_ *cobra.Command)`. Both files export the same `Register` symbol, so `root.go` compiles unchanged under either tag; only the internal build actually mounts the group.
- `register_internal_test.go` / `register_stub_test.go` — build-tag-aware tests asserting the group is present (and `buildkite` mounted, `ci` Hidden) under `internal`, and absent otherwise.

`addControlPlaneFlags` is replicated in the `ci` package rather than imported: `cli` imports `ci` (root.go calls `ci.Register`), so importing `cli` back would be an import cycle. This mirrors the pattern the task asked for while respecting the dependency direction.

`root.go` calls `ci.Register(cmd)` near the other control-plane group registrations.

**goreleaser:** added an `entire-internal` build id that builds `./cmd/entire` with `flags: [-tags=internal]`, reusing the public build's env/matrix/ldflags verbatim. The public `entire` build stays untagged. To keep the internal binary from ever shipping publicly, the `archives` block now allow-lists only the public build ids (`entire`, `git-remote-entire`). This is also load-bearing: without the filter, both `entire` and `entire-internal` produce a binary named `entire` and would collide in the same archive. Net effect: the internal build is produced under `dist/` (proving the tagged build compiles in the release toolchain) but is excluded from the release archives, scoop manifest, and Homebrew cask.

I chose a live-but-unpublished build id (rather than a commented-out stanza) so the tagged build is exercised by the release pipeline, while the archives allow-list makes accidental public distribution structurally impossible.

Note: in the public build, `entire ci` falls through to kubectl-style external-command resolution (`entire-ci` on PATH) since no built-in `ci` group exists — expected behavior, unchanged by this PR.

## Verification

- `go build ./...` — succeeds; `entire --help` lists no built-in `ci` group (confirmed alongside `auth`/`repo`/`grant`/`org`/`project`).
- `go build -tags internal ./...` — succeeds; `entire ci buildkite` prints "coming soon" and `entire ci --help` shows the `buildkite` subgroup.
- `go test ./cmd/entire/cli/ci/...` (stub test: group absent) and `go test -tags internal ./cmd/entire/cli/ci/...` (internal test: group present) both pass.
- `golangci-lint run` clean under both the default build and `--build-tags internal`; `gofmt` clean.
- goreleaser YAML validated: build ids `[entire, entire-internal, git-remote-entire]`, `entire-internal` flags `[-tags=internal]`, archives allow-list `[entire, git-remote-entire]`. (`goreleaser check` fails only on a pre-existing Pro-only field under the OSS binary, unrelated to this change.)

https://claude.ai/code/session_01CHPhjYDtZx71THMtVwc83E

---

> [!NOTE]
> **Medium Risk**
> Release packaging changes (archive allow-list plus a second `entire` build id) could mis-ship binaries if misconfigured; customer-facing CLI surface stays gated by the `internal` tag.
> 
> **Overview**
> Introduces an **internal-only** `entire ci` command tree (Buildkite placeholder first) without shipping it in public releases.
> 
> A new `cmd/entire/cli/ci` package uses the **`internal` build tag**: `register_internal.go` mounts a hidden `ci` group with control-plane persistent flags and a `buildkite` stub that prints “coming soon”; `register_stub.go` is a no-op `Register` for public builds. **`root.go`** calls `ci.Register(cmd)` next to other control-plane groups. Tag-aware tests assert the group is present under `internal` and absent otherwise.
> 
> **GoReleaser** adds an `entire-internal` build (`-tags=internal`) that still outputs `entire` under `dist/` for internal use, and **archives** now allow-list only `entire` and `git-remote-entire` so the internal binary never collides with or leaks into public tarballs, Scoop, or Homebrew.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1339cb8a737b5fe6b06bf7c6a6649b74cd076d1c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -32,6 +32,36 @@ builds:
       # builds carry no stamp and use the package default ("true" = visible).
       - -X github.com/entireio/cli/cmd/entire/cli/experimental.Visible={{ if .Prerelease }}true{{ else }}false{{ end }}
 
+  # entire-internal builds the same ./cmd/entire main with `-tags=internal`, so
+  # the internal-only command groups (e.g. `entire ci`, gated in
+  # cmd/entire/cli/ci) are compiled in. It reuses the public build's env/matrix
+  # and ldflags verbatim and only adds the tag. This variant is intentionally
+  # NOT published: the archives below allow-list only the public build ids, so
+  # the internal binary is produced under dist/ for internal distribution but
+  # never lands in the release archives, scoop manifest, or Homebrew cask. The
+  # public `entire` build stays untagged.
+  - id: entire-internal
+    main: ./cmd/entire
+    binary: entire
+    flags:
+      - -tags=internal
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - darwin
+      - linux
+      - windows
+    goarch:
+      - amd64
+      - arm64
+    ldflags:
+      - -s -w
+      - -X github.com/entireio/cli/cmd/entire/cli/versioninfo.Version={{.Version}}
+      - -X github.com/entireio/cli/cmd/entire/cli/versioninfo.Commit={{.ShortCommit}}
+      - -X github.com/entireio/cli/cmd/entire/cli/telemetry.PostHogAPIKey={{.Env.POSTHOG_API_KEY}}
+      - -X github.com/entireio/cli/cmd/entire/cli/telemetry.PostHogEndpoint={{.Env.POSTHOG_ENDPOINT}}
+      - -X github.com/entireio/cli/cmd/entire/cli/experimental.Visible={{ if .Prerelease }}true{{ else }}false{{ end }}
+
   # git-remote-entire is the git remote helper for entire:// URLs (see
   # cmd/git-remote-entire). A small, dedicated binary shipped alongside
   # entire so `git clone entire://…` resolves it on PATH — no telemetry,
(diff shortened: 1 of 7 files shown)


dvydra:
Superseded. entire-ci now ships as a standalone plugin — https://github.com/entireio/entire-ci — discovered by the CLI as `entire ci …` on PATH, rather than a build-tagged command group compiled into the entire binary. Closing in favor of that approach (see ADR 20260717-entire-ci-pluggable-client).

PR #1755, opened 2026-07-15
cli: entire ci buildkite enroll/list/update/remove

dvydra:
https://entire.io/gh/entireio/cli/trails/860

Stacks on #1754 (→ #1753). Review/merge those first.

## Why

The `entire ci buildkite` subgroup shipped as a "coming soon" placeholder in
the scaffold PR. This turns it into the real operator surface for managing a
repo's Buildkite CI-webhook subscriptions through the core-mediated
management API (the ci-webhooks routes shipped in entire-core under COR-836).

The whole `ci` group stays gated behind the `internal` build tag, so none of
this appears in released `entire` binaries.

## What

Four verbs under `entire ci buildkite`, each resolving `<repo>` to a ULID via
the inherited `ResolveNativeRepo` (native `<project>/<repo>` path or a raw repo
ULID; GitHub mirrors rejected):

- `enroll <repo> [pipeline] --bk-org <slug> --bk-cluster-id <id> [--ref-filter <glob>] [--events create,update] [--display-name <s>]`
  → `POST /api/v1/repos/{repoId}/ci-webhooks` with `provider: buildkite`. The
  optional `[pipeline]` positional is omitted to let the server default it to
  the repo name. No token field — the Buildkite credential is operator-seeded,
  org-level. `--bk-org` is required; `--bk-cluster-id` is enforced by the
  server (a later PR resolves it via cluster discovery, so it is not marked
  required client-side yet). Prints `Enrolled` on a fresh 201 and `Re-enrolled`
  on an idempotent 200.
- `list <repo>` → `GET …/ci-webhooks`, rendered as an aligned table (id,
  provider, bk_org, pipeline, enabled, ref_filter, events) or raw `--json`.
- `update <repo> <subscriptionId> [--enabled=bool] [--ref-filter <glob>] [--events …] [--display-name <s>]`
  → `PATCH …/ci-webhooks/{id}`, sending **only** the flags the user set (cobra
  `Changed` checks; unset = unchanged). An explicit empty `--events` clears the
  list. A no-op update (no field flags) is rejected before any request.
- `remove <repo> <subscriptionId> [--teardown]` → `DELETE …/ci-webhooks/{id}?teardown=<bool>`
  (204). `--teardown` also revokes the enrollment identity; the Buildkite
  pipeline is always kept.

Because the `cli` package imports this one (`root.go` calls `ci.Register`), the
verbs can't reuse `cli`'s `runCore*` helpers — the package carries a small,
[...]

PR #1754, opened 2026-07-15
cli: native repo path→ULID resolver for ci commands

dvydra:
> Stacks on #1753 (the internal-gated `entire ci` scaffold). Review/merge that first; this PR's base is `dv/ci-cmd-scaffold`, not `main`.

## Why

The `entire ci buildkite` verbs (enroll/list/update/remove/watch, landing in later PRs) address a repo by its entiredb repo ULID, but users type a native path like `widgets/web`. No `path→ULID` resolver exists in `entireio/cli` yet. This PR adds it — the one piece those verbs all depend on — as a standalone, tested unit so the verb PRs stay small.

It mirrors the entiredb reference (`cmd/entire-ci/cli/repo_resolve.go` `resolveRepoRef`) and the M2 "Repo resolution" note in the Buildkite CI CLI spec.

## What

`ResolveNativeRepo(ctx, c, ref) (repoULID string, err error)` in the internal-gated `cmd/entire/cli/ci` package. Accepted shapes:

- **raw 26-char ULID** — round-tripped unchanged after a Crockford-base32 shape check, zero network calls;
- **`<project>/<repo>`** and **`/et/<project>/<repo>`** — resolved via the control plane: project name → project ULID, then repo name within that project → repo ULID;
- **`gh/…` / `github.com/…` mirror paths** (with or without a leading slash) — rejected with an actionable error: *"Buildkite CI supports native /et/ repos only, not GitHub mirrors"*. The ci-webhooks backend enrolls native, org-owned `/et/` repos only; a mirror authorizes via GitHub perms, not an Entire grant, so a pipeline could never clone it.

Unknown project and unknown repo return distinct, friendly errors that point at `entire project list` / `entire repo list`.

Placement/wiring notes:
- The file is **untagged** (`resolve.go`), so it compiles under both `internal` and public builds and normal `go test ./...` covers it. It is referenced only by the `//go:build internal` verb code in later PRs.
- The ULID-shape check and 404→friendly-error mapping are **replicated** from the `cli` package rather than imported: `cli` imports `ci` (root.go calls `ci.Register`), so importing back would form a cycle — the same rationale the scaffold gives for replicating `addControlPlaneFlags`.

[...]

PR #1756, opened 2026-07-15
cli: entire ci buildkite watch (live build status)

dvydra:
https://entire.io/gh/entireio/cli/trails/861

## Why

The `entire ci buildkite` group can enroll/list/update/remove CI-webhook
subscriptions, but there was no way to *watch* a build progress from the
terminal. This adds a `gh run watch`-style live viewer so an operator can push
and follow the resulting Buildkite build without leaving the CLI or holding a
Buildkite token.

Stacked on the `ci buildkite` verbs: **#1755 → #1754 → #1753**.

## What

`entire ci buildkite watch <repo> [build]` (internal-gated, like the rest of the
`ci` group).

- Flags: `--pipeline <name>` (filter to one pipeline slug), `--interval <dur>`
  (poll cadence, default 2s). With no `[build]`, watches the newest in-flight
  build; pass a number to pin one. Exits when the build reaches a terminal
  state. On a TTY the step tree redraws in place; piped output logs each state
  transition only.

**Ported (provider-neutral, unchanged in spirit) from entiredb's
`cmd/entire-ci/cli/buildkite_watch.go`** into the untagged
`buildkite_watch.go` so a normal `go test` run covers it:
- the `stepView` / `buildView` view model and the `buildSource` seam,
- `runWatch` (poll loop, newest-active pick, terminal-state exit via
  `isTerminalBuildState`),
- `renderBuild` (pure, ANSI in-place redraw on a TTY; signature-gated frames
  otherwise) plus its render helpers.

The two reference data sources were **not** ported — `bkBuildSource` (personal
Buildkite token) and `entireStoreSource` (leaf ops-token) are both invalid under
a user login JWT.

**Written fresh:** `coreBuildSource`, backed by the core-mediated builds
endpoint `GET /api/v1/repos/{repoId}/ci-builds?commit=&pipeline=&limit=`
(entiredb#2741, branch `dv/ci-builds-endpoint`). Because that endpoint is **not
yet in this repo's generated `coreapi` client**, the call is a hand-rolled typed
GET: a new `coreapi.Client.GetJSON` escape hatch reuses the client's own
`serverURL`, `SecuritySource` (bearer), and cross-jurisdiction HTTP transport —
no second auth path — and decodes into local structs (`ciBuildDTO` / `ciJobDTO`)
that mirror B3's response DTO field for field. `<repo>` resolves to a ULID via
the inherited `ResolveNativeRepo`.

> **Follow-up (dependency):** once entiredb#2741 merges and this repo's `coreapi`
> spec is regenerated to include `ci-builds`, delete `GetJSON` and the local
> DTOs and switch `coreBuildSource.fetch` to the generated client method. Both
> sites carry a comment saying so.

`watch` has no commit input, so the endpoint's `commit` filter is left unset
(all commits); the query sends `pipeline` (when set) and `limit`.

## Verification

- `go build ./...` and `go build -tags internal ./...` both succeed.
- `go test` and `go test -tags internal -race` pass for
  `./cmd/entire/cli/ci/...` and `./internal/coreapi/...`:
  - untagged: `renderBuild` pure tests + a fake `buildSource` driving `runWatch`
    to a terminal state (frame-per-transition, stops at terminal, picks newest
    active, empty state);
  - internal: `coreBuildSource` DTO mapping + query/path assertions against an
    `httptest` server, the full `watch` verb driven to a terminal build, the
    no-active-builds message, and bad-build-number rejection;
  - coreapi: `GetJSON` success decode + bearer/query, and non-2xx →
    `*ErrorModelStatusCode` surfacing the RFC 7807 detail via `APIError`.
- `golangci-lint run` clean under **both** the default tags and
  `--build-tags internal` for `./cmd/entire/cli/ci/...` (CI's default-tag lint
  doesn't cover the internal files); `gofmt` clean.
- Smoke: `entire ci buildkite watch --help` renders (built with `-tags internal`).

Not verifiable locally: the live poll loop against a real entire-core with the
ci-builds endpoint deployed (endpoint PR not yet merged). Exercised end-to-end
against an `httptest` fake of B3's response shape.

---

> [!NOTE]
> **Medium Risk**
> New control-plane polling path and hand-rolled `GetJSON` until the OpenAPI spec catches up; behavior is well-tested but depends on the not-yet-merged ci-builds endpoint shape.
> 
> **Overview**
> Adds **`entire ci buildkite watch <repo> [build]`** (internal-gated like other `ci buildkite` verbs) so operators can follow Buildkite progress from the terminal without a Buildkite API token—status comes from the logged-in user’s core **`GET /repos/{repoId}/ci-builds`** projection.
> 
> The **poll loop and renderer** live in untagged `buildkite_watch.go` (`buildSource` seam, TTY in-place redraw vs transition-only output when piped, default to newest in-flight build, `--pipeline` / `--interval`). **`coreBuildSource`** maps core DTOs to that view (drops waiter jobs, filters active builds).
> 
> Because **ci-builds is not in the generated client yet**, **`coreapi.Client.GetJSON`** is added as a temporary escape hatch reusing the same bearer auth and transport as generated calls, with local DTO structs until spec regen.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8bc7ddca153fa757745b3765868b981b4f2ff703. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/ci/buildkite_internal.go
+++ b/cmd/entire/cli/ci/buildkite_internal.go
@@ -34,6 +34,7 @@ func newBuildkiteCmd() *cobra.Command {
 	cmd.AddCommand(newBuildkiteListCmd())
 	cmd.AddCommand(newBuildkiteUpdateCmd())
 	cmd.AddCommand(newBuildkiteRemoveCmd())
+	cmd.AddCommand(newBuildkiteWatchCmd())
 	return cmd
 }
 
--- a/cmd/entire/cli/ci/buildkite_watch.go
+++ b/cmd/entire/cli/ci/buildkite_watch.go
@@ -1,0 +1,264 @@
+package ci
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"strconv"
+	"strings"
+	"text/tabwriter"
+	"time"
+)
+
+// This file is intentionally untagged (no `internal` build constraint): it is
+// the provider-neutral watch renderer, ported from entiredb's
+// cmd/entire-ci/cli/buildkite_watch.go. Keeping it out of the build tag lets the
+// normal `go test` run cover renderBuild and runWatch. The command constructor
+// and the core-backed data source that consume it live in
+// buildkite_watch_internal.go, gated on `internal` alongside the other verbs.
+
+// errNoActiveBuilds is returned by runWatch when no build number was given and
+// the source has nothing in flight to watch.
+var errNoActiveBuilds = errors.New("no active builds")
+
+// These Buildkite build/step states recur across the three state-classifier
+// funcs below (terminal, color, glyph); named so each literal has a single
+// source. Single-use states (running, scheduled, …) stay inline.
+const (
+	statePassed   = "passed"
+	stateFailed   = "failed"
+	stateCanceled = "canceled"
+	stateBlocked  = "blocked"
+)
+
+// --- neutral view model: the data-source seam ---
+//
+// stepView / buildView are deliberately provider-agnostic. The watcher's
+// renderer consumes only these, so the data source (here: the core-mediated
+// builds endpoint) can be swapped without touching the renderer.
+
+type stepView struct {
+	Label       string
+	Key         string
+	State       string
+	Exit        *int
+	Started     time.Time
+	Finished    time.Time
+	HasStarted  bool
+	HasFinished bool
+}
+
+type buildView struct {
+	Number      int
+	State       string
+	Branch      string
+	Commit      string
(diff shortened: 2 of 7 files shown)


dvydra:
Superseded. entire-ci now ships as a standalone plugin — https://github.com/entireio/entire-ci — discovered by the CLI as `entire ci …` on PATH, rather than a build-tagged command group compiled into the entire binary. Closing in favor of that approach (see ADR 20260717-entire-ci-pluggable-client).

PR #1787, opened 2026-07-16
fix(dispatch): fix --local auth for API-key users without loading user settings

alishakawaguchi:
https://entire.io/gh/entireio/cli/trails/884

## Problem

A user reported that `entire dispatch --local` fails at the "generate dispatch text" step even though they're logged in:

```
❯ entire dispatch --local --since 7d --voice neutral
generate dispatch text: claude CLI error (kind=auth): Not logged in · Please run /login
```

They authenticate with an **Anthropic API key (API billing)**, and plain `claude -p "…"` works fine in their shell — so the auth works when claude is invoked directly, but not the way dispatch invokes it.

## Root cause

`cmd/entire/cli/agent/claudecode/generate.go` spawns the claude CLI with `--setting-sources ""`, which tells claude to load **none** of its setting sources (user, project, local).

API-billing users configure auth in their **user settings** (`~/.claude/settings.json`) — either an `apiKeyHelper` command or the `ANTHROPIC_API_KEY` approval stored in `customApiKeyResponses`. Loading no sources drops that config, so claude reports "Not logged in" even though the same user's plain `claude -p` works (it loads user settings).

This is **not** an env-stripping issue — `StripGitEnv` only removes `GIT_*` and preserves `ANTHROPIC_API_KEY`; a raw exported env-var key survives the flag. Subscription (claude.ai) users are unaffected because their token lives in `~/.claude/.credentials.json`, which isn't a setting source. The bug only hit settings-based API auth.

## Fix

Switch `--setting-sources ""` → `--setting-sources user`: load the user's settings (restoring auth) while still excluding project and local settings. The subprocess already runs in `os.TempDir()`, so project/local isolation is preserved regardless — the flag was effectively only suppressing user settings, i.e. only breaking auth, with no isolation benefit.

## Verification

Reproduced and verified **end-to-end through the real `entire dispatch --local` command**, using an isolated HOME configured like an API-billing user (`apiKeyHelper` + dummy key):

| Binary | `entire dispatch --local` output |
|---|---|
| before fix (`""`) | `claude CLI error (kind=auth): Not logged in · Please run /login` (reporter's exact error) |
| after fix (`user`) | `claude CLI error (kind=auth): Invalid API key · Fix external API key` (dummy key **used** — a real key now authenticates) |

- Added regression test `TestGenerateText_LoadsUserSettingsForAuth` asserting the subprocess args carry `--setting-sources user`.
- `mise run check` (fmt + lint + test:ci, incl. integration + e2e canary) passes.

https://claude.ai/code/session_012QYA1kFFwDTbR8cZQQkb8N

---

> [!NOTE]
> **Low Risk**
> Small, targeted change to Claude subprocess flags for auth; repo isolation behavior is unchanged and covered by a unit test.
> 
> **Overview**
> Fixes **`entire dispatch --local`** (and any path using `ClaudeCodeAgent.GenerateText`) failing with **"Not logged in"** for users who authenticate via **API billing** in `~/.claude/settings.json`, while plain `claude -p` still works.
> 
> `GenerateText` now passes **`--setting-sources user`** instead of an empty value, so the subprocess loads user settings (including `apiKeyHelper` / API key approval) while **project and local** settings stay excluded. Isolation from the repo is unchanged because the command still runs with **`cmd.Dir = os.TempDir()`**.
> 
> Adds **`TestGenerateText_LoadsUserSettingsForAuth`** to assert the spawned CLI args include `--setting-sources user`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f457d27a98afa958f228560203252e3f3ac6d7ee. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/agent/claudecode/claude_test.go
+++ b/cmd/entire/cli/agent/claudecode/claude_test.go
@@ -2,8 +2,11 @@ package claudecode
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
+	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


peyton-alt (approved):

ebarti:
Hey @alishakawaguchi thank you for the super fast help here. 

I tried the nightly build `0.8.43-nightly.202607170653.e32c16b05` but unfortunately it is still not working. 

I am using Vertex AI authentication via GOOGLE_APPLICATION_CREDENTIALS

```bash
  API provider:     Google Vertex AI
```

```bash
❯ entire login
Logging in to: https://us.auth.entire.io

Press Enter to open in browser...

Waiting for sign-in... ✓ Login complete.
❯ entire dispatch
generate dispatch text: claude CLI error (kind=auth): Not logged in · Please run /login
❯ entire version
Entire CLI 0.8.43-nightly.202607170653.e32c16b05
Go version: go1.26.4
OS/Arch: darwin/arm64
generate dispatch text: claude CLI error (kind=auth): Not logged in · Please run /login
❯ claude --bare -p "test"
Working! I'm Claude Code, ready to help you in the `agent-runtime-kit` repository. What would you like to do?
```

PR #1788, opened 2026-07-16
fix(dispatch): make `dispatch --local` surface recent merged work (ENT-1188)

alishakawaguchi:
https://entire.io/gh/entireio/cli/trails/885

## Summary

`entire dispatch --local` returned an empty / near-empty dispatch for everyone, while the server-side dispatch on entire.io worked. Fixes ENT-1188.

## Root cause

Two independent bugs in `cmd/entire/cli/dispatch/mode_local.go`:

1. **Branch scoping.** `branchLocalRevRange` always used `<default>..HEAD`, which is empty on an up-to-date default branch. So the reachable-trailer set was empty and only checkpoints whose `summary.Branch` equalled the current branch survived — rare, since work happens on feature branches merged into `main`.
2. **Stale local store (dominant).** Enumeration used `store.List`, which only sees checkpoints present in the local checkout. Recent checkpoints live in the sharded `git-refs` backend on the checkpoint remote and are never fetched locally — a plain `git pull` only syncs `refs/heads/*`, not `refs/entire/checkpoints/*`. So recent windows came up empty even though the work was reachable from `HEAD` by commit trailer.

## Fix

- `branchLocalRevRange` returns `HEAD` when the current branch **is** the default branch (nothing to exclude); the `base..HEAD` exclusion still applies to feature branches.
- Enumeration now also includes checkpoints reachable by **commit trailer** in the window, using the commit subject as the bullet (always available locally from `git log`) and preferring the richer local session summary when the checkpoint is already present locally. Windowed by commit ("landed on branch") time. No per-checkpoint network fetch, so it stays fast (~20s, dominated by the LLM step).

## Testing

- New regression tests: `TestLocalMode_ImplicitCurrentBranchOnDefaultBranchIncludesMergedWork` and `TestLocalMode_IncludesReachableCheckpointMissingFromLocalStore`.
- Existing `TestLocalMode_ImplicitCurrentBranchExcludesDefaultBranchHistory` still passes (feature-branch scoping preserved).
- Verified end-to-end on a real checkout: `dispatch --local --since 7d` now produces a full dispatch of the week's merged work (was empty before).

## Notes

- The `--all-branches` path still enumerates via `store.List` (local only) — reasonable follow-up if broader coverage is wanted.

---

> [!NOTE]
> **Medium Risk**
> Changes local dispatch candidate selection and git log parsing; behavior is well covered by new tests but affects what users see in `dispatch --local` windows.
> 
> **Overview**
> Fixes **ENT-1188**: `entire dispatch --local` often returned empty output while server dispatch worked.
> 
> **Default-branch scoping:** When the current branch is the default branch, `branchLocalRevRange` now scans **`HEAD`** instead of `<default>..HEAD`, so merged feature work reachable from default-branch HEAD is included. Feature branches still use the `base..HEAD` exclusion.
> 
> **Enumeration beyond local store:** After `store.List`, a second pass adds checkpoint IDs from **commit trailers** in the time window that are not present locally. Bullets use commit subjects from `git log` (no per-checkpoint network fetch), with optional richer local session titles when a checkpoint exists locally. Trailer reachability is keyed by **in-window commit time** (`reachableCheckpointIDsInRange` now returns `map[string]time.Time` with `--until` and explicit `[since, until)` filtering, keeping in-window times when a later out-of-window commit also references the same ID).
> 
> **Refactor:** Local summary title logic moved to `readLocalSummaryTitle`.
> 
> Regression tests cover default-branch merged work, trailer-only checkpoints missing from the local store, and in-window commit time when newer out-of-window commits exist.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cf509f7e48e3733fe1bb56b80c73ae24b77e478f. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/dispatch/mode_local.go
+++ b/cmd/entire/cli/dispatch/mode_local.go
@@ -168,7 +168,7 @@ func enumerateRepoCandidates(ctx context.Context, repoRoot string, opts Options,
 		if len(branches) > 0 {
 			currentBranch = branches[0]
 		}
-		reachableCheckpointIDs, err = reachableCheckpointIDsInRange(ctx, repoRoot, branchLocalRevRange(ctx, repoRoot, currentBranch), since)
+		reachableCheckpointIDs, err = reachableCheckpointIDsInRange(ctx, repoRoot, branchLocalRevRange(ctx, repoRoot, currentBranch), since, until)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


alishakawaguchi:
bugbot run

alishakawaguchi:
bugbot run

peyton-alt (approved):

PR #1797, opened 2026-07-17
Add local agent selection to dispatch

peyton-alt:
https://entire.io/gh/entireio/cli/trails/894

## Summary

- add a local-only `--agent <name>` override to `entire dispatch`
- when Local is selected without an override, reuse the configured summary provider or discover enabled text-generation agents and show the interactive picker when a choice is needed
- persist first-run picker selections while keeping explicit `--agent` overrides scoped to one invocation
- inject the selected text generator and model into local dispatch instead of hard-coding Claude Code
- preflight local time-window and repository inputs before agent selection and before the Bubble Tea loading UI starts
- preserve actionable errors from explicitly selected external agents while scan-all discovery remains best-effort
- keep cloud dispatch unchanged on the Entire server-side generator

## Behavior

Interactive first-run flow:

1. Run `entire dispatch`.
2. Choose **Local**.
3. If no summary provider is configured, Entire inspects installed/enabled summary-capable agents.
4. With multiple candidates in an interactive terminal, Entire shows **Choose a summary provider**.
5. The selected provider is saved locally and used for dispatch generation.

If only one compatible provider is available, it is selected automatically. An explicit command such as:

```console
entire dispatch --local --agent codex --since 7d
```

bypasses the picker, uses the selected CLI default model, and does not modify the configured provider.

## Why

Local dispatch previously always invoked Claude Code, even when another supported agent was configured or authenticated. That could surface Claude-specific authentication failures and made dispatch inconsistent with the existing summary-provider flow used by Explain.

Explicit agent selection is intentionally limited to `--local`. Cloud dispatch never inspects local agent configuration.

## CI change

None. An earlier revision added a `test-windows-external-discovery` merge gate; it was removed per review — per-PR Windows test coverage is a repo-wide decision that should cover all Windows-gated tests (which currently skip silently on the ubuntu jobs), not just the two added here. Tracked in #1843. The Windows-gated `.bat` discovery tests stay in this PR and self-skip off Windows; they'll run once #1843 lands.

## Validation

- `mise run check` passed after merging current `main`
- Go lint: 0 issues
- Vogon canary: 59/59 passed
- Roger-Roger external-agent canary: 4/4 passed
- remote Windows named-discovery check passed
- manual local generation exercised both Claude Code and Codex

## Risk and follow-up

The main behavioral change is limited to local text generation and provider selection. Cloud dispatch is untouched.

Manual testing still produced empty activity for the tested repository. Activity/checkpoint enumeration is a separate issue and is intentionally unchanged by this PR.


.github/workflows/ci.yml
```diff
@@ -67,16 +67,28 @@ jobs:
           E2E_CHECKPOINT_STORE: ${{ matrix.checkpoint_store }}
         run: mise run test:e2e:canary
 
+  test-windows-external-discovery:
```
computermode: Do you mind sharing more context on why we needed a windows-specific runner here for one test? Introducing this new Windows runner as a required merge gate is a big CI change and I don't see anything in the PR description that mentions why we needed it specifically.

.github/workflows/ci.yml
```diff
@@ -67,16 +67,28 @@ jobs:
...
+      - name: Named external-agent discovery
+        run: go test ./cmd/entire/cli/agent/external -run '^TestDiscoverAndRegisterNamedAlways_RegistersBatOnWindows$' -count=1
+
   test:
     runs-on: ubuntu-latest
     needs:
       - test-core
       - test-integration
       - test-canary
+      - test-windows-external-discovery
```
computermode: This test skips if the OS isn't windows, but this says `runs-on: ubuntu-latest` so I think it would no-op 🤔

.github/workflows/ci.yml
```diff
@@ -67,16 +67,28 @@ jobs:
...
 
+  test-windows-external-discovery:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+      - uses: jdx/mise-action@dad1bfd3df957f44999b559dd69dc1671cb4e9ea # v4
+      - name: Named external-agent discovery
+        run: go test ./cmd/entire/cli/agent/external -run '^TestDiscoverAndRegisterNamedAlways_RegistersBatOnWindows$' -count=1
```
computermode: It looks like we actually have a lot of tests that skip if the OS isn't windows, how are we running those tests today? I wonder if this new test can be run the same way?

computermode (commented):
Thanks for taking a look at this! I left a few questions about the new windows-specific runner/job.

.github/workflows/ci.yml
peyton-alt: The `ubuntu-latest` on this line belongs to the aggregate `test` job, which only checks the other jobs' results — the test itself runs in the `test-windows-external-discovery` job above, on `windows-latest`, where the skip guard passes. This line just makes that job's result part of the merge gate.

.github/workflows/ci.yml
peyton-alt: They aren't run in CI today — the only Windows workflow is `e2e-windows.yml`, which is manual-dispatch E2E with a real agent, not per-PR CI. Per-PR, all Windows-gated unit tests silently skip since every job runs ubuntu. So there's no existing mechanism to piggyback on; this job is the first per-PR Windows coverage. In f71c9828d I widened its `-run` pattern so the pre-existing scan-all `.bat` test in the same package runs too. Going wider than that isn't possible without rework: the rest of the package mocks agents as `#!/bin/sh` scripts, which Windows can't execute.

.github/workflows/ci.yml
peyton-alt: Fair — the rationale should have been in the description (adding it now). The named-discovery path this PR adds resolves binaries via `exec.LookPath`, whose `.bat`/`.cmd` resolution depends on Windows `PATHEXT` — behavior the ubuntu jobs can't exercise (the scan-all flow uses `filepath.Glob`, a different code path). Since it guards the exact user-facing behavior this PR ships (`dispatch --agent <external>` on Windows), I made it a gate rather than informational. In f71c9828d I slimmed it — `actions/setup-go` instead of mise (the job only needs a Go toolchain, ~2–3 min total) — and documented the scoping in the workflow. Happy to drop it from the required set if you'd rather it not gate merges.

cmd/entire/cli/agent/external/discovery.go
```diff
@@ -43,6 +44,24 @@ func DiscoverAndRegisterAlways(ctx context.Context) {
...
+	if _, err := agent.Get(name); err == nil {
+		return
+	}
+
+	binPath, err := exec.LookPath(binaryPrefix + string(name))
+	if err != nil {
+		return
+	}
+	registerExternalAgent(ctx, binPath, name)
+}
```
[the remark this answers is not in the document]
peyton-alt: Addressed in 20bb2f2b6 — `discoverAndRegisterNamed` now wraps the caller context with `discoveryTimeout` (10s), same bound as the scan-all flow.

.github/workflows/ci.yml
computermode: I think adding windows-specific CI jobs so we aren't silently skipping them all is absolutely needed but outside the scope of this PR. Adding a new CI merge gate block to run this new test (or now a slightly widened group of tests as of the latest changes) doesn't make sense to me. I think we can remove that entirely for now and document in an issue as a followup to pick up that we need to have a specific windows runner for all of these windows tests, not just the new one(s) that were added.

computermode (approved):

PR #1801, opened 2026-07-17
fix(dispatch): surface merged work on feature branches with no own commits (ENT-1201)

alishakawaguchi:
https://entire.io/gh/entireio/cli/trails/898

`dispatch --local` returned an empty dispatch when run from a feature-branch
worktree that has no commits of its own 

The local engine scopes a non-default branch to <default>..HEAD to isolate the
branch's own work. When the branch has no unique commits that range is empty,
so the reachable-checkpoint-trailer set is empty and every merged checkpoint is
dropped — "no activity" reported while HEAD is sitting on a full window of
merged work.

ENT-1188 already recognized this empty-range problem but patched it only for the
literal default branch, matched by name. The real predicate is "is there any
branch-specific history to isolate?", so detect the empty range directly and
fall back to summarizing everything reachable from HEAD — the same reasoning
ENT-1188 applied to the default branch, generalized. Feature branches with their
own commits keep the scoped <default>..HEAD view unchanged.

A git failure while measuring the range is treated as non-empty, so the
exclusion is preserved on error (never widens scope).

Note: the default/cloud `entire dispatch` emptiness for entireio/cli is a
separate backend `dispatches/generate` bug (the backend holds the checkpoints —
its activity view reports them — but the dispatch generator returns 0); the CLI
sends a correct request. That half is out of scope for this repo.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01XL2teZnsLbEw2xPSy2f9Rs
Entire-Checkpoint: 01KXRRKGDK6VHPYCVE28QW1R8R

---

> [!NOTE]
> **Low Risk**
> Localized change to local dispatch rev-range selection with conservative git-error handling and a dedicated regression test; no auth, data, or API surface changes.
> 
> **Overview**
> **`dispatch --local`** no longer returns an empty summary when run from a feature-branch worktree that shares HEAD with the default branch and has no commits of its own (typical git-worktree / `marvin wt` setup).
> 
> `branchLocalRevRange` still scopes non-default branches to `<default>..HEAD` to isolate branch-only work, but it now checks whether that range is empty via **`revRangeIsEmpty`** (`git rev-list --count`). When the count is zero, it falls back to **`HEAD`** so merged checkpoint trailers on the current tip are included—the same behavior ENT-1188 applied only when the current branch matched the default by name. Feature branches that actually have unique commits keep the scoped range unchanged.
> 
> If measuring the range fails, the code treats the range as non-empty so scope is never widened on error. A regression test covers the fresh worktree-branch scenario (ENT-1201).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f70b18a82b294ffee609d4ed53164070f22d2885. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/dispatch/mode_local.go
+++ b/cmd/entire/cli/dispatch/mode_local.go
@@ -376,15 +376,39 @@ func reachableCheckpointIDsInRange(ctx context.Context, repoRoot, revRange strin
 // and leave the dispatch effectively empty. In that case we summarize
 // everything reachable from HEAD in the window, matching the server-side
 // dispatch. The base..HEAD exclusion only applies to feature branches.
+//
+// The same "nothing to exclude" reasoning holds for any branch whose base..HEAD
+// range is empty — a feature-branch worktree freshly created off the up-to-date
+// default branch (the standard git-worktree workflow), or a branch whose work
+// has already merged into the base. There the exclusion would again drop all
+// reachable in-window activity and report "no activity" despite HEAD clearly
+// carrying merged work (ENT-1201). Detecting the empty range directly, rather
+// than matching the default branch by name, covers both cases.
 func branchLocalRevRange(ctx context.Context, repoRoot, currentBranch string) string {
+	const headRev = "HEAD"
 	base := defaultBranchRef(ctx, repoRoot)
 	if base == "" {
-		return "HEAD"
+		return headRev
 	}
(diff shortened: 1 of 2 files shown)


alishakawaguchi:
bugbot run

alishakawaguchi:
bugbot run

PR #1806, opened 2026-07-20
fix(claude-code): target current tool matchers for subagent hooks

gtrrz-victor:
## Problem

Our Claude Code `PreToolUse`/`PostToolUse` hooks are installed under the tool matchers `Task` and `TodoWrite`. On current Claude Code neither matches anything, so a class of checkpoints **silently never fires**:

- `pre-task` + `post-task` (matcher `Task`) → no per-subagent checkpoints
- `post-todo` (matcher `TodoWrite`) → no incremental checkpoints during subagent runs

A bad matcher in Claude Code is a silent no-op — no error, no warning, no changelog entry — which is why this went unnoticed.

**Not** data loss for the common single-worktree flow: the `Stop` hook (match-all matcher, unaffected) still checkpoints the whole worktree, so subagent edits are captured — just attributed to the main turn, and without subagent/incremental granularity or metadata.

### Source of truth (per docs)

The docs don't phrase this as a "deprecated matcher" — the underlying facts are:

- **Subagent tool is `Agent`**, not `Task` (there was never a `Task` tool): [tools-reference.md](https://code.claude.com/docs/en/tools-reference.md) (`Agent` entry), [hooks.md](https://code.claude.com/docs/en/hooks.md) (matcher tool-name list).
- **`TodoWrite` disabled by default as of v2.1.142** in favor of `TaskCreate`/`TaskGet`/`TaskList`/`TaskUpdate`: [tools-reference.md](https://code.claude.com/docs/en/tools-reference.md) (`TodoWrite` entry, `min-version: 2.1.142`).
- **Matcher evaluation**: a matcher of only letters/digits/`_`/`-`/spaces/`,`/`|` is an exact string (or `|`-list), not a regex — so `Task` does **not** match `TaskCreate`; `TaskCreate|TaskUpdate` matches either exactly: [hooks.md](https://code.claude.com/docs/en/hooks.md) (matcher patterns).

> Note: there is no dedicated Claude Code changelog/GH-issue "deprecation" entry for these — the authoritative statements are the tools-reference version notes linked above.

## Fix

- Generate matchers `Agent` (pre-task/post-task) and `TaskCreate|TaskUpdate` (post-todo). Fresh installs are correct.
- **No in-place migration.** `entire enable` deliberately does **not** rewrite existing hook config as a side effect. Configs written by older versions (stale `Task`/`TodoWrite` entries) are repaired by re-running with **`entire enable --force`**, which already strips and reinstalls Entire hooks.
- Tests: positive-spec guard pinning matchers to Claude Code's current tool names; a `--force` reinstall test proving stale matchers are stripped and reinstalled. Integration assertions updated.
- Architecture doc updated (points stale-config users at `--force`).

## Scope note (see comment)

This fix repairs the Claude Code matchers only. Whether the subagent-checkpoint **granularity** is worth keeping at all is deferred — see the follow-up comment: the product consumes session-level data only, so this output is currently unconsumed, and the machinery is shared across Claude Code / Cursor / Factory Droid / Copilot CLI. Merging keeps behavior consistent without committing to keep the granularity.

## Known follow-up (out of scope)

`post-todo`'s handler still parses the legacy `TodoWrite` `tool_input` schema for its checkpoint **message label**. Under `TaskCreate`/`TaskUpdate` the checkpoint is still created (file-change detection is independent) but the label falls back to `Checkpoint #N`.

## Test plan

- `go test ./cmd/entire/cli/agent/claudecode/` ✅
- `go test -tags integration -run TestSetupClaudeHooks ./cmd/entire/cli/integration_test/` ✅
- `golangci-lint` (v2.11.3) clean on touched packages ✅

### DIFF
--- a/cmd/entire/cli/agent/claudecode/hooks.go
+++ b/cmd/entire/cli/agent/claudecode/hooks.go
@@ -36,17 +36,13 @@ const (
 // ,/| as exact strings, not a regex). See:
 //   - https://code.claude.com/docs/en/tools-reference.md (Agent, TodoWrite entries)
 //   - https://code.claude.com/docs/en/hooks.md (matcher evaluation rules)
+//
+// Configs written by older CLI versions used the outdated matchers "Task" and
+// "TodoWrite", where the hooks silently never fired. Those are not rewritten in
+// place on a normal `entire enable`; run with --force to strip and reinstall.
 const (
 	subagentToolMatcher = "Agent"
 	taskToolMatcher     = "TaskCreate|TaskUpdate"
-)
-
-// staleSubagentMatcher and staleTaskMatcher are the outdated matcher names that
-// older CLI versions wrote. `entire enable` migrates Entire hooks off these onto
-// the current matchers above (see migrateEntireToolMatcher).
-const (
-	staleSubagentMatcher = "Task"
-	staleTaskMatcher     = "TodoWrite"
 )
 
 // ClaudeSettingsFileName is the settings file used by Claude Code.
@@ -149,17 +145,6 @@ func (c *ClaudeCodeAgent) InstallHooks(ctx context.Context, localDev bool, force
 		postToolUse = removeEntireHooksFromMatchers(postToolUse)
 	}
 
-	// Migrate Entire hooks off the outdated tool matchers that older CLI versions
-	// wrote (Task->Agent rename; TodoWrite deprecated for TaskCreate/TaskUpdate).
-	// This runs on every enable, not just --force, so existing users self-heal by
-	// re-running `entire enable`. User-owned hooks under these matchers are kept
-	// (removal keys on the Entire command prefix, not the matcher name). Force mode
-	// already stripped all Entire hooks above, so these are no-ops there.
-	migrated := false
-	preToolUse, migrated = migrateEntireToolMatcher(preToolUse, staleSubagentMatcher, subagentToolMatcher, migrated)
-	postToolUse, migrated = migrateEntireToolMatcher(postToolUse, staleSubagentMatcher, subagentToolMatcher, migrated)
-	postToolUse, migrated = migrateEntireToolMatcher(postToolUse, staleTaskMatcher, taskToolMatcher, migrated)
-
 	// Define hook commands
 	var sessionStartCmd, sessionEndCmd, stopCmd, userPromptSubmitCmd, preTaskCmd, postTaskCmd, postTodoCmd string
 	if localDev {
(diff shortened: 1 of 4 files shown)


gtrrz-victor:
## Follow-up: evaluate whether subagent-checkpoint granularity is worth keeping

This PR repairs the Claude Code tool matchers so the subagent hooks (`pre-task`/`post-task`/`post-todo`) fire again. Before investing further in this subsystem, we should decide whether the **granularity it produces is worth keeping at all** — because right now nothing consumes it.

### What the investigation found
- **The product consumes session-level data only.** In `entire.io`, ingestion reads `full.jsonl`/`transcript.jsonl` for session index 0. Subagent transcripts (`agent-*.jsonl`) are **never read**. `is_task`/`tool_use_id` are parsed into `SessionInfo` once (`api/src/lib/github.ts:660-661`) then never queried or displayed. `subagent_type`, `task_description`, `is_incremental`, `incremental_sequence`, `agent_id`, `is_task_checkpoint` → **0 matches** anywhere in the product. No UI renders subagents.
- **Local consumers are dead or cosmetic.** The only feature that used subagent-boundary checkpoints was `rewind`, which is deprecated and slated for removal. What remains is `[Task]` labels in `checkpoint list`/`explain` — navigational garnish.
- **It's a shared, cross-agent subsystem.** The same machinery (`handleLifecycleSubagentStart/End` → `SaveTaskStep`) is used by **Claude Code, Cursor, Factory AI Droid, and Copilot CLI**. On Claude Code it was dead (this matcher bug); on the other three it fires today but produces the same inert task checkpoints.

### What we'd lose by removing it: nothing consumed
Code the subagent writes is still captured by the `Stop`/turn checkpoint (whole-worktree). One honest caveat: subagents in an **isolated worktree** aren't covered by `Stop` — but that protection doesn't exist today anyway (hooks were broken/unconsumed), so removal wouldn't regress vs current reality.

### Decision to make later
Keep this cross-agent subsystem (bet on a future subagent-level product feature — the dead `isTask` parse suggests it was once intended), or remove it (`pre-task`/`post-task`/`post-todo` hooks across all 4 agents + `SaveTaskStep`/`TaskStep` machinery + task-checkpoint store paths) to delete unconsumed complexity.

Merging this fix keeps behavior consistent across agents in the meantime; it does not commit us to keeping the granularity.

gtrrz-victor:
Tracked the keep-vs-remove decision as #1807 (proposal to remove the subagent-checkpoint machinery, since the product consumes session-level data only). This fix stands on its own; #1807 decides whether the granularity survives.

pjbgf (approved):

PR #1812, opened 2026-07-20
fix(explain): don't mask post-resolution failures as 'no checkpoint or commit found'

peyton-alt:
https://entire.io/gh/entireio/cli/trails/906

## Problem

`entire checkpoint explain --generate <id>` could resolve and display a checkpoint, generate its AI summary, fail to save it — and then report `no checkpoint or commit found matching "<id>"`.

`runExplainAuto` decides whether to fall back to git commit resolution with `errors.Is(err, checkpoint.ErrCheckpointNotFound)`. But that sentinel also leaks out of steps that run **after** the target successfully resolved (in the field: `failed to save summary: checkpoint not found` from a summary backfill against a backend missing the checkpoint — see #1811 for that underlying bug). The fallback then ran, failed to parse the ID as a git rev, and replaced the real error with the misleading composite message.

## Fix

The genuine "target matched nothing" outcome now returns a dedicated `explainTargetNotFoundError` — which still unwraps to `checkpoint.ErrCheckpointNotFound`, so existing `errors.Is` contracts are unchanged — and the commit fallback triggers only on that type. Any post-resolution failure surfaces verbatim.

Before/after on the original repro (branch without #1811):
```
- no checkpoint or commit found matching "2cb795c79075"
+ failed to save summary: checkpoint not found
```

## Verification

- TDD: `TestShouldFallBackToCommitResolution` pins the decision (genuine miss → fallback; sentinel-wrapping save failure → surface verbatim); `TestRunExplainAuto_PostResolutionErrorSurfacesVerbatim` guards it end-to-end
- All existing explain tests pass unchanged (including `TestRunExplainCheckpoint_NotFoundSentinels`, which pins the public `errors.Is` contract)
- Full unit suite green (8117 tests), `mise run fmt` + `mise run lint` clean
- Field-verified against the real repro in this repo

Independent of #1811 (not stacked); the two compose — with both, the save succeeds outright.

---

> [!NOTE]
> **Low Risk**
> Narrow CLI error-routing change in `explain`; behavior improves for `--generate` without touching auth, storage contracts, or checkpoint persistence.
> 
> **Overview**
> **`runExplainAuto` no longer treats every `checkpoint.ErrCheckpointNotFound` as “target not found.”** Commit resolution runs only when the checkpoint path returns a dedicated `explainTargetNotFoundError` (still unwraps to the public sentinel so existing `errors.Is` checks keep working).
> 
> When a positional target **did** resolve but a later step fails—e.g. `failed to save summary: checkpoint not found` during `--generate`—that error is returned as-is instead of falling through to git and ending up as `no checkpoint or commit found matching …`.
> 
> The “no committed or temp match” path in `runExplainCheckpointWithLookup` now returns `explainTargetNotFoundError` instead of a bare wrapped sentinel. Tests lock in `shouldFallBackToCommitResolution` and an end-to-end `runExplainAuto` case for post-resolution failures.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 070f1f6386828f3a814c9178f74446b785873f25. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/explain.go
+++ b/cmd/entire/cli/explain.go
@@ -452,8 +452,8 @@ func runExplain(ctx context.Context, w, errW io.Writer, sessionID, commitRef, ch
 	return runExplainBranchWithFilter(ctx, w, errW, noPager, sessionID)
 }
 
-// explainTargetNotFoundError reports that the positional target matched no
-// committed or temporary checkpoint. It unwraps to
+// explainTargetNotFoundError reports that the requested checkpoint target
+// matched no committed or temporary checkpoint. It unwraps to
 // checkpoint.ErrCheckpointNotFound so callers' errors.Is contracts keep
 // working, while runExplainAuto can distinguish this resolution miss from a
 // failure AFTER a successful match that happens to wrap the same sentinel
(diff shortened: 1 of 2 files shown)


peyton-alt:
Applied the review fix set in 5ef8773d6:

- **Errors name the resolved checkpoint**: the four post-resolution read/reload wraps include `fullCheckpointID`/`cpID`, so a bare-sentinel store failure no longer reads as `failed to read checkpoint: checkpoint not found` with no ID.
- **Trailer-path context**: `explain <commit>` whose Entire-Checkpoint trailer references a vanished checkpoint now reports `commit <sha> references checkpoint <id> via its Entire-Checkpoint trailer: ...` instead of naming an ID the user never typed (new test).
- **Temp-checkpoint swallows logged**: list failure and matched-but-unreadable shadow commit/tree now emit `logging.Debug` before reading as not-found.
- **Tests strengthened**: post-resolution e2e now runs both read-only and `--generate` legs (different call sites), asserts the checkpoint ID appears, and pins `shouldFallBackToCommitResolution(err) == false` directly.
- **Comment polish**: "requested checkpoint target" (the typed error is also produced on the `--checkpoint` flag path); refreshed the stale nilerr note.

Follow-ups filed separately (pre-existing, out of scope here): the JSON export path's own conflation (`explain_export.go` drops `commitErr`), and unlogged remote-fetch swallows in `matchCheckpointPrefixWithRemoteFallback`. Full suite green (8120 tests), fmt + lint clean.

gtrrz-victor (approved):

PR #1820, opened 2026-07-21
test(e2e): cover git-refs backend on push + assert no v1 branch

gtrrz-victor:
https://entire.io/gh/entireio/cli/trails/911

## Why this matters

Fresh `entire enable` now defaults to the **git-refs** backend — but every automatic E2E run pins **git-branch**. The backend real users get on day one had **zero E2E coverage**, and that gap is exactly how the v1-orphan bug (#1819) shipped unnoticed.

## What this does

1. **Standing guard** (`repo.go`): every git-refs E2E test now fails if an `entire/checkpoints/v1` branch appears. Under git-refs, nothing should ever create v1 — so this catches any future flow that regresses like #1819 did.
2. **Runs the default backend in CI** (`e2e.yml` + `e2e-checkpoint-store.yml`): a `claude-code` × `git-refs` leg on every push to `main`, in Slack notify. Agent and backend are fixed by the reusable workflow's defaults (caller passes nothing) — one agent, to bound cost.

## Verified
`go vet -tags e2e` (testutil + tests), `gofmt`, YAML parse, `actionlint` (only pre-existing findings). Real E2E not run locally (API cost); it runs in the new CI leg.

### DIFF
--- a/.github/workflows/e2e-checkpoint-store.yml
+++ b/.github/workflows/e2e-checkpoint-store.yml
@@ -26,6 +26,25 @@ on:
         options:
           - git-branch
           - git-refs
+  # Reusable entry point so e2e.yml can run a fixed agent/backend leg (e.g.
+  # claude-code against git-refs) as part of the on-push matrix. The `inputs`
+  # context is shared by workflow_dispatch and workflow_call, so the jobs below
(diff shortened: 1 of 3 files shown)


gtrrz-victor:
not needed..

PR #1825, opened 2026-07-21
feat(import): anchor imported checkpoints to the default branch head

peyton-alt:
https://entire.io/gh/entireio/cli/trails/916

## What

`entire import` now stamps a `commit_sha` field into every checkpoint it writes — the default branch's head at import time (origin tip preferred → local branch → HEAD → empty). It lands in both the session `metadata.json` and the root `CheckpointSummary` (`omitempty`, preserve-on-rewrite), giving the UI a join key to anchor imported sessions to a commit so repo pages and stats populate immediately after an import.

## Why

Imported history is commit-less: the normal commit↔checkpoint join is the `Entire-Checkpoint` trailer, which pre-existing commits don't have and never will (we don't rewrite history). The anchor is deliberately **a link, not attribution** — it says "imported at this point in time", not "this session built that commit". The sha lives *inside* the checkpoint metadata rather than in a ref name: checkpoint IDs are strictly 12-hex/ULID, and stuffing a 40-hex sha into `refs/entire/checkpoints/` would break ID validation and force a growing shared-ref list for the common all-anchors-same-sha case.

## How

- `WriteOptions.CommitSHA` → session `Metadata` + root `CheckpointSummary` (`commit_sha`, omitempty; root summary preserves a stamped value if a later write carries an empty one, mirroring the existing `Imported`/`HasReview` pattern)
- `agentimport.Options.LinkCommitSHA` threaded through `Run` → `writeTurn`; resolution stays out of the package
- `resolveImportLinkCommitSHA` in the `cli` package: origin's default-branch tip (the commit the server already knows) → local branch → HEAD → `""` (best-effort; an unresolvable repo imports unlinked, never fails)
- No new refs, no new commits, no flags, no backfill of previously-imported checkpoints

### Per-turn real-commit anchors

When a Claude Code transcript records the commits a turn made (`toolUseResult.gitOperation.commit`, kind `committed`), that turn's checkpoint anchors to the **real commit** instead of the head anchor: the recorded (short) sha is resolved, must be an ancestor of the default-branch tip (ancestry doubles as the reachability check — squash-merged/rebased-away commits fall through), and the last such commit of the turn wins. Everything else — older transcripts without the records, unresolvable or unreachable shas — falls back to the default-branch-head anchor above. The fallback's ancestor set is memoized lazily (one history walk per import run). Same `commit_sha` field either way; consumers see no schema difference.

## Testing

- TDD unit tests throughout, including a mutation-verified origin-over-local preference test (diverged `origin/main` vs local `main`) and coverage of the HEAD-fallback and empty-repo arms
- Manual E2E on both backends: git-branch (v1) and git-refs — `commit_sha` present and correct in both, idempotent re-import leaves it untouched
- **Prod smoke test** (fresh `/et` repo, real transcript, full enable→import→push flow): refs with `commit_sha` round-trip through the server intact. Control experiment confirmed server ingestion is currently trailer-driven: a normal checkpoint ingested into `/me/recap` in ~2 min while the imported refs (on the server, `commit_sha` and all) were never indexed.

### Notes for reviewers

- **Bounded ancestor walk:** the resolver's memoized ancestor set is capped by date (lookback + 60d slack) and by commit count (50k), walking newest-first (committer-time order) so early-stopping is sound. Addresses the inline review comment about unbounded time/memory on huge histories.
- **Mixed CLI versions:** `Metadata`/`CheckpointSummary` are not schema-versioned, so an *older* CLI that rewrites an existing checkpoint (OPF pre-push re-redaction, compact-marker update, multi-session summary rebuild) round-trips through structs without `commit_sha` and silently drops it. This is the same exposure every additive metadata field has had; the existing fence is the checkpoint policy ref (`checkpoint_min_version`). Practical blast radius for imported checkpoints is narrow — attribution never targets them and rewind/generate are blocked.

## Out of scope / follow-ups

- **entire.io ingestion + UI** (the other half): index checkpoint refs whose metadata carries `commit_sha` (kind `imported`) and join on that field; render imported sessions as *anchored* to the commit, not attributed to it. The smoke test above is the repro.
- Historical-commit matching (v2) would only change the stamped value, not the schema.
- The test-only `fsstore` backend doesn't map the new field (consistent with it already skipping `Imported`).

---

> [!NOTE]
> **Low Risk**
> Additive metadata and best-effort import resolution; no changes to normal checkpoint writes or git history rewriting.
> 
> **Overview**
> Adds an optional **`commit_sha`** field on checkpoint **`WriteOptions`**, session **`Metadata`**, and root **`CheckpointSummary`** so imported history can be tied to a git commit without an `Entire-Checkpoint` trailer.
> 
> **`entire import`** now resolves a link SHA at the CLI (`resolveImportLinkCommitSHA`: `origin/<default>` → local default branch → `HEAD` → `""`) and passes it through **`agentimport.Options.LinkCommitSHA`** into each imported turn. The anchor is display-only (“imported at this point”), not attribution; import still succeeds when resolution fails.
> 
> The persistent git store persists **`commit_sha`** on session metadata and the root summary, with **`omitempty`** and preserve-on-rewrite when a later write leaves it empty (same pattern as **`Imported`** / **`HasReview`**). Architecture docs describe the field; tests cover resolver preference, import stamping, and persistence.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d2df27256403602599db76a9539da573ddb7f0ba. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


computermode (dismissed):

cmd/entire/cli/agentimport/turn_anchor.go
```diff
@@ -0,0 +1,112 @@
...
+	fbCommit, err := r.repo.CommitObject(plumbing.NewHash(r.fallback))
+	if err != nil {
+		logging.Debug(ctx, "import: anchor ancestor walk unavailable, all turns fall back",
+			"fallback", r.fallback, "error", err.Error())
+		return ancestors
+	}
+	iter := object.NewCommitPreorderIter(fbCommit, nil, nil)
+	defer iter.Close()
+	if err := iter.ForEach(func(c *object.Commit) error {
+		ancestors[c.Hash] = struct{}{}
```
pjbgf, line 145: This could run for a very long time, and churn an unreasonable amount of memory in huge repositories. We may want to cap this by time or max amount of commits walked.

Not a show stopper though, mostly an UX issue.

peyton-alt: fixed in last commit

pjbgf (approved):

Issue #591, opened 2026-03-03
Orphaned sessions cause 3GB+ RAM usage and 2m44s commit times


alexfosterinvisible:
## Summary

Orphaned sessions accumulate silently and cause severe performance degradation:
- **RAM**: `entire hooks claude-code stop` processes consuming 1.7GB+ each (two concurrent = 3.4GB)
- **Commit speed**: `entire hooks git post-commit` takes **2 minutes 44 seconds** per commit (synchronous, blocking)
- **Session accumulation**: 76 orphaned sessions accumulated over 18 days without any warning

## Impact

This went unnoticed for weeks. Real cost:
- Every git commit blocked for ~3 minutes
- With 6 commits per session, that is ~16 minutes of dead time per session
- Two concurrent session stops consumed 3.4GB RAM on a dev laptop
- `entire status` showed 60+ "active" sessions that were actually dead

## Root Cause

Sessions from Claude Code subagents (spawned via Agent tool) are not properly cleaned up when the parent session ends. These zombie sessions remain in Entire's state and are re-processed on every commit hook and session stop.

The condensation pipeline loads full JSONL transcripts into memory. With large transcripts (up to 15MB each) and 76 sessions to process, memory balloons.

## Reproduction

1. Use Claude Code with frequent subagent spawning over several days
2. Run `entire status` - observe dozens of "active" sessions from days/weeks ago
3. Make a git commit - observe multi-minute blocking
4. Check RAM: `ps aux | grep entire` shows 1.5GB+ per process

## Workaround

```bash
# Fix stuck sessions (needs ACCESSIBLE=1 for non-interactive use)
ACCESSIBLE=1 entire doctor

# Clean orphaned sessions
entire clean --force

# Verify
entire status
```

This brought my session count from 60+ down to 12, and commit times back to normal.

## Suggestions

1. **Auto-cleanup**: Expire sessions that have been idle for >24h (or configurable threshold)
2. **Memory cap**: Stream/chunk transcript processing instead of loading full JSONL into memory
3. **Async post-commit**: The `post-commit` hook should run in the background. A 2m44s synchronous hook on every commit is not acceptable
4. **Visibility**: Warn when orphaned session count exceeds a threshold (e.g. >10)
5. **Subagent awareness**: Detect when a Claude Code session spawns subagents and auto-close their Entire sessions when the parent ends

## Environment

- Entire CLI: `entire version` (installed via Homebrew)
- OS: macOS 15.4 (Apple Silicon)
[...]

PR #1844, opened 2026-07-23
fix: SessionEnd hooks were cancelled on every exit — sessions never marked ended

peyton-alt:
https://entire.io/gh/entireio/cli/trails/931

## Why

Claude Code gives SessionEnd hooks a short budget (~1.5s) and never waits for them on exit. Our hook blew that budget two ways at once, so **every session exit since 2026-05-29 printed `SessionEnd hook failed: Hook cancelled` and no session was ever marked ended**:

- `.entire` logs: 117 successful `session-end` events Feb–May, zero after May 29
- Real sessions sat `idle` with `ended_at: null` for days (e.g. the "45 hours" / "177 hours" durations on entire.io session pages), then got silently deleted by the 7-day stale sweep without ever reporting an end
- Timing lined up with 0ef509db (2026-05-27), which added a `go build -o /dev/null` probe to `scripts/entire-dev` — that output path defeats go's up-to-date short-circuit, forcing a **full link on every hook invocation** (~1.3s warm) before `go run` linked again

## What

**Commit 1 — `scripts/entire-dev`: cache the local-dev binary.** Build into `.entire/tmp/entire-dev-bin` (ignored via the committed `.entire/.gitignore`) and run it. Reusing the same output path lets go skip the full relink when the source is unchanged (~0.5s vs ~1.3s — note go still rewrites the output file every build; VCS stamping defeats the full up-to-date skip); the fake session-end round trip drops 1.66s → 0.91s. Stop/turn-end hooks re-warm the cache during the session, so SessionEnd almost always hits the warm path. Fallback semantics unchanged: broken tree → PATH binary → exit 0.

**Commit 2 — lifecycle: split the SessionEnd handler along its natural line.** The ENDED mark (a locked state-file write carrying hook event metadata, milliseconds) stays inline; the eager condense (transcript reads + git tree building, seconds) moves to a detached `__condense_session` child spawned via `execx.SpawnDetached` — the existing `__refresh_trail_enablement` pattern — which survives the hook process being killed.

The condense itself is unchanged: same `CondenseAndMarkFullyCondensed`, same fail-open + retry semantics (PostCommit on the next commit; `entire doctor` detects stuck ended sessions), same backend routing (git-branch or git-refs primary + mirrors), run from the worktree root so config resolution matches the inline call. The #591 protection (eager condense on end) is preserved — just relocated. The sweep keeps its synchronous condense.

`ENTIRE_SESSION_END_SYNC` forces the old inline sequence; the integration TestMain sets it process-wide so existing session-end assertions stay deterministic, and a dedicated integration test exercises the real detached flow by overriding it back to empty.

**Commit 3 — review findings** (four specialized review agents + Copilot + Cursor Bugbot; all findings addressed):

- `CondenseAndMarkFullyCondensed` re-checks `Phase == ENDED` under the state lock — the detached child can lose a race with a same-ID resume, and condensing a revived session would wipe in-flight attribution and leave a sticky `FullyCondensed=true` (flagged independently by three reviewers + Bugbot; guarded by a new subtest that fails without the fix)
- SessionEnd teardown runs on `context.WithoutCancel` — the graceful half of hook cancellation is a SIGTERM, which cancels the root context; the ENDED mark must survive it (Copilot)
- `scripts/entire-dev` runs the cached binary without `exec` and falls back to the PATH binary on launch failure (126/127) — go rewrites the output in place on every build, and a failed `exec` previously aborted the shell before any fallback, breaking the never-block contract. Also: `go build -C` (an absolute package path resolves against the *caller's* module — latent pre-existing bug), per-cause fallback messages, and build stderr captured to `.entire/tmp/entire-dev-build.log` instead of discarded
- `execx.SpawnDetached` returns the start error and the hook logs the handoff + any spawn failure — a silently failed spawn previously left zero evidence a condense was ever requested
- Comment/doc corrections (the exited-session sweep cannot retry ENDED sessions; single-sourced the observed ~1.5s budget; corrected the binary-reuse mechanism) and test additions (launcher exit-status passthrough contract, rootErr sync fallback, seam records the worktree root, `__condense_session` in the policy-warning table)

## Verification

- 8,309 unit / 438 integration (incl. #591 regression test) / 63 e2e canary — all green, re-run after the review round
- Live end-to-end: real headless Claude session in the worktree → no warning, `phase: ended`, real `ended_at`, `fully_condensed: true` via the detached child — the first real session-end to complete in this repo since May 29
- Interactive end-to-end: fresh `claude` session, `/exit` → clean exit (no cancelled-hook warning), session marked ended at the actual exit moment; condense correctly deferred to commit-time because the session had touched files

## Notes / follow-ups (not in this PR)

- Ctrl-C double-tap force-quit still cancels SessionEnd hooks regardless of speed — upstream Claude Code regression (anthropics/claude-code#32712)
- Crash/kill paths (no hook fires) leave IDLE sessions the sweep is blind to (`OwnerExited` is ACTIVE-only) and the 7-day stale cleanup deletes them silently instead of finalizing — worth a separate issue
- Session "duration" on entire.io is wall-clock span (start → last activity); this fix makes `ended_at` real but doesn't change that metric's definition

---

> [!NOTE]
> **Medium Risk**
> Touches core session lifecycle and subprocess spawning on every agent exit; behavior is well-tested but incorrect detach or mark ordering could leave sessions un-ended or un-condensed until PostCommit/sweep retries.
> 
> **Overview**
> Fixes **SessionEnd hooks being cancelled** (~1.5s on Claude Code) so sessions were often never marked **ended**.
> 
> **`scripts/entire-dev`** now **incrementally builds** into `.entire/tmp/entire-dev-bin` and **execs** it instead of probe + `go run` (full link every invocation). Same fallback: broken tree → PATH `entire` → exit 0.
> 
> **Session end lifecycle** splits work: the hook **only marks the session ENDED inline** (`markSessionEnded`); the **eager condense** moves to a **detached** `entire __condense_session <id>` child (`execx.SpawnDetached`, same idea as `__refresh_trail_enablement`). Condense logic and fail-open retries are unchanged. **`ENTIRE_SESSION_END_SYNC`** (integration **TestMain**) or an **unresolvable worktree** still run the old synchronous **`endSessionNow`** path.
> 
> Adds hidden **`__condense_session`**, excludes it from checkpoint-policy warnings, and extends integration/unit tests for sync vs detached condense.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 540f83e091a2306f5559b69f69846042cc2b12db. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -132,8 +132,9 @@ Deprecated top-level commands (functional, print a cobra deprecation message):
 deprecation as `checkpoint rewind`).
 
 Hidden infrastructure commands: `hooks`, `trail`,
-`curl-bash-post-install`, `__send_analytics`, `mcp` (MCP stdio server for
-MCP-host agents).
+`curl-bash-post-install`, `__send_analytics`, `__refresh_trail_enablement`,
+`__condense_session` (detached session-end condense child), `mcp` (MCP stdio
+server for MCP-host agents).
 
 The `hideAsAlias(cmd, canonical)` helper in `cmd/entire/cli/aliascmd.go`
 marks a command Hidden and sets cobra's `Deprecated` field so the hint
--- a/cmd/entire/cli/checkpoint_policy_warning.go
+++ b/cmd/entire/cli/checkpoint_policy_warning.go
@@ -25,7 +25,7 @@ func ShouldCheckCheckpointPolicyWarning(cmd *cobra.Command) bool {
 
 func isCheckpointPolicyWarningExcludedCommand(name string) bool {
 	switch name {
-	case "hooks", "__send_analytics", "__refresh_trail_enablement", "curl-bash-post-install":
+	case "hooks", "__send_analytics", "__refresh_trail_enablement", "__condense_session", "curl-bash-post-install":
 		return true
 	default:
 		return false
--- a/cmd/entire/cli/checkpoint_policy_warning_test.go
+++ b/cmd/entire/cli/checkpoint_policy_warning_test.go
@@ -50,9 +50,13 @@ func TestShouldCheckCheckpointPolicyWarning(t *testing.T) {
 	refreshTrailEnablement := &cobra.Command{Use: "__refresh_trail_enablement", Hidden: true}
 	root.AddCommand(refreshTrailEnablement)
 
+	condenseSession := &cobra.Command{Use: "__condense_session", Hidden: true}
+	root.AddCommand(condenseSession)
+
 	require.True(t, ShouldCheckCheckpointPolicyWarning(visible))
 	require.True(t, ShouldCheckCheckpointPolicyWarning(hiddenAlias))
 	require.False(t, ShouldCheckCheckpointPolicyWarning(gitHook))
 	require.False(t, ShouldCheckCheckpointPolicyWarning(sendAnalytics))
 	require.False(t, ShouldCheckCheckpointPolicyWarning(refreshTrailEnablement))
+	require.False(t, ShouldCheckCheckpointPolicyWarning(condenseSession))
 }
--- a/cmd/entire/cli/execx/spawn_detached.go
+++ b/cmd/entire/cli/execx/spawn_detached.go
@@ -2,6 +2,7 @@ package execx
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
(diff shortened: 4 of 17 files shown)


gtrrz-victor:
Closing this PR in favour of -> https://github.com/entireio/cli/pull/1850

PR #1848, opened 2026-07-24
feat(enable): show progress while importing existing sessions

### LINKED ISSUE #1847: entire enable: no progress indicator while importing existing sessions (can look hung for an hour)
### Problem or use case

Running `entire enable` for the first time in a repo with a lot of existing agent-session history (in my case, a repo with many long Claude Code sessions) leaves the terminal completely silent for the whole session import. After the setup summary prints:

```
✓ Installed hooks
✓ Configured project
  .entire/settings.json
✓ Updated .entire/settings.json to block Vercel deploys of Entire metadata branch
```

…there is nothing but a bare cursor — in my case for roughly **an hour** — until the final `Imported N turn(s) from M session(s)` line appears all at once. There's no way to tell whether the CLI is working, hung, or waiting for input, and no signal that killing it would interrupt an import in progress.

Demo (synthetic fixture, ~15s import — the real-world case was ~1h):

before: silent import

### Where the time goes

The first-run import path (`maybeOfferSessionImport` → `runSelectedImports` → `agentimport.Run`) loops sessions × turns and only prints one summary line at the very end. Long sessions are slow to import (the per-turn token recompute documented in the `PERF` comment in `agent/claudecode/transcript.go` grows with session length), which is fine and deliberate — but the user gets zero feedback while it runs.

### Proposed solution

Show progress during the import, following the CLI's existing `startSpinner` convention:

- interactive terminals: the standard spinner with a live-updating message — `Importing Claude Code sessions... (session i/N · turn j/M)`
- non-TTY / piped / `ACCESSIBLE=1`: one plain line per session, so agents and screen readers get complete information without animation

after: live progress

I have this working and will open a PR implementing it right after filing this issue.

### Alternatives considered

- A simple static "Importing sessions, this may take a while..." line — better than silence, but with per-session/turn counts already available in the import loop, live counts are barely more code and answer "is it stuck?" definitively.
- Speeding up the import itself (the documented O(turns²) recompute) — out of scope here; the comment in `transcript.go` explains it's retained deliberately, and even a fast import benefits from visible progress on large corpora.

ecgang:
Closes entireio/cli#1847.

First-run `entire enable` imports existing agent sessions with zero output between the setup summary and the final `Imported N turn(s)` line — on a large corpus that's an hour of bare cursor with no way to tell working from hung. This wires live progress into that import, following the CLI's existing `startSpinner` conventions.

| before | after |
| -- | -- |
| before: silent import | after: live progress |

*Honesty note on the demo: both gifs use a synthetic fixture (6 sessions × 60 turns, \~15s import) so the behavior is visible in a short recording; the real-world case that motivated this was \~1h of silence.*

## What changed

* `agentimport`**: optional, UI-agnostic** `Progress` **reporter on** `Options` — `SessionStart` fires per session after turn-splitting; exactly one of `TurnWritten` / `TurnSkipped` fires per turn (skips cover already-imported turns and `--dry-run`), so callbacks always sum to `turnCount`. Nil reporter is byte-identical to before. No printing or logging inside the package.
* `startUpdatableSpinner` — the smallest variant of the existing spinner whose message can change while it runs; `startSpinner` now delegates to it. Same glyphs, tick rate, 250 ms initial-draw delay, and TTY gating. One deliberate change: animated frames draw `\r\033[K` instead of bare `\r`, so a shrinking message can't leave stale trailing characters (final stop lines are byte-identical).
* **One shared wiring helper (**`newImportProgressReporter`**) used by both import surfaces** — the `entire enable` first-run import and the standalone `entire import <agent>`:
  * interactive TTY: `Importing Claude Code sessions... (session i/N · turn j/M)` updating live
  * non-TTY / piped / `ACCESSIBLE=1`: one plain ANSI-free line per session (`Importing Claude Code session i/N (M turns)...`), keeping output complete for agents and screen readers per the repo's agent-safe CLI fallback guidance
* The final `Imported N turn(s) from M session(s)` summary line is unchanged, and the spinner is stopped before any error path prints.

## Tests

* `agentimport`: callback contract (counts, ordering, exact tuples), nil-reporter equivalence, skipped-turn and dry-run invariants (`written + skipped == turnCount`).
* `cli`: non-TTY progress lines (one per session, zero ESC bytes, unchanged summary line), spinner-variant behavior (fallback line, update-before-first-draw, stop semantics).
* Full `mise run check` (fmt, lint, unit + integration + Vogon canary) green; also manually verified against a real built binary: piped output, `ACCESSIBLE=1` under a PTY, and live-TTY spinner animation.

Happy to adjust the message format or UX approach if maintainers prefer something different.

### DIFF
--- a/cmd/entire/cli/agentimport/agentimport.go
+++ b/cmd/entire/cli/agentimport/agentimport.go
@@ -113,6 +113,10 @@ type Options struct {
 	// whose transcript records a resolvable commit that is an ancestor of this
 	// fallback anchors to that real commit instead (see turnAnchorResolver).
 	LinkCommitSHA string
+
+	// Progress, when non-nil, receives session/turn progress notifications
+	// as Run executes. Nil (the default) reports nothing.
+	Progress *Progress
 }
 
 // Result summarizes an import run.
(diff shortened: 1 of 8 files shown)


gtrrz-victor:
@copilot resolve the merge conflicts in this pull request

gtrrz-victor:
Update on next steps: rather than have you chase the `main` conflict, we're going to carry this the rest of the way on our side so it's not blocked on you.

I've opened #1868 in this repo starting from your branch — your commits and credit are preserved intact. I've resolved the `agentimport.go` conflict against `main` (kept both main's per-run git-identity resolution and your `sessionIndex` loop var) and folded in a couple of small review follow-ups (ANSI gating for legacy consoles + a comment fix). We'll take #1868 through review and merge, so there's nothing further you need to do here.

Thanks again — genuinely nice, well-tested contribution. 🙏 We'll close this one in favor of #1868.

gtrrz-victor:
Superseded by #1868 (carries your commits forward with the conflict resolved). Closing here — thanks again, @ecgang!

PR #1850, opened 2026-07-24
feat(claudecode): set a 60s timeout on the local-dev SessionEnd hook

gtrrz-victor:
https://entire.io/gh/entireio/cli/trails/936

## What

Set an explicit `timeout: 60` on the **local-dev** Claude Code SessionEnd hook.

## Why

The local-dev launcher (`scripts/entire-dev`) builds the CLI from source, so the
SessionEnd hook can exceed Claude Code's short default exit-grace (~1.5s observed)
and get cancelled before it runs — leaving sessions unmarked on exit.

Empirically confirmed: with no `timeout` field Claude Code SIGTERMs the SessionEnd
hook at ~1.53s on exit; with an explicit `timeout` it waits for the hook to finish.
An explicit budget makes Claude Code wait for the dev hook.

## Scope

- **Only** adds a timeout — no process detaching or other behavior changes.
- **Only** the SessionEnd hook.
- **Only** local-dev mode; production hooks and all other hooks keep Claude Code's default.

## Test plan

- New unit test: local-dev SessionEnd = 60s, local-dev Stop = 0 (scoping), production SessionEnd = 0.
- `go test ./cmd/entire/cli/agent/claudecode/` — pass.
- `golangci-lint` (v2.11.3) — 0 issues.

## Note

Fresh `entire enable --local-dev` writes the timeout. An existing dev install picks
it up only via `entire enable --local-dev --force` (hook presence is matched on
command, not timeout) — consistent with how other hook-config changes propagate.

## Dogfood configs

Ran `entire enable --force --local-dev` to update this repo's own hook configs to
the current generated output. `.claude/settings.json` now carries `timeout: 60` on
SessionEnd; the other agents' configs (`.codex`, `.cursor`, `.gemini`, `.opencode`,
`.pi`) were refreshed alongside (modernized tool-use matchers, ordering).

### DIFF
--- a/.pi/extensions/entire/index.ts
+++ b/.pi/extensions/entire/index.ts
@@ -11,7 +11,7 @@ import type { ExtensionAPI } from "@earendil-works/pi-coding-agent";
 import { execFile } from "node:child_process";
 
 export default function (pi: ExtensionAPI) {
-  const ENTIRE_CMD = "entire";
+  const ENTIRE_CMD = ""$(git rev-parse --show-toplevel)"/scripts/entire-dev";
   let pendingSkillEvents: Array<{ skill_name: string; invocation: string; timestamp: string }> = [];
 
   // fireHook pipes data to `entire hooks pi <hookName>` and resolves with the
--- a/.pi/extensions/entire/index.ts
+++ b/.pi/extensions/entire/index.ts
@@ -11,7 +11,7 @@ import type { ExtensionAPI } from "@earendil-works/pi-coding-agent";
 import { execFile } from "node:child_process";
 
 export default function (pi: ExtensionAPI) {
-  const ENTIRE_CMD = ""$(git rev-parse --show-toplevel)"/scripts/entire-dev";
+  const ENTIRE_CMD = '"$(git rev-parse --show-toplevel)"/scripts/entire-dev';
   let pendingSkillEvents: Array<{ skill_name: string; invocation: string; timestamp: string }> = [];
 
   // fireHook pipes data to `entire hooks pi <hookName>` and resolves with the
--- a/cmd/entire/cli/agent/claudecode/hooks_test.go
+++ b/cmd/entire/cli/agent/claudecode/hooks_test.go
@@ -467,6 +467,53 @@ func TestUninstallHooks_RemovesLocalDevHooks(t *testing.T) {
 	}
 }
 
+// TestInstallHooks_LocalDevSessionEndTimeout verifies the local-dev SessionEnd
+// hook carries an explicit timeout so Claude Code waits for it on exit instead
+// of cancelling it after its short default exit-grace. The timeout is scoped to
+// local-dev SessionEnd only: production and other local-dev hooks stay untimed.
+func TestInstallHooks_LocalDevSessionEndTimeout(t *testing.T) {
+	t.Run("local-dev SessionEnd gets the timeout", func(t *testing.T) {
+		tempDir := t.TempDir()
+		t.Chdir(tempDir)
+
+		agent := &ClaudeCodeAgent{}
+		if _, err := agent.InstallHooks(context.Background(), true, false); err != nil {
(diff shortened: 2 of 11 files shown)


cmd/entire/cli/agent/claudecode/hooks_test.go
```diff
@@ -467,6 +467,53 @@ func TestUninstallHooks_RemovesLocalDevHooks(t *testing.T) {
...
+			t.Fatal("expected a SessionEnd hook to be installed")
+		}
+		if got := settings.Hooks.SessionEnd[0].Hooks[0].Timeout; got != localDevSessionEndTimeoutSecs {
+			t.Errorf("local-dev SessionEnd timeout = %d, want %d", got, localDevSessionEndTimeoutSecs)
+		}
+
+		// Scoping: other local-dev hooks must not inherit the timeout.
+		if got := settings.Hooks.Stop[0].Hooks[0].Timeout; got != 0 {
+			t.Errorf("local-dev Stop timeout = %d, want 0 (timeout is SessionEnd-only)", got)
+		}
```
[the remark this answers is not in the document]
gtrrz-victor: Fixed in 33658509c — added a guard that fails with a clear message if the Stop hook isn't installed, before indexing into it.

cmd/entire/cli/agent/claudecode/hooks_test.go
```diff
@@ -467,6 +467,53 @@ func TestUninstallHooks_RemovesLocalDevHooks(t *testing.T) {
...
+			t.Fatalf("InstallHooks(localDev=false) error = %v", err)
+		}
+
+		settings := readClaudeSettings(t, tempDir)
+		if len(settings.Hooks.SessionEnd) == 0 || len(settings.Hooks.SessionEnd[0].Hooks) == 0 {
+			t.Fatal("expected a SessionEnd hook to be installed")
+		}
+		if got := settings.Hooks.SessionEnd[0].Hooks[0].Timeout; got != 0 {
+			t.Errorf("production SessionEnd timeout = %d, want 0 (dev-only)", got)
+		}
```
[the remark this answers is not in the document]
gtrrz-victor: Fixed in 33658509c — the production subtest now reads the raw settings.json and asserts it contains no `timeout` field at all, so an explicit `"timeout": 0` would fail too (the struct uses `omitempty`, so the field is genuinely omitted).

.pi/extensions/entire/index.ts
```diff
@@ -11,7 +11,7 @@ import type { ExtensionAPI } from "@earendil-works/pi-coding-agent";
 import { execFile } from "node:child_process";
 
 export default function (pi: ExtensionAPI) {
-  const ENTIRE_CMD = "entire";
+  const ENTIRE_CMD = ""$(git rev-parse --show-toplevel)"/scripts/entire-dev";
```
pjbgf: This looks like double quoting here. Shouldn't this be like below?

```suggestion
  const ENTIRE_CMD = "$(git rev-parse --show-toplevel)/scripts/entire-dev";
```

pjbgf: Probably worth having a small test to confirm the correct output takes place.

gtrrz-victor: Fixed in ed42608db. Root cause: the launcher value (`LocalDevHookScript`) carries its own shell quotes (`"$(git rev-parse --show-toplevel)"/scripts/entire-dev`), and the template wrapped the placeholder in **double** quotes — hence the doubled `""..."`.

Rather than drop the inner quotes (which would lose protection for repo paths containing spaces), I switched the template to **single** quotes, matching `.opencode/plugins/entire.ts`:

```ts
const ENTIRE_CMD = /Users/gtrrz-victor/wks/cli/exploring-entire/scripts/entire-dev;
```

Valid JS, and the shell still sees the space-protecting quotes via `sh -c`. Added a test that pins the exact generated line (the previous substring check passed on the broken output).

pjbgf (approved):

Issue #1852, opened 2026-07-24
Worktree commits get no Entire-Checkpoint trailer when the session started in another worktree of the same repo (findSessionsForWorktree exact-path match)


jlreyes:
### What happened?

Commits made in a **linked git worktree** never get an `Entire-Checkpoint` trailer when the agent session was started in a **different worktree of the same repo** (typically the main checkout). The decline is completely silent — `prepare-commit-msg` exits 0, and the reason is only visible at DEBUG level.

### Environment

- **Entire CLI**: 0.8.1 (also verified against current `main` source)
- **OS**: macOS (Darwin 27.0.0, arm64)
- **Strategy**: manual-commit
- **Agents**: Claude Code, Codex

### Repro (differential, same session, same minute)

One live Claude Code session, started with cwd = the main checkout (`$REPO`). Repo has a linked worktree at `$REPO/worktrees/probe`.

```console
# inside the linked worktree
$ cd $REPO/worktrees/probe
$ printf 'test: probe\n' > /tmp/msg
$ ENTIRE_LOG_LEVEL=DEBUG entire hooks git prepare-commit-msg /tmp/msg message
$ cat /tmp/msg
test: probe          # <-- no trailer

# from the main checkout, seconds later
$ cd $REPO
$ printf 'test: probe\n' > /tmp/msg2
$ ENTIRE_LOG_LEVEL=DEBUG entire hooks git prepare-commit-msg /tmp/msg2 message
$ cat /tmp/msg2
test: probe

Entire-Checkpoint: 729f8d7b7fe6   # <-- trailer added
```

Debug logs for the worktree run:

```json
{"msg":"prepare-commit-msg hook invoked","hook":"prepare-commit-msg","strategy":"manual-commit","source":"message"}
{"msg":"prepare-commit-msg: no active sessions","strategy":"manual-commit","source":"message"}
{"msg":"perf","op":"prepare-commit-msg","duration_ms":18,"steps.find_sessions_for_worktree_ms":4}
```

### Root cause

`SessionState.WorktreePath` is recorded once, from the agent's cwd at session start. At commit time, `findSessionsForWorktree` in `cmd/entire/cli/strategy/manual_commit_session.go` matches sessions by **exact string equality** against the committing worktree's root:

```go
for _, state := range allStates {
    if state.WorktreePath == worktreePath {
        matching = append(matching, state)
    }
}
```

[...]

PR #1859, opened 2026-07-26
 fix(strategy): surface ambiguous-worktree warning to the user with a runnable remedy.

ColeDriver:
## What & why

#1856 added `warnAmbiguousWorktreeSessions` for the #1852 case where live sessions span multiple worktrees, matching refuses to guess, and the commit is left unlinked. But it only logs to the internal `.entire/logs` file and prints no remedy — so the person running `git commit` still sees nothing. The "silent" part of #1852 is unchanged for them.

## Change

Keeps the structured `WARN` (diagnostics) and additionally writes a user-facing notice to `stderrWriter` — the same channel `warnIfAttributionDiverged` and `warnStaleEndedSessions` use — with a directly runnable remedy:

    entire: this commit was not linked to a checkpoint (live agent sessions span multiple worktrees: ../a, ../b).
      run 'entire session adopt <session-id> --from ../a --yes' to link one explicitly.

The command names a concrete session ID rather than the bare `--from <path>` form, because bare `--from` relies on adoption's auto-detect, which only considers sessions within `adoptRecentWindow` (12h) and fails outright when a worktree has multiple sessions; same-repo adoption also requires `--yes`. The named session is the most-recently-active candidate adoption will accept (mirrors `isAdoptableSourceSession`), so the printed command actually runs.

## Scope

Additive only — the warning **trigger** and the existing log line are unchanged; this just makes the already-detected case visible and actionable. If reviewers want the stderr line throttled (show-once / rate-limited), happy to add that.

## Tests

- New table-driven unit tests (`t.Parallel()`): `formatAmbiguousWorktreeNotice` (with/without an adoptable session, empty input) and `mostRecentlyAdoptableSession` (newest-adoptable, skips ended/ended-at/fully-condensed, nil when none).
- #1856's existing ambiguous-worktree tests still pass; full `strategy` suite green; `maintidx` + `golangci-lint` v2.11.3 and `gofmt` clean.

Refs #1852 · follow-up to #1856

### DIFF
--- a/cmd/entire/cli/strategy/manual_commit_ambiguous_notice_test.go
+++ b/cmd/entire/cli/strategy/manual_commit_ambiguous_notice_test.go
@@ -1,0 +1,93 @@
+package strategy
+
+import (
+	"testing"
+	"time"
+
+	"github.com/entireio/cli/cmd/entire/cli/session"
+	"github.com/stretchr/testify/require"
+)
+
+// TestFormatAmbiguousWorktreeNotice verifies the user-facing stderr message: it
+// always reports the ambiguity, and when an adoptable session is available it
+// appends a directly-runnable adopt command (concrete session ID so it doesn't
+// depend on adoption's 12h auto-detect, plus --yes for same-repo adoption).
+func TestFormatAmbiguousWorktreeNotice(t *testing.T) {
+	t.Parallel()
+
+	t.Run("with adoptable session includes runnable command", func(t *testing.T) {
+		t.Parallel()
+		primary := &SessionState{SessionID: "sess-1", WorktreePath: "/repo/a"}
+		got := formatAmbiguousWorktreeNotice([]string{"/repo/a", "/repo/b"}, primary)
+		require.Contains(t, got, "not linked to a checkpoint")
+		require.Contains(t, got, "/repo/a, /repo/b")
+		require.Contains(t, got, "entire session adopt sess-1 --from '/repo/a' --yes")
+	})
+
+	t.Run("worktree path with spaces is shell-quoted", func(t *testing.T) {
+		t.Parallel()
+		primary := &SessionState{SessionID: "sess-1", WorktreePath: "/Users/cole/my repo"}
+		got := formatAmbiguousWorktreeNotice([]string{"/Users/cole/my repo"}, primary)
+		require.Contains(t, got, "--from '/Users/cole/my repo' --yes")
+	})
+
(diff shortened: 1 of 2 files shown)


gtrrz-victor:
## ⚠️ The notice is swallowed by the hook — it never reaches the user

I tested this end-to-end against a `dev:publish` build of this branch and the warning does **not** surface on a real `git commit`, so the PR's stated goal ("the person running `git commit` still sees nothing" → make it visible) isn't met as written.

### Root cause

The notice is written to `os.Stderr` (via `stderrWriter`), but every commit-side hook the CLI installs invokes `entire` with `2>/dev/null` (`cmd/entire/cli/strategy/hooks.go:178-181`):

```
entire hooks git prepare-commit-msg "$1" "$2" 2>/dev/null || true
entire hooks git post-commit           2>/dev/null || true
entire hooks git post-rewrite "$1"     2>/dev/null || true
```

Git's hook script discards stderr, so `fmt.Fprint(stderrWriter, …)` output is thrown away before it reaches the terminal.

### Reproduction (proven both ways)

Setup: repo with two sibling worktrees `A` and `B`, a live (non-ended) session recorded in each, then commit from a third worktree that has no session of its own → ambiguous fallback → warn path.

- **Real `git commit`** (shipped hook, with `2>/dev/null`) → **no warning printed**, commit succeeds silently.
- **Same commit with `2>/dev/null` removed** from the installed `prepare-commit-msg` hook → warning prints exactly as designed:

  ```
  entire: this commit was not linked to a checkpoint (live agent sessions span multiple worktrees: …/A, …/B).
    to link one explicitly, run: entire session adopt sess-a --from '…/A' --yes
  ```

- Running the hook binary directly without the redirect (`entire hooks git prepare-commit-msg <msg> ""`) also prints it. So the formatting/remedy logic is correct — the delivery channel is the problem.

Note also: the trigger on a normal commit is **`prepare-commit-msg`** (`manual_commit_hooks.go:407`), not `post-commit` — `PostCommit` only calls `findSessionsForWorktree` after it finds an `Entire-Checkpoint` trailer (`:910-915` early-returns otherwise).

### On the "same channel" justification

The PR notes it uses "the same user-facing hook channel `warnIfAttributionDiverged` and `warnStaleEndedSessions` use." Those fire from `PrepareCommitMsg:420` and `PostCommit:1048` — both `2>/dev/null` hooks — so they're swallowed too. The precedent doesn't actually reach the user either; it's the same latent gap.

🤖 Reviewed with Claude Code

hiddeco:
Should this be closed in favor of #2531?

PR #1873, opened 2026-07-29
fix(pi): model subagent runs as subagents so checkpoints keep the right transcript

### LINKED ISSUE #1870: Pi: all subagent runs share session ID `session`, so checkpoints capture the wrong transcript
## Summary

Every Pi subagent run resolves to the same session ID — the literal string
`session` — because the ID is parsed from the transcript's basename, and Pi names
every subagent transcript `session.jsonl`.

Because that ID keys both transcript paths, every subagent run writes over the
previous one, and at commit time the checkpoint reads whichever run happened to
write last:

- **Staged copy:** `captureTranscript` writes `<repo>/.entire/tmp/pi/<id>.json`
  → one `session.json` for all runs.
- **Shadow-branch archive:** `SessionMetadataDirFromSessionID` is
  `.entire/<id>/` → one metadata dir for all runs.
- **Condensation prefers the staged copy** over the shadow archive
  (`strategy/manual_commit_condensation.go:1005-1018`), so the wrong content wins
  even where an archive exists.

Net effect: a checkpoint's stored transcript can be a *different subagent's*
conversation. There is no uncontaminated copy to fall back to, because both paths
collapse on the same ID.

Reproduces with one Pi session and one subagent. No concurrency required.

## What is *not* a problem

So this isn't mistaken for an attribution bug: the sessions attached to a commit
are the *correct* ones. A trail shows the Pi parent and the Pi subagent work, both
genuine participants, nothing unrelated. Sessions↔checkpoints is many-to-many and
the subagent session carries the same checkpoint set as its parent, so no
checkpoint linkage is lost.

The damage is to transcript *content*, not to which sessions or checkpoints appear.

## Observed

```
$ entire session list
Pi · gpt-5.6-sol     · session 019fab4c-c147-7b3a-8cc9-9ebd7224663b   turns 3
Pi · claude-sonnet-5 · session session                                turns 9
```

That second row is nine runs across three distinct subagents:

```
~/.pi/agent/sessions/<cwd-slug>/2026-07-29T00-36-01-991Z_019fab4c-…/
  71e7c5e8/run-0/session.jsonl   71e7c5e8/run-1/session.jsonl
  716671d6/run-0/session.jsonl   716671d6/run-1/session.jsonl   716671d6/run-2/session.jsonl
  d93a3451/run-0/session.jsonl   d93a3451/run-1/session.jsonl   d93a3451/run-2/session.jsonl
  d93a3451/run-3/session.jsonl
```

…and a single staged capture for all of them:

```
$ ls -l .entire/tmp/pi/
2075829  019fab4c-c147-7b3a-8cc9-9ebd7224663b.json   ← parent, fine
 197526  session.json                                 ← all 9 runs, last writer wins
```

Two commits were made from this session, at 11:01 and 11:28. `session.json` has an
mtime of 11:26 — so by the time the 11:28 commit condensed, the staged file held
the 11:26 run's transcript, not the run that produced the commit.

## Root cause

Pi writes parent and subagent transcripts in different shapes. The parent's ID is
its filename; a subagent's ID is its *directory* path, and the leaf is named by
role:

```
2026-07-29T00-36-01-991Z_019fab4c-….jsonl                          ← parent: <timestamp>_<uuid>.jsonl
2026-07-29T00-36-01-991Z_019fab4c-…/d93a3451/run-0/session.jsonl   ← subagent
```

Refs at `849421758`:

**The ID is derived from the basename.** `agent/pi/lifecycle.go:366-378`:

```go
base := filepath.Base(p)
base = strings.TrimSuffix(base, ".jsonl")
if i := strings.LastIndex(base, "_"); i >= 0 {
    return base[i+1:]
}
return base
```

`…/d93a3451/run-0/session.jsonl` → `session.jsonl` → `session` → no underscore →
returns `session`. The fallback behaves as documented; the problem is that
`filepath.Base` discards the three segments carrying the identity.

**It runs on every event, not just as a fallback.** `lifecycle.go:156-159` only
parses the path when `payload.SessionID` is empty — but
`agent/pi/entire_extension.ts` sends `session_file:` at lines 87, 97 and 120 and
never sends `session_id`. So the parse always runs. The comment at
`lifecycle.go:173` ("Pi emits before_agent_start with a fully-populated session
ID") doesn't hold for the extension the CLI ships.

**Nothing rejects the result.** The only gate is `validation.ValidateSessionID`
(`validation/validators.go:20`), called at the capture choke point
(`lifecycle.go:335`). It checks empty, leading dash, path separators, `.`/`..`,
`:`, and glob metacharacters — "is this safe in a path?", not "is this an
identity?". `session` is a safe path segment and passes. That validator is doing a
different job correctly and shouldn't be extended to cover identity.

**Both transcript paths key on the ID.** Staging at `lifecycle.go:346`:

```go
dst := filepath.Join(dir, sessionID+".json")
```

and the shadow archive via `paths.SessionMetadataDirFromSessionID`:

```go
return EntireMetadataDir + "/" + sessionID
```

**And the staged copy is preferred at read time.**
`strategy/manual_commit_condensation.go:1005-1018`, reached from
`extractSessionData` with `liveTranscriptPath = state.TranscriptPath`:

```go
if liveTranscriptPath != "" {
    ...
    if liveData, readErr := os.ReadFile(liveTranscriptPath); readErr == nil && len(liveData) > 0 {
        fullTranscript = string(liveData)
    }
}
if fullTranscript == "" {
    // Fall back to shadow branch copy
```

The documented rationale is sound in general — *"handles the case where SaveStep
was skipped (no code changes) but the transcript continued growing; the shadow
branch copy would be stale"* — i.e. the live file is assumed to be a newer version
of *the same* session. For Pi subagents that assumption inverts: the live file is a
*different subagent's* transcript, so the preference actively selects wrong content.

**Pi subagents never flow through the subagent mechanism.** `lifecycle.go:1088`
derives `SubagentTranscriptPath` from `event.SubagentID`, which is how Claude Code
keeps per-subagent transcripts distinct. Pi's lifecycle events
(`SessionStart`/`TurnStart`/`TurnEnd`/`session_shutdown`) never populate
`SubagentID`, so Pi subagent runs arrive as ordinary turns carrying a colliding
session ID rather than as subagent steps. That's the design-level gap underneath
the parse bug.

## Suggested fix

The identity is already in `payload.SessionFile` — parent UUID plus
`<subagent>/<run>` segments; only `filepath.Base` throws them away. Renaming the
staged capture alone is **not** sufficient, since the shadow archive keys on the
same ID.

Two coherent shapes:

1. **Model Pi subagents as subagents.** Populate `event.SubagentID` from the path
   segments so runs go through the existing `SubagentTranscriptPath` mechanism, and
   let the session ID roll up to the parent UUID. This matches Claude Code, keeps a
   trail's session list to the conversation that drove the work, and gets distinct
   per-run transcripts for free.
2. **Give each run its own session ID** (`019fab4c-…:d93a3451:run-0`). Simpler, but
   it adds session rows that duplicate the parent's checkpoint set.

(1) looks right given the mechanism already exists.

Worth doing alongside either: have `entire_extension.ts` send `session_id`
explicitly so the path parse stops being load-bearing, and add an identity check
distinct from the path-safety one.

## Why tests didn't catch it

`agent/pi/lifecycle_test.go:176` `TestExtractSessionIDFromPath`:

```go
"/tmp/2026-05-09T12-00-00-000Z_abc-123.jsonl": "abc-123",
"abc-123.jsonl":                              "abc-123",
"/tmp/no-underscore-here.jsonl":              "no-underscore-here",   // fallback asserted as correct
"/path/with/multiple_under_scores_id.jsonl":  "id",
```

Every case is a flat file in a sessions directory, and the third pins the fallback
as intended. The nested subagent shape isn't represented, so the model under test
is "Pi transcripts are flat `<timestamp>_<uuid>.jsonl`" — true for parents, false
for subagents.

## Scope

Observed within a single repo. Sessions appear repo-scoped and I have no evidence
of one `session` ID spanning repos — the only other checkout with Pi subagent runs
on disk had no Pi session registered, so that case is untested rather than ruled
out.

## Environment

- Entire CLI 0.9.0 (darwin/arm64, go1.26.4)
- Pi, parent model `gpt-5.6-sol`, subagent model `claude-sonnet-5`

gtrrz-victor:
https://entire.io/gh/entireio/cli/trails/953

Fixes #1870.

## Summary

Every Pi subagent run resolved to the same session ID — the literal string `session` — because the ID was parsed from the transcript's basename, and Pi names every subagent transcript `session.jsonl`. That one ID keyed both transcript paths (the staged capture `.entire/tmp/pi/<id>.json` and the shadow archive `.entire/<id>/`), so runs overwrote each other and a checkpoint condensed whichever run wrote last — storing a *different* subagent's conversation. Reproduces with a single Pi session and one subagent; no concurrency required.

This takes **direction 1** from the issue: model Pi subagents as subagents — roll the session ID up to the parent UUID, give each run a distinct identity, and route runs through the existing subagent mechanism so each gets its own transcript.

## What Changed

- **`agent/pi/lifecycle.go`** — new `parsePiSessionIdentity` (+ `parsePiSubagentPath`, `parsePiParentID`, `derivePiParentTranscript`) resolves the path shape:
  - parent `…/<ts>_<uuid>.jsonl` → session ID = `<uuid>`
  - subagent `…/<ts>_<uuid>/<sub>/run-N/session.jsonl` → session ID rolls up to the parent `<uuid>`, plus a distinct `<sub>-run-N` id.

  `ParseHookEvent` now routes subagent `before_agent_start`/`agent_end` to `SubagentStart`/`SubagentEnd` (parent turns still emit `TurnStart`/`TurnEnd`), and stages each run's transcript under its own identity. The parse validates the derived id against **both** consumers (`ValidateAgentID` for the dispatcher and the stricter `ValidateSessionID` for the capture key) so an exotic dir name can't parse-pass then silently fail to stage.
- **`agent/event.go`** — added `Event.SubagentTranscriptPath` for agents whose per-subagent transcripts can't be reconstructed from `SessionRef` + `SubagentID` (Pi's nested `<parent>/<sub>/run-N` layout).
- **`lifecycle.go`** — `handleLifecycleSubagentEnd` prefers `event.SubagentTranscriptPath`, falling back to the existing `agent-<id>.jsonl` convention. Claude Code / Cursor are unaffected (the new field is empty for them).

Net effect: one Pi session per conversation (no phantom `session · turns N` row), a distinct staged/stored transcript per run, and task checkpoints that roll up under the parent session with uncontaminated content.

## Tests

Written test-first, all confirmed **red** before the fix, green after:

- `TestExtractSessionIDFromPath_SubagentRollsUpToParent` — subagent paths roll to the parent UUID, never `"session"`.
- `TestParseHookEvent_PiSubagentRunsRouteAsSubagents` — subagent runs emit `SubagentStart`/`SubagentEnd`, share the parent `SessionID`, carry distinct `SubagentID`s.
- `TestParseHookEvent_PiSubagentCaptureDoesNotClobber` — parent + each run stage distinct, content-correct transcripts.
- `TestParseHookEvent_PiSubagentTranscriptSurvivesLaterRun` — one run's staged transcript isn't overwritten when a later run ends (the exact corruption).
- `TestExtractSessionIDFromPath` — extended the canonical table with the nested subagent shape and dropped the case that pinned the wrong fallback.

## Verification

- `mise run fmt`, lint (golangci-lint v2.11.3, pinned) — clean.
- `mise run test:ci` (full race suite + e2e canary) — green; `pi`, `agent/*`, `checkpoint/*`, `strategy`, `cli` suites green.
- **End-to-end smoke test** with the real binary in a scratch repo, driving a parent turn + two subagent runs that each modify a file:
  - `entire session list` → one Pi session with the real parent UUID, no phantom `session` row.
  - `.entire/tmp/pi/` → three distinct captures (`019fab4c-….json`, `d93a3451-run-0.json`, `716671d6-run-1.json`), no shared `session.json`.
  - Shadow branch → task checkpoints under `…/<parentUUID>/tasks/<run>/agent-<run>.jsonl`, each storing its **own** run's content (run-0 → run0 marker, run-1 → run1 marker), zero cross-contamination.

## Deferred

The issue's optional "have `entire_extension.ts` send `session_id` explicitly" was intentionally dropped: the path parse is now correct and load-bearing, so it's unnecessary.

---

> [!NOTE]
> **Medium Risk**
> Changes Pi hook routing and subagent checkpoint transcript resolution; mistakes could mis-attribute sessions or corrupt task checkpoints, but scope is Pi-specific with a clear fallback for other agents.
> 
> **Overview**
> Fixes **#1870**: Pi subagent hooks used to derive session ID from `session.jsonl` basenames, so every run shared the ID `session` and **staged transcripts / task checkpoints overwrote each other**.
> 
> **Pi lifecycle** now parses nested paths (`…/<ts>_<uuid>/<sub>/run-N/session.jsonl`), rolls **SessionID** up to the parent UUID, assigns a distinct **`<sub>-run-N`** subagent ID, and emits **`SubagentStart` / `SubagentEnd`** instead of normal turns. Each run is staged via `captureTranscript` under its own cache key.
> 
> **Framework**: `agent.Event` gains **`SubagentTranscriptPath`**; `handleLifecycleSubagentEnd` prefers it over the `agent-<id>.jsonl` convention beside the main transcript.
> 
> Other agents are unchanged when the new field is empty. Regression coverage lives in `subagent_1870_test.go`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a9b8d49248cc78b2ac4ea1aecc961a711a89343e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/agent/event.go
+++ b/cmd/entire/cli/agent/event.go
@@ -111,6 +111,14 @@ type Event struct {
 	// SubagentID identifies the subagent instance (for SubagentEnd events).
 	SubagentID string
 
+	// SubagentTranscriptPath, when set, is the path to the subagent's own
+	// transcript file (SubagentEnd events). Agents that keep subagent
+	// transcripts at a location the framework cannot reconstruct from
+	// SessionRef + SubagentID (e.g. Pi, which stores each run under a nested
+	// <parent>/<sub>/run-N/ directory) set this directly; the SubagentEnd
+	// handler prefers it over the AgentTranscriptPath naming convention.
+	SubagentTranscriptPath string
+
 	// ToolInput is the raw tool input JSON (for subagent type/description extraction).
 	// Used when both SubagentType and TaskDescription are empty (agents that don't provide
 	// these fields directly parse them from ToolInput).
--- a/cmd/entire/cli/agent/pi/lifecycle.go
+++ b/cmd/entire/cli/agent/pi/lifecycle.go
@@ -153,10 +153,15 @@ func (a *PiAgent) ParseHookEvent(ctx context.Context, hookName string, stdin io.
 	}
 	payload := *parsed
 
+	identity := parsePiSessionIdentity(payload.SessionFile)
 	sessionID := payload.SessionID
 	if sessionID == "" {
-		sessionID = extractSessionIDFromPath(payload.SessionFile)
+		sessionID = identity.SessionID
 	}
+	// subagentID is non-empty only for Pi subagent runs, whose transcript path
+	// carries <parent>/<sub>/run-N identity that filepath.Base would otherwise
+	// collapse to the role-named "session" leaf (issue #1870).
+	subagentID := identity.SubagentID
 
 	now := time.Now()
 
@@ -170,6 +175,25 @@ func (a *PiAgent) ParseHookEvent(ctx context.Context, hookName string, stdin io.
 		}, nil
 
 	case HookNameBeforeAgentStart:
+		if subagentID != "" {
+			// A Pi subagent run is starting. Route it through the subagent
+			// mechanism (issue #1870) so pre-task state is captured and the run
+			// is attributed to the parent session rather than a phantom
+			// "session"-named session. No context injection: that is for the
+			// parent turn only.
+			return &agent.Event{
+				Type:       agent.SubagentStart,
+				SessionID:  sessionID,
+				SubagentID: subagentID,
+				ToolUseID:  subagentID,
+				// SessionRef is the parent conversation on both Pi subagent
+				// events (Start/End). The Start handler only logs it and
+				// snapshots pre-task git state, so the value is informational
+				// here; keeping it identical to End avoids a confusing mismatch.
+				SessionRef: derivePiParentTranscript(payload.SessionFile),
+				Timestamp:  now,
+			}, nil
+		}
 		// Pi emits before_agent_start with a fully-populated session ID, but
 		// we cache it anyway to support the agent_end fallback below.
 		if sessionID == "" {
@@ -192,6 +216,29 @@ func (a *PiAgent) ParseHookEvent(ctx context.Context, hookName string, stdin io.
 	case HookNameAgentEnd:
 		if sessionID == "" {
 			sessionID = readCachedSessionID(ctx)
+		}
+		if subagentID != "" {
+			// A Pi subagent run ended. Stage its transcript under the run's own
+			// identity (never the shared "session.json"), attach it to the
+			// parent session as a subagent transcript, and route as SubagentEnd
+			// so distinct runs no longer overwrite one another (issue #1870).
+			subagentRef := captureTranscript(ctx, subagentID, payload.SessionFile)
+			return &agent.Event{
+				Type:       agent.SubagentEnd,
+				SessionID:  sessionID,
+				SubagentID: subagentID,
+				ToolUseID:  subagentID,
+				// SessionRef points at the raw parent transcript, not the staged
+				// .entire/tmp/pi/<parentUUID>.json copy: the parent's agent_end
+				// (which stages that copy) fires last, after every subagent run,
+				// so the staged copy does not exist yet here. The raw path still
+				// exists mid-session and is read best-effort at checkpoint-write
+				// time, so the parent content lands in the shadow tree then.
+				SessionRef:             derivePiParentTranscript(payload.SessionFile),
+				SubagentTranscriptPath: subagentRef,
+				Model:                  extractModelFromPiSessionFile(subagentRef),
+				Timestamp:              now,
+			}, nil
 		}
 		// Capture the Pi JSONL into <repo>/.entire/tmp/pi/<id>.json so the
 		// strategy has a stable transcript reference even if the user later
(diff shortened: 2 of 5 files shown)


gtrrz-victor:
Closing as not needed. This fix keys on Pi's nested `<parent>/<sub>/run-N/session.jsonl` layout, but current Pi (0.82.1) writes each subagent as its own top-level `<ts>_<uuid>.jsonl` session with a `parentSession` field — unique UUIDs, no `session.jsonl`, so the `session` collision this fixes cannot reproduce, and the fix path stays dormant.

Verified on two real runs + committed checkpoints: transcript content is correct, shards distinct, no last-writer-wins.

Full write-up + open version questions: entireio/cli#1870 (comment https://github.com/entireio/cli/issues/1870#issuecomment-5117660831).

If parent↔subagent linking on modern Pi is wanted, that is a separate enhancement keyed on `parentSession`, not this data-loss fix.

PR #1877, opened 2026-07-30
fix(attach): fetch on-demand transcripts for untracked sessions

### LINKED ISSUE #1876: session attach cannot find transcripts for OpenCode sessions not spawned from a hooked terminal
## Problem

`entire session attach <id> --agent opencode` fails with `transcript not found` for OpenCode sessions that Entire never tracked — e.g. sessions spawned by an external session host rather than a hooked terminal — even though `opencode export <id>` succeeds.

## Root cause

`resolveAndValidateTranscript` (cmd/entire/cli/attach.go) only calls `PrepareTranscript` when the transcript file *already exists* under `.entire/tmp`:

```go
// Only call PrepareTranscript when the file already exists — it flushes
// in-progress writes, but can't conjure a file that was never started.
```

That assumption holds for file-based agents, but not for OpenCode: its transcript is produced on demand by `opencode export` (see `fetchAndCacheExport` in cmd/entire/cli/agent/opencode/lifecycle.go), which works for **any** session in the opencode store, including sessions no hook ever saw. OpenCode also has no `SessionBaseDirProvider`, so the project-dir fallback search can't help either (`fallback transcript search not supported for agent "opencode"`).

Result: there is no way to attach the correct session when hooks didn't capture one. In our case (a commit authored by an externally-hosted OpenCode session), the checkpoint-trailer guard required a session, and the only attachable one belonged to unrelated work — mis-attribution.

## Proposed fix

Add a `TranscriptFetcher` capability (agent can materialize a transcript on demand), implement it on `OpenCodeAgent` via `fetchAndCacheExport`, and consult it in `resolveAndValidateTranscript` after the initial stat fails, before the project-dir search. PR coming up referencing this issue; also tracked in a downstream integration repository.

Note: `session list` enumeration from the opencode store is a separate, larger follow-up — the fix targets the attach path, which is the workflow blocker.

Legonaftik:
Fixes #1876.

## Problem

`entire session attach <id> --agent opencode` fails with `transcript not found` for OpenCode sessions Entire never tracked — e.g. sessions spawned by an external session host rather than a hooked terminal. `resolveAndValidateTranscript` only calls `PrepareTranscript` when the transcript file *already exists* under `.entire/tmp`, on the assumption that preparation "can't conjure a file that was never started". For OpenCode that assumption is false: its transcript is produced on demand by `opencode export` (`fetchAndCacheExport`), which works for any session in the opencode store, tracked or not. OpenCode also lacks a `SessionBaseDirProvider`, so the project-dir fallback can't help. Net effect: no way to attach the correct session when hooks didn't capture one.

## Fix

- New optional `TranscriptFetcher` agent capability (`agent.go` + `AsTranscriptFetcher`, type-assertion gated like `SidecarImageProvider`): the agent can materialize a session transcript on demand, including for sessions with no hook-cached file.
- `OpenCodeAgent.FetchTranscript` implements it via the existing `fetchAndCacheExport` (which already honors `ENTIRE_TEST_OPENCODE_MOCK_EXPORT`).
- `resolveAndValidateTranscript` consults the capability after the initial stat fails, before the project-dir search fallback. File-based agents are unaffected.

Out of scope (noted in the issue): `session list` enumeration from the opencode store — the attach path was the workflow blocker.

## Testing

- New regression test `TestAttach_OpenCodeFetchesTranscriptForUntrackedSession`: a session with no hook-cached export attaches successfully via the fetch path (mock-export mode), producing session state with a checkpoint id.
- `TestFetchTranscript_ValidatesSessionID` / `TestFetchTranscript_AttemptsExport` for the new method.
- `go test ./cmd/entire/cli/...` and `golangci-lint run` clean.

## Verification (downstream)

The same scenario was exercised end-to-end against a real externally-hosted OpenCode session (`opencode export` succeeds, attach previously failed); with this change the fetch path materializes the transcript and attach proceeds.


MuskanPaliwal:
I opened #1996 as a follow-up to this PR. It preserves the `FetchTranscript` failure when no fallback transcript is found and turns common `opencode export` failures into shorter user-facing messages.

Because it is stacked on #1877, it currently inherits this PR’s conflict with `main`. Once #1877 merges, I’ll update the branch so #1996 contains only the follow-up change against the latest `main`.

Soph:
Force-pushed a rebase of this branch onto current `main`, and folded in #1996.

**What changed on the branch** (was `36724c8`, now `76c5d8e`):

1. `a1ab538` — this PR's original commit, rebased. One conflict, in `cmd/entire/cli/agent/capabilities.go`: `main` has since added `AsTranscriptSanitizer` at the same insertion point. Resolved by keeping both and rewriting `AsTranscriptFetcher` in `main`'s current idiom — `builtinCapability[TranscriptFetcher](ag)` instead of the hand-rolled nil-check — plus adding `TranscriptFetcher` to the `DeclaredCaps` exclusion list that `main`'s commit had just updated.
2. `6ef8c49` — @MuskanPaliwal's commit from #1996, unchanged. That PR asked for exactly this merge order; #1996 is now closed as folded in.
3. `76c5d8e` — new: `runOpenCodeExportToFile` stages the export in a sibling temp file and renames over the target only on success.

**Why (3) is here.** `runOpenCodeExportToFile` opened `<repo>/.entire/tmp/<id>.json` with `O_TRUNC` and `os.Remove`d it when the export failed. That file is often the only local copy of a session — the turn-end hook writes it, and nothing condenses it into a checkpoint until the user commits. Both callers re-export over a possibly-populated path (`PrepareTranscript` on every turn end via `cli/lifecycle.go:711`, and the new `FetchTranscript` on attach), so a missing `opencode` binary, a rejected session, or a 30s timeout destroyed the transcript it was asked to refresh — and attach then reported "transcript not found" for a session whose transcript it had just deleted. The staging name starts with `.` and does not end in `.json` so `.entire/tmp` scanners skip it, mirroring `makeInstallTmpPath` in `cli/plugin_store.go`.

Regression test verified both ways: without the fix, `TestRunOpenCodeExportToFile_FailedExportPreservesExistingFile` reports `cached transcript was destroyed by the failed export`; with it the cached bytes are untouched. `mise run check` passes.

The PR is now `MERGEABLE`. Review comments to follow separately — the rebase and this fix are mechanical, not the review.

Soph:
Pushed `f1f9946` — gates auto-detection so it no longer exports on the user's behalf.

`detectAgentByTranscript` probes every registered agent through `resolveAndValidateTranscript`, so making OpenCode a `TranscriptFetcher` meant any failed attach — of any agent — ran `opencode export <that-session-id>` as a side effect of detection: a subprocess with a 30s timeout ceiling plus a `MkdirAll` of `<repo>/.entire/tmp`, during what reads as a read-only probe. `agent.List()` is alphabetical, so opencode was probed before pi and vogon.

This is the same reason `PrepareTranscript` sits behind an `os.Stat` (the comment there about sparing Cursor a 3s poll) — agent-side materialization is strictly more expensive than a flush.

`resolveAndValidateTranscript` now takes an explicit `transcriptLookup`: `lookupAllowFetch` for the agent the user named, `lookupLocalOnly` while probing the others. Since detection no longer exports, the transcript-not-found error names the agents that could have, so omitting or mistyping `--agent` yields `--agent opencode` instead of "is the session ID correct?" — without that, gating would have quietly removed a path that used to work.

`TestAttach_TranscriptNotFound`: 1.05s → 0.13s.

Full `mise run test:ci` green on this head with nothing cached — 90 packages including `integration_test` and `strategy`, plus the canary at 56/56 (vogon) and 4/4 (roger-roger).

Soph:
Pushed `3a39273`. My `76c5d8e` was an incomplete fix — flagging it rather than amending it away.

**What was still broken:** `fetchAndCacheExport`'s `json.Valid` gate ran *after* the rename, so staging only ever protected the non-zero-exit path. An `opencode export` that exits 0 with truncated output still replaced the cached transcript. That's worse than a missing file — attach.go:817 takes the `os.Stat` branch, swallows `PrepareTranscript`'s error, and silently uses the corrupt path, where a missing one would be re-fetched.

**Fix:** staging moves up into `fetchAndCacheExport` — export → read → validate → *then* rename — plus the `Sync()` the runner was missing (`jsonutil.WriteFileAtomic` documents why). A failed rename now keeps the validated export and names its path. Since the runner is the stubbed seam, stub tests exercise staging for free: three new cases (partial-write-then-error, truncated exit-0, empty exit-0) fail against `76c5d8e`. The old success-path test passed against *pre-fix* code and is gone.

**Windows:** `MoveFileEx(MOVEFILE_REPLACE_EXISTING)` must delete the destination and Go opens files without `FILE_SHARE_DELETE`, so a concurrent reader breaks the replace where in-place writing worked. `renameOverExisting` retries briefly, then names the contended file; `isRenameContention` is build-tagged to false on POSIX. My earlier comment claiming rename "replaces the destination on both POSIX and Windows" was wrong and is corrected.

**`entire clean`:** ENOENT in `deleteTempFiles` is no longer counted as a failure — `.entire/tmp` now holds transient files, so `clean --all` racing a turn-end export would report `failed to delete N item(s)`. Verified RED against the old code.

**Two pre-existing test bugs, both writing into the developer's own repo:** `TestPrepareTranscript_ErrorOnBrokenSymlink` ran the real `opencode` binary there — its premise is wrong, `os.Stat` follows the link and reports ENOENT, so the guard never fires and it falls through to a live export (this is what rewrote `.opencode/package-lock.json`, the behaviour #1992 documents). `TestPrepareTranscript_AlwaysRefreshesTranscript` lacked a `t.Chdir`.

Verified: `mise run fmt`/`lint` clean, `GOOS=windows` build + vet clean, full `test:ci` green with nothing cached (86 packages, canary 56/56 and 4/4).

Follow-ups, out of scope here: this is now the fifth atomic-replace copy in the repo and wants a streaming sibling to `WriteFileAtomic`; and `capabilities.go`'s `DeclaredCaps` exclusion list omits `HookFreshness`, `SidecarImageProvider`, and `SubagentSessionResolver`.

MuskanPaliwal:
hi @Soph , let me know if this pr is good to be merged or anything else is pending.

Soph:
@MuskanPaliwal yeah I'm waiting for someone from my team to approve.

gtrrz-victor (approved):

gtrrz-victor (approved):

PR #1895, opened 2026-08-04
Use login JWTs directly for cell auth

jagregory:
https://entire.io/gh/entireio/cli/trails/972

---

> [!NOTE]
> **High Risk**
> This changes authentication for entire-api cell calls and all git smart-HTTP traffic; backends must accept login JWTs where they previously expected jurisdiction tokens, and misconfiguration could cause widespread clone/API failures until aligned.
> 
> **Overview**
> **Cell API clients** (`NewEntireAPICellClient` / `CellClientFactory`) no longer RFC 8693–exchange the login JWT for a jurisdictional identity token. They still resolve the target cell URL as before, but send the **refreshed login JWT** as the `Authorization` bearer. Per-jurisdiction mint caching and single-flight exchange logic are removed.
> 
> **Git remote helper** drops the entire `jurisdictionauth` stack (keychain-persisted jurisdiction tokens, exchange routing, `jurisdiction_audience` requirement). Interactive auth uses **`NewRefreshingLoginCredential`** with a **401 → `ForceRefresh`** retry path; **`ENTIRE_TOKEN`** is sent verbatim after the existing cluster core trust gate, without requiring `jurisdiction_audience` on the cluster.
> 
> **`entire auth token --jurisdiction`** (`JurisdictionToken`) is unchanged: it still performs the jurisdiction identity exchange when explicitly requested.
> 
> Docs/comments for **`ENTIRE_TOKEN`** are updated to describe direct bearer use on control/git paths vs. the separate jurisdiction-mint command.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6556fb75aa425ea716c9c0cec7e274a6acc68d94. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


khaong:
👋 Claude here — I'm building ENT-1329 (sandbox clones from the entiredb mirror instead of github.com) with @khaong, and it depends on this PR, so I've read it closely. Two questions below plus one observation. **These are mine, not Alex's** — he hasn't reviewed them.

Both are about `ENTIRE_TOKEN` changing meaning rather than about the diff itself.

**1. `cmd/entire/cli/auth/env_token.go` isn't in the diff, so its docs now describe the deleted mechanism.**

`EnvTokenVar`'s doc still says the value is *"used verbatim as the login JWT for repo-scoped token exchange"*, and `CoreURLFromEnvToken`'s SECURITY comment still says *"the returned URL becomes the host the env token is POSTed to as a subject_token during exchange"* — nothing is POSTed once this lands. Its rejection message (*"must be a login or sa-session JWT"*) will also misdescribe any terminal token. Should those move with the behaviour?

**2. Does `coreTrusted` still assert the right thing?**

It asks whether the token's aud is one of the cluster's trusted *login servers* — exactly right for a token about to be exchanged **at** that server, but an end-use bearer's aud names the **resource server**. Concretely, eukanuba advertises:

```
core_urls:            [eu.auth.partial.to, us.auth.partial.to, us.partial.to]
jurisdiction_audience: eu.partial.to
```

`eu.partial.to` is not in that list — so a *correctly*-audienced end-use token for eukanuba would be refused by eukanuba's own gate.

This is live for us. The runner injects a core-issued automation token, and it clears the gate only because `ExchangeAutomationToken` currently forces `audience == loginAud`, which makes its aud a core. That coupling is what COR-1132 is meant to fix — so a *more correct* automation token could start failing here.

**Observation, not a request:** deleting `crossJurisdictionHint` also removes the thing that actually stopped a cross-jurisdiction token. Its own doc notes the gate alone doesn't stop one — the exchange did, via `invalid_target`. Presumably intended, since authorization is resolved live at the data plane, but the description's *"jurisdiction routing/trust checks remain intact"* reads as covering more than routing.

Not asking for changes — I mainly want to know whether the gate's predicate is meant to follow the variable's new meaning, since we're about to build on it.

jagregory:
Thanks for the close read. I addressed points 1 and 3 in e44b72f.

jagregory:
bugbot run

khaong (approved):

PR #1918, opened 2026-08-07
Don't prompt from git hooks on a capturing client's own pty

gtrrz-victor:
https://entire.io/gh/entireio/cli/trails/987

The raw-mode check in 508c438e3 assumed a TUI git client keeps the user's
terminal while running `git commit`. lazygit doesn't: it gives the command a
fresh pty in a new session and only renders what comes back. That pty opens
cleanly and sits in canonical mode, so the hook concluded it could prompt,
painted "Link this commit to session context?" into a terminal nobody reads,
and blocked the commit — lazygit froze. Reproduced with real lazygit 0.48 on
Linux driven through a pty, which is where the report came from.

Job control is the discriminator. An interactive shell puts each foreground
command in its own process group and stays session leader (sid != pgrp), while
a command handed a private session leads that session itself (sid == pgrp).
Measured against real lazygit (sid == pgrp == the git pid, on a pty distinct
from lazygit's own) and against interactive bash and busybox sh.

Scoped to git-hook context via MarkGitHookContext: one-shot invocations like
`ssh -t host entire enable` and `docker run -it … entire enable` share the
private-session shape but do forward keystrokes, so they keep prompting.

The pty's window size looked like an easier signal — 0x0 under lazygit 0.48 —
but lazygit master presets 80x24 to "avoid a zero-size pty", so it would have
broken on the next release. Session shape follows from the setsid a private
controlling terminal requires, so it doesn't drift with rendering details. The
raw-mode check stays as the layer for clients that reuse the user's terminal
rather than allocating one (gitui, tig).

End-to-end with a live agent session, real hooks and real lazygit: before, the
prompt painted into lazygit's pty and no commit was created; after, the commit
lands with its Entire-Checkpoint trailer auto-linked, while a human committing
in a real terminal still gets the prompt and can answer it.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Entire-Checkpoint: 01KZBGWJP37Z5JFDH5W1HKH1JA

---

> [!NOTE]
> **Medium Risk**
> Changes when commit-linking and other hook prompts run; misclassification could suppress prompts on ssh -t/script-style sessions (documented trade-off: auto-link default).
> 
> **Overview**
> Fixes **commits hanging in lazygit** (and similar clients) when Entire tried to prompt during git hooks on a **private PTY** the UI only displays—it never forwards keystrokes.
> 
> **`CanPromptInteractively()`** gains a new step (before the raw-mode `/dev/tty` check): when **`MarkGitHookContext()`** was set on the `entire hooks git` tree and **`ttyIsPrivateSession()`** reports `sid == pgrp` (Unix), prompting is disabled so hooks follow the **non-interactive default** (e.g. auto-link commits). Shell-driven `git commit` keeps `sid != pgrp` and still prompts. Direct CLI use is unchanged because hook context is not marked.
> 
> Adds **`githookcontext.go`**, **`jobcontrol_unix.go`** / stub for other OSes, re-exec tests for session shapes, and **CLAUDE.md** documentation for the updated precedence.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a636839a22e24c9d727eb396c857d35712984a14. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -299,12 +299,27 @@ Tests that spawn the real `entire` or `git` binary need the child to be non-inte
 2. `testing.Testing()` → false. In-process `go test` runs are non-interactive by default; no per-test `t.Setenv("ENTIRE_TEST_TTY", "0")` is needed.
 3. Agent sentinels (`GEMINI_CLI`, `COPILOT_CLI`, `PI_CODING_AGENT`, `GIT_TERMINAL_PROMPT=0`) → false.
 4. `CI=<non-empty-non-false>` → false.
-5. `/dev/tty` probe, plus its terminal mode → a terminal held in raw mode
-   (canonical input off) belongs to a full-screen TUI that spawned us, not to a
-   shell we can prompt: TUI git clients (lazygit, gitui, tig) run `git commit`
-   as a child while owning the screen, so the hook inherits a `/dev/tty` it
-   must not prompt on. Fails open when the mode can't be read. See
+5. Git-hook context with a private terminal session → false. TUI git clients run
+   `git commit` on a pty of their own (lazygit does) and only render what comes
+   back, so the hook inherits a `/dev/tty` that opens, sits in canonical mode,
+   and that nothing forwards keystrokes to — prompting there hangs the commit and
+   freezes the client. Job control is the tell: a shell gives each foreground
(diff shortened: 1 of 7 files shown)


gtrrz-victor:
Claude hallucination

PR #1925, opened 2026-08-07
fix(enable): stop the enable-time history import on Ctrl-C, and take it out of --yes

Soph:
https://entire.io/gh/entireio/cli/trails/991

A user ran `entire enable` in a repo that never had Entire, hit Ctrl-C, and got a flood of `resolve push queue failed; ref not enqueued` warnings on their terminal.

First-run enable defaults to the **git-refs** backend and then offers to import pre-existing agent history. Ctrl-C didn't stop that import: neither `agentimport.Run` nor `gitRefsStore.writeSession` observed cancellation, and nothing below them does either (go-git writes and `CreateCommit` ignore ctx). The *only* step that did observe it was the push-queue resolution — it shells out to git — so every post-cancel write logged a warning and left its ref out of the queue. That's worse than noisy: the queue is the sole push-discovery mechanism and the writers are idempotent, so those checkpoints were stranded locally forever.

The warnings reached the terminal at all because `entire enable` never called `logging.Init`, so every `logging.*` call under setup fell back to `slog.Default()`.

### Changes

- **git-refs store** rejects writes on a canceled context; the push-queue record is written with cancellation stripped, since the ref is already on disk by then
- **`agentimport.Run`** checks ctx per session and per turn
- **Interrupted imports** report as resumable (`Import interrupted after N turn(s)`) and stop instead of moving on to the next agent's history
- **`entire enable`** routes its logging to `.entire/logs`
- **`--import-history`** replaces `--yes` as the way to opt into importing history
- **Perf:** memoize the git common dir — the fork was ~9.5ms of a ~12.9ms checkpoint write (73%, 4.5x), on the live hook path too

### Breaking

`entire enable --yes` no longer imports agent history. Pass `--import-history`, or run `entire import <agent>`. `--yes` means "accept defaults", and the interactive default is to import nothing — so it was doing the opposite of the attended path, and in a non-repo dir a single flag both created a GitHub repo and ingested a month of local transcripts destined for it.

### Follow-ups (not in this PR)

- An interrupted enable still exits 0 and prints "Ready.", which opts out of `main.go`'s deliberate re-raise-the-signal contract. `entire import` gets this right; enable doesn't. Changing an exit code felt like its own decision.
- `GitStore.writeSession` (git-branch) has the identical missing guard — verified, pre-existing. A guard in `kindRoutingStore.Write` would close both.

Each fix has a test verified to fail without it.

---

> [!NOTE]
> **Medium Risk**
> Breaking change to `entire enable --yes` behavior and checkpoint/import paths on cancellation; git-refs write and push-queue semantics affect data sync if mis-handled, though changes are guarded by tests.
> 
> **Overview**
> Fixes **Ctrl-C during enable-time or bulk import** continuing to mint git-refs checkpoints while spamming push-queue warnings, and changes how history import is opted in.
> 
> **Cancellation:** `agentimport.Run` checks `ctx` before each session and turn; `gitRefsStore.writeSession` refuses canceled writes. Push enqueue uses `context.WithoutCancel` so refs already written still land in the push queue (missing queue entries strand checkpoints locally). Interrupted runs print a **resumable** message instead of raw `context canceled`, and enable-time import **stops** instead of moving to the next agent.
> 
> **Enable UX (breaking):** New **`--import-history`** is the explicit first-run opt-in; **`entire enable --yes` no longer imports** agent history (aligned with interactive default: import nothing). Non-first-run `--import-history` is reported as not applicable with a pointer to `entire import`.
> 
> **Logging & perf:** `entire enable` now calls **`logging.Init`** so setup warnings go to `.entire/logs`. **`resolveGitCommonDir`** is memoized per worktree root (successes only) to avoid repeated `git rev-parse` on every ref write.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1049be5af250c63ca64aae5d4d505573c00c087a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/agentimport/agentimport.go
+++ b/cmd/entire/cli/agentimport/agentimport.go
@@ -215,10 +215,7 @@ func Run(ctx context.Context, repo *git.Repository, imp Importer, opts Options) 
 	authorName, authorEmail := cp.GetGitAuthorFromRepo(repo)
 
 	for sessionIndex, sf := range files {
-		// Ctrl-C must stop the import. Nothing else on this path observes
-		// cancellation — go-git object/ref writes and CreateCommit all ignore
-		// ctx — so without these checks an interrupted import runs to
-		// completion, silently minting checkpoints the user asked to stop.
+		// Stop before reading and splitting the next transcript.
 		if err := ctx.Err(); err != nil {
 			return res, err //nolint:wrapcheck // propagate context cancellation
 		}
@@ -240,8 +237,12 @@ func Run(ctx context.Context, repo *git.Repository, imp Importer, opts Options) 
 		var red redact.RedactedBytes
 		redacted := false
 		for turnIndex, turn := range turns {
-			// Per-turn, not just per-session: one session can carry hundreds of
-			// turns, and each turn is a checkpoint write.
+			// Ctrl-C must stop the import, and per turn rather than per session:
+			// one session can carry hundreds of turns, each a checkpoint write.
+			// Nothing below this observes cancellation — go-git object/ref
+			// writes and CreateCommit all ignore ctx, and neither git store
+			// guards its create path — so without this an interrupted import
+			// runs to completion, minting checkpoints the user asked to stop.
 			if err := ctx.Err(); err != nil {
 				return res, err //nolint:wrapcheck // propagate context cancellation
 			}
--- a/cmd/entire/cli/agentimport/agentimport.go
+++ b/cmd/entire/cli/agentimport/agentimport.go
@@ -239,10 +239,9 @@ func Run(ctx context.Context, repo *git.Repository, imp Importer, opts Options) 
 		for turnIndex, turn := range turns {
 			// Ctrl-C must stop the import, and per turn rather than per session:
 			// one session can carry hundreds of turns, each a checkpoint write.
-			// Nothing below this observes cancellation — go-git object/ref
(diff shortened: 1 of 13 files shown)


Soph:
Both addressed in a3c5b126e and 513009238.

**Bugbot — rejected enable now mutates repositories.** Confirmed and reproduced: `entire enable --local --project` in an untouched repo left `?? .entire/` behind, because `logging.Init` creates `.entire/logs/` and ran before the checks that reject the invocation. `--agent <typo>` did the same. Fixed by resolving `--agent` before logging starts and initializing after every check that can bail — the placement is now load-bearing in both directions (after the git-repo check so it can't write outside a repo, after validation so it can't seed one enable refuses to configure). Regression test covers both rejection paths.

**Trail finding — guard applied only to the git-refs backend.** Correct, and it matched this branch's own review. `GitStore.writeSession` had no guard while its backfill siblings all check `ctx.Err()`, so branch-backed repos had no store-level brake on Ctrl-C — and git-branch is still the runtime fallback with no `checkpoints` config. I'd listed it as a follow-up; it's three lines, so it's in. Test verified to fail without the guard.

Also corrected two comments that described the branch-store gap as open.

Follow-up still open: an interrupted enable exits 0 and prints "Ready.", bypassing `main.go`'s deliberate re-raise-the-signal contract. Left out as its own decision.

gtrrz-victor (approved):

PR #1926, opened 2026-08-07
fix(checkpoint): preserve concurrent persistent ref updates

### LINKED ISSUE #1917: Concurrent git-refs checkpoint writes can silently lose completed data!
### What happened?

The persistent `git-refs` checkpoint backend has a read-modify-write race.

When two independent writers update the same checkpoint, both can read the same
checkpoint tip, build different child commits, and successfully update the
checkpoint ref. Because the final ref update is unconditional and does not
verify the expected old commit, the later writer can silently hide the earlier
writer's mutation.

I reproduced this deterministically on the official stable release `v0.9.0`
and on the latest fetched public `main`. Both ref writes returned success, but
the first writer's summary disappeared while the second writer's attribution
remained visible.

This can silently lose checkpoint data such as summaries, transcripts,
attribution, assets, or attached sessions even though the writers report
success.

### Steps to reproduce

I added the following local regression test to
`cmd/entire/cli/checkpoint/refs_store_test.go`:

`TestGitRefsStore_ForcedConcurrentBackfillsLoseOneMutation`

The test performs this deterministic interleaving:

1. Create a checkpoint.
2. Read its current parent once.
3. Build a summary mutation from that parent.
4. Build an attribution mutation from the same parent.
5. Write the summary commit successfully.
6. Write the attribution commit successfully afterward.
7. Read the checkpoint.
8. Observe that attribution remains but the summary is absent.

Run:

`go test ./cmd/entire/cli/checkpoint -run '^TestGitRefsStore_ForcedConcurrentBackfillsLoseOneMutation$' -count=1 -v`

The test passes because it successfully observes the lost-update behavior; it
does not mean the defect is fixed.

### Entire CLI version

Entire CLI v0.9.0

### OS and architecture

Windows amd64 (build 26200.8875)

### Agent

Codex

### Terminal

PowerShell 5.1 through the Codex desktop integrated shell

### Logs / debug output

```bash
=== RUN   TestGitRefsStore_ForcedConcurrentBackfillsLoseOneMutation
--- PASS: TestGitRefsStore_ForcedConcurrentBackfillsLoseOneMutation (1.88s)
PASS
ok   github.com/entireio/cli/cmd/entire/cli/checkpoint   5.479s
```

### Additional context

Validated against official stable release `v0.9.0`:

`8b77ad43132d18f7958825c9dcd26544ab8f5d92`

Also validated against the latest fetched public `main` revision:

`dd2e87a02ef853b32a85e7094e1944fe320739a3`

Environment details:

- Stable test Go version: `1.26.4`
- Latest-main test Go version: `1.26.5`
- Git: `2.46.0.windows.1`
- Shell: PowerShell `5.1.26100.8875`
- Validation worktree: detached worktree at the exact revision above
- Production source changes: none; only the local regression test was added

The likely cause is that `gitRefsStore.setRef` advances the checkpoint ref with
an unconditional `SetReference`. The read/build/set sequence has no
checkpoint-scoped lock and does not use an expected-old-value compare-and-swap.
Two writers can therefore build valid commits from the same parent while the
last ref update silently makes the first mutation unreachable through normal
checkpoint reads.

Expected behavior: both independent mutations should remain reachable, or one
writer should receive an explicit conflict and retry or fail clearly.

ChetanReddyC:
## Summary

Fixes entireio/cli#1917.

Concurrent checkpoint writers could read the same persistent ref tip, build<br>independent commits, and successfully update the ref without preserving both<br>mutations. The later update made the earlier completed data unreachable.

This follows the reproduction and guidance from entireio/cli#1920.

## Changes

* Add a cross-process lock scoped to each persistent ref.
* Replace unconditional ref updates with native Git CAS using:<br>`git update-ref <ref> <new> <old>`.
* Retry from a freshly read parent after CAS conflicts.
* Rebuild the tree and commit on every retry.
* Apply the same protection to:
  * `git-branch` writes
  * `git-refs` writes
  * persistent-ref initialization
  * git-branch to git-refs migration
* Add focused CAS retry and cleanup tests.
* Add the concurrent two-Stop-hook integration regression for both backends.

## Testing

* Soph's sequential/concurrent integration test passed for both<br>`git-branch` and `git-refs`.
* Five-repeat stress run passed all 20 integration scenarios.
* Focused checkpoint, migration, dispatch, and enqueue tests passed.
* `go vet ./cmd/entire/cli/checkpoint` passed.
* `gofmt -s` and `git diff --check` passed.
* Rebased and verified against current `main`.
* Added coverage proving that cancellation before CAS prevents the ref write.
* Added coverage proving that successful CAS followed by cancellation still
  enqueues the ref for push.
* The sequential and concurrent #1920 scenarios pass for both `git-branch`
  and `git-refs`, with both session updates preserved.

The full checkpoint package was also run. Seven existing Windows tests are<br>environment-sensitive because they access the real Git configuration or<br>AppData paths; the remaining package tests passed.

## Notes

The implementation uses the same lock/CAS/retry pattern already used by the<br>shadow-branch store. The retry rebuilds from the new tip rather than reusing a<br>commit built from the stale parent.

## Behavior and scope

`casUpdateRef` uses `exec.CommandContext`, so cancellation before the CAS
prevents `git update-ref` from running and the checkpoint ref is not written.
This is intentional: no partial checkpoint is stranded, `FullyCondensed`
remains false, and PostCommit can retry the work later.

Cancellation after a successful CAS does not prevent push discovery. Enqueueing
is context-free, so a ref that was successfully written is still recorded in
the push queue.

The lock-plus-CAS guarantee in this PR covers writers in the `checkpoint`
package. Pre-existing unconditional writes to the same refs from `strategy`
remain follow-up scope.

Lock granularity differs by backend:

* `git-refs` locks each checkpoint ref independently.
* `git-branch` locks the shared v1 branch, so all checkpoint writes in a
  repository are serialized.

Note: This commit does not include an Entire-Checkpoint trailer because the implementation session was not captured by Entire.

### DIFF
--- a/cmd/entire/cli/checkpoint/persistent_ref_update.go
+++ b/cmd/entire/cli/checkpoint/persistent_ref_update.go
@@ -68,12 +68,12 @@ func persistentRefLockPath(commonDir string, refName plumbing.ReferenceName) (st
 	return filepath.Join(lockDir, safe+".lock"), nil
 }
 
-func withPersistentRefFlock(commonDir string, refName plumbing.ReferenceName, fn func() error) error {
+func withPersistentRefFlock(ctx context.Context, commonDir string, refName plumbing.ReferenceName, fn func() error) error {
 	path, err := persistentRefLockPath(commonDir, refName)
 	if err != nil {
 		return err
 	}
-	release, err := flock.Acquire(path)
+	release, err := flock.AcquireContext(ctx, path)
 	if err != nil {
 		return fmt.Errorf("acquire persistent ref flock %s: %w", refName, err)
 	}
@@ -90,7 +90,7 @@ func updatePersistentRef(ctx context.Context, repo *git.Repository, refName plum
 		return fmt.Errorf("resolve repository directories: %w", err)
 	}
 
-	return withPersistentRefFlock(commonDir, refName, func() error {
+	return withPersistentRefFlock(ctx, commonDir, refName, func() error {
 		for attempt := range shadowRefMaxRetries {
 			newHash, expectedHash, buildErr := build()
 			if buildErr != nil {
--- a/cmd/entire/cli/checkpoint/persistent_ref_update.go
+++ b/cmd/entire/cli/checkpoint/persistent_ref_update.go
@@ -1,0 +1,123 @@
+package checkpoint
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/entireio/cli/cmd/entire/cli/internal/flock"
+	"github.com/entireio/cli/cmd/entire/cli/logging"
+
+	"github.com/go-git/go-git/v6"
+	"github.com/go-git/go-git/v6/plumbing"
+)
+
+type persistentRefBuilder func() (newHash, expectedHash plumbing.Hash, err error)
+
+func repositoryDirs(ctx context.Context, repo *git.Repository) (worktreeRoot, commonDir string, err error) {
+	wt, err := repo.Worktree()
+	if err != nil {
+		return "", "", fmt.Errorf("open worktree: %w", err)
+	}
+	worktreeRoot = wt.Filesystem().Root()
+	if worktreeRoot == "" {
+		return "", "", errors.New("repository worktree filesystem has no root path")
+	}
+	commonDir, err = resolveGitCommonDir(ctx, repo)
+	if err != nil {
+		return "", "", err
+	}
+	return worktreeRoot, commonDir, nil
+}
+
+// casUpdateRef atomically updates refName through native Git's lock protocol.
+// Pass ZeroHash as expectedHash to require that the ref does not exist.
+func casUpdateRef(ctx context.Context, repoRoot string, refName plumbing.ReferenceName, newHash, expectedHash plumbing.Hash) error {
+	newValue := newHash.String()
+	oldValue := strings.Repeat("0", newHash.HexSize())
+	if expectedHash != plumbing.ZeroHash {
+		oldValue = expectedHash.String()
+	}
+
+	cmd := exec.CommandContext(ctx, "git", "update-ref", refName.String(), newValue, oldValue)
+	cmd.Dir = repoRoot
+	cmd.Env = append(os.Environ(), "LC_ALL=C", "LANG=C")
+	output, err := cmd.CombinedOutput()
+	if err == nil {
+		return nil
+	}
+
+	out := string(output)
+	if strings.Contains(out, "cannot lock ref") || strings.Contains(out, "but expected") {
+		return ErrShadowRefBusy
+	}
+	return fmt.Errorf("git update-ref %s: %s: %w", refName, strings.TrimSpace(out), err)
+}
+
+func persistentRefLockPath(commonDir string, refName plumbing.ReferenceName) (string, error) {
+	lockDir := filepath.Join(commonDir, "entire-persistent-ref-locks")
+	if err := os.MkdirAll(lockDir, 0o750); err != nil {
+		return "", fmt.Errorf("create persistent ref lock directory: %w", err)
+	}
+	safe := strings.NewReplacer("/", "_", "\\", "_", ":", "_").Replace(refName.String())
+	return filepath.Join(lockDir, safe+".lock"), nil
+}
+
+func withPersistentRefFlock(commonDir string, refName plumbing.ReferenceName, fn func() error) error {
+	path, err := persistentRefLockPath(commonDir, refName)
+	if err != nil {
+		return err
+	}
+	release, err := flock.Acquire(path)
+	if err != nil {
+		return fmt.Errorf("acquire persistent ref flock %s: %w", refName, err)
+	}
+	defer release()
+	return fn()
+}
+
+// updatePersistentRef serializes Entire writers for one ref and retains a CAS
+// retry for native Git or other external writers. build runs again after every
+// conflict, so each retry reconstructs its tree and commit from the fresh tip.
+func updatePersistentRef(ctx context.Context, repo *git.Repository, refName plumbing.ReferenceName, build persistentRefBuilder) error {
+	repoRoot, commonDir, err := repositoryDirs(ctx, repo)
+	if err != nil {
+		return fmt.Errorf("resolve repository directories: %w", err)
+	}
+
+	return withPersistentRefFlock(commonDir, refName, func() error {
+		for attempt := range shadowRefMaxRetries {
+			newHash, expectedHash, buildErr := build()
+			if buildErr != nil {
+				return buildErr
+			}
+
+			refErr := casUpdateRef(ctx, repoRoot, refName, newHash, expectedHash)
+			if refErr == nil {
+				return nil
+			}
+			if newHash != expectedHash {
+				tryDeleteLooseObject(commonDir, newHash)
+			}
+			if !errors.Is(refErr, ErrShadowRefBusy) {
+				return fmt.Errorf("update persistent ref %s: %w", refName, refErr)
+			}
+
+			if backoffErr := shadowRefBackoff(ctx, attempt); backoffErr != nil {
+				return backoffErr
+			}
+		}
+
+		logging.Warn(logging.WithComponent(ctx, "checkpoint"),
+			"persistent ref CAS retry budget exhausted",
+			slog.String("ref", refName.String()),
+			slog.Int("retries", shadowRefMaxRetries),
+		)
+		return fmt.Errorf("failed to update persistent ref %s after %d CAS retries: %w", refName, shadowRefMaxRetries, ErrShadowRefBusy)
+	})
+}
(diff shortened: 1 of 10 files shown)


cmd/entire/cli/checkpoint/persistent_ref_update.go
```diff
@@ -0,0 +1,123 @@
...
+	cmd.Dir = repoRoot
+	cmd.Env = append(os.Environ(), "LC_ALL=C", "LANG=C")
+	output, err := cmd.CombinedOutput()
+	if err == nil {
+		return nil
+	}
+
+	out := string(output)
+	if strings.Contains(out, "cannot lock ref") || strings.Contains(out, "but expected") {
+		return ErrShadowRefBusy
```
Soph: Reusing the shadow-store error and knobs is right — no argument there — but the names and their doc comments are now wrong for half their callers. `ErrShadowRefBusy` is documented as "shadow branch ref moved" (`shadow_ref.go:21`), `shadowRefMaxRetries` as bounding "the `WriteTemporary` retry loop" (`:27`), and `shadowRefBackoff`/`shadowRefMaxJitter` similarly. All three now serve the persistent path too.

A rename to something backend-neutral (`ErrRefBusy`, `refCASMaxRetries`, `refCASBackoff`) plus a doc touch-up would keep the next reader from concluding this path is shadow-branch machinery.

cmd/entire/cli/checkpoint/persistent_ref_update.go
```diff
@@ -0,0 +1,123 @@
...
+	safe := strings.NewReplacer("/", "_", "\\", "_", ":", "_").Replace(refName.String())
+	return filepath.Join(lockDir, safe+".lock"), nil
+}
+
+func withPersistentRefFlock(commonDir string, refName plumbing.ReferenceName, fn func() error) error {
+	path, err := persistentRefLockPath(commonDir, refName)
+	if err != nil {
+		return err
+	}
+	release, err := flock.Acquire(path)
```
Soph: `flock.Acquire` takes the unbounded blocking path, and there's a `ctx` in scope one frame up. `flock.AcquireContext(ctx, path)` behaves identically when `ctx` carries no deadline (it takes the same kernel `LOCK_EX` path), so this is a free improvement for the callers that do have one.

It matters because of what's inside this critical section. For git-branch the ref is the single shared v1 branch, so this serializes every checkpoint write in the repo behind tree flattening, transcript chunking, blob writes, and — on the `TranscriptPath` fallback in `writeTranscript` — a full redaction pass. Codex declares `agent.SessionEndBudgeter` with a 3s cap and its process tree is killed on expiry, and CLAUDE.md already documents a blocking flock acquire pushing session-end past that cap as a known hazard. This adds a second one on the condense path.

cmd/entire/cli/checkpoint/persistent_ref_update.go
```diff
@@ -0,0 +1,123 @@
...
+
+	out := string(output)
+	if strings.Contains(out, "cannot lock ref") || strings.Contains(out, "but expected") {
+		return ErrShadowRefBusy
+	}
+	return fmt.Errorf("git update-ref %s: %s: %w", refName, strings.TrimSpace(out), err)
+}
+
+func persistentRefLockPath(commonDir string, refName plumbing.ReferenceName) (string, error) {
+	lockDir := filepath.Join(commonDir, "entire-persistent-ref-locks")
```
Soph: On the git-refs backend this creates one lock file per checkpoint ref, and nothing ever removes them, so the directory grows for the life of the repo. `entire-shadow-locks` has the same no-cleanup story but is bounded by distinct base commits; this one is unbounded in checkpoints.

They're tiny and harmless individually, so this isn't urgent — but worth either reaping them somewhere (the `cleanup.go` discovery paths already walk per-checkpoint state) or noting here that the growth is known and accepted, so the next person doesn't have to work out which it is.

cmd/entire/cli/checkpoint/persistent_ref_update_test.go
```diff
@@ -0,0 +1,118 @@
...
+		if err != nil {
+			return plumbing.ZeroHash, plumbing.ZeroHash, err
+		}
+
+		if calls == 1 {
+			externalTip, err = CreateCommit(ctx, repo, commit.TreeHash, parent, "external writer", "Test", "test@test.com")
+			if err != nil {
+				return plumbing.ZeroHash, plumbing.ZeroHash, err
+			}
+			if err := repo.Storer.SetReference(plumbing.NewHashReference(refName, externalTip)); err != nil {
```
Soph: Both tests simulate the competing writer with an in-process go-git `Storer.SetReference`, but with the flock held our own writers can't collide — the only writer that can actually cause a CAS conflict in production is native `git update-ref` from another process. So the retry loop's real precondition isn't covered: that after an external `git update-ref`, this process's go-git `repo.Reference` read returns the *new* value. If it returned a cached one, `build()` would rebuild from a stale tip, the CAS would fail again, and this would burn all 16 retries and fail the write.

I checked separately and go-git does observe a native `update-ref` (loose refs are re-read, not cached), so the loop terminates correctly — but it's an assumption the tests currently leave implicit. Driving the external write through `casUpdateRef` here instead would pin it.

Soph (commented):
Thanks for working on this!

Two things:

**1. The write no longer lands when ctx is canceled. I think that's correct, but it needs a different test.** `main` has since added `TestGitRefsStore_EnqueuesForPushDuringShutdown` (`refs_store_test.go:780`), which calls the `setRef` this deletes. `casUpdateRef` uses `exec.CommandContext`, so on a canceled ctx the subprocess never runs and the ref isn't written:

```
casUpdateRef with canceled ctx -> err = ...: context canceled
ref after canceled-ctx write   -> reference not found
```

Keep that behavior. It's consistent with where the package already went: `writeSession` and all three backfills early-return on `ctx.Err()`, and that guard exists because `entire import` kept minting checkpoints after Ctrl-C. It also matches the shadow store, which has always used `exec.CommandContext`. Nothing partial is written, so nothing can be stranded: `FullyCondensed` stays false and PostCommit retries. The cost is redone work, not lost data.

Worth knowing that the enqueue half of that test's invariant is *not* affected: `Enqueue` is ctx-free and `resolveGitCommonDir` is memoized, so whenever the CAS succeeds the enqueue still lands. So please replace the test rather than delete it, pinning "CAS succeeded, ctx then canceled, still enqueued", and mention the ref-write change in the PR body.

**2. FYI, not a request: the same refs are still written unconditionally from `strategy`.** Pre-existing and out of scope here, so we'll take it as a follow-up. Recording it so the boundary is on the record: this PR's lock-plus-CAS guarantee covers `checkpoint` writers only, not the refs in general.

It affects both backends, very unevenly. `fetchAndRebaseRefCommon` (`push_common.go:470`) reads the tip at :530 and writes it back unconditionally at :541, and it's reached with per-checkpoint refs on git-refs (`manual_commit_push.go:533`, the per-ref retry after a rejected batch push) as well as with the v1 branch. git-refs needs a remote divergence on that exact checkpoint to get there, so it's narrow. git-branch is the wide one: ten more sites target the single shared `refs.Primary` (`cleanup.go:395`, `common.go:756/775/787/854/954/1060`, `git_operations.go:539`, `resume.go:891`, `doctor.go:505`), and any of them racing any checkpoint write in the repo loses data.

Also worth a line in the description: on git-branch the lock is the whole v1 branch, so every checkpoint write in the repo now serializes. Correct trade, but a real change; git-refs is per-checkpoint and unaffected.

We'll also pick up a few naming and comment tidy-ups in the new file post-merge, nothing for you to action (except if you really want).

**Rebase.** Two conflicts (`persistent.go`, `refs_store.go`), both from `main` dropping `applySessionWrite`'s `taskMetadataPath` return and `buildCommitMessage`'s second argument. Two breaks git won't flag: `persistent.go:90` needs the same change but sits just outside the conflict region, and `refs_store_test.go:780` is item 1 above. With those three resolved I get a green checkpoint package and a green #1920 repro against current `main`, so nothing deeper has rotted.

ChetanReddyC:
Thanks, addressed both follow-ups:
- withPersistentRefFlock now uses flock.AcquireContext(ctx, path), with coverage for stopping while waiting on the lock.
- Both CAS retry tests now use native casUpdateRef for the competing writer, covering an external Git update-ref and go-git rereading the changed ref.
I also updated the PR description with the cancellation behavior and scope details. The focused tests pass 10 consecutive runs, and go vet passes. Included in 45afce62.

Soph:
One change before merge.

`tryDeleteLooseObject` at `persistent_ref_update.go:105` runs *before* the `!errors.Is(refErr, ErrShadowRefBusy)` check at `:107`, so it fires on hard errors too. The path it's modelled on does the opposite - `ephemeral.go:158` checks busy first, then deletes at `:163` (same at `:339`/`:342`).

That matters because `main.go:32-46` installs a SIGINT/SIGTERM handler that cancels the root context, so this is ordinary use rather than theoretical:

1. Ctrl-C, or a supervisor SIGTERM, during a checkpoint write
2. ctx cancels, `exec.CommandContext` kills `git update-ref`
3. the kill lands *after* git renamed the lock file into place, so the ref is already updated
4. non-zero exit with empty output, classified non-busy
5. `:105` deletes the loose object the ref now points at

The result is a checkpoint ref pointing at a missing commit, which is less recoverable than the lost update this PR fixes. Moving the delete below the busy check is the whole fix.

Thanks for doing this!

ChetanReddyC:
Fixed in 14cb36ba7. I moved loose-object cleanup below the ErrShadowRefBusy check, matching ephemeral.go, and reran the focused and integration coverage successfully.

Soph (approved):
Thanks a lot for doing this!

PR #1900, opened 2026-08-05
Make git-refs the sole enable default; drop backend picker

gtrrz-victor:
https://entire.io/gh/entireio/cli/trails/975

## Summary
- `entire enable` no longer prompts new users to choose between the git-refs and branch checkpoint backends — a storage-topology decision they can't meaningfully answer during first-run setup.
- git-refs (already the recommendation) is now written silently on first run.
- The branch backend stays fully reachable non-interactively via `--checkpoint-backend branch`.

## Changes
- Removed `promptCheckpointBackend` / `checkpointBackendChoices` and the `huh` picker from `checkpoint_backend.go`.
- Simplified `resolveFirstRunCheckpointBackend`: flag wins, else git-refs default on first run.
- `ENTIRE_CHECKPOINTS_PRIMARY` env override and the config-less git-branch runtime fallback are unchanged, so existing repos are untouched.
- Dropped the picker unit test and the redundant "without --yes" subtest (only covered the deleted prompt branch).

## Test plan
- `go build ./cmd/...`
- `go test ./cmd/entire/cli/ -run 'TestRunEnableInteractive_FirstRunDefaultsToGitRefs|TestResolveCheckpointBackendType|TestApplyCheckpointBackend|TestUpdateCheckpointBackend'` — pass
- golangci-lint v2.11.3 on `cmd/entire/cli/...` — 0 issues

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Setup UX and default-selection logic only; env override, configure path, and legacy config-less behavior are explicitly preserved.
> 
> **Overview**
> First-time **`entire enable`** no longer asks users to pick checkpoint storage in an interactive form. **`git-refs`** is applied silently on first run (same as the former recommendation), while **`--checkpoint-backend branch`** still selects the shared branch backend.
> 
> The **`huh`** picker (`promptCheckpointBackend` / `checkpointBackendChoices`) is removed from `checkpoint_backend.go`, and **`resolveFirstRunCheckpointBackend`** is reduced to: explicit flag wins, else first-run default via **`firstRunCheckpointBackendDefault()`** (still empty when **`ENTIRE_CHECKPOINTS_PRIMARY`** is set so settings are not written). Config-less repos keep the **git-branch** runtime fallback; **`entire configure --checkpoint-backend`** is unchanged.
> 
> Tests for the picker contract and the headless “without `--yes`” branch that only exercised the deleted prompt are dropped.
> 
[...]

PR #1929, opened 2026-08-08
fix(import): sync imported checkpoints once logged in

### LINKED ISSUE #1773: entire enable/import while logged out imports locally with no way to sync after login
### What happened?

Received via support. Running `entire enable` (or `entire import`) while **logged
out** imports agent history into the **local** `entire/checkpoints/v1` store and
prints success ("Imported N turn(s)"), but nothing syncs to the account and the
sessions never appear in the dashboard.

Logging in afterward does **not** recover them. Because import is idempotent on
*local* existence, a second run reports "Imported 0 turn(s) (N already imported)"
and makes no attempt to sync — leaving the user with imported-but-unsyncable
history and no first-class way out.

(Root-caused from source below; this is an ordering/UX + missing-sync bug, not a
one-off.)

### Steps to reproduce

1. In a repo, **before** `entire login`, run `entire enable` (accept the history
   import) — or `entire import <agent>`. It prints "Imported N turn(s)".
2. Open the dashboard → the sessions are absent.
3. Run `entire login`, then re-run enable/import → "Imported 0 turn(s)
   (N already imported)". Still absent; no way to push the already-imported turns.

### Entire CLI version

Not captured (received via support). Root-caused from `main`; reproduces
independent of version.

### OS and architecture

Reporter on macOS; not OS-specific (logic is platform-independent).

### Agent

Agent-agnostic — reproduces for any importer (Claude Code, Cursor, Codex,
Copilot CLI, Gemini CLI, Pi, Factory Droid).

### Additional context

**Root cause (verified in code).** Import is a local-only, idempotent operation
with no auth awareness and no sync:

- `agentimport.Run` writes each turn as a read-only checkpoint on the **local**
  `entire/checkpoints/v1` store (`cmd/entire/cli/agentimport/agentimport.go:112`,
  `writeTurn` → `stores.Persistent.Write` at `:249`) — it never contacts the
  account.
- Idempotency is keyed on **local existence**: existing checkpoints are listed
  and any whose deterministic ID already exists are skipped
  (`agentimport.go:123-128, 147-151`), so a logged-in retry skips everything and
  never retries a sync.
- Neither `entire import` (`cmd/entire/cli/import_cmd.go`) nor the enable import
  offer (`cmd/entire/cli/setup_import.go:48` `maybeOfferSessionImport`, `:224`
  `runSelectedImports`) gates on or checks login. The enable offer is also
  first-run-only (`setup_import.go:49`), so a later logged-in `enable` won't
  re-offer.

Sync happens **only** via the git pre-push hook, gated on a `git push`:

- `PrePush` ships the checkpoint refs/branch alongside the user's push
  (`cmd/entire/cli/strategy/manual_commit_push.go:39`, `:152`). Default backend
  is git-branch (`cmd/entire/cli/checkpoint/open.go:36`, `PrimaryIsRefs(nil)==false`),
  so the whole `entire/checkpoints/v1` branch — including imported commits —
  ships on push.
- Publication is **deferred on an empty remote** until the user's own branch
  exists there (`manual_commit_push.go:76, 186`).
- `enable` only *reports* the enablement to the backend best-effort, and that
  report is **silently skipped when logged out** (`cmd/entire/cli/setup.go:972-1032`
  `reportRepoEnabled`).

Net: three stacked gaps — (1) import writes locally with no login signal;
(2) there is no push trigger for commit-less imports on the default git-branch
backend (the pre-push hook on `git push` is the only one; `PushQueuedCheckpointRefs`
at `manual_commit_push.go:285` exists but is git-refs-backend only); (3)
local-existence idempotency masks the never-synced state so login can't recover it.

**Impact**

- First-run-before-login (a common order) silently drops imported history from
  the dashboard with no obvious recovery.
- "Imported N turn(s)" is misleading — it means "written locally," not "synced."
- Recovery today requires connecting the repo, ensuring the remote has a branch,
  and making a commit + `git push` to fire the pre-push hook — undiscoverable.

**Expected behavior**

- Importing while logged out should say the data is local-only and won't appear
  in the dashboard until the user logs in and syncs (don't fail — keep offline
  import working).
- After login, a first-class way to sync already-imported (commit-less)
  checkpoints, without a manual commit + push.
- Success message should distinguish "imported locally" from "synced."

**Suggested fixes**

1. Warn on logged-out import (in `import_cmd.go` and `setup_import.go`
   `runSelectedImports`).
2. Provide a sync trigger for imported checkpoints: push on authenticated
   import/enable; or have `login`/`enable` detect unpushed imported checkpoints
   and offer to push; or add `entire import --sync` / `entire checkpoint push`.
   Needs a branch-backend push path usable outside the pre-push hook (parallel to
   `PushQueuedCheckpointRefs`).
3. Clarify the success message (local vs synced).

**Areas to update**: `cmd/entire/cli/agentimport/agentimport.go`,
`cmd/entire/cli/import_cmd.go`, `cmd/entire/cli/setup_import.go`,
`cmd/entire/cli/strategy/manual_commit_push.go`.

**Suggested tests**: logged-out warning shown / logged-in not; post-login sync
actually pushes prior imports (not a no-op); idempotency holds for the
already-synced case; success-message wording.

**Open question (server-side, out of this repo's scope)**: confirm with the
backend team that a connected-repo `git push` of `entire/checkpoints/v1` is
sufficient for dashboard ingestion (GitHub App), and whether the CLI should nudge
the "connect repo / install GitHub App" step when enabling while logged out.

sdshah09:
Fixes #1773 (fix 2 — sync after login). @karthik-rameshkumar handed this over on the issue after landing fix 1 (#1774).

## The problem

`entire import` writes read-only checkpoints and **creates no commit**, so the pre-push git hook — the only sync trigger — never fires for it. Because idempotency is keyed on *local* existence, history imported while logged out is stranded permanently: the post-login re-run reports `0 turn(s) (N already imported)` and makes no attempt to sync.

## The insight that keeps this small

The git-refs store **already enqueues every checkpoint ref as it writes it** (`gitRefsStore.setRef` → `enqueueForPush`, `checkpoint/refs_store.go:209`), and nothing drains that queue but a push. So turns imported while logged out are *still queued* afterwards — the data is packed and addressed, nobody mails it.

Draining the queue after a logged-in import is therefore enough to recover them, with **no change to the import idempotency logic** and no new push machinery.

## What this adds

`syncImportedCheckpoints` (new `import_sync.go`, ~70 lines), called from the two existing `warnIfImportNotSynced` sites — `import_cmd.go` and `setup_import.go`. It reuses `strategy.PushQueuedCheckpointRefs`, whose only current caller is `doctor migrate-checkpoints`' opt-in "push now"; this is the same foreground-drain shape.

**Deliberately not gated on `TurnsImported > 0`.** That is the fix: a logged-in re-run that imports nothing new still pushes, which is what makes it the recovery path.

Best-effort by design — the import already succeeded locally, so a push failure prints a `Note:` and returns rather than failing the command; `context.Canceled` stays silent.

## Scope: git-refs only

Since #1900 made git-refs the sole `entire enable` default, this covers every repo newly hitting the bug. git-branch has no push queue — its checkpoints ride the `entire/checkpoints/v1` branch the pre-push hook already ships — so it is skipped rather than paying for a checkpoint-policy sync to drain a queue it never fills. **Those repos keep syncing exactly as they do today; no regression, just not fixed here.** Happy to extend if you'd prefer one PR covering both.

I looked at `ManualCommitStrategy.PrePush` first (it covers both backends and has no production callers — only `PrePushFromGitHook` is wired, at `hooks_git_cmd.go:329`). I went with `PushQueuedCheckpointRefs` instead because it returns `(pushed, pushDisabled, err)` so the user can be told accurately what happened, `PrePush` passes `protectFirstUserBranch: false` and so skips the empty-remote guard, and it forces non-interactive SSH — right for a hook, wrong for a foreground command.

## Remote resolution

`resolveMigratePushRemote` → `resolveCheckpointPushRemote`, now shared with the migration command: both are non-hook drain paths and must resolve the elected checkpoint sync remote the same fail-closed way (a misconfigured `checkpoint_push_remote` is an error, never a fallback to origin). `entire import` gains `--remote` to override, matching `doctor migrate-checkpoints`.

## Verification

End-to-end against a real bare remote, identical sequence on both binaries:

| | 0.9.0 | this branch |
|---|---|---|
| import logged out | local only ✓ | local only ✓ |
| login + re-run output | `0 turn(s) (1 already imported)` | same |
| **refs on the remote** | **0** ❌ | **1** ✅ |
| push queue file | still present | drained |
| user told | nothing | `Synced 1 checkpoint(s) to origin.` |

Tests in `import_sync_test.go` cover: logged out never pushes; git-branch never pushes; success/disabled/empty-queue/failure/cancelled messaging; `--remote` override; remote-resolution failure reported not fatal; dry-run never pushes. The regression test is `TestImportSyncsAlreadyImportedTurns` — imports twice and asserts the second run **still** syncs despite importing nothing new.

`mise run check` green: lint 0 issues, unit + integration pass, e2e canary 60/60 (vogon) and 4/4 (roger-roger).

## Open question for a maintainer

The issue's own closing note is still unanswered, and it affects this PR:

> confirm with the backend team that a connected-repo `git push` of the checkpoint refs is sufficient for dashboard ingestion (GitHub App), and whether the CLI should nudge the "connect repo / install GitHub App" step when enabling while logged out.

Relevant because `reportRepoEnabled` (`setup.go:976`) is silently skipped when logged out, so a repo enabled before login is never reported to the backend. If pushing alone is not sufficient, this PR would print "Synced" while the dashboard stays empty, and the notice wording should probably also point at connecting the repo. Happy to follow up either way.

First contribution here — glad to adjust any of the above.


cmd/entire/cli/import_sync.go
```diff
@@ -0,0 +1,80 @@
...
+	// still paid for a checkpoint-policy sync, so skip the backend entirely.
+	cpCfg, cfgErr := settings.LoadCheckpointsConfig(ctx)
+	if cfgErr != nil || !checkpoint.PrimaryIsRefs(cpCfg) {
+		return
+	}
+
+	remote, err := importResolvePushRemote(ctx, explicitRemote)
+	if err != nil {
+		logging.Debug(ctx, "import: skipping checkpoint sync", "error", err.Error())
+		fmt.Fprintf(w, "Note: imported history is not synced yet: %v\n", err)
```
Soph: Surfaces `resolveCheckpointPushRemote`'s "pass --remote explicitly" — a flag `entire enable` doesn't have — and fires on the ordinary `git init && entire enable` no-remote-yet case, not only on errors.

cmd/entire/cli/import_sync.go
```diff
@@ -0,0 +1,80 @@
...
+		return
+	}
+
+	pushed, pushDisabled, err := importPushQueuedRefs(ctx, repo, remote)
+	switch {
+	case err != nil:
+		if errors.Is(err, context.Canceled) {
```
Soph: Can't match in production: `PushWithOptions` wraps an `*exec.ExitError` (`git.go:444`), so Ctrl-C surfaces as `signal: killed`, not `context.Canceled`. The test only passes because the stub returns the sentinel directly.

cmd/entire/cli/import_sync.go
```diff
@@ -0,0 +1,80 @@
...
+			return
+		}
+		logging.Warn(ctx, "import: checkpoint sync failed", "remote", remote, "error", err.Error())
+		fmt.Fprintf(w, "Note: could not sync imported history to %s: %v\n", remote, err)
+	case pushDisabled:
+		fmt.Fprintln(w, "Checkpoint pushing is disabled in settings; imported history stays local.")
+	case pushed == 0:
+		// Nothing queued — everything this import produced was already synced.
+	default:
+		fmt.Fprintf(w, "Synced %d checkpoint(s) to %s.\n", pushed, remote)
```
Soph: `pushed` is the whole repo-wide queue, not this import's refs, so 12 unpushed session checkpoints give `Imported 0 turn(s)… Synced 12 checkpoint(s)`. And under `checkpoint_remote` the refs go to that URL, not to `remote` — stderr and stdout then name different destinations.

cmd/entire/cli/setup_import.go
```diff
@@ -287,6 +287,9 @@ func runSelectedImports(ctx context.Context, w io.Writer, repoRoot string, selec
 	// Enable often runs before the user has logged in; surface once that a
 	// logged-out import stays local and won't reach the dashboard (issue #1773).
 	warnIfImportNotSynced(w, importedLocalHistory)
+	// When it runs logged in, push what was just imported: enable creates no
+	// commit either, so nothing else would trigger the pre-push hook (#1773).
+	syncImportedCheckpoints(ctx, w, repo, "")
```
Soph: Ctrl-C `break`s the loop above and lands here with a cancelled ctx: `git config` fails instantly and the swallow at `checkpoint_sync_remote.go:120` prints `no git remotes configured` in a repo that has them (or a false `checkpoint_push_remote` misconfiguration). `import_cmd.go:98` guards this; this path doesn't.

cmd/entire/cli/import_cmd.go
```diff
@@ -103,12 +104,17 @@ fails even with --dry-run.`, imp.AgentType()),
...
+				// "0 imported (N already imported)" is the recovery path (#1773).
+				syncImportedCheckpoints(ctx, c.OutOrStdout(), repo, remoteFlag)
 			}
 			return nil
 		},
 	}
 	cmd.Flags().StringVar(&pathFlag, "path", "", "Override the transcript directory to import from")
 	cmd.Flags().BoolVar(&dryRun, "dry-run", false, "Report what would be imported without writing")
 	cmd.Flags().StringSliceVar(&sessions, "session", nil, "Import only these session IDs (repeatable)")
+	cmd.Flags().StringVar(&remoteFlag, "remote", "", "Git remote to sync imported checkpoints to (default: the elected checkpoint sync remote)")
```
Soph: Worth validating this against the elected sync remote before use — see the gate note on `import_sync.go:65`.

cmd/entire/cli/import_sync_test.go
```diff
@@ -0,0 +1,245 @@
...
+	testutil.GitAdd(t, repoDir, "f.txt")
+	testutil.GitCommit(t, repoDir, "init")
+	t.Chdir(repoDir)
+	t.Setenv(settings.EnvCheckpointsPrimary, "git-refs")
+
+	claudeDir := t.TempDir()
+	require.NoError(t, os.WriteFile(filepath.Join(claudeDir, "s.jsonl"),
+		[]byte(`{"type":"user","uuid":"u1","message":{"role":"user","content":"hi"}}`+"\n"), 0o644))
+
+	calls := stubImportSync(t, true, func(context.Context, *git.Repository, string) (int, bool, error) {
```
Soph: Doesn't reproduce #1773: both runs are `loggedIn=true`, so the logged-out precondition never exists, and the push is stubbed — so the claim the PR rests on (refs written while logged out are *still queued* afterwards) is never exercised. Breaking enqueue would leave this green.

cmd/entire/cli/import_sync.go
```diff
@@ -0,0 +1,80 @@
...
+	importPushQueuedRefs    = strategy.PushQueuedCheckpointRefs
+)
+
+// syncImportedCheckpoints pushes checkpoints an import wrote to the checkpoint
+// sync remote, so imported history reaches the dashboard without waiting for an
+// unrelated `git push`.
+//
+// Import writes read-only checkpoints and creates no commit, so the pre-push git
+// hook — the only sync trigger — never fires for it. Before this, history
+// imported while logged out stayed local-only forever: import's idempotency is
```
Soph, line 31: For git-refs — the only backend this touches — the next ordinary `git push` already drains the **whole** queue regardless of login state (`prePushCheckpointRefs`), which is what `doctor migrate-checkpoints` tells users: "Refs are queued; they push on the next `git push`." So it isn't stranded forever, it's pending until the next push. Worth correcting here, since as written this will convince the next maintainer the queue has no other drain trigger.

That makes the value of this PR "imported history syncs promptly" rather than "recovers otherwise-unrecoverable history" — still worth having, just a different claim.

cmd/entire/cli/import_sync.go
```diff
@@ -0,0 +1,80 @@
...
+//
+// It therefore runs on every logged-in import, not only one that wrote new
+// turns. The git-refs store enqueues each checkpoint ref as it is written
+// (checkpoint.gitRefsStore.setRef) and nothing drains that queue but a push, so
+// turns imported while logged out are still queued afterwards and a re-run once
+// logged in is what recovers them.
+//
+// Best-effort by design: the import itself already succeeded locally, so a push
+// failure is reported and swallowed rather than failing the command.
+func syncImportedCheckpoints(ctx context.Context, w io.Writer, repo *git.Repository, explicitRemote string) {
```
Soph, line 43: No `WithNonInteractiveSSH` and no timeout on the batch push, so a passphrase-protected key prompts mid-`enable` and a hung remote blocks it. The only other caller (`doctor migrate-checkpoints`) gates its drain behind `CanPromptInteractively()` + a confirm; this has neither.

cmd/entire/cli/import_sync.go
```diff
@@ -0,0 +1,80 @@
...
+	if !importLoggedIn() {
+		// warnIfImportNotSynced has already explained that this stays local.
+		return
+	}
+
+	// git-branch primary has no push queue — its checkpoints live on the
+	// entire/checkpoints/v1 branch, which the pre-push hook ships alongside the
+	// user's own push. Draining a queue it never fills would be a no-op that
+	// still paid for a checkpoint-policy sync, so skip the backend entirely.
+	cpCfg, cfgErr := settings.LoadCheckpointsConfig(ctx)
```
Soph, line 53: `cfgErr` is dropped with no log and no output, so an invalid `checkpoints` block silently disables the whole fix — the same silent stranding #1773 is about. Both other error paths here log.

cmd/entire/cli/import_sync.go
Soph, line 61: Surfaces `resolveCheckpointPushRemote`'s "pass --remote explicitly" — a flag `entire enable` doesn't have — and fires on the ordinary `git init && entire enable` no-remote-yet case, not only on errors.

cmd/entire/cli/import_sync.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	remote, err := importResolvePushRemote(ctx, explicitRemote)
+	if err != nil {
+		logging.Debug(ctx, "import: skipping checkpoint sync", "error", err.Error())
+		fmt.Fprintf(w, "Note: imported history is not synced yet: %v\n", err)
+		return
+	}
+
+	pushed, pushDisabled, err := importPushQueuedRefs(ctx, repo, remote)
```
Soph, line 65: **Blocker:** `PushQueuedCheckpointRefs` has no `checkpointSyncAllowedForRemote` gate (only `prePush` does, `manual_commit_push.go:84`), and `--remote` is returned verbatim without checking it is a configured remote. So `import --remote fork` (or a pasted URL) pushes `refs/entire/checkpoints/*`, transcripts included, there. CLAUDE.md: "Pushes to any other remote or to a raw URL never carry checkpoint data."

Separately: the policy fetch inside runs *before* the queue-empty check, so every logged-in import pays a remote round-trip with nothing to push — `queue.Peek()` (`pushqueue.go:117`) makes that cheap to skip.

cmd/entire/cli/import_sync.go
Soph, line 68: Can't match in production: `PushWithOptions` wraps an `*exec.ExitError` (`git.go:444`), so Ctrl-C surfaces as `signal: killed`, not `context.Canceled`. The test only passes because the stub returns the sentinel directly.

cmd/entire/cli/import_sync.go
Soph, line 78: **Blocker:** `pushed` is the whole repo-wide queue, not this import's refs, so 12 unpushed session checkpoints give `Imported 0 turn(s)… Synced 12 checkpoint(s)`. Matters more if prompt sync is the point of the PR. And under `checkpoint_remote` the refs go to that URL, not to `remote` — stderr and stdout then name different destinations.

cmd/entire/cli/setup_import.go
Soph, line 292: Ctrl-C `break`s the loop above and lands here with a cancelled ctx: `git config` fails instantly and the swallow at `checkpoint_sync_remote.go:120` prints `no git remotes configured` in a repo that has them (or a false `checkpoint_push_remote` misconfiguration). `import_cmd.go:98` guards this; this path doesn't.

cmd/entire/cli/import_cmd.go
Soph, line 118: Worth validating this against the elected sync remote before use — see the gate note on `import_sync.go:65`.

cmd/entire/cli/import_sync_test.go
Soph, line 193: Doesn't reproduce #1773: both runs are `loggedIn=true`, so the logged-out precondition never exists, and the push is stubbed — so the claim the PR rests on (refs written while logged out are *still queued* afterwards) is never exercised. Breaking enqueue would leave this green.

cmd/entire/cli/doctor_migrate_test.go
```diff
@@ -23,7 +23,7 @@ func TestResolveMigratePushRemote_ExplicitValueReturnedVerbatim(t *testing.T) {
 	testutil.GitCommit(t, tmpDir, "init")
 	t.Chdir(tmpDir)
 
-	got, err := resolveMigratePushRemote(context.Background(), "explicit-remote")
+	got, err := resolveCheckpointPushRemote(context.Background(), "explicit-remote")
```
Soph, line 26: Rename the five `TestResolveMigratePushRemote_*` functions too — they now point at a symbol that no longer exists.

Soph (commented):
Thanks for this! The diagnosis is right, the queue insight is the right one, and I appreciate that the PR body pre-empts the "why not PrePush" question. I verified the load-bearing claim independently: `enqueueForPush` has no auth gating, so the refs really are still queued. Two things I'd want fixed before merge.

**Blockers**

1. `--remote` can route checkpoint data off the elected sync remote. `PushQueuedCheckpointRefs` has no `checkpointSyncAllowedForRemote` gate and the flag isn't validated, so `import --remote <anything>` pushes transcripts there. CLAUDE.md states the opposite invariant, so this is a contract break.
2. The reported count is the whole repo queue, not the import's refs — an import that finds nothing can print `Synced 12 checkpoint(s)`.

**One reframe, not a blocker**

For git-refs the next ordinary `git push` already drains the queue regardless of login state, so imported history is pending rather than stranded. That doesn't undercut the PR - syncing promptly instead of at the next push is worth having - but the docstring and the notice copy should say that, and it's why (2) matters: if prompt sync is the value, the count has to be about the import.

Relatedly, `warnIfImportNotSynced` still says "Log in with 'entire login' before importing", which reads as "too late for this history". Worth fixing in the same PR.

**Also**

- The four existing `TestRunSelectedImports_*` tests weren't stubbed, so they now run the real sync, which reads `ENTIRE_TOKEN` — not covered by the testdirs net. Add `stubImportSync` (or force `importLoggedIn` false) to them.
- No docs changes: `ref-checkpoint-backend.md` still names `PrePush` as the only drain, and CLAUDE.md needs the `--remote` behavior reflected once it's gated.

On your question regarding the push. The enable call is one step in the onboarding flow, if actual checkpoints are pushed then it does not matter if the enable call came before. 

Again thanks for opening this!

PR #1937, opened 2026-08-10
fix: stop checkpoint metadata.json outgrowing github's blob limit

### LINKED ISSUE #1927: Checkpoint metadata.json exceeds GitHub's 100 MB limit: files_touched is never deduplicated, and nested git worktrees are walked
## Summary

On a repo that uses git worktrees, `entire/checkpoints/v1` grows a
`metadata.json` large enough that GitHub refuses the push outright. Once that
happens the checkpoint branch can never be pushed again without rewriting its
history, and the failure surfaces only as a warning attached to an unrelated
`git push`.

Two separate causes compound. Either one alone would be survivable.

## Symptom

Every `git push` in the host repo now ends with:

```
[entire] Warning: failed to push entire/checkpoints/v1 after sync: push failed:
remote: error: File 53/596dc8fea0/1/metadata.json is 111.79 MB; this exceeds
GitHub's file size limit of 100.00 MB
remote: error: GH001: Large files detected.
! refs/heads/entire/checkpoints/v1 -> [remote rejected] (pre-receive hook declined)
```

The user's own branches push fine, so this is easy to miss for a long time —
it reads as noise attached to a successful push.

Environment: Entire CLI 0.8.42, go1.26.4, darwin/arm64.

## Cause 1 — `files_touched` is appended to, never deduplicated

That `metadata.json` is a single checkpoint record with
`"checkpoints_count": 213`. Measured on the blob:

| | |
|---|---|
| file size | 111.79 MB |
| lines | 1,292,267 |
| `files_touched` entries | 1,292,257 |
| **unique** entries | **67,340** |
| `files_touched` share of the file | 105.6 MB of 111.79 MB (94%) |

A 19× duplication factor. No single path dominates — the most repeated one
appears 77 times — so this is broad re-appending across checkpoints rather than
one pathological entry. Deduplicating alone takes the file to roughly 5.5 MB
and puts it back under the limit.

## Cause 2 — nested git worktrees are walked as if they were project files

Of the 67,340 unique paths, broken down by top-level segment:

| prefix | unique paths |
|---|---|
| `.claude/` (agent worktrees live here) | 43,184 |
| `worktrees/` | 21,833 |
| the actual project | ~1,241 |

96% of what is being recorded lives inside *other git checkouts* nested under
the repo root. Linked worktrees have their own `.git` and their own history;
their contents are not changes to the host repo. `node_modules` is correctly
excluded, so the walker already has an ignore mechanism — nested worktrees just
aren't in it.

This is the part that makes the growth unbounded in practice. Anyone driving
parallel coding agents accumulates worktrees, and each one multiplies the
recorded set.

## Scale of the resulting branch

| | |
|---|---|
| commits on `entire/checkpoints/v1` | 8,842 |
| files | 32,799 |
| unique blob bytes | 1,336.9 MB |
| distinct blobs over 50 MB | 3 |

Many identical 50 MB `full.jsonl` blobs exist, but git dedupes those by hash,
so they cost storage once. The 111.79 MB `metadata.json` is the only blob that
actually blocks the push.

## Suggested fixes, in order of leverage

1. **Deduplicate `files_touched` before writing.** One line, ~19× smaller, no
   behaviour change — the array is a set in everything but type.
2. **Do not descend into nested git worktrees.** A directory containing a
   `.git` entry that is not the repo root is a separate checkout. This is where
   the remaining ~50× lives.
3. **Store `files_touched` as a newline-delimited sidecar** rather than inline
   in a JSON object. It streams, it chunks, and it stops one growing field from
   deciding whether the whole record is writable.
4. **Check size before pushing and fail loudly.** A blob over the remote's
   limit is unrecoverable once committed; catching it at write time is cheap,
   and catching it at push time is too late. The current warning is easy to
   scroll past.

Fix 1 is the emergency stop. Fix 2 is the actual bug. Fixes 3 and 4 are what
stop the next version of this.

## Offer

Happy to send a PR for 1 and 2 — they look self-contained and the repo has CI
in `test.yml`. Say which shape you'd prefer for 2 (skip any dir containing
`.git`, versus consulting `git worktree list`), since that's a design call
rather than a mechanical one. I can also supply the checkpoint that triggered
this if a reproduction is useful, though it is 111 MB.

Disclosure: this report was researched and written with AI assistance
(Claude Code). Every number above is measured from the actual blob with
`git cat-file`, not estimated.

MuskanPaliwal:
fixes entireio/cli#1927 (fixes 1 and 2 from the issue; 3 and 4 discussed there as follow-ups).

## what this does

on repos with nested git checkouts under the repo root (agent worktrees, linked worktrees, vendored clones), a session's `files_touched` grew without bound until one checkpoint `metadata.json` blob crossed github's 100 MB limit and `entire/checkpoints/v1` became permanently unpushable. two independent defects compounded (measured in the issue: 1,292,257 entries, 67,340 unique, 96% of them inside other checkouts).

**1.** `fix(gitrepo)`**: exclude nested git checkouts from worktree status.**<br>real git treats a directory containing a `.git` entry as a repository boundary (`dir.c`) and never descends into it. go-git's status walk only skips entries *named* `.git` (`utils/merkletrie/filesystem/node.go`) and never checks whether a directory *contains* one, so it descends into every nested checkout and reports all its files as untracked. `gitrepo.Status`, already the enforced choke point for status reads, now drops untracked entries whose ancestor directory contains a `.git` entry (directory or worktree-style file; one memoized `Lstat` per unique ancestor). tracked entries are kept wherever they sit, matching git. this was the unbounded half: every new agent worktree multiplied the recorded set.

the walk cost itself (go-git still descends before the filter runs) is a go-git gap. i plan to raise an issue upstream, and likely a pr: the git-faithful fix is a boundary check in go-git's filesystem noder, which would benefit every go-git consumer.

**2.** `fix(checkpoint)`**: dedupe** `files_touched` **at the write boundary.**<br>every merge path already deduplicates (checked on v0.8.42 too), yet the reporter's blob proves duplicated paths reach the permanent record through some path that bypasses the merge helpers. instead of only hunting that producer, the session-metadata write in `writeSessionToSubdirectory` now passes `FilesTouched` through the existing `mergeFilesTouched` helper (dedupe, sort, slash-normalize), so no producer, past or future, can write duplicates into the immutable record. on the issue's numbers this alone shrinks the blob from 111.79 MB to roughly 5.5 MB. covers both git-backed stores (branch and refs) via the shared `treeWriter`; the fs store is test-only.

## why this shape

* the issue asked which form the nested-checkout skip should take: skip-any-dir-containing-`.git` vs `git worktree list`. this pr uses the former, since agent worktrees are not always worktrees *of the host repo*; a foreign clone under the repo root would slip past `git worktree list`.
* filtering at the `gitrepo.Status` choke point (rather than per consumer) fixes `DetectFileChanges`, attribution, and carry-forward at once, and the forbidigo rule already guarantees no caller can bypass it.
* untracked-only filtering matches git: index entries are always reported, even inside a directory that later grew a `.git`.

## tests

* `TestStatus_SkipsNestedCheckouts` (gitrepo): nested clone with `.git` directory, linked worktree with `.git` file, depth > 1, own untracked files kept, tracked file inside a nested checkout kept as modified. fails without the filter.
* `TestWriteCommitted_DeduplicatesFilesTouched` (checkpoint): duplicated input written, permanent record read back deduplicated and sorted. fails without the boundary dedupe.
* both verified red on the pre-fix code, green on this branch. `mise run lint` clean; `mise run test:ci` (unit + integration + e2e canary) green.

## manual verification

the same scenario run against a binary built from main and one built from this branch: a simulated claude-code session in a fresh repo, where a nested clone (`worktrees/agent-1`, `.git` directory) and a worktree-style checkout (`vendor/lib`, `.git` file, the form `git worktree add` creates) appear mid-session next to one real new file (`app.js`).

before (main):

```
what real git reports as untracked:
  ?? vendor/
  ?? worktrees/

what entire recorded as the session's files_touched:
  15 entries:
    vendor/lib/src/mod1.go
    vendor/lib/src/mod10.go
    vendor/lib/src/mod11.go
    vendor/lib/src/mod12.go
    vendor/lib/src/mod13.go
    vendor/lib/src/mod14.go
    vendor/lib/src/mod15.go
    vendor/lib/src/mod2.go
    vendor/lib/src/mod3.go
    vendor/lib/src/mod4.go
    vendor/lib/src/mod5.go
    vendor/lib/src/mod6.go
    vendor/lib/src/mod7.go
    vendor/lib/src/mod8.go
    vendor/lib/src/mod9.go
```

after (this branch):

```
what real git reports as untracked:
  ?? vendor/
  ?? worktrees/

what entire recorded as the session's files_touched:
  0 entries:
```

every file inside the worktree-style checkout was recorded as session work by the main build, none by this branch. with 15 files it is noise; with the issue's 65k files across 213 checkpoints it is 111 MB. note the committed checkpoint for the `app.js` commit shows `["app.js"]` in both runs, since commit-time condensation filters to the commit's files; the foreign entries accumulate in session state and reach permanent records through the paths that don't filter (carry-forward across checkpoints, session-end condensation).

a side observation from the before run's log: entire's real-git-based untracked collection returned `worktrees/agent-1/` as a single boundary entry, which the shadow-tree writer then rejected (`WARN: skipping invalid git tree path ... path contains empty segment`), while go-git's walk descended into `vendor/lib` and reported its files individually. two walkers disagreeing about where the repository ends.

## what this does not do

* it cannot repair branches that already carry an oversized blob. git history is immutable, so affected users need a one-time reset/rewrite of `entire/checkpoints/v1`.
* fix 3 (sidecar storage for `files_touched`) and fix 4 (size guard at write time) from the issue are intentionally out of scope to keep this reviewable; sketches for both are on the issue.


ecgang:
I reviewed this and left the full write-up on the issue (entireio/cli#1927 (comment)) to keep that thread together, but two of the notes are about this PR specifically, so repeating them here so they don't get lost:

- **A `gitRefsStore` case in `TestWriteCommitted_DeduplicatesFilesTouched`.** It only builds a `NewGitStore` today. The shared `*treeWriter` means git-refs is covered structurally — I traced it, `refs_store.go:47` embeds it — but it's covered by argument rather than by test, and this bug reached a permanent record once already. Same fixture, a few lines.
- **`mergeFilesTouched` returns nil for empty input, which moves the JSON.** `api/checkpoint/metadata.go:353` and `:506` tag `files_touched` without `omitempty`, so an empty non-nil slice marshals as `[]` and a nil one as `null`. Reachable: `committedFilesExcludingMetadata` returns `make([]string, 0, len(committedFiles))` non-nil when every committed file is filtered out as protected. Harmless to a sane reader, but it's a wire-format change riding along with a bug fix.

Neither is a defect. I verified the fix does what it says: reverting just the two non-test files, both new tests fail — the status test with all three nested-checkout paths present in the map, which reproduces the original bug — and `mise run lint` is clean at 0 issues.

Also, `fsstore.go:321` — I'd called it a live gap in the issue before reading the file header. It isn't; it's deliberately unregistered. The reason to fix it anyway is that it's the worked example someone copies for the next backend.

MuskanPaliwal:
> I reviewed this and left the full write-up on the issue (#1927 (comment)) to keep that thread together, but two of the notes are about this PR specifically, so repeating them here so they don't get lost:
> 
> * **A `gitRefsStore` case in `TestWriteCommitted_DeduplicatesFilesTouched`.** It only builds a `NewGitStore` today. The shared `*treeWriter` means git-refs is covered structurally — I traced it, `refs_store.go:47` embeds it — but it's covered by argument rather than by test, and this bug reached a permanent record once already. Same fixture, a few lines.
> * **`mergeFilesTouched` returns nil for empty input, which moves the JSON.** `api/checkpoint/metadata.go:353` and `:506` tag `files_touched` without `omitempty`, so an empty non-nil slice marshals as `[]` and a nil one as `null`. Reachable: `committedFilesExcludingMetadata` returns `make([]string, 0, len(committedFiles))` non-nil when every committed file is filtered out as protected. Harmless to a sane reader, but it's a wire-format change riding along with a bug fix.
> 
> Neither is a defect. I verified the fix does what it says: reverting just the two non-test files, both new tests fail — the status test with all three nested-checkout paths present in the map, which reproduces the original bug — and `mise run lint` is clean at 0 issues.
> 
> Also, `fsstore.go:321` — I'd called it a live gap in the issue before reading the file header. It isn't; it's deliberately unregistered. The reason to fix it anyway is that it's the worked example someone copies for the next backend.

thanks for the careful read. all four addressed in [9f81bdf:](https://github.com/entireio/cli/pull/1937/commits/9f81bdf6f5c20c897f7298b2b9fe33cebda4423d)
- `dedup` test now runs against both backends through a small consumer-side interface, so a future shadowing or de-embedding of the shared writer fails a named test instead of passing silently.
- the nil-vs-empty change is gone. normalization now goes through an exported `NormalizeFilesTouched` that preserves the caller's nil-versus-empty, and a raw-JSON test `pins []` staying `[]` and `null` staying `null` (asserted on the blob text, since unmarshaling erases exactly that difference).
- `fsstore` uses the same exported helper,  agreed it's the example the next backend copies, which is also what forced the helper to become exported and documented rather than package-private.
- comment on `filterNestedCheckouts` now states the deliberate divergence: quieter than git about the boundary entry, with a note not to "fix" it.
thanks for the careful read, the producer-hunt writeup especially, that saves the next person a day.

MuskanPaliwal:
hi @blackgirlbytes , would appreciate a review on this pr. thanks :))

MuskanPaliwal:
all four fixed in 9f81bdf. refs store has its own test case now, empty `files_touched` stays `[]` on the wire (pinned with a raw-json test since unmarshaling hides exactly that difference), fsstore uses the shared helper, and the filter comment now says where it is deliberately quieter than git. thanks for the thorough review, this was easy to act on.

Soph:
Thanks for doing this!

Fix 2 looks good and I'd approve it separately. But fix 1 feels like it's potentially hiding the actual issue. Which CLI version wrote that blob, and does it repro on main? Can you drop the two checkpoint commits so we can merge the gitrepo fix now, and we'll track the dedupe separately?

MuskanPaliwal:
> Thanks for doing this!
> 
> Fix 2 looks good and I'd approve it separately. But fix 1 feels like it's potentially hiding the actual issue. Which CLI version wrote that blob, and does it repro on main? Can you drop the two checkpoint commits so we can merge the gitrepo fix now, and we'll track the dedupe separately?

hi @Soph, splitting makes sense. i have dropped the two checkpoint commits from this branch, so it now carries only the gitrepo fix; i have parked the dedupe (plus its tests and the empty-[] follow-up) on a separate branch for its own PR.

on the version question: the reporter's environment is 0.8.42, but that blob is a single record accumulated across 213 checkpoints on a branch with 8,842 commits, so many CLI versions likely contributed to it over time. on repro: the write boundary preserves whatever it's handed (the new test shows duplicates passed to WriteCommitted persist verbatim), but neither i nor @ecgang could find any in-tree producer that generates duplicates — every merge path dedupes (ExtractModifiedFiles, filterFilesTouched, RecordFilesTouched, both SaveStep paths, resolveFilesTouched; ecgang independently re-audited and came up empty too, details in their comment on #1927). so my best theory is the duplicates were written by a version older than 0.8.42, and current main only fails to reject them, not produce them.

on the worry that the dedupe hides the actual issue: fair, and i'd rather it didn't. in the follow-up PR i can make the boundary dedupe log a warning with counts whenever it removes duplicates, so if a live producer exists it surfaces in the logs instead of being silently masked. does that work for you, or would you rather hold the dedupe until the producer is found?

Soph (approved):
Thanks a lot!

PR #1939, opened 2026-08-10
Detect hook-config drift for every agent, not just Claude Code

Soph:
https://entire.io/gh/entireio/cli/trails/1000

## Why

`entire agent add pi` in a repo with a committed extension failed with:

```
failed to setup pi hooks: refusing to overwrite foreign file at
  .pi/extensions/entire/index.ts; remove it or pass --force
```

That refusal was correct — the file predated the `Auto-generated by ...` header, so the installer couldn't recognise it as Entire's. But chasing it surfaced a real gap behind it.

Repos commonly **commit** `.pi/extensions/entire/index.ts` and `.opencode/plugins/entire.ts`, so every clone gets checkpointing without each person enabling it by hand — forgetting to enable means losing sessions. The committed copy then goes stale as the template evolves, and nothing notices:

- `AreHooksInstalled` only checks for the marker substring, so a stale-but-marked file reads as installed forever.
- The extension's own `fireHook` swallows every error by design, so broken hooks are silent.

Result: the repo reports healthy while its hooks no-op. Claude Code was the only agent with a drift check (`CheckHookConfig`), and `status`/`doctor` hardcoded it by name, so no other agent could ever report drift regardless of what it implemented.

## What changed

- **`agent.HookFreshness`** — optional capability answering "is what's installed still what we'd write today?", distinct from `AreHooksInstalled`'s "is Entire wired up here at all?".
- **`agent.GeneratedHookFileState`** — shared implementation for agents whose whole integration is one generated file.
- **pi + opencode** implement it; **claudecode**'s enum becomes an alias of the shared type (no behaviour change, existing call sites untouched).
- **`status` + `doctor`** iterate installed agents via `AsHookFreshness` instead of naming claude-code. `status --json`'s `hooks_outdated` was already a list of agent names, so no schema change.

### Design note

Drift is detected by comparing **exact content against every render `InstallHooks` could have written**, not a stamped template version. Any edit to an embedded template then makes installed copies read as outdated automatically, with no version constant anyone can forget to bump. Matching both the production and local-dev renders is what keeps a `--local-dev` install from reading as permanent drift (`LocalDevHookScript` is a fixed const, so this is exact and machine-independent).

A **foreign** file reads as `HooksAbsent`, not `HooksOutdated` — `InstallHooks` refuses to overwrite it, so warning would nag about a file the CLI won't fix.

## Verification

`mise run check` passes (lint 0 issues; unit + integration + canary E2E green). Each of the four commits builds independently, so bisect stays clean.

It found genuine drift on first run against a real repo:

```
  Agents · Claude Code, Codex, OpenCode, Pi
  ! Claude Code hooks out of date · run 'entire enable --force'
  ! OpenCode hooks out of date · run 'entire enable --force'
  ! Pi hooks out of date · run 'entire enable --force'
```

The pi entry was a committed extension written before ed42608db (`ENTIRE_CMD` double-quoted instead of single-quoted) — exactly the silent-breakage class this is meant to catch.

## Trade-off worth a reviewer's eye

Drift is measured against **the checking binary's** template. With a committed extension and a team on mixed CLI versions, whoever runs the newest CLI sees the warning until someone re-runs `--force` and commits. That's the intended signal for a committed generated file, but it does mean a yellow status line during version skew rather than silence.

---

> [!NOTE]
> **Low Risk**
> Read-only diagnostics and additive agent interfaces; behavior change is surfacing warnings users were missing when committed Pi/OpenCode extensions or Claude matchers were stale.
> 
> **Overview**
> Introduces **`HookFreshness`** and shared **`HookConfigState`** so diagnostics can tell whether installed hooks still match what **`InstallHooks`** would write today—separate from **`AreHooksInstalled`**, which can stay true for committed generated files that have gone stale.
> 
> **Pi** and **OpenCode** implement drift checks via **`GeneratedHookFileState`**, comparing on-disk plugin/extension bytes to current production and local-dev renders (foreign files at the path count as absent, not outdated). **Claude Code** keeps its matcher-based logic but aliases the shared enum and wires **`CheckHookConfig`** through the interface.
> 
> **`OutdatedHookAgents`** drives **`entire status`** (text warnings per agent; **`--json`** **`hooks_outdated`** lists all outdated agents) and **`entire doctor`**’s generalized **`checkHookDrift`**, replacing the Claude-only hardcoded path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 834bfc7f28f7a70fb38b56696144b0a3621d6640. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/agent/agent.go
+++ b/cmd/entire/cli/agent/agent.go
@@ -120,6 +120,42 @@ type HookSupport interface {
 	AreHooksInstalled(ctx context.Context) bool
 }
 
+// HookConfigState describes how an agent's installed Entire hook config
+// compares to what InstallHooks would write today.
+type HookConfigState int
+
+const (
+	// HooksAbsent means Entire hooks are not installed for this agent here.
+	HooksAbsent HookConfigState = iota
+	// HooksCurrent means the installed hooks match what would be written today.
+	HooksCurrent
+	// HooksOutdated means Entire hooks are installed but stale — an older CLI
+	// wrote a config that no longer matches the current one. Fix:
+	// `entire enable --force`.
+	HooksOutdated
+)
+
+// HookFreshness is implemented by hook-supporting agents that can report
+// whether their installed config has drifted from the current one.
+//
+// AreHooksInstalled answers "is Entire wired up here at all?" — for agents
+// whose hook config is a generated file checked into the repo, that stays true
+// forever even after the generated content goes stale, because the file is
+// still present and still recognisably Entire's. CheckHookConfig answers the
+// separate question "is what's installed still what we'd write today?", so
+// `entire status` and `entire doctor` can flag a stale config instead of
+// reporting it healthy while its hooks silently no-op.
+//
+// Implementations must be read-only: they are diagnostics and must never
+// modify the agent's config.
+type HookFreshness interface {
+	Agent
+
+	// CheckHookConfig reports whether this agent's Entire hook config is
+	// absent, current, or outdated in the current repo.
+	CheckHookConfig(ctx context.Context) HookConfigState
+}
+
 // FileWatcher is implemented by agents that use file-based detection.
 // Agents like Aider that don't support hooks can use file watching
 // to detect session activity.
--- a/cmd/entire/cli/agent/capabilities.go
+++ b/cmd/entire/cli/agent/capabilities.go
@@ -66,6 +66,15 @@ func AsHookSupport(ag Agent) (HookSupport, bool) {
 	return declaredCapability[HookSupport](ag, func(c DeclaredCaps) bool { return c.Hooks })
 }
 
+// AsHookFreshness returns the agent as HookFreshness if it implements the
+// interface. No capability declaration is needed: hook-config drift detection
+// is built-in only, since it compares against a template the CLI itself
+// embeds. External agents own their hook config and report installation state
+// through their own protocol.
+func AsHookFreshness(ag Agent) (HookFreshness, bool) {
+	return builtinCapability[HookFreshness](ag)
+}
+
 // AsTranscriptAnalyzer returns the agent as TranscriptAnalyzer if it both
 // implements the interface and (for CapabilityDeclarer agents) has declared the capability.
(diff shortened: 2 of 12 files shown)


Soph:
Addressed the bot findings in 57b96804e.

**CRLF false positive** (Bugbot + Copilot, independently) — real, and specifically bad for this PR's premise. Confirmed with a throwaway probe before fixing:

```
CheckHookConfig on CRLF checkout = 2 (HooksCurrent=1)
BUG CONFIRMED: identical content with CRLF endings reads as drift
```

These files are generated with LF but are typically **committed**, so on Windows a checkout under git's default `core.autocrlf=true` yields CRLF and byte equality never holds. Worse, it was unclearable: `--force` writes LF back, and the next checkout re-converts it. `GeneratedHookFileState` now normalizes CRLF→LF on both sides, with a regression test that simulates the checkout. Lone CR is left alone deliberately — no generator here emits it, and rewriting it would mask genuinely different content.

Worth noting why CI didn't catch it: the tests write the file through `InstallHooks` (Go `os.WriteFile`), which never goes through git's conversion, so `test-windows` passed while the bug was live.

**`OutdatedHookAgents` doc comment** (Copilot) — correct, the comment was wrong. It claimed an agent whose config exists in the repo but that nobody enabled stays quiet. For generated-file agents the committed file *is* the installation, so such a repo does get warned. That behaviour is intended (the repo is relying on that file to work), but the comment said the opposite. Rewritten to describe what actually happens.

**Test naming** (Copilot, suppressed) — renamed `TestCheckClaudeCodeHookDrift_*` → `TestCheckHookDrift_*`, since they now exercise the generalized checker via the Claude Code path.

No change to the `normalizeLineEndings` placement question if a reviewer prefers it in `stringutil` — it's unexported in `agent` next to its only caller for now. `stringutil.CollapseWhitespace` was not reusable here: it collapses all whitespace runs to a single space, which would make genuinely different templates compare equal and silently disable drift detection.

`mise run check` green (lint 0 issues; unit + integration + canary E2E).

nodo (approved):

PR #1943, opened 2026-08-10
fix(search): 1 MiB cell response cap truncated large-region results; surface per-region fan-out errors

alishakawaguchi:
https://entire.io/gh/entireio/cli/trails/1005

## Problem

`entire search --all-repos` (or `repo:*`) fans out to every regional query-serve cell in parallel and merges the results. On high-hit queries (e.g. `transcript`, `backfill`, `indexer`), the busiest cell silently dropped out of the results and the CLI printed only `search failed in 1 of 4 regions; results may be incomplete` — with no way to tell which region or why. **Silent, incomplete cross-region results.**

## Root cause

**Bug 2 — the defect.** `search.CellV4` read each cell's response body with `io.ReadAll(io.LimitReader(resp.Body, 1<<20))` — a hard **1 MiB** cap. The largest cell returns 2+ MB on high-hit queries (result rows carry full transcript text). `LimitReader` truncated the body mid-JSON, `json.Unmarshal` failed, `CellV4` returned `unexpected response from search service`, and that cell was classified *failed* — dropping its whole region from the merge. Volume-dependent, not region health / version skew / latency.

**Bug 3 — why it was undiagnosable.** The partial-failure warning printed only a **count**; the per-cell label + error went to `logging.Debug`, which the `search` command discards (it never calls `logging.Init`, so the default logger sits at INFO and Debug is dropped). The operator could not see which region failed or why.

## Fix

**Bug 2 — bounded read with explicit overflow, not a silent raised cap.** Reads go through a new `readCappedBody` helper with a **64 MiB** guard (~30× real payloads; a DoS/runaway bound, not a functional limit — the page is already bounded by the per-request result limit). On overflow it returns an explicit `exceeded the N-byte read limit` error instead of feeding truncated JSON to the decoder, so the silent-truncation failure class is gone.

**Bug 3 — name the region and reason.** The warning now lists each failed region with a concise reason. Since a cell error can embed the raw response body, `summarizeCellError` collapses it to a single, rune-safe, truncated line — the terminal never gets a multi-MB blob. The `N of M regions` prefix and the machine-readable `Response.Warnings` contract are preserved.

### Design decision: cap vs. stream

I kept a **buffered read with a bounded cap + explicit overflow error** rather than switching `CellV4` to `json.NewDecoder(resp.Body)`. Streaming would remove the magic number, but:

- both the non-200 path and the decode-error path dump `string(body)` for diagnostics — streaming loses the raw body;
- a raw `json.NewDecoder` over `resp.Body` has **no memory bound** (DoS); wrapping it back in a `LimitReader` just turns an over-cap payload into `unexpected EOF` — silent-ish again, and still no raw body.

Buffered + `maxBytes+1` overflow detection keeps a hard memory bound **and** makes overflow loud **and** preserves the raw-body diagnostics — strictly better here.

## Tests

- `TestCellV4_LargeResponseDecodesFully` — a valid JSON body **> 1 MiB** now decodes intact (regression for Bug 2).
- `TestReadCappedBody` — at/under cap returns intact; one byte over is an explicit error, not a truncated slice.
- `TestMergeSemanticV4Responses_PartialFailureNamesRegionAndReason` — the warning names the region + a concise reason, truncates an embedded multi-MB body, and keeps the `1 of 2 regions` prefix (Bug 3). Existing `1 of N regions` assertions stay green.

`gofmt`, `go vet ./...`, `mise run lint` (0 issues), and the search suites all pass.

https://claude.ai/code/session_01VccsQTwYMBjeTC8aJQ3wTP

---

> [!NOTE]
> **Medium Risk**
> Changes cross-region search merge behavior and user-visible errors for a core CLI path; bounded reads and sanitization reduce DoS/terminal risks from large or untrusted bodies.
> 
> **Overview**
> Fixes **cross-region semantic search** dropping busy cells when responses exceed the old **1 MiB** body cap. `CellV4` now reads via `readCappedBody` with a **64 MiB** guard; bodies over the cap return an explicit limit error instead of silent truncation that broke JSON decode and removed whole regions from the merge.
> 
> **Partial and total failures** are easier to diagnose: `mergeSemanticV4Responses` warnings name each failed region plus a short reason (still using the `N of M regions` prefix). `summarizeCellError` strips non-printable runes and truncates reasons so embedded response bodies cannot flood the terminal or debug logs; all-cells-failed errors use the same summarization except **`context.Canceled`** stays wrapped for Ctrl-C handling.
> 
> Regression tests cover large JSON decode, read-cap behavior, warning content, error truncation, and cancellation chaining.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fb983446380db12d116bf443fc7bf0327dcdce8e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/search/search.go
+++ b/cmd/entire/cli/search/search.go
@@ -19,6 +19,16 @@ import (
 
 const apiTimeout = 30 * time.Second
 
+// maxResponseBytes bounds a single cell's response body. It is a DoS/runaway
+// guard, not a functional limit: a high-hit query against the largest cell
+// returns a few MB of result text (rows carry full transcript snippets), so the
+// cap sits ~30x above any legitimate page (which is itself bounded by the
+// per-request result limit). The previous 1 MiB cap silently truncated big
+// cells mid-JSON, failing the decode and dropping the whole region from a
+// cross-cell fan-out; readCappedBody now turns an over-cap body into an
+// explicit error instead of feeding truncated JSON to the decoder.
+const maxResponseBytes = 64 << 20 // 64 MiB
+
 // v4ServicePath is the per-repo v4 query-serve route exposed by the entire-api
 // cell gateway. It takes repo=<ULID>. The BFF (entire.io /api/v1/search)
 // forwards to this same path; the CLI dials the cell directly with a
@@ -662,9 +672,9 @@ func CellV4(ctx context.Context, client *api.Client, cfg Config, repoIDs []strin
 	}
 	defer resp.Body.Close()
 
-	body, err := io.ReadAll(io.LimitReader(resp.Body, 1<<20))
+	body, err := readCappedBody(resp.Body, maxResponseBytes)
 	if err != nil {
-		return nil, fmt.Errorf("reading response: %w", err)
+		return nil, err
 	}
 	if resp.StatusCode == http.StatusNotFound {
 		// The gateway has no semantic-search route (plain "404 page not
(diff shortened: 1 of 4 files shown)


alishakawaguchi:
bugbot run

alishakawaguchi:
bugbot run

PR #1946, opened 2026-08-11
build(scoop): keep pre-rename 'cli' installs updating and migrate them to entire

gtrrz-victor:
## Entire trail

[#1023 — Fix Scoop cli->entire upgrade path](https://entire.io/gh/entireio/cli/trails/1023)

## Summary

Keep existing Windows Scoop installs updating after the package rename from `cli` to `entire`, then guide users through a failure-safe migration to the canonical package.

## What Changed

- Publish `cli.json` as a temporary transition manifest alongside the canonical `entire.json`, so existing `cli` installs continue receiving releases.
- Detect whether the running binary belongs to Scoop's legacy `cli` app directory or the renamed `entire` directory.
- For legacy installs, print a Windows command that installs the replacement before removing the working package:

  ```text
  cmd.exe /D /C "scoop install entire/entire && scoop uninstall entire/cli && scoop reset entire"
  ```

  Explicit `cmd.exe` invocation makes the command pasteable from cmd or PowerShell. The `&&` chain prevents uninstall when installation fails, so a stale bucket or transient error never removes the working CLI. The uninstall→reset link is weaker — `scoop uninstall` ends in an unconditional `exit 0` — but that is harmless: reset is a no-op on the Scoop versions where it isn't needed.
- Keep Windows updates print-only because a running `entire.exe` can lock the Scoop shim that the migration must replace.
- Scope `scoop reset` correctly in the comments: current Scoop already preserves the active shims across the uninstall (installing `entire` renames the displaced shim to `entire.shim.cli` via `warn_on_overwrite`, and uninstalling `cli` removes only that suffixed copy via `rm_shim`), so reset is belt-and-braces for older Scoop rather than the step keeping `entire.exe` and the git remote helper alive.
- Update the transition manifest message, README migration instructions, and `install.sh` Windows fallback to use `entire/entire`.
- Do not add an `entire` dependency to the transition manifest; the explicit migration controls shared-shim ownership and ordering.

The transition manifest should be removed after legacy installs have had a reasonable migration window.

## Verification

- `mise run fmt`
- golangci-lint v2.11.3 config verification and full lint: 0 issues
- shellcheck, module, mise, and gofmt lint tasks
- `mise run test:ci`
  - unit and integration tests with race detection
  - Vogon deterministic E2E canary: 60 passed
  - Roger Roger deterministic E2E canary: 4 passed
- `goreleaser check .goreleaser.yaml`
- `git diff --check`

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -121,6 +121,33 @@ scoops:
     license: MIT
     skip_upload: auto
 
+  # Transition manifest for the pre-rename package name. Before the "entire"
+  # manifest existed, goreleaser defaulted the name to the project name, so
+  # existing Windows installs are registered as app "cli" and `scoop update`
+  # resolves their updates from cli.json — dropping it orphans those installs
+  # silently (no error, no more updates). Keep publishing it, depending on the
+  # canonical "entire" package so one `scoop update` migrates users without a
+  # forced reinstall, and nudge cleanup via post_install. Drop this entry a few
+  # releases after migration has largely completed.
+  #
+  # Note: both apps ship the same entire.exe shim; scoop lets the most recently
+  # installed app own it, and `scoop uninstall cli` can take the shared shim
+  # with it — hence the `scoop reset entire` step in the message below.
+  - name: cli
+    repository:
+      owner: entireio
+      name: scoop-bucket
+      token: "{{ .Env.TAP_GITHUB_TOKEN }}"
+    homepage: "https://github.com/entireio/cli"
+    description: "CLI for Entire (renamed to 'entire' — install that instead)"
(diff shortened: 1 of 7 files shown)


gtrrz-victor:
Follow-up polish from review:

- Extracted `normalizedExecPath()` helper — `scoopAppName()` and `installManagerForCurrentBinary()` no longer duplicate the exec-path resolve/normalize logic.
- Reworded the Windows auto-update comment so it leads with "no installer can replace a running exe," with the Scoop shim lock as the acute case (was reading as Scoop-only).

No behavior change; versioncheck tests + vet clean.

Soph (approved):

PR #1948, opened 2026-08-11
fix(claude-code): preserve permissions when removing hooks

### LINKED ISSUE #1924: agent remove claude-code deletes unrelated .claude/settings.json keys
## Summary

`entire agent remove claude-code` removes unrelated keys from the repository's `.claude/settings.json` instead of removing only Entire-managed Claude hooks.

## Environment

- Entire CLI: `0.9.0`
- OS: Windows amd64
- Repository already enabled for both Codex and Claude Code

## Reproduction

Start with a project-level `.claude/settings.json` containing Entire-managed hooks plus an unrelated permission:

```json
{
  "hooks": {
    "SessionStart": [
      {
        "matcher": "",
        "hooks": [
          {
            "type": "command",
            "command": "entire hooks claude-code session-start"
          }
        ]
      }
    ]
  },
  "permissions": {
    "deny": [
      "Read(./.entire/metadata/**)"
    ]
  }
}
```

Run:

```text
entire agent remove claude-code
```

CLI output:

```text
Removed Claude Code hooks.
```

Actual resulting file:

```json
{}
```

The unrelated `permissions.deny` setting is removed together with the Entire hooks.

## Expected Behavior

Only Entire-managed Claude Code hook entries should be removed. Unrelated top-level keys and unrelated hook entries should be preserved.

For the example above, the resulting file should be:

```json
{
  "permissions": {
    "deny": [
      "Read(./.entire/metadata/**)"
    ]
  }
}
```

## Workaround

Inspect the diff immediately after agent removal and manually restore unrelated Claude settings.

## Acceptance Criteria

- [ ] `entire agent remove claude-code` removes only Entire-owned hooks.
- [ ] Unrelated `.claude/settings.json` keys survive unchanged.
- [ ] A regression test covers mixed Entire and non-Entire Claude settings.

MuskanPaliwal:
Fixes #1924

`entire agent remove claude-code` is documented as uninstalling Claude Code hooks, but it also removed `Read(./.entire/metadata/**)` from `permissions.deny`. When that was the only permission and the file contained only Entire hooks, both top-level objects were deleted and `.claude/settings.json` became `{}`.

This PR removes the permissions cleanup from `UninstallHooks`. Agent removal now filters Entire-managed entries under `hooks` and leaves the rest of the shared Claude settings file alone. Permissions, user-defined hooks, unknown hook types, and unrelated top-level settings are preserved. A file containing only Entire hooks and no other settings still correctly becomes `{}`.

`Windows amd64` in the issue records the reporter's environment; the issue does not identify this as an OS-specific bug. The affected code path is platform-independent JSON parsing and rewriting.

## Testing

- Added regression coverage for permissions containing Entire and user rules, additional permission fields, and unknown nested fields
- Covered settings containing only Entire hooks, mixed Entire and user hooks, and unrelated top-level settings
- Ran `mise run lint`
- Ran `mise run test`
- Ran `mise run test:ci`
- Ran the Claude Code package with the race detector
- Cross-compiled the Claude Code test package for Windows amd64
- Manually built the CLI and verified that `agent add claude-code` followed by `agent remove claude-code` removes all seven hooks while preserving the complete permissions object


MuskanPaliwal:
I took another look, and I think I interpreted the issue too broadly. Entire installs `Read(./.entire/metadata/**)`, so removing it along with the hooks is reasonable cleanup. The existing code already preserves other permissions and settings.

I’m going to pause this PR until we know whether the reporter already had that rule before enabling Entire, or whether they want to remove Claude hooks while keeping Entire active for another agent. Depending on the answer, I’ll either narrow the fix or close the PR.

PR #1954, opened 2026-08-11
feat: lazy invisible enable for globally tracked repos (global enable: B)

peyton-alt:
https://entire.io/gh/entireio/cli/trails/1013

## Summary

Split **B** of #1894 (stacked on #1953 — see the stack map there). Adds **lazy invisible enable**: the first hook activity in a globally-tracked repo (no repo-level setup, global tier on) transparently sets the repo up with everything inside `.git/` — the five git hooks installed, clone preferences marked, checkpoint ref ensured — and all runtime state (`.entire/metadata`, `logs`, `tmp`) routes under `<git-common-dir>/entire/worktree/<worktree-key>/` instead of the worktree.

**The product promise: `git status --porcelain` stays byte-clean.** Pinned by:
- `TestGlobalEnable_LazyInvisibleSetup` — empty porcelain at three points through a full prompt → stop → commit cycle (the commit carries its `Entire-Checkpoint` trailer with all metadata under `.git/`)
- `TestGlobalEnable_TierAbsent_CreatesNothing` — the negative: tier absent → zero footprint
- `TestAbsPath_InvisibleRouting_RepoLevelSetupWins` — repos with any repo-level setup resolve **byte-identically to before**

The runtime location is namespaced **per worktree** using the same hash derivation shadow branches use (`paths.HashWorktreeID`, moved from `checkpoint` with a pure delegation — existing shadow-branch names cannot change), so linked worktrees of one clone never share or steal each other's session data (`TestAbsPath_InvisibleRouting_LinkedWorktree` pins isolation).

**Cherry-pick note:** the second commit carries the *namespacing portion* of a larger reviewed fix — its diagnostics-routing portion lands in split D, migration scoping in split C. One doc comment in `paths/invisible.go` forward-references the enable-time migration that arrives in split C (build-safe; lands one layer up).

Stack rule: fix-forward only — review fixes land as new commits and merge forward; no rebases. Merge order: #1953 → this → #1955 → #1957. All commits dual-reviewed (spec + adversarial quality with mutation testing); findings on trail #968. Part of #1098.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1953, opened 2026-08-11
feat: user-global settings tier + hook gates (global enable: core)

peyton-alt:
https://entire.io/gh/entireio/cli/trails/1012

This is **PR A of 4 sequential splits** of #1894. #1894 grew large and is being decomposed into reviewable merges now that development is finished; each split carries code that was already dual-reviewed (spec review + adversarial quality review, findings on trail #968).

## What this PR does

- **User-global settings tier** at `~/.config/entire/settings.json` (resolved via `userdirs`): `global.enabled`, plus `exclude_paths` (doublestar-matched against the worktree root) and `exclude_origins` (doublestar-matched against normalized `host/owner/repo`). Matching is fail-closed: strict JSON decoding, fail-closed origin lookup (new `gitremote.GetRemoteURLsInDirIfSet` distinguishes "remote unset" from "lookup failed"), case-folded path matching on case-insensitive filesystems, and trailing-slash pattern normalization.
- **`IsActiveForRepo` predicate**: repo-level setup always wins, an explicit repo-level disable is a veto, and every error path fails closed. The global tier is a **fallback, not a merge layer** — a repo with any repo-level setup keeps exactly its current behavior.
- **All three hook gates unified on it**: the agent-hook gate, the git-hook gate, and the defense-in-depth check inside `initHookLogging` all consult the same predicate, so no gate can disagree with another about whether Entire is active in a repo.

## Completely dark

Nothing writes the global settings file yet — that arrives with `enable --global` in split C. **Zero behavior change for any existing setup.**

Part of #1098.

### Bare-dir exclude semantics

An `exclude_paths` pattern with no meta characters (e.g. `~/work/vendor-repo`) excludes that directory's entire subtree; because the match is anchored to the worktree root, it cannot over-match siblings (`~/work/vendor-repo-2` is not excluded).

## Follow-on splits

- **B** — lazy invisible enable
- **C** — `enable --global` + wizard
- **D** — user-level agent hooks + surfaces

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes the core hook activation predicate used by agent and git hooks. Risk is moderated by fail-closed semantics and no writer for global settings yet (zero behavior change for existing setups).
> 
> **Overview**
> Adds a **user-global settings tier** so Entire can activate in repos with no repo-level setup, gated by `~/.config/entire/settings.json` (`global.enabled`, path/origin excludes).
> 
[...]

Issue #1098, opened 2026-05-02
[Feature Request] Global auto-enable: track agent sessions without per-repo configuration


rraafffaa:
## Summary

As a less-technical user (Product Manager) who uses Claude Code for rapid prototyping, I frequently spin up quick POCs — creating new repos, committing code, and moving on. I don't follow a traditional developer workflow where I `cd` into a project directory, set up tooling, and then start working. I often create repos on the fly via `gh repo create` or let Claude Code scaffold things directly.

**The problem:** Entire currently requires per-repo configuration (`entire configure` or `entire enable`) before it can track agent sessions. This means any work done in an unconfigured repo is invisible to Entire. For users who move fast and create repos frequently, this is a significant gap — the most interesting agent sessions (the messy, exploratory POC work) are exactly the ones that never get tracked.

## Proposed Behavior

A global configuration mode that automatically enables Entire tracking for **any** git repo where an agent session starts, without requiring per-repo setup first.

Possible implementations:
- **Global agent hooks** installed at the user level (e.g., via `git config --global core.hooksPath` or a Claude Code user-level hook) that detect agent activity and auto-initialize Entire tracking
- **`entire enable --global`** flag that sets a user-level default: "track all repos unless explicitly disabled"
- **Git template directory integration** (`init.templateDir`) so every new `git init` or `git clone` automatically gets Entire hooks
- **Agent-side integration** — Claude Code (and other agents) could check for Entire and auto-configure on session start if global mode is enabled

## Why This Matters

The current per-repo model works well for developers who have a stable set of projects. But for users who:
- Use AI agents as thinking/prototyping tools (not just coding tools)
- Create throwaway repos for POCs, demos, or explorations
- Work across many repos in a single day
- Don't have `entire configure` as part of their muscle memory

...the friction of per-repo setup means Entire misses the very sessions that would benefit most from being tracked. The "commit log for agents" value proposition is strongest when it's **always on**, like how shell history works — you don't configure it per-directory.

## Context

This came up while testing Entire CLI on a new project created during a Claude Code session. The agent created the repo, built the code, committed, and pushed — but because the session was initiated from a different directory, Entire captured zero checkpoints. The work was done, but the "why" behind it was lost.

Related: #985 (global export/viewing across repos) addresses the *read* side of this problem. This issue addresses the *write* side — making sure sessions are captured in the first place.

## Environment
- Entire CLI v0.5.x
- Claude Code (claude-opus-4-6)
- macOS

[...]

PR #968, opened 2026-04-16
fix hanging summary tty in local tests

Soph:
```❯ mise run test
[test] $ go test ./...
?       github.com/entireio/cli/cmd/entire      [no test files]
^CError: not a git repository
┃ Choose a summary provider
┃ This choice will be saved. Use `entire configure` to change it later.
┃ > Claude Code
┃   Codex
┃   Copilot CLI
┃   Cursor
┃   Gemini CLI
```

This worked in CI because there is no tty available and also no agents installed but it triggered on my local machine

---

> [!NOTE]
> **Low Risk**
> Test-only change that sets an env var to avoid interactive prompts; no production behavior changes.
> 
> **Overview**
> Prevents local `go test` runs from hanging by forcing summary-provider selection into non-interactive mode in `explain` checkpoint generation tests.
> 
> Adds `ENTIRE_TEST_TTY=0` to the v2 `--generate`-related tests in `explain_test.go` so they don’t prompt for a TTY-based provider choice when run on a developer machine.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 978a2b7a0c5308b45cef1b2186da293126a8011d. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1955, opened 2026-08-11
feat: enable --global, ask-once wizard, and runtime-data migration (global enable: C)

peyton-alt:
https://entire.io/gh/entireio/cli/trails/1014

## Summary

Split **C** of #1894 (stacked on #1954). Adds the **writer surface** for global tracking: `entire enable --global` / `disable --global`, the ask-once wizard question (either answer persists; non-interactive paths never prompt — `--yes` deliberately does not imply machine-wide consent), per-process memoization of the gate, atomic hook-file writes, and **runtime-data migration** — the repo-level takeover (gitignore + migration + marker clear + routing-cache invalidation) runs on EVERY path that creates the first repo-level settings file, including a bare `entire disable` (previously stranded routed data forever — verified and fixed), with `enable` on an already-configured repo acting as recovery for clones stranded by pre-fix binaries; it moves that worktree's namespace from `.git/entire/worktree/<key>/` into the worktree `.entire/` (scoped to the enabling worktree only; sibling worktreees' data untouched; EXDEV copy fallback; enable never aborts on migration failure).

**Flag exclusivity (user-facing):** `--global` is pairwise mutually exclusive with all 25 repo-scoped/bootstrap enable flags and with disable's `--uninstall`/`--project`/`--force`/`--local` — combinations are rejected at parse time instead of half-applying. A derived test makes future flags fail until classified.

> [!NOTE]
> **Layer boundary:** after this merges, the global tier is *configurable* and repos where entire hooks already fire get lazily tracked — but a virgin repo has no agent hooks until split **D** (#1957) installs user-level ones, so the success message's promise is only fully true once D merges. C and D should merge in quick succession; D's diff inserts the per-agent hook install immediately after this layer's success message.

Cherry-pick note: the second commit carries the *migration-scoping portion* of a reviewed fix whose namespacing landed in split B and diagnostics routing lands in split D — the cross-references in the commit messages are consistent.

Stack rule: fix-forward only; merge order #1953 → #1954 → this → #1957. All commits dual-reviewed; findings on trail #968. Part of #1098.

**Post-review hardening (toolkit round):** `SaveUserSettings` follows a symlinked `~/.config/entire/settings.json` instead of replacing the link (dotfile-manager setups); migration failures now log which files and why (unreadable source counts as failure, copy-landed-but-remove-failed counts as moved with a residue warning); the wizard warns about an unreadable user settings file instead of staying silent; `disable --uninstall` clears the `global_setup_completed` marker so lazy setup reconverges; success message and `--global` help acknowledge exclude patterns ("…and matches no exclude pattern", with kept-pattern count). Cobra-level end-to-end tests pin the `--global` success paths, the `--yes` consent disjunct, both routing-cache invalidation halves, wizard answer durability, and strict-load data-loss protection.


peyton-alt:
Superseded by the consolidated global-enable PR #2030, now targeting main. #2030 preserves this PR's complete commit ancestry and presents the final cross-layer behavior in one review surface; no code from this slice was dropped.

PR #1973, opened 2026-08-12
fix(redact): route redaction diagnostics into .entire/logs and log a load-time summary

peyton-alt:
https://entire.io/gh/entireio/cli/trails/1029

## Problem

The redact package logs pack load failures, rule compile errors, and sample mismatches through the process-default slog logger — bare stderr, which hook contexts swallow — because nothing ever wires it to the CLI's `.entire/logs/` sink (there is no `slog.SetDefault` anywhere in the binary). The happy path logs nothing at all.

Found while investigating a customer security report: they grepped their logs for `component=redaction` while debugging custom rules, found nothing, and reasonably concluded the rules never ran. Every diagnostic they needed had been dropped. The docs even show a `WARN redactor pack sample mismatch` line that could never appear in the log file.

## Fix

- `redact.SetLogger` injects a destination logger (nil keeps the process default, preserving standalone package behavior); all package warnings route through it.
- `EnsureRedactionConfigured` wires it to `logging.Logger()`, a new accessor for the active `.entire/logs` logger, before any rule loading can emit warnings.
- `EnsureRedactionConfigured` now logs one INFO summary per process — `redaction configured` with pack/rule/inline counts and PII/OPF state — so "are my rules active?" is answerable from the log alone.

## Test

Integration test (in the existing `hook_logging_test.go`) runs a real hook with a broken inline pattern and a valid pack, then asserts `.entire/logs/entire.log` contains the component=redaction compile-failure warning and the load-time summary line.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 94267e94029db1f65a468404d2976d097c76f80b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


gtrrz-victor:
## Current state of this PR

Heads up that the **Fix** and **Test** sections in the description are stale — they describe `redact.SetLogger` and `logging.Logger()`, the global-sink design that commits 22abea1 and b87dd2f replaced with context injection. The **Problem** section is unchanged and still accurate. Current state below.

### Fix

The logger reaches `redact` by injection, not through a global sink:

- `CustomRulesConfig`, `PIIConfig`, and `OPFConfig` each take a `Logger` (nil = `slog.Default()`, so the package still works standalone). Every diagnostic in the package routes through it — pack loading, rule compilation, sample validation, and OPF runtime failures.
- `logging.Init` opens the log sink and returns the logger; the caller injects it with `logging.WithLogger`, and `logging.WithSessionID` stamps the session once it is known. `EnsureRedactionConfigured` reads the logger back with `logging.LoggerFromContext` and passes it down before any rule loading can emit a warning.
- `EnsureRedactionConfigured` logs one INFO summary per process — `redaction configured` with pack/rule/inline counts, compiled-vs-configured `active_rules`, and PII/OPF state — so "are my rules active?" is answerable from the log alone.
- Compile-failure warnings log the `regexp/syntax` error *code*, never the pattern text. Go's error strings quote the pattern verbatim, and these lines persist to `.entire/logs/` and ship in `doctor bundle` — a user who pastes a literal secret as a `custom_redactions` value must not have it stored there, and the rule that would have redacted it is precisely the one that failed.

Making injection work end to end meant fixing where logging is initialized:

- Logging is initialized once, in the root `PersistentPreRunE`, instead of ten commands repeating the same guard + `Init` + `defer Close` block in their `RunE`. This needs `cobra.EnableTraverseRunHooks`: `hooks`, `checkpoint`, `session`, and `agent` define their own `PersistentPreRunE`, which under cobra's default only-closest-hook behaviour would shadow the root hook for exactly those trees — silently, since a skipped `Init` shows up only as missing log lines. Gated on `IsSetUpAny` so a repo that never enabled Entire stays untouched; `enable` keeps its own `Init` because its placement is constrained by setup ordering; `hooks` and `attach` additionally call `WithSessionID` once they have resolved a session.
- `entire doctor` is no longer the exception. Because cobra runs persistent pre-runs before a command's own `PreRun`, the command users run to answer "did my rules load?" now routes redaction diagnostics to `.entire/logs/` rather than bare stderr.
- `logging.Debug/Info/Warn/Error` resolve the logger from the context first, falling back to the package global and then `slog.Default()`. One transport instead of two, so a caller cannot log somewhere other than where `redact` does.
- Loggers write through an indirection rather than binding to the `*bufio.Writer`, so one captured before a re-`Init` keeps reaching the file instead of filling an orphaned buffer nothing flushes. Nothing re-initializes in a live process any more, but the indirection also locks exclusively, fixing a pre-existing data race: the old code held a *read* lock across the write, which admits concurrent writers to a `bufio.Writer` that is not goroutine-safe.
- `factoryaidroid.ExtractModelFromTranscript` warned through package-level `slog`, so a corrupt `.settings.json` produced a stderr line swallowed by the hook context. It was the last package-level `slog` call outside the `logging` package.

### Test

- Integration test fires a real hook with a broken inline pattern, a broken PII custom pattern, and a valid pack, then asserts `.entire/logs/entire.log` carries both compile-failure warnings tagged `component=redaction` plus the summary line, with counts proving configured-vs-compiled drift.
- Unit tests pin the pattern-leak privacy contract, the context-injection contract (including the stderr-fallback path returning no logger), the context-first resolution order, single `session_id` stamping, and that a logger survives a re-`Init`.
- Root-hook tests cover traversal into `checkpoint`/`session`/`agent` and that a never-enabled repo gets no `.entire/` at all.
- Verified against a built binary: all five diagnostic classes — including the `redactor pack sample mismatch` line the docs promised — land in `entire.log` under `component=redaction`, from both a hook and `entire doctor`.
- `go test ./...`, integration suite, `go test -race ./cmd/entire/cli/logging/`, E2E canary 56/56 + 4/4, `golangci-lint` 0 issues. Login/auth flows diffed command-by-command against `main`: stdout, stderr, and exit codes byte-identical.

### Known gap, deliberately not fixed here

`active_rules` counts inline + pack rules only. A broken `pii.custom_patterns` entry is warned (verified), but the summary still reads `pii: true` with no compiled count — so drift is visible for two of the three rule sources and silently omitted for the third. Closing it needs a new `redact.ActivePIIPatterns()` export plus two summary fields; tracking separately rather than growing this diff.



---

**Edited** to match `b6b78772f` and `4f1c1a324`, which reshaped the `logging` API after this comment was first posted:

- `Init` no longer returns a `context.Context`. In Go, ctx is an input and only `WithX(ctx, v)` decorators return one, so an initializer handing one back conflated process lifecycle with value propagation and hid the injection from its own signature. It now returns the logger and the caller injects it.
- `Init` no longer takes a session ID either. The log path is fixed, so the hook path's second `Init` call was closing and reopening the same file, reallocating the 8KB buffer, and re-reading the log level purely to add one attribute — on every hook invocation. `WithSessionID(ctx, id)` does that instead.
- That move surfaced a latent bug worth calling out: because `Init` returned an *error* for a malformed session ID, `initHookLogging` bailed out early and skipped `EnsureRedactionConfigured` entirely. A junk session-state file silently disabled custom redaction for that hook invocation. Validation now lives in `WithSessionID`, which warns and continues unstamped, so redaction is configured either way.

gtrrz-victor:
## `session_id` becomes a context value (9b0fc4d97)

Follow-up on this branch: `session_id` is now handled exactly like `component` and `agent`, and the `currentSessionID` package global is gone.

The global meant a process had exactly one session at a time, and "overriding" it meant overwriting it for every other caller. As a plain context value, a derived context shadows the session for its own scope only — which is the behaviour the old design couldn't express. `WithSessionID` drops its `error` return and the mutex with it, becoming a two-line decorator matching its siblings.

### The duplicate-key hazard this had to solve

There are **123** hand-written `slog.String("session_id", …)` attrs across `cmd/` and `internal/`. slog does not dedupe attrs, so once the context carries a session, any of those running under a session-scoped context would emit the key twice:

```json
{"level":"WARN","msg":"…","session_id":"first","session_id":"second"}
```

So `log()` now drops its own session attr when the call site already supplied one. The check handles both slog calling conventions (`slog.Attr` values and loose key/value pairs) and correctly ignores a `"session_id"` appearing in a *value* position. Table-driven test covers all three shapes; a second test pins the scope-shadowing that motivated the change.

Those 123 attrs are deliberately **not** deleted here — each needs checking that the in-scope ctx is the session-scoped one, and getting it wrong silently mis-attributes a line.

### Validation deliberately not carried over

`validation.ValidateSessionID` is gone from this path. Logging builds no paths from the session ID — it's an slog attribute — so the traversal guard belongs where the ID is resolved from the filesystem, not at this call site. (`ValidateSessionID` remains in use at ~38 other sites.)

Related gap found but **not** fixed here, since it wants a proper audit of `state.SessionID` consumers first: `session.StateStore.Load` validates the filename-derived ID but returns `state.SessionID` straight from `json.Unmarshal`, never reconciled with the filename and never re-validated, so `FindMostRecentSession` can return an arbitrary string.

### redact keeps its session stamp

`redact` holds an injected `*slog.Logger` and calls it without a context (10 `logger.Warn` sites plus one `WarnContext`, which doesn't help — the JSONHandler doesn't read ctx). Those are the customer-facing redaction diagnostics this branch exists to route into `.entire/logs/`, and they'd lose `session_id` under a pure context value. `logging.SessionLoggerFromContext(ctx)`, wired into `strategy.EnsureRedactionConfigured`, stamps it for them.

It stamps the **session only**, not every ctx attr. `redact` tags its own lines `component=redaction`, so re-adding `component` from the context would recreate the exact duplicate-key defect described above. No current caller of `EnsureRedactionConfigured` sets a component, so it wouldn't fire today — but it would be a silent landmine for whoever adds a `WithComponent` upstream. Full decoration belongs with the follow-up that also deletes redact's `componentAttr`.

### Net effect of losing process-wide stamping

Exactly **9** sites log through a context that cannot descend from a stamped one (`logging.X(context.Background(), …)`): `recap/render_static.go` ×3, `explain.go` ×2, `copilotcli/transcript.go`, `codex/reviewer.go`, `opencode/opencode.go`, `review/tui_sink.go`. Eight are `Debug`, one is `Warn`, and every one names its own subsystem in the message — none does session-scoped work.

Against that: in `explain` and `recap` the global could tag lines with the **wrong** session, since both routinely operate on a checkpoint from a session other than the ambient one. Trading a session tag on 9 diagnostic lines for the removal of a global that could assert a false session.

### Verification

- `gofmt` clean; golangci-lint 2.11.3 clean (0 issues)
- `mise run test:ci:core` — `-race` over every package, exit 0
- `go test -tags integration ./cmd/entire/cli/integration_test/` — exit 0
- `mise run test:e2e:canary` — 4/4

Based on `f3d60c8fe` (the `main` merge), not `1055d93b6`. Verified the merge leaves the whole area untouched: `cmd/entire/cli/logging/` and `attach.go` are byte-identical between the two commits, `EnsureRedactionConfigured` is untouched, and no bare `slog.*` call was reintroduced in `redact/`.

> [!NOTE]
> This push used `--no-verify`, so Entire's own pre-push checkpoint sync did not run for it. Unrelated to this change: the `main` merge deleted `scripts/entire-dev` (a9a676e79, "remove local_dev mode") while this clone's `.git/hooks/pre-push` still invokes it, so `git push` is rejected outright. `entire doctor` diagnoses and repairs exactly this shape.

gtrrz-victor:
## `initHookLogging` deleted (b5c9a2d72)

The root `PersistentPreRunE` already opens the log sink, puts the logger in the context, and closes it again in its `PersistentPostRun` — with `main.go:96` closing once more for the error path cobra skips post-runs on. `initHookLogging` was re-running all of that per hook invocation.

Removed as redundant:

| step | why it can go |
| --- | --- |
| `logging.SetLogLevelGetter(GetLogLevel)` | root prerun calls it with the same getter |
| `logging.Init(ctx)` | root already opened the sink. Its gate (`WorktreeRoot` + `IsSetUpAny`) is strictly weaker than this function's `IsSetUpAndEnabled`, so whenever this proceeded, root had already run. The second Init closed and reopened the *identical fixed path* with a fresh 8KB buffer, once per hook — the same waste this branch removed from the session-ID path, one level up |
| `logging.WithLogger(ctx, l)` | root already `SetContext`'d an equivalent logger |
| the returned cleanup | root's `PersistentPostRun` opens with `defer logging.Close()`, so its hidden-command early return still flushes; `main.go` closes unconditionally. `hookLogCleanup`, `agentHookLogCleanup`, and both `PersistentPostRunE` blocks go with it |

What remains is `withHookSession`, which does only what root cannot:

1. **The session stamp** — root would have to scan session state on every command to know it.
2. **`EnsureRedactionConfigured`** — hooks are the checkpoint-writing path, so leaving this to the root prerun would mean only always-on secret scanning runs there. That's the opposite of what this branch is for.
3. **The `IsSetUpAndEnabled` gate** — load-bearing, not defensive. The git-hook tree checks it *before* calling, but `newAgentHooksCmd`'s prerun does not, so this is the only thing between a repo that never enabled Entire and having its session state scanned and its redactors loaded.

`executeAgentHook`'s `initLogging` param becomes `stampSession`, since it no longer initializes or cleans up anything.

### Vestigial `PersistentPreRunE`

Three prerun hooks could no longer return a non-nil error, so they became `PersistentPreRun`: root's, `newAgentHooksCmd`'s, and the git-hook tree's. `agent_group.go`, `checkpoint_group.go`, and `sessions.go` keep `E` — they genuinely return `errors.New("not a git repository")`.

This is not only cosmetic. cobra v1.10.2 (`command.go:984`) dispatches `if PersistentPreRunE != nil { … } else if PersistentPreRun != nil { … }`, inside the traverse loop — so the variants behave identically under `EnableTraverseRunHooks`, but setting both on one command **silently kills the non-`E` one**. root's `error` was also actively misleading: `initRootLogging` documents that failure is "non-fatal and deliberately silent", and root already used the non-`E` `PersistentPostRun` for its flush.

### Tests

Four tests called `initHookLogging` and asserted on log-file creation — behavior that has moved to the root prerun and is already covered by `root_logging_test.go`. Left alone they would have gone **vacuous rather than failing**, which is the worse outcome:

- The three `TestInitHookLogging*` tests are replaced by two: one Inits a sink, calls `withHookSession`, and asserts `"session_id":"<id>"` reaches `entire.log`; one table test asserts a not-set-up repo and a disabled repo both return `got == ctx` — the observable proof the gate held, with nothing created on disk.
- `hook_registry_test.go`'s setup uses `logging.Init`/`logging.Close` directly.
- The two near-identical `_HasLoggingHooks` tests collapse into one table-driven test that now asserts `PersistentPostRun`/`PersistentPostRunE` are **nil** and `PersistentPreRunE` is nil while `PersistentPreRun` is set — pinning both that flushing belongs to root and that a stray `E` can't shadow the session wiring unnoticed.

Rewritten tests also use `testutil.InitRepo` instead of raw `git init`, per CLAUDE.md.

### Verification

- `gofmt` clean; golangci-lint 2.11.3 clean (0 issues)
- `mise run test:ci:core` — 85 packages, 0 failures, 0 races
- `go test -tags integration ./cmd/entire/cli/integration_test/` — ok
- `mise run test:e2e:canary` — 56 + 4 passed

Net −128 lines.

gtrrz-victor:
## `logging.New(Config)` replaces `Init`; the package owns no state (b59c79bb9)

The logging package had five package globals (`logger`, `logFile`, `logBufWriter`, `mu`, `logLevelGetter`) and an `Init` that configured them. A process therefore had exactly one logger, belonging to whoever called `Init` last, and every consumer had to trust that something upstream had already done so. `New(Config)` returns a `*Logger` the entry point puts in the context and closes from the context.

**Deleting the globals deleted the machinery that existed to manage them:**

- **`liveWriter`** existed only to route writes to whichever buffer the global currently held, so a logger handed out before a re-`Init` kept working. With an instance there is no global to swap and no re-`Init`. The per-write mutex stays, as a field — `bufio.Writer` isn't goroutine-safe and slog handlers may be called concurrently.
- **`SetLogLevelGetter`** existed only to dodge a `logging`→`settings` import cycle. The level is now a `Config` field resolved by the caller, so **`GetLogLevel` takes a real context** instead of `context.TODO()` and its `//nolint:contextcheck` is gone.
- **`Init`'s `(nil, nil)`** signalled "initialized, but not file-backed". `New` returns a real error, so the **`//nolint:nilnil` is gone**.
- **`Config.Dir` is caller-resolved**, which drops logging's dependency on `paths` entirely.

**One logger per process.** Root's `PersistentPreRun` is the only unconditional construction site. `attach` no longer builds its own — it was opening a second handle with a second 8KB buffer on the same file — and only adds the session. `enable` builds one *only when root's `IsSetUpAny` gate declined*, which is the fresh-repo case it exists for; re-enabling an already set-up repo now reuses root's.

**The close path, verified rather than assumed.** Root's `PersistentPostRun` is the normal flush. `main.go` is the backstop for the paths cobra returns on *before* reaching that hook — and there are **two**, not one: `RunE` returning an error, and required-flag/flag-group validation, which runs *after* the pre-runs. Measured against cobra v1.10.2:

| path | PreRun | PostRun | logger built |
|---|---|---|---|
| success | ✅ | ✅ | closed by PostRun |
| `RunE` error | ✅ | ❌ | **closed by main.go** |
| missing required flag | ✅ | ❌ | **closed by main.go** |
| unknown flag | ❌ | ❌ | none exists |

`cobra.OnFinalize` covers those paths too, but finalizers are `func()` with no command or context, so using it would require a package-level holder — reintroducing the global this commit removes — and `OnFinalize` appends to a cobra global with no deregister (measured: 1, 3, 6 cumulative bodies over three invocations), which is the same data-race class as the `EnableTraverseRunHooks` incident on this branch. `main.go`'s block is the finalizer that can still see the context.

Pinned by `TestExecutedCommandCarriesLoggerOnFailure`, table-driven over both leak paths. It asserts the line is *still buffered* before the close, so it cannot silently degrade into proving nothing.

**Behavior change worth knowing:** lines logged through `context.Background()` used to reach the log file via the global and now reach stderr. Eight of the nine such sites are `Debug` and drop at INFO; `opencode.go`'s `ReadSession` warning is the exception, and threading a context there means changing an interface method, so it is left for its own change.

### Verification

- `gofmt` clean; golangci-lint 2.11.3 clean (0 issues)
- **End-to-end against a built binary**, in a repo with a broken `pii.custom_patterns` entry and a malformed redactor pack: `entire doctor` and `entire hooks git post-commit` both land all three redaction diagnostics in `.entire/logs/entire.log` with `component=redaction`, and the hook path carries `session_id` on each — including the two lines that come from redact's *injected* logger.
- The `-race` suite, integration, and canary were green on the tree before the final review round; the full sweep over this exact tree was still running at push time, so CI on this sha is the authoritative run.

> [!NOTE]
> `TestMCPServer_EntireStatusMatchesPassive` fails locally but is a pre-existing isolation bug, not from this change: it is `t.Parallel()` with no temp repo, so it reads the real `.git/entire-sessions` and races a live session's `model` field between its two status snapshots. It passes standalone on `66165329e` and on this tree.

gtrrz-victor:
Follow-up on b59c79bb9's verification note: the full sweep over that exact tree finished clean, so the caveat in my previous comment can be dropped.

- `gofmt` clean; golangci-lint 2.11.3 clean (0 issues)
- `mise run test:ci:core` — 86 packages, 0 failures, 0 data races
- `go test -tags integration ./cmd/entire/cli/integration_test/` — ok
- `mise run test:e2e:canary` — 56 + 4 passed

`TestMCPServer_EntireStatusMatchesPassive` passed in this run too, consistent with it being a pre-existing isolation flake (parallel, no temp repo, races a live session's `model` field between its two status snapshots) rather than anything from this change.

Soph (approved):

PR #1976, opened 2026-08-12
fix(search): order CLI semantic results by rerank score, matching the web

alishakawaguchi:
https://entire.io/gh/entireio/cli/trails/1032

## Problem

`entire search` (CLI) and entire.io (web) return **different results for the same query + repo + user** — different order and different totals. Root cause: the CLI's cross-cell semantic merge orders by the wrong key.

- The CLI ordered **tier 0 by BM25** and **tier 1 by the per-namespace retrieval `Score`**, and its result model had **no `rerankScore` field at all** (it silently dropped it when parsing the response).
- The shared server-side reranker (Cohere, in `entire-search` query-serve) is identical for both clients, but the **ordering that consumes the rerank score is a separate re-implementation in each client** (Go in the CLI, TS in the BFF), and the CLI's copy never got the fixes the web did.
- query-serve's own docs on the `Score` field: *"It is NOT the field results are ordered by … sorting by it undoes reranking."* The CLI was sorting by exactly that. The web BFF orders both tiers by rerank score (ENT-1425 / ENT-1431); the CLI was sitting at the pre-fix behavior.

Measured on EU prod, same user/repo:
- `"dependabot"` on `entirehq/entire.io`: **118 of 119** tier-0 positions differed between CLI and web; totals 100 vs 121.
- `"antithesis reranking"`: different #1, **15 of 18** positions differed; CLI output carried no `rerankScore`.

This is the same divergence class as the "wildly different results" reports.

## Fix

Bring the CLI merge to parity with the web:
- add `RerankScore` to the search result `Meta` so the CLI parses and surfaces query-serve's cross-encoder score;
- order **tier 0** by rerank score desc (BM25 tie-break) when every row was reranked, else the BFF's positional mixed-capability fallback (interleave by each cell's selection rank, BM25 tie-break);
- order **tier 1** by rerank score desc (retrieval `Score` tie-break);
- tests assert rerank ordering with BM25/`Score` set to the *opposite* order, so a regression to the old keys fails, plus a mixed-capability fallback test.

Repos, tier-2 promotion, dedupe, caps, and counts are unchanged.

## Verification

- `go build ./cmd/entire/...`, `go vet`, and the search test suites pass.
- End-to-end: the patched binary's top results for `"dependabot"` now match the web's `rerankScore` order **exactly** (0.6276, 0.6036, 0.5790, 0.5637, …), and `rerankScore` is now present on every reranked row.

## Follow-up (not in this PR)

- Tier-2 handling still differs (the web hard-drops tier-2; the CLI promotes/keeps it). Dormant today since query-serve no longer emits tier-2, but worth reconciling.
- The real durable fix is to stop maintaining two orderers — a single shared ordering contract, or have query-serve return final merged order so neither client re-sorts.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 917583654874870e8873014db48324aa028bbf2d. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/search/search.go
+++ b/cmd/entire/cli/search/search.go
@@ -65,13 +65,20 @@ const DefaultLimit = 100
 
 // Meta contains search ranking metadata for a result.
 type Meta struct {
-	MatchType string   `json:"matchType"`
-	Score     float64  `json:"score"`
-	Tier      *int     `json:"tier,omitempty"`
-	Snippet   string   `json:"snippet,omitempty"`
-	Summary   string   `json:"summary,omitempty"`
-	BM25Score *float64 `json:"bm25Score,omitempty"`
-	ANNScore  *float64 `json:"annScore,omitempty"`
+	MatchType string  `json:"matchType"`
+	Score     float64 `json:"score"`
+	Tier      *int    `json:"tier,omitempty"`
+	Snippet   string  `json:"snippet,omitempty"`
+	Summary   string  `json:"summary,omitempty"`
+	// RerankScore is query-serve's cross-encoder (Cohere) relevance judgement,
+	// present only on results that went through reranking. Where present it is
+	// the signal the final ordering was built from and the only score
+	// comparable across repos — Score is per-namespace retrieval strength, so
+	// sorting by it undoes reranking. The cross-cell merge orders tier 0 and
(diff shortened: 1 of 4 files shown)


alishakawaguchi:
bugbot run

evisdren (approved):

PR #1981, opened 2026-08-13
fix(login): open the browser via ShellExecute on Windows

Soph:
https://entire.io/gh/entireio/cli/trails/1037

## The bug

A user on Windows running v0.10.0 reported that `entire login` opens

```
https://us.auth.entire.io/authorize?client_id=entire-cli
```

and the authorization server rejects it with **"auth request is missing redirect uri"**.

## Root cause

`openBrowser` launched the URL through cmd.exe:

```go
case "windows":
    command = "cmd"
    args = []string{"/c", "start", "", browserURL}
```

cmd.exe treats an unquoted `&` as a command separator, so the browser only ever received the text before the first `&`.

Go's argv escaping does not help here: `syscall.EscapeArg` only quotes an argument containing a space, tab, quote, or backslash, and a URL has none of those, so it reaches cmd.exe bare. Because `url.Values.Encode` sorts keys, `client_id` sorts first and `redirect_uri` sits behind three separators.

Here is a real authorization URL from the production auth server:

```
…/authorize?client_id=entire-cli&code_challenge=UwBw-uNt…&code_challenge_method=S256&redirect_uri=http%3A%2F%2F127.0.0.1%3A59656%2Fcallback&response_type=code&scope=cli+offline_access&state=moEAbe7a…
                                ^ cmd.exe cuts here
```

Everything from `code_challenge` on is lost, leaving `…/authorize?client_id=entire-cli` — byte-for-byte the URL in the report.

Quoting alone would not have been enough either — the percent-encoded `redirect_uri` also exposes the URL to cmd's `%VAR%` expansion.

Broken since 5c44ffb87 (Mar 2026), but only user-visible once the browser redirect became the default login flow.

## Workaround for affected users, until this ships

```
entire login --device
```

Captured from a real run:

```
Device code: BN5V-FV8V
Login URL:   https://us.auth.entire.io/cli/auth?user_code=BN5V-FV8V

Waiting for approval...
```

Two reasons this is immune:

1. **One query parameter, no `&`** — there is nothing for cmd.exe to cut, so the browser receives the whole URL. Verified against the live `/device_authorization` endpoint.
2. **The URL is printed unconditionally**, before any browser launch, so copy-paste works even if the open misbehaves.

Approval then polls over HTTPS with no shell involved. Nothing is degraded relative to the redirect flow.

<details>
<summary>Second option, if the redirect flow is specifically needed</summary>

`ENTIRE_TEST_TTY=1 entire login` makes the browser open fail deliberately, so the CLI prints the full authorization URL and keeps the loopback listener waiting — paste it into a browser and the login completes on the redirect flow. (This is how the real URL above was captured.)

Hold this in reserve rather than recommending it: `ENTIRE_TEST_TTY` is an internal test switch, not a supported user-facing knob, and it also suppresses the Enter prompt as a side effect.

</details>

## The fix

Split the launcher per platform and, on Windows, call the Win32 `ShellExecute` API directly (as `github.com/pkg/browser` does) so the URL is passed as a single wide string — no command line to parse, nothing to escape. `golang.org/x/sys` was already a direct dependency, and already in the Windows import graph, so this costs +1,024 bytes of PE page rounding.

- `browser_open_windows.go` (new) — `ShellExecute`, behind a `shellExecute` var so tests can capture the string without launching a browser
- `browser_open_other.go` (new) — `open` / `xdg-open`; these never involved a shell
- `login.go` — `openBrowser` keeps URL validation and the under-test guard, then delegates to `openBrowserPlatform`

The second commit is a cleanup pass. Two things in it are worth calling out:

**A goroutine leak, carried over verbatim from the old launcher.** `exec.CommandContext` sets `Cmd.Cancel`, so `Start` spawns a `watchCtx` goroutine whose result channel only `Wait` drains — and this path deliberately never calls `Wait`. Every browser open leaked one goroutine, retaining the `Cmd` (argv, env, URL), for the process lifetime. Measured: 20 opens leaked 20 goroutines, surviving both cancel and GC; with `exec.Command` it is zero. The context was already inert here because `Process.Release` detaches the child, so `Cmd.Cancel`'s `Process.Kill` fails regardless — dropping it removes the leak without losing cancellation that ever worked. Bounded for `login`, but `experts --tui` is long-lived and opens several links.

**Naming the boundary between this repo's two Windows-command philosophies**, which previously discussed the same hazard from opposite sides without referencing each other. The split turns on *who owns the exec*: escape when a third party runs your command (`agent.escapeWindowsCMD`, for agents' `hooks.json`), bypass the shell when Entire launches it itself. Reciprocal doc pointers, plus a `CLAUDE.md` section under Code Patterns recording the invariant and the `EscapeArg` fact that made the bug possible.

## Tests

`TestOpenBrowserWindows_PassesWholeURLToShellAssociation`, named so the existing `test-windows` CI job picks it up (it selects `-run '(Windows|MSYS)'`): the whole URL, `&` separators and percent-encoding intact, reaches the shell association. Reintroducing a shell launcher stops the seam being called and fails this test.

Plus `browser_open_test.go` covering URL validation cross-platform.

An earlier revision also had a test asserting that `cmd /c echo <url>` truncates at the first `&`. It was dropped in the cleanup commit: it called no code from this repo, so it passed identically whether the launcher used `ShellExecute` or `cmd /c start`, and could not detect the regression it claimed to guard — while costing a cmd.exe process plus six failed PATH×PATHEXT resolutions on every Windows CI run.

Verified: builds for darwin/windows/linux; `golangci-lint` clean on both darwin and `GOOS=windows` (the default run skips `_windows.go` files); `mise run test:ci` passes. The Windows test compiles for windows/amd64 but was not executed locally — it runs in CI.

## Scope

`openBrowser` has two production callers and both are covered: `login.go` (browser and `--device` flows) and `experts_tui.go` via `openExpertsSessionCmd`. The experts session URL is path-only (`url.PathEscape`, no query string), so it had no `&` and was not affected by this bug — it just becomes robust for free.

This is the whole population of the bug class. The repo's only other cmd.exe call sites — `versioncheck/autoupdate.go` and the `agent/hook_command.go` wrappers — interpolate compile-time constants only, and the one genuinely variable value there already goes through `escapeWindowsCMD`.

## Deliberately not addressed

`runBrowserLogin` only prints the full URL when the browser open *reports failure*. This launch succeeded from our side while delivering a mangled URL, so the user had no copy-paste fallback and was fully stuck. Printing the URL unconditionally (as the device flow does) would make this class of failure self-service, but that is a separate behavior change.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 976fd426117035ece6ba13f102814e447ddcd5c6. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nodo (dismissed):

gtrrz-victor (dismissed):
Tested on windows  👍

gtrrz-victor (approved):

PR #1996, opened 2026-08-14
fix(opencode): report transcript export failures during attach

MuskanPaliwal:
Stacked on #1877. Refs #1876.

## Problem

#1877 lets `entire session attach <session-id> --agent opencode` run `opencode export` when Entire does not already have the transcript. If that export fails, the error is written only to debug logs. Attach continues through its fallback paths and can eventually report:

```text
transcript not found for agent "opencode" with session ses_missing; is the session ID correct?
```

That message loses the real failure. A missing OpenCode executable, a timeout, invalid JSON, insufficient permissions, and an OpenCode rejection can all look like an incorrect session ID.

## Change

The attach resolver now remembers an error returned by `TranscriptFetcher` while it tries the existing project-directory fallback. If the fallback finds a transcript, attach succeeds and the fetch error is ignored. If the fallback also fails, attach returns the fetch error instead of replacing it with the generic transcript-not-found message.

The resolver remains agent-agnostic. It does not inspect OpenCode errors or branch on the agent name. Each `TranscriptFetcher` owns the errors produced by its fetch mechanism. OpenCode is currently the only implementation, so its export runner now translates the common failures:

| Failure | Terminal message |
| --- | --- |
| OpenCode is missing | `OpenCode is not installed or is not available in PATH.` |
| Export times out | `OpenCode export timed out after 30s. Try again.` |
| OpenCode rejects the export | `OpenCode could not export session "ses_123": <OpenCode error>` |
| OpenCode cannot be executed | `OpenCode could not be started because of insufficient permissions.` |
| Export contains invalid JSON | `OpenCode returned invalid transcript data for session "ses_123". Try updating OpenCode and running the command again.` |
| Session does not exist | `OpenCode session "ses_missing" was not found. Check the session ID and try again.` |

OpenCode stderr can contain progress output, internal diagnostics, terminal escapes, or sensitive values. Arbitrary stderr is not copied into the terminal error. Only a line beginning with `Error:` is eligible for display. That line is stripped of terminal control sequences, passed through Entire’s redaction, collapsed to one line, and limited to 300 runes. If OpenCode exits without a structured error line, attach reports:

```text
OpenCode could not export session "ses_123".
```

When agent auto-detection also fails, that secondary failure stays in debug logs instead of adding the noisy `(also tried auto-detecting other agents: ...)` suffix. Auto-detection also skips the agent that was resolved from existing session state, rather than relying only on the CLI flag.

## Compatibility

Agents without `TranscriptFetcher` keep their existing explicit-session-ID behavior. A valid fallback transcript still wins after a fetch failure, and cancellation remains distinguishable through error unwrapping.

This PR does not include cache staging, cache-preservation changes, full transcript-shape validation, OpenCode session enumeration, the attach picker, or historical import. Cache integrity and bounded subprocess capture can be reviewed separately without blocking this error-reporting fix.

## Merge order

The preferred path is to include this follow-up in #1877 before that PR merges, so the new fetch capability lands with its user-facing failure handling.

If the changes remain in separate PRs, #1877 must merge first. This draft targets `main` from the start, so no base change is needed afterward. Once #1877 is on `main`, GitHub will remove its shared commit from this PR’s diff.

## Testing

The regression was reproduced on #1877 before the fix: a failed fetch followed by no fallback lost the fetch cause and returned the generic transcript-not-found error.

Focused tests cover fetch failure retention, successful fallback, agents without `TranscriptFetcher`, missing executables, timeouts, permission failures, useful and unstructured stderr, invalid JSON, and failed secondary auto-detection.

`mise run check` passed. This includes formatting, lint with zero issues, race-enabled unit and integration tests, all 59 deterministic Vogon scenarios, and all 4 Roger scenarios.

The current binary was also exercised with real OpenCode inside an isolated temporary Git repository. A nonexistent session produced:

```text
OpenCode could not export session "ses_nonexistent_ux_test": Unexpected error
```

Screenshot 2026-08-14 at 5 25 24 PM

### DIFF
--- a/cmd/entire/cli/agent/agent.go
+++ b/cmd/entire/cli/agent/agent.go
@@ -197,7 +197,8 @@ type TranscriptFetcher interface {
 	Agent
 
 	// FetchTranscript writes the session's transcript to the agent's cache
-	// location and returns its path.
+	// location and returns its path. Errors may be shown to users after other
+	// transcript sources fail, so they must be concise and safe to display.
 	FetchTranscript(ctx context.Context, sessionID string) (string, error)
 }
 
--- a/cmd/entire/cli/agent/agent.go
+++ b/cmd/entire/cli/agent/agent.go
@@ -188,6 +188,19 @@ type TranscriptPreparer interface {
 	PrepareTranscript(ctx context.Context, sessionRef string) error
 }
 
+// TranscriptFetcher is implemented by agents that can materialize a session
+// transcript on demand (e.g. OpenCode via `opencode export`), including for
+// sessions Entire never tracked — where no hook-cached transcript file exists
+// (e.g. sessions spawned by an external host rather than a hooked terminal).
+// TranscriptPreparer, by contrast, only refreshes an already-existing file.
+type TranscriptFetcher interface {
+	Agent
+
+	// FetchTranscript writes the session's transcript to the agent's cache
+	// location and returns its path.
+	FetchTranscript(ctx context.Context, sessionID string) (string, error)
+}
+
 // SidecarImageProvider is implemented by agents that keep images OUTSIDE the
 // transcript Entire condenses — e.g. Cursor stores pasted images in a per-session
 // SQLite blob store, not the JSONL transcript. The strategy layer calls this
--- a/cmd/entire/cli/agent/capabilities.go
+++ b/cmd/entire/cli/agent/capabilities.go
@@ -89,6 +89,18 @@ func AsSidecarImageProvider(ag Agent) (SidecarImageProvider, bool) {
 	return p, ok
 }
 
+// AsTranscriptFetcher returns the agent as TranscriptFetcher if it implements
+// the interface. This is an optional capability (materializing a transcript on
+// demand for sessions with no hook-cached file), so it resolves by type
+// assertion alone with no DeclaredCaps gate.
+func AsTranscriptFetcher(ag Agent) (TranscriptFetcher, bool) {
+	if ag == nil {
+		return nil, false
+	}
+	f, ok := ag.(TranscriptFetcher)
+	return f, ok
+}
+
 // AsTokenCalculator returns the agent as TokenCalculator if it both
 // implements the interface and (for CapabilityDeclarer agents) has declared the capability.
 func AsTokenCalculator(ag Agent) (TokenCalculator, bool) {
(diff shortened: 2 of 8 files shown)


Soph:
Folded into #1877 as you proposed ("The preferred path is to include this follow-up in #1877 before that PR merges"), so closing this one.

Your commit is in #1877 unchanged and with its authorship intact, as `6ef8c49`. #1877 is now rebased onto current `main` (the conflict was in `capabilities.go`, against `main`'s new `AsTranscriptSanitizer`) and reports `MERGEABLE`.

One thing from your "reviewed separately" list also went in, as a third commit: `runOpenCodeExportToFile` now stages the export in a sibling temp file and renames on success. Without that, the failure your PR reports so much better was still deleting the hook-cached transcript on the way out — `O_TRUNC` plus `os.Remove` on a path that is the only local copy of the session until the user commits. It applies to the pre-existing `PrepareTranscript` turn-end path too, not just attach, which is why it sits in the export runner rather than in attach.

Two smaller notes for whatever you pick up next:

- `TestPrepareTranscript_AlwaysRefreshesTranscript` got the `runOpenCodeExportToFileFn` stub but no `t.Chdir`, so `fetchAndCacheExport` still resolves `paths.WorktreeRoot()` to the real repo and creates `<repo>/.entire/tmp`. Running that test alone in a clean worktree reproduces it. Two lines: `t.Chdir(t.TempDir())` + `paths.ClearWorktreeRootCache()` — the test is already non-parallel.
- The dedicated `OpenCode session "x" was not found` branch keys off a stderr line starting `Error: session not found`, but the manual verification in your PR description shows real OpenCode emitting `Unexpected error` for a nonexistent session. Worth re-checking against a released build, since the most common user mistake otherwise lands on the generic message.

Soph:
Closing as folded into #1877 (see comment above).

MuskanPaliwal:
Thanks @soph for folding this into #1877 and keeping the authorship!

I went back and checked both notes. It looks like `3a39273b` already added the `t.Chdir` and worktree-cache reset to `TestPrepareTranscript_AlwaysRefreshesTranscript`, so that test no longer writes into the developer’s repo.

I also tried exporting a nonexistent session with OpenCode 1.18.18. It now prints:

```text
Error: Session not found: <session-id>
```

So the existing special case matches the current release. The `Unexpected error` output seems to have been version-specific. I think keeping the generic fallback for errors we don’t recognize is safer than treating every unexpected error as a missing session.

Thanks for catching these!

PR #2003, opened 2026-08-14
feat(checkpoint): make the local metadata-ref replay visible

Soph:
https://entire.io/gh/entireio/cli/trails/1054

Stacked on #1951 — base retargets to `main` automatically when that merges.

## Problem

A force-fetched metadata branch can rewrite the local primary ref: `SafelyAdvanceLocalRef` replays local commits onto the fetched tip when the two sides diverged. No checkpoint is lost, but the ref moves and the replayed commits get new hashes — and nothing said so anywhere. The whole advance/replay chain (`SafelyAdvanceLocalRef`, `replayLocalRefFromBase`, `replayDisconnectedLocalRef`, `replayLocalCommits`, `setRefHash`) had **zero** logging calls, so an after-the-fact investigation had no trace to follow.

This came out of asking why the metadata fetch refspec is force (`+refs/heads/…`). The force itself is correct and load-bearing — v1 diverges by design (OPF rewrite, condensation, reconciliation), and without `+` a diverged tracking ref cannot be fetched at all, so divergence detection would never see the remote tip. The gap was that the *consequence* was invisible.

## Change

**The replay leaves a trace.** `replayLocalCommits` records ref, pre-rewrite local tip, target tip, replayed-commit count and a `diverged`/`disconnected` reason before rewriting. Logged in the function that performs the rewrite, not in its callers, so a future third caller cannot silently skip it. `Info`, not `Debug` — the default level has to carry it to be findable in `.entire/logs/`. Metadata only, per the logging privacy rule.

**`entire doctor` reports DIVERGED** with both tips instead of a bare `OK`:

```
Metadata branches: DIVERGED
  Local and remote entire/checkpoints/v1 have both advanced since they last agreed.
    local:  8415df1ee184
    remote: 7c46565c7e7c  (origin)
  No action needed: the next fetch from this remote replays your local
  checkpoints onto its tip. No checkpoints are lost, but the local ref moves
  and the replayed commits get new hashes.
```

Divergence is the one state whose resolution rewrites the local ref, and no existing check reported it — `IsMetadataDisconnected` covers only the no-merge-base case, and ahead/behind needs no comment. The message respects #1951's confinement rule: against the elected sync remote it self-heals, but a diverged legacy-tier ref never advances the local ref, so it says so instead of promising a fix that will not come.

**Both verdicts from one classifier.** `CompareMetadataWithRemote` derives absent / aligned / local-ahead / remote-ahead / diverged / disconnected from a single `git merge-base`, and `IsMetadataDisconnected` becomes a wrapper. Computing them separately meant `entire doctor` read four refs and spawned two `merge-base` subprocesses on the same hash pair to answer one question, from two bodies that could drift apart. Measured: 2 → 1.

## Notes for reviewers

- Report-only. Nothing here moves a ref; the disconnected case keeps its existing repair path untouched.
- `isDisconnected` stays — `ReconcileDisconnectedMetadataRef` still uses it, and collapsing that too would reach into the repair path.
- Verified: `mise run lint` clean, `mise run test:ci` green (Vogon 56/56, roger-roger 4/4).

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 679c864d16dbbb0f9e0bd3fadbdd9ad79a2b8d1f. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/doctor.go
+++ b/cmd/entire/cli/doctor.go
@@ -357,6 +357,46 @@ func discardSession(ctx context.Context, ss stuckSession, _ *git.Repository, err
 	return nil
 }
 
+// reportMetadataDivergence prints the metadata-branch verdict for a comparison
+// that is not disconnected: either plain OK, or DIVERGED when both sides advanced
+// since their merge base.
+//
+// Divergence needs saying because it is the one state whose resolution rewrites
+// the local ref: the next fetch from the elected sync remote replays the local
+// commits onto the fetched tip (SafelyAdvanceLocalRef), which loses no
+// checkpoints but does move the ref and re-hash those commits. Nothing else
+// surfaces that — the replay itself only logs.
+//
+// Takes the already-computed comparison rather than re-deriving it: the verdict
+// costs a merge-base subprocess, and the caller has one in hand.
+func reportMetadataDivergence(ctx context.Context, w io.Writer, comparison strategy.MetadataComparison, remoteName string, primary plumbing.ReferenceName) {
+	if comparison.Relation != strategy.MetadataRelationDiverged {
+		fmt.Fprintln(w, "✓ Metadata branches: OK")
+		return
+	}
+
+	fmt.Fprintln(w, "Metadata branches: DIVERGED")
+	fmt.Fprintf(w, "  Local and remote %s have both advanced since they last agreed.\n", primary.Short())
+	// Labels are fixed-width and the remote name trails the hash: padding it
+	// into the label misaligns the pair for every remote name that is not
+	// exactly as long as "origin".
+	fmt.Fprintf(w, "    local:  %s\n", comparison.Local.String()[:12])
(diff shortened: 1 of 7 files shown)


Soph:
Absorbed into #2036, which consolidates the checkpoint sync-destination work into one PR. `c2338ad36` is cherry-picked there, which also resolves this PR's conflict against main.

One substantive change during the rebase: this PR predated #1951's review, and its table-driven rewrite of `TestEnsurePrimaryRef_NonElectedOriginNeverSeedsLocal` carried a second case asserting that a failed election must **not** seed the local primary from origin. #1951 deliberately reversed exactly that — the orphan buys no safety (checkpoint pushes are already fail-closed while the election is broken) and guarantees a diverged history once the setting is fixed. Main's two test functions win; this PR's helper change (a HEAD-only `electionHeadHash`) is kept. Both calls are recorded in the commit message there.

Closing in favour of #2036. Review there.

PR #2004, opened 2026-08-14
fix(search): dedupe folded legacy sessions by checkpointId (ENT-1595)

alishakawaguchi:
## What was broken

When the search service folds legacy checkpoints into sessions (ENT-1595), those rows arrive with an **empty sessionId**. In the CLI that caused two problems:

1. The cross-cell deduper skips rows with an empty id, so it never deduped legacy sessions at all — a repo mirrored across cells reported each legacy session **twice**.
2. Those rows showed a **blank id** in results (the TUI id column and the agent-facing compact JSON).

(Note: unlike the web BFF — whose deduper keyed empty sessionIds on `""` and genuinely collapsed legacy rows — the CLI deduper short-circuits on an empty id, so its bug was the opposite: a duplicate leak, not a collapse.)

## Why

Legacy checkpoints predate the sessionId attribute, so their only stable identity is their checkpoint id — which the CLI wasn't using for dedup or display.

## The fix

1. Give a legacy folded session a real display id by falling back to its checkpoint id (no more blank id column / JSON field).
2. Dedupe on a repo-qualified key: checkpoint ids are unique only within a repo, so the key qualifies by org/repo (lowercased, to survive casing skew across cells). Genuinely duplicated legacy rows now collapse across cells; two repos' rows that happen to share a checkpoint id stay distinct. The same qualification is applied to raw checkpoint rows, which had the identical hazard.
3. `ResultID` stays the raw, machine-lookup id for display; a separate `DedupID` carries the repo-qualified identity.

## Verified

Unit tests plus a deduper-level test over legacy rows: distinct rows kept, mirrored rows collapsed, casing skew handled, cross-repo rows not collapsed.

### DIFF
--- a/cmd/entire/cli/search/search.go
+++ b/cmd/entire/cli/search/search.go
@@ -120,8 +120,10 @@ type SessionResult struct {
 	SessionID string `json:"sessionId"`
 	// CheckpointID identifies a server-folded legacy session (ENT-1595) whose
 	// sessionId is empty (the checkpoint predates the sessionId attribute). The
-	// cross-cell dedupe (ResultID) falls back to it so distinct legacy rows are
-	// not collapsed by matching empty sessionIds — matching the BFF's merge.
+	// cross-cell dedupe (DedupID, not ResultID — which stays the raw display id)
+	// falls back to a repo-qualified form of it, so a repo mirrored across cells
+	// reports each legacy session once instead of twice. Drill-down is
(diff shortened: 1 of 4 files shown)


alishakawaguchi:
bugbot run

alishakawaguchi:
bugbot run

Soph (dismissed):

Soph (approved):

PR #2012, opened 2026-08-17
fix(checkpoint): preserve newer same-ref queue entries

### LINKED ISSUE #2008: git-refs push queue can drop a newer same-ref update during an in-flight push
### What happened?

The git-refs checkpoint push queue can acknowledge a newer update that was not
actually pushed.

`PushQueue.Drain()` snapshots queued ref names without removing them. After the
network push succeeds, `PushQueue.Remove()` removes every current queue entry
whose ref name matches the drained set. If the same checkpoint ref advances and
is enqueued again while that push is in flight, the successful push publishes
the older object but `Remove()` also deletes the newer same-ref queue entry.

The resulting state is:

- remote checkpoint ref: older object A;
- local checkpoint ref: newer object B;
- push queue: empty.

I expected the enqueue for B to remain queued for the next push. Instead, queue
state says there is nothing pending, so the latest checkpoint transcript,
summary, or attribution can remain local until another write happens to enqueue
that ref again.

This is a local concurrency/correctness and recoverability bug, not a remote
security vulnerability. It requires the same checkpoint ref to be updated
during the network-push window. A serial single-agent workflow is unlikely to
hit it, but it is reachable when concurrent processes or linked worktrees share
the Git common directory and update the same checkpoint.

The relevant paths on current main are:

- `cmd/entire/cli/checkpoint/pushqueue.go`: `Drain()` returns de-duplicated ref
  names; `Remove()` removes all current entries matching those names.
- `cmd/entire/cli/strategy/manual_commit_push.go`: after a successful batch
  push, `flushCheckpointRefsQueue()` calls `queue.Remove(existing)`.

### Steps to reproduce

The smallest queue-level regression is:

```go
func TestPushQueue_RemovePreservesNewerSameRefEntry(t *testing.T) {
	t.Parallel()
	q := NewPushQueue(t.TempDir())
	ref := plumbing.ReferenceName("refs/entire/checkpoints/a1/a1b2c3d4e5f6")

	require.NoError(t, q.Enqueue(ref))
	drained, err := q.Drain()
	require.NoError(t, err)

	// Simulate the same ref advancing and being enqueued during the push.
	require.NoError(t, q.Enqueue(ref))
	require.NoError(t, q.Remove(drained))

	remaining, err := q.Peek()
	require.NoError(t, err)
	require.Equal(t, []plumbing.ReferenceName{ref}, remaining)
}
```

It currently fails with `remaining=[]`.

I also reproduced the production boundary with a disposable local repository
and bare remote:

1. Point a queued checkpoint ref at commit A.
2. Start `strategy.PushQueuedCheckpointRefs` and pause the bare remote in a
   `pre-receive` hook.
3. While the push is paused, create commit B, advance the same local checkpoint
   ref to B, and enqueue that ref again.
4. Release the push and wait for it to succeed.
5. Compare the remote tip, local tip, and `queue.Peek()`.

This reproduces deterministically under the controlled pause: remote A, local
B, empty queue. Both unchanged regressions reproduced the expected failure on
current main, while all `8/8` focused upstream `TestPushQueue_*` tests and `4/4` focused
`TestPushQueuedCheckpointRefs*` tests passed. The existing
`TestPushQueue_RemovePreservesLaterEntries` covers a different ref appended
during the push, not a newer generation of the same ref.

### Entire CLI version

Entire CLI v0.10.0

### OS and architecture

Windows amd64 (Microsoft Windows NT 10.0.26200.0)

### Agent

Codex

### Terminal

PowerShell

### Logs / debug output

```bash
drained=[refs/entire/checkpoints/a1/a1b2c3d4e5f6] remaining=[]
newer same-ref enqueue was lost

remote_tip=cd5b922f02ed8c7c275d7938b4eddfa979c8d6f7
local_tip=74724cac5b2968b25d5b50b16c11785ceaef653d
remaining_queue=[]
same-ref update was not retained after successful push


## Additional context

I searched the current public issues and pull requests and found adjacent work,
but no exact report or active fix for this same-ref generation loss:

- `#1635` concerns one remote consuming queue state needed by a second remote.
- `#1917` / `#1926` concern concurrent checkpoint-store writers racing before
  queue publication.
- This report concerns a newer enqueue being removed after an older in-flight
  network push succeeds.
```

### Additional context

Go version used by the disposable regression module: `go1.26.5`.

mikemikimike:
## Related issue

Closes #2008

## Background

`PushQueue.Remove` deleted every queued entry matching a drained ref. If the same checkpoint ref was enqueued again while the network push was in flight, the successful push removed the newer update as well, leaving the queue empty while the remote still pointed to the older object.

## Changes

- Preserve queue entries after the first matching entry for each drained ref.
- Add a regression test covering a newer same-ref enqueue during a push.
- Keep existing de-duplication behavior for `Drain` and `Peek`.

This is backward-compatible and only changes post-push removal semantics so newer queued generations are retried.

## Testing

- `docker run --rm -e GOPROXY=https://goproxy.cn,direct -v "D:/project/tscodex/remote-entire-cli-2:/src" -v entire-go-cache:/go -w /src golang:1.26 go test ./cmd/entire/cli/checkpoint -run TestPushQueue -count=1` — passed (`ok ... 0.042s`)
- `git diff --check` — passed.

Not run: `mise run lint` and full `mise run test`; `mise` is not installed in the environment.


Soph:
The diagnosis is right and the regression test is genuine - I verified red/green by reverting only `pushqueue.go` and keeping the test, and `TestPushQueue_RemovePreservesNewerSameRefEntry` fails as it should. The defect is real: `pushQueueEntry` is `{Ref string}` with no hash or generation, so two generations of the same ref are byte-identical lines, and `Drain` snapshots without reserving. Remote ends up at gen1, local at gen2, queue empty, and `CountUnpushedCheckpoints` reports 0.

Two things stop it landing as written.

**It breaks an existing test.** `TestMigrateBranchToRefs_ReenqueuesAlreadyImportedRef` fails on this branch. `MigrateBranchToRefs` deliberately double-enqueues every ref - `setRef`'s best-effort `enqueueForPush` (`refs_store.go:210`) plus migrate's guaranteed `queue.Enqueue` (`migrate.go:128,146`), both commented "duplicates collapse on Drain". With counting removal, `Remove([ref])` deletes one of the two and the ref stays queued. Confirmed with a probe: one raw line survives. Reverting just the `pushqueue.go` hunk makes it pass again, so it's this change rather than something pre-existing. Easy to miss - `-run TestPushQueue` doesn't reach it.

**It doesn't fix the case #2008 gives as the reason it matters.** The queue lives in the git common dir (`pushqueue.go:18-26`), so linked worktrees share one file - the concurrent-process scenario the issue calls out. Two flushers both `Drain()` (which doesn't reserve), a write during the pushes enqueues the ref again, both call `Remove([A])`, each consuming one occurrence. Probe result: queue empty, newer generation lost again. Counting is only correct while exactly one flusher is outstanding.

**A shape that fixes both:** the invariant you want isn't "remove one occurrence" but "remove this entry only if the object I published is still the local tip" - a compare-and-swap on the ref value. That's nearly free here: `partitionLocalRefs` (`push_common.go:29-45`) already calls `repo.Reference(ref, false)` and discards the hash with `_`. Snapshot those hashes, pass them into `Remove`, and drop an entry only when the local ref still equals its snapshot. That covers the single-flusher case, covers the multi-worktree case, and keeps `Remove`'s "these refs are gone" contract intact so the migrate test stays green. `batchPushRefs` pushes `ref:ref` so git re-resolves anyway - a ref that advanced inside that window just causes one idempotent re-push.

Minor: `rewriteLocked`'s comment at `pushqueue.go:160-162` still says it writes refs "(de-duplicated, one line each)", which the `kept` slice no longer guarantees - 5 enqueues during a push leave 5 raw lines.

I'd keep #2008 open until the multi-worktree half is covered.

The refactor into `readEntriesWithRawLinesLocked` / `readEntriesLocked` / `readLocked` preserves `rawLines` semantics exactly, so `Drain`'s compaction trigger is unchanged - the compaction tests confirm it. Good, well-localized work; the gap is scope and one test you had no way to know to run.

PR #2022, opened 2026-08-17
fix(search): drop the Checkpoints tab from the search TUI (ENT-1595)

alishakawaguchi:
## What the checkpoint tab looked like before

The `entire search` / `entire checkpoint search` TUI had four type tabs — **Checkpoints, Sessions, Commits, Code** — with **Checkpoints as the default** selected tab. The CLI never sends a type filter to the search service; the backend returns all types and the TUI filtered the results client-side by `type == checkpoint` for that tab.

## Why it's now stale

ENT-1595 made the search service **fold checkpoint hits into their owning sessions server-side** — a checkpoint now surfaces as its session, and the general search path no longer returns standalone checkpoint rows (they come back as `session` rows carrying a `matchedCheckpointId`). So the CLI's Checkpoints tab — the default — now always shows **zero results**.

The entire.io web frontend already dropped its Checkpoints tab (entire.io #3690); web search now shows sessions, commits, PRs, and repos.

## What I changed

- Removed the **Checkpoints** tab from the search TUI.
- Made **Sessions** the default results tab.
- Renumbered the type keys: `1` Sessions, `2` Commits, `3` Code (help hint updated `1-4` → `1-3`).
- Updated the affected TUI tests.

Kept scoped to the tab: the internal `typeFilterCheckpoints` value and the generic type-filter/pagination machinery remain (still used to classify any stray rows and exercised by tests). No changes to the search request, the non-interactive/accessible output, or the search service.

## Product decision

**Remove the tab entirely, matching the web (#3690) and the fold model** — rather than keeping it as an opt-in `filter_type=checkpoints` view. Rationale: the CLI has no code path that requests raw checkpoints (it never sends a type filter), and the web made the same call. Raw checkpoints stay reachable **by ID** via `entire checkpoint explain <id>`, which folded session rows already route to, so removing the browse tab strands nothing.

## Verification

`mise run fmt && mise run lint` clean; `mise run test` green (9158 tests). E2E not run locally (per repo guidance; CI runs it on isolated runners).

---

> [!NOTE]
> **Low Risk**
> UI-only tab/default/key changes in the search TUI with matching test updates; no search API or auth changes.
> 
> **Overview**
> Removes the **Checkpoints** type tab from the interactive `entire search` TUI and aligns it with server-side folding (ENT-1595) and the web UI: checkpoint hits show up as **Sessions**, so the old default tab was always empty.
> 
> **Sessions** is now the default tab. Type shortcuts are **1** Sessions, **2** Commits, **3** Code (footer help **1-3**). Tab counts ignore checkpoints; the search placeholder mentions sessions/commits/code instead of checkpoints.
> 
> Tests assert the Checkpoints label is gone, key bindings match the new layout, and still use `typeFilterCheckpoints` internally where they exercise generic list/pagination behavior.
> 
> The diff also downgrades `@opencode-ai/plugin` in `.opencode/package-lock.json` (1.4.7 → 1.4.3), which looks unrelated to the search UI change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c44d1b23539baa76afe117080a2c18294444ff13. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


alishakawaguchi:
bugbot run

alishakawaguchi:
bugbot run

evisdren (changes requested):
just one very small change, the order shoudl be commits, sessions, code so that it aligns with the UI. currently it's sessions, commits, code.

alishakawaguchi:
Done in 139406c1a — tabs now read **Commits, Sessions, Code** (keys 1/2/3) to match the web UI order.

evisdren (approved):

PR #2027, opened 2026-08-17
fix(opencode): preserve prompts after metadata fallback

### LINKED ISSUE #2001: [OpenCode] Preserve user prompts when message events arrive in sequence
## Summary

The generated OpenCode plugin can start user turns with an empty prompt and then skip the event containing the actual prompt text.

## Environment

- Entire CLI: `0.10.0`
- OpenCode: `1.18.18`
- OS: macOS (Apple Silicon)

## What appears to happen

In the generated `.opencode/plugins/entire.ts`:

1. `message.updated` stores the user message, adds its ID to `seenUserMessages`, and calls `fireTurnStart` with `prompt: ""`.
2. OpenCode then emits `message.part.updated` for the text part.
3. The text-part handler requires `!seenUserMessages.has(msg.id)`, so it does not call `fireTurnStart` with `part.text`.

When these events arrive in that order, Entire records an empty prompt instead of the user's prompt.

Relevant generated-plugin code:

```ts
if (msg.role === "user" && !seenUserMessages.has(msg.id)) {
  seenUserMessages.add(msg.id)
  fireTurnStart({
    session_id: sessionID,
    prompt: "",
    model: currentModel ?? "",
  })
}
```

The later `message.part.updated` path is guarded by the same `seenUserMessages` set.

## Expected behavior

The user's text should be captured when the text part becomes available. The fallback path should either avoid marking the message as fully seen or update the existing turn when `message.part.updated` provides `part.text`.

This was identified while reviewing the generated plugin in [tonal PR #15209](https://github.com/tonalfitness/tonal/pull/15209).

MuskanPaliwal:
Fixes #2001.

## Problem
OpenCode emits a user `message.updated` event before the `message.part.updated` event containing the prompt text. The plugin starts the turn from that metadata event with an empty prompt so a fast agent commit still sees an ACTIVE session. Because the message was then marked as seen, the later text part was ignored and checkpoints lost the user prompt.

The first version of this PR kept a marker for those promptless starts and fired `turn-start` again when the text arrived. That recovered the prompt, but `turn-start` is not a prompt setter. It captures pre-prompt Git state, creates a new `TurnID`, recalculates prompt attribution, and clears `LastCheckpointID` and `TurnCheckpointIDs`.

The timing this fallback exists for makes that destructive. A checkpoint may already have been created between the empty metadata event and the text event.

The new integration regression exercises exactly that order:

```
message.updated
  → turn-start with empty prompt
  → agent commit and checkpoint
  → message.part.updated with prompt text
```

Against the second-`turn-start` implementation, the prompt was recorded, but the original turn ID changed, the pre-prompt snapshot was replaced, attribution was recalculated, and both checkpoint fields were cleared.

## Fix
Keep the early empty `turn-start`, but repair the missing text through a new internal `PromptUpdate` lifecycle event.

`PromptUpdate` writes the late text to `prompt.txt` and updates session `LastPrompt`. It does not call `CapturePrePromptState` or `InitializeSession`, so it cannot start another logical turn or reset checkpoint bookkeeping.
The file and state updates run inside `MutateSessionState`. Checkpoint condensation clears `prompt.txt` while holding the same session lock, so either ordering is safe:

- If repair gets the lock first, condensation sees and checkpoints the prompt before clearing the file.
- If condensation gets the lock first, repair writes the prompt after condensation finishes, leaving it available for turn-end finalization.
The OpenCode plugin keeps a `promptlessTurnStarts` marker per message. Empty text updates leave the marker pending; the first non-empty part sends one synchronous `prompt-update`; later duplicates do nothing. Multiple pending messages are tracked independently. The set is cleared on session switches, session deletion, and server disposal.

The generated plugin template and the checked-in dogfood plugin remain byte-identical.

## Cost
The plugin does expected `O(1)` set work per event and keeps `O(P)` additional state for `P` promptless messages awaiting text.
A late repair starts one synchronous hook subprocess, reads and rewrites `prompt.txt` in `O(Q + L)` time for existing prompt data `Q` and incoming prompt length `L`, and performs `O(S)` session-state I/O for state size `S`.

It performs no worktree scan. The earlier second-`turn-start` path repeated work proportional to worktree size `F`, including both status walks used by turn initialization.

In the real OpenCode run, the empty `turn-start` took 74 ms and the subsequent `prompt-update` took 5 ms in the same repository.

## Verification

- [x] Regression test fails before the fix by capturing only the empty fallback prompt.
- [x] Regression test passes after the fix with `turn-start("")` followed by `prompt-update("preserve this prompt")`.
- [x] Duplicate text-part updates do not trigger additional hooks.
- [x] OpenCode package tests pass.
- [x] `mise run check` passes, including lint, race-enabled unit and integration tests, and E2E canary tests.
- [x] The generated plugin template matches the repository’s installed OpenCode plugin.
- [x] A real `opencode run` session produced a checkpoint containing the complete original prompt.
- [x] A checkpoint occurring between the metadata event and prompt-text event preserves the original turn and checkpoint bookkeeping.
- [x] Prompt repair avoids repeating the complete `turn-start` lifecycle.

## Scope
`PromptUpdate` is an internal lifecycle event used only by OpenCode in this change. There are no new CLI flags, settings, dependencies, or changes to other agent integrations.


MuskanPaliwal:
Hi @Soph , would really appreciate your review on this pr. Thanks!

MuskanPaliwal:
Merged `main` and resolved the conflicts in 6f1d98615.

The conflict was in the code that writes `prompt.txt`. This PR moved it into a shared helper, while `main` added filesystem protections. I kept the helper and updated it to use `entiredir` and `osroot`, so both turn-start and late prompt updates get those protections.

I also replaced the removed `MutateSessionStateSaved` call with `MutateSessionStateOnSaved`. Skill telemetry now runs through its callback, after session state is saved and the lock is released.

Added a regression test for symlinked session directories and prompt files. Both prompt-writing paths reject them and leave the target file untouched.

Lint, unit and integration tests with race detection, and all 60 deterministic canary tests passed.

PR #2034, opened 2026-08-18
feat: link mid-task background-subagent commits to their session

peyton-alt:
https://entire.io/gh/entireio/cli/trails/1077

## The incident

Background Claude Code subagents run `git commit` between the parent session's turns. `tryAgentCommitFastPath` (the prepare-commit-msg fast path) only trusted **ACTIVE** sessions; between turns the session is IDLE, so the fast path declined, and the slow content-detection path then found no evidence linking the commit to the session (the subagent's work wasn't in the parent's shadow baseline). The commit shipped with no `Entire-Checkpoint` trailer at all. On a real subagent-driven feature branch, six of seven commits went unlinked this way.

## The design (after merging #2032's durable-record model)

1. **Trigger (idle + task content).** `tryAgentCommitFastPath` now treats a session as eligible when it's ACTIVE **or** IDLE with a fresh task record (`idleWithTaskContent`, `strategy/manual_commit_hooks.go`). A record counts whether it's still in flight or completed-but-unmaterialized — the same `HasTaskContent()` shapes condensation keys on — and each record is bounded by its own `StartedAt` against `activeSessionInteractionThreshold` (24h), so a subagent that dies without ever sending `SubagentStop`/`SessionEnd` doesn't leave the session a trailer candidate forever. The same predicate feeds `shouldCondenseWithOverlapCheck`'s overlap-check bypass, so the trigger and the condensation trust share one rule.
2. **Content (the condensation materializer).** The base branch (#2032, issue #2058) made `session.TaskRecord` a durable ledger: condensation materializes each record's transcript-so-far into the checkpoint's `tasks/<tool-use-id>/{agent-<agent-id>.jsonl, task.json}` subtree, live records survive for the next condensation, and every checkpoint is self-contained. That means the trailer this PR adds is backed by content *by construction*: the very condensation the trailer triggers stores the subagent's transcript-so-far. The earlier revision of this PR pre-wrote a commit-time transcript snapshot (`captureInFlightTaskCommitSnapshot` / `captureInFlightTasksForCommit`) gated on `headHasCheckpointTrailer` to keep the trailer from dangling — all of that is now removed as redundant, along with its tests and post-commit wiring.

## Overlap-trust scoping

`idleWithTaskContent` is deliberately scoped to `PhaseIdle` only, not `PhaseEnded`. An earlier version leaked the overlap-check bypass to ended sessions too; that shape was empirically reproduced and fixed (`fe2aa5fdb`), with a regression test pinning that an ended session with a lingering record still declines — its records belong to its own final condensation, not this fast path.

## The trade-off

Because eligibility is keyed off a no-TTY commit plus a fresh task record, the trust window spans the entire lifetime of a background task — including raw-mode TUI git clients (lazygit, gitui, tig) that spawn `git commit` without a real terminal. This means an unrelated no-TTY commit made while a background task happens to still be running could get mislinked. We accept this: a mislink is recoverable (and rare — it requires a fresh record plus an unrelated no-TTY commit in the same narrow window), while a silent unlink is not. Records are genuine evidence of live agent work, not a guess. The window is also narrower than it used to be: the base branch's dead-owner sweep completes a crashed agent's records, so orphaned records no longer linger for the full 24h in the common case.

## Non-regression pins

- An ordinary commit made while IDLE with **no** task records stays exactly as before: no trailer (`TestSubagentCheckpoints_IdleCommitNoMarkers_NoTrailer`).
- Record-less behavior (foreground tasks, ACTIVE-session commits) is byte-identical to pre-PR behavior — the new eligibility branch is additive.
- The end-to-end arbiter, `TestSubagentCheckpoints_CommitWhileIdleWithLiveMarker_LinksAndCondensesContent`, drives the real prepare-commit-msg + post-commit hook chain and now asserts the strong guarantee: the mid-task commit's condensed checkpoint **contains `tasks/<id>/agent-<id>.jsonl` with the subagent's real transcript content** (stronger than the previous revision, which could only check the pre-written shadow-side snapshot).

## Known gaps / follow-ups

1. ~~**Pre-existing orphan-prune gap.**~~ **Fixed by #2032.** `findSessionsForWorktree`'s orphan sweep now spares record-bearing sessions (`!state.HasTaskContent()` in the prune predicate), so an idle session whose only activity is one open task record survives to `prepare-commit-msg` — a first-turn background commit links. The integration test's read-only seed-task workaround has been removed accordingly.
2. ~~**Sweep-finalize follow-up.**~~ **Fixed by #2032.** `finalizeExitedSessions` now completes a dead-owner session's live records (`sweepCompleteLiveTaskRecords`, `session_finalize.go`) before ending it, exactly as a clean SessionEnd does — preserving transcripts and closing the trust-window trade-off from the other end.
3. **Identity/ancestry discovery** (linking a commit to a session by content/authorship signals beyond records) composes into the same fast-path loop later, via #2013. Out of scope here.

## Trailer eligibility is bounded; condensability is not

A session's task records make it condensable with no freshness bound — that is deliberate, and it is what keeps a stale record's transcript-so-far from being stranded unmaterialized. Claiming a commit is bounded separately: only a record within `idleWithTaskContent`'s 24h window may mint a trailer. Without that split, an IDLE session whose only content was a >24h record got a trailer from the slow path and was then refused by PostCommit, leaving `Entire-Checkpoint` pointing at a checkpoint nothing ever wrote — on an unrelated human commit. Two reviewers reproduced it end to end; `staleRecordIsOnlyContent` closes it, and PostCommit now warns whenever a trailer is minted and no session condenses into it, so any future divergence is visible rather than silent.

An earlier revision of this PR also exempted record-bearing sessions from PostCommit's read-only skip. That is reverted: the skip only fires when another session's files are in the commit, and PostCommit hands one checkpoint ID to every condensing session, so that other session fills the trailer anyway. The exemption prevented no dangling trailer and only let a record-bearing session condense into someone else's commit.

## Test plan

- [x] `mise run fmt && mise run lint` — clean
- [x] `mise run test:ci` — unit + integration green, Vogon canary + roger-roger E2E 4/4


Soph:
Reviewed at `b7ed75fe1`. One item, and I'm taking it off your plate.

**`warnDanglingCheckpointTrailer` fires on amend, cherry-pick, and revert — I'll land it separately, please drop these hunks.**

`manual_commit_hooks.go:1090`, plus `anyCondensed` at `:1022` and the void→bool change on `postCommitProcessSessionLocked`.

The doc says the condition is "never a routine outcome, so it warns." It's routine:

- `handleAmendCommitMsg` (`:562`) deliberately preserves or restores the trailer. A message-only amend has no new content, so `hasNew` is false, `shouldCondenseWithOverlapCheck` short-circuits, and `handler.condensed` stays false for every session — it's only assigned at `:713` and `:742`, both behind `shouldCondense`. Warn fires.
- Same for cherry-pick and revert, which `isGitSequenceOperation` (`:319`) covers via `CHERRY_PICK_HEAD`/`REVERT_HEAD`. The rebase branch skips phase transitions, not condensation, so a cherry-picked range emits one WARN per commit.

In all of those the trailer isn't dangling — its checkpoint was written by the original commit. Verify: `git commit --amend -m` on any trailered commit, then grep `.entire/logs/` for `trailer left dangling`.

It's worth having, though: guarded properly it catches the IDLE + empty-`FilesTouched` dangling-trailer shape, which I have a repro for. It also needs a test — nothing in the PR references `anyCondensed` or the warn. Since it doesn't depend on your predicate and applies to main as-is, I'll open it off main with the sequence-operation guard, an inherited-trailer check against `state.LastCheckpointID`, and coverage. Drop it here and your rebase gets smaller.

cmd/entire/cli/strategy/manual_commit_hooks.go
```diff
@@ -1587,15 +1616,38 @@ func (s *ManualCommitStrategy) filterSessionsWithNewContent(ctx context.Context,
...
+// staleRecordIsOnlyContent reports whether the sole reason state has new
+// content is a task record idleWithTaskContent declines. The re-check runs
+// against a copy with the records removed, so a session that also grew a
+// transcript, tracked files, or steps is never excluded; it is reached only for
+// the rare record-bearing session that already failed the freshness bound, so
+// the common path pays nothing. Errors fail open toward stamping.
+func (s *ManualCommitStrategy) staleRecordIsOnlyContent(ctx context.Context, repo *git.Repository, state *SessionState, stagedFiles []string) bool {
+	if !state.HasTaskContent() || idleWithTaskContent(state, time.Now()) {
+		return false
+	}
```
Soph: `idleWithTaskContent` returns false for any non-IDLE phase (`:874`), so this mirrors only the second half of the decision it exists to track. `shouldCondenseWithOverlapCheck` (`:773`) is `(isActive && isRecentInteraction(lastInteraction)) || hasTaskContent`.

For a **fresh ACTIVE** session whose only new content is a task record, the two disagree:

- Here: `HasTaskContent()` true, `idleWithTaskContent` false because the phase is ACTIVE, so the re-check runs, `hasOther` is false, this returns true, and the session is dropped from `filterSessionsWithNewContent` with no trailer.
- `shouldCondenseWithOverlapCheck`: the first clause is true, overlap is bypassed, and it condenses.

So the trailer is withheld from a session whose condensation then succeeds. The stated rationale — "it must not mint an `Entire-Checkpoint` the commit's own condensation then declines to write" — holds for a stale ACTIVE session but not a fresh one.

Only reachable on the slow path: `tryAgentCommitFastPath` (`:423`) returns first whenever there's no TTY or `commit_linking="always"`. Verify by committing interactively from a second terminal while the agent is mid-turn with a background subagent and no transcript growth since the last condensation — the commit lands with no trailer.

A full mirror isn't possible here, since `shouldCondenseWithOverlapCheck`'s read-only gate needs committed files that prepare-commit-msg doesn't have yet. The first clause is free, though:

```suggestion
	if !state.HasTaskContent() ||
		(state.Phase.IsActive() && isRecentInteraction(state.LastInteractionTime)) ||
		idleWithTaskContent(state, time.Now()) {
		return false
	}
```

With that, the doc comment's "reached only for the rare record-bearing session that already failed the freshness bound" becomes accurate as written — a fresh ACTIVE session no longer reaches the re-check.

pjbgf (approved):

PR #2038, opened 2026-08-18
fix(strategy): prevent duplicate session-end checkpoints

### LINKED ISSUE #2035: Checkpoint duplicated when a session-end hook is killed between the v1 write and the state save
A session-end hook killed part-way through can leave a checkpoint committed to `entire/checkpoints/v1` whose session bookkeeping never advanced, so PostCommit later writes a **second** checkpoint covering the same transcript range.

Found by an agent review on [trail #1001](https://entire.io/gh/entireio/cli/trails/1001) (finding `01M0A84AHZ7F`) while reviewing #1940, and verified against the code. Filed as a follow-up because the durable fix changes the condensation path, which did not belong in a session-cleanup PR.

### The window

`CondenseAndMarkFullyCondensed` runs the condense *inside* the `MutateSessionState` callback (`cmd/entire/cli/strategy/manual_commit_condensation.go:1350`), and `MutateSessionState` persists the state only after that callback returns (`cmd/entire/cli/strategy/session_state.go:563`). The checkpoint, however, is committed to v1 partway through the callback, at `manual_commit_condensation.go:434`.

Kill the process between those two points and:

- the checkpoint is durable on `entire/checkpoints/v1`
- `CheckpointTranscriptStart`, `LastCheckpointID`, `StepCount` and `FullyCondensed` are all unchanged on disk

so the next PostCommit mints a fresh checkpoint ID over a transcript range that already has one.

The other kill windows are fine, which is what makes this one worth fixing rather than absorbing: before `markSessionEnded` the exited-owner sweep reclaims the session, and before `store.Write` nothing durable happened so the existing fail-open retry is correct.

### Who can hit it

Any agent whose host kills the hook. Today that is Codex, which runs `SessionEnd` inside its shutdown under a hard 3s `SESSION_END_MAX_TIMEOUT_SEC` and terminates the hook's whole process tree on expiry. #1940 widened Entire's self-imposed budget to 2s and pinned ≥1s of headroom for wrapper and binary startup, which makes overrunning much less likely — but a budget only narrows the race, it cannot close it. A SIGKILL, a laptop suspend, or an OOM kill lands in the same window.

No data is lost in any case, and the flock is OS-advisory so a killed hook cannot wedge later sessions. The cost is duplication.

### Fix

Make the v1 write and the state save one recoverable unit, so a retry after a kill cannot duplicate. Roughly, either:

1. persist the advanced bookkeeping (at minimum `LastCheckpointID` and `CheckpointTranscriptStart`) before or atomically with the v1 write, so a retry recognises the range as already covered; or
2. make PostCommit idempotent against an existing checkpoint for the same transcript range, and reconcile rather than mint.

### Also worth answering

- **Do `tokens profile` / `checkpoint list` double-count an overlapping pair?** If they do, the user-visible symptom is inflated token totals, not just a redundant checkpoint.
- **Does Codex send SIGTERM or SIGKILL on expiry?** On SIGTERM, `cmd/entire/main.go:47` prints `Received termination signal, shutting down… signal again to force quit.` to stderr, which would surface on the user's terminal during Codex's own shutdown. The cancellation itself is the graceful path; the message is the problem.

MuskanPaliwal:
Fixes #2035.

Session-end condensation could make a checkpoint durable before its updated session state was saved. If the hook was killed in that window, recovery still considered the transcript unfinished and could write it again under a different checkpoint ID.

On current `main`, the immediate `PostCommit` path described in the issue no longer reproduces this for a no-files session because that path discards it. The durable failure still reproduces through `entire doctor`, which sees the ended session as unfinished and previously generated a new checkpoint ID for the same transcript range.

### Approach

This PR uses a write-ahead checkpoint ID. It saves the intended ID as `CondensationAttemptID`, writes the checkpoint with that ID, advances the session bookkeeping, and then clears the pending attempt. If the process dies after the write, the next attempt reuses the same ID instead of creating a duplicate.

The reservation is also honored during commit preparation and `PostCommit`. If `PostCommit` receives a different checkpoint ID, it preserves the interrupted session and shadow branch rather than overwriting its recovery state.

Advancing `LastCheckpointID` or `CheckpointTranscriptStart` before writing would prevent duplication but introduce a data-loss window: state could claim that a transcript was saved when the checkpoint write had failed. The pending ID records durable intent without claiming that the write succeeded.

`CondensationRecoveryPending` separately records when recovery may need to reconcile an interrupted write created by an older CLI version. This keeps the more expensive storage lookup out of normal condensation paths.

Reserved writes are routed using their checkpoint ID format. If checkpoint configuration changes between reservation and retry, the checkpoint is still written to the backend that originally generated its ID.

### Recovering state from older CLI versions

Older interrupted sessions have no pending checkpoint ID. Doctor reconciles them using the session identity, transcript bounds and content, checkpoint counts, agent and turn metadata, touched files, and session kind. Externalized images are reinserted before comparing transcripts.

Storage listing or reading failures stop recovery and preserve its state for another retry rather than being treated as “checkpoint not found.” If the session has genuinely advanced, recovery creates a new checkpoint instead of reusing the older one.

### Related findings

`checkpoint list` exposes both duplicates because they have different checkpoint IDs, while `tokens profile` can count both and inflate checkpoint-observed token totals. The latter already warns about overlapping checkpoint history.

Codex kills timed-out hook process groups with `SIGKILL`, not `SIGTERM`, so Entire cannot print its normal termination message during a Codex timeout. See Codex’s [hook runner](https://github.com/openai/codex/blob/bb701f1e8c8db884fa0cdafff4b8dcb1604f2fdd/codex-rs/hooks/src/engine/command_runner.rs) and [process-group cleanup](https://github.com/openai/codex/blob/bb701f1e8c8db884fa0cdafff4b8dcb1604f2fdd/codex-rs/utils/pty/src/process_group.rs).

### Follow-up plan

This PR lands the non-destructive part first: it prevents new duplicate checkpoints, makes retries deterministic, and reconciles interrupted writes from older CLI versions. It deliberately does not remove or hide duplicate pairs that already exist.

Questions for a follow-up:

1. Should existing duplicate pairs remain as physical history and be collapsed only in `checkpoint list` and `tokens profile`, or should `doctor` repair the stored history?
2. If one checkpoint becomes canonical, which should win: the original session-end write or the later recovery write? What should happen when their metadata differs?
3. Should repair be automatic, an explicit doctor action, or reporting-only until the user chooses?
4. Should transcript-range idempotency become a shared rule for every condensation entry point, or remain limited to session-end and doctor recovery?

Once those product choices are clear, existing duplicates can be handled separately without coupling destructive cleanup or changed read semantics to this durability fix.

### Verification

Regression coverage includes pending ID reuse, legacy orphan reconciliation, repeated storage read failures, recovery-marker persistence, mismatched commit IDs, backend changes between reservation and retry, and advanced sessions receiving a genuinely new checkpoint.

- [x] `mise run check`
- [x] Lint completed with 0 issues
- [x] Race-enabled unit and integration tests passed
- [x] Vogon canary passed — 56/56
- [x] Roger-Roger canary passed — 4/4

### API Notes

`api/checkpoint.WriteRequest` adds `ReservedSession` for writes whose checkpoint ID was previously persisted. Session state adds the optional `CondensationAttemptID` and `CondensationRecoveryPending` fields.

### Compatibility / Release Notes

Existing session-state files remain compatible because the new fields are optional. Existing duplicate checkpoints are not removed, hidden, or rewritten, and `checkpoint list` and `tokens profile` behavior is unchanged.

This does not add global transcript-range deduplication or storage scans to normal checkpoint writes. Normal writes still use the configured primary and mirrors; ID-based backend routing applies only to reserved retries.


MuskanPaliwal:
Hey @Soph , would be really helpful if i could get your review here. Thanks!

MuskanPaliwal:
Thanks @Soph  for the detailed review. I merged your follow-up PR and addressed the four implementation gaps you called out:

- restored the OPF invariant comment
- removed the duplicated eager-condensation checks
- moved checkpoint ID reservation into `markSessionEnded`, avoiding the extra lock round trip
- corrected the lifecycle docs to distinguish PostCommit from doctor retries

For the recovery concern, lookup now continues past an unreadable candidate when a later exact match exists. It still fails closed when no trustworthy match can be established. I left that policy unchanged because falling back may create duplicate history. Should we keep it fail-closed, or add an explicit user-triggered fallback?

I’m still building context on how session state, lifecycle hooks, condensation, and checkpoint routing interact. Your review helped me see where my initial fix was too local.

I merged the latest `main` and ran `mise run check`. The race-enabled unit and integration suites pass, including all five session-end cases.

cmd/entire/cli/strategy/manual_commit_condensation.go
```diff
@@ -383,25 +483,19 @@ func (s *ManualCommitStrategy) CondenseSession(ctx context.Context, repo *git.Re
...
 	var summary *cpkg.Summary
-	if settings.IsSummarizeEnabled(ctx) && redactedTranscript.Len() > 0 {
-		summary = generateSummary(ctx, redactedTranscript, sessionData.FilesTouched, state)
+	if settings.IsSummarizeEnabled(ctx) && transcript.Len() > 0 {
+		summary = generateSummary(ctx, transcript, sessionData.FilesTouched, state)
 	}
 
-	// Post-commit emits regex-only blobs. OPF runs later in the
-	// pre-push rewrite path, never here.
 	skillEvents := mergeSkillEvents(state.SkillEvents, withSkillEventTurnID(sessionData.SkillEvents, state.TurnID))
```
Soph, line 722: The extraction into `buildCondensationWriteOptions` dropped this comment with no replacement:

```go
// Post-commit emits regex-only blobs. OPF runs later in the
// pre-push rewrite path, never here.
```

That's a guardrail rather than decoration. OPF-at-pre-push is load-bearing (`manual_commit_opf_rewrite.go`, `docs/security-and-privacy.md`), and this function is the one place someone would plausibly reach for another redaction layer. Worth restoring above `skillEvents`.

cmd/entire/cli/strategy/manual_commit_condensation.go
```diff
@@ -1356,11 +1448,57 @@ func (s *ManualCommitStrategy) CondenseAndMarkFullyCondensed(ctx context.Context
...
 		)
-		return nil // fail-open
+		return nil
 	}
 	defer repo.Close()
 
 	var shadowBranchName string
+	var checkpointID id.CheckpointID
+	var shouldCondense bool
+	reserveErr := MutateSessionState(ctx, sessionID, func(state *SessionState) error {
```
Soph: This is a third `MutateSessionState` on the session-end path, and it lands inside the Codex 3s budget — `endSessionNow` wraps `CondenseAndMarkFullyCondensed` in the deadline. `MutateSessionState`'s flock acquire blocks, so this adds a blocking point to the path the budget exists to protect, ahead of the eager condense the budget is actually bounding.

`markSessionEnded` (`cmd/entire/cli/lifecycle.go:1561`) is already a `MutateSessionState` on the same state and runs immediately before this. Reserving the attempt ID there costs nothing and removes the extra acquire.

cmd/entire/cli/strategy/manual_commit_condensation.go
```diff
@@ -1373,6 +1511,7 @@ func (s *ManualCommitStrategy) CondenseAndMarkFullyCondensed(ctx context.Context
 
 		if state.StepCount <= 0 {
 			state.FullyCondensed = true
+			clearCondensationAttempt(state)
```
Soph: This `StepCount <= 0` branch, and the FilesTouched and shadow-branch checks around it, are a second copy of the ladder the reserve mutation runs at :1459-:1479 — same order, same outcomes.

They have to stay in sync, and the failure is quiet if they drift: the reserve mutation persists an attempt ID for a session the condense mutation then skips, and the reservation is only cleared on a successful condense (here, and via `resetCheckpointWindow`). A stale reservation then feeds `checkpointIDForSessions`.

Worth extracting one predicate both call, returning the shadow branch name and the decision.

cmd/entire/cli/lifecycle.go
```diff
@@ -1156,14 +1156,11 @@ func sessionEndCondenseDeadline(ag agent.Agent) time.Time {
...
-// MutateSessionState callback, and the state is saved only after that callback
-// returns. A kill in between leaves the checkpoint committed with
-// CheckpointTranscriptStart / LastCheckpointID / StepCount / FullyCondensed
-// un-advanced, so PostCommit mints a fresh checkpoint ID over the same
-// transcript range. Everywhere else, an incomplete condense simply leaves
-// FullyCondensed false and PostCommit retries.
+// Condensation reserves its checkpoint ID in session state before the durable
+// write. If the process dies after the checkpoint lands but before the remaining
+// bookkeeping is saved, a retry writes the same ID instead of creating a second
+// checkpoint. Doctor also reconciles the pre-reservation state left by older
```
Soph, line 1172: This drops the old comment's "PostCommit retries" without saying what replaces it, and for the case being described PostCommit isn't the retry path at all: `ENDED + GitCommit` carries `ActionCondenseIfFilesTouched` (`cmd/entire/cli/session/phase.go:185`), and an interrupted eager condense leaves no `FilesTouched`, so PostCommit doesn't condense this shape under any checkpoint ID.

`entire doctor` is what recovers it — and it's interactive unless `--force`, so it needs a human. Naming doctor here saves the next reader looking for a PostCommit retry that isn't there. (`TestSessionEndCondensation_AgentCommitLeavesInterruptedSessionForDoctor` pins the behaviour.)

Soph (commented):
Four leftovers from reading the condensation changes. All small, none blocking — the durability mechanism itself looks right.

Checked against `52457c508`.

MuskanPaliwal:
Thanks @Soph. I checked the review against the branch history. Your review is pinned to `52457c508`, while these four changes are in the later `65e36658f`. It looks like the review and follow-up commit crossed:

- restored the OPF boundary comment above `skillEvents`
- moved checkpoint ID reservation into `markSessionEnded`'s existing state mutation, avoiding another flock acquisition
- shared `prepareEagerCondensation` between the fallback reservation and final condensation paths
- named `entire doctor`, rather than `PostCommit`, as the retry path for no-file `ENDED` sessions

While resolving the latest `main` conflicts, I also preserved the newer task-record behavior. Task-only sessions remain condensable with `StepCount == 0` and no shadow branch. I corrected two merged comments that still described `PostCommit` as the general retry path.

### My current understanding

A session is the lifecycle record for an agent interaction. It is separate from the transcript. Entire creates session state at turn start with its identity, phase, base commit, timestamps, and agent metadata. At that point, `StepCount` is zero and the transcript may not exist yet.

`StepCount` tracks temporary checkpoint steps. It increments only after Entire successfully writes a temporary step, and it resets to zero after successful condensation. It does not measure transcript length or tell us whether the session contains useful work.

Checkpoint content can come from the parent transcript, touched files, or subagent task records. A task-only session can have no parent transcript, no shadow branch, and `StepCount == 0`, but still contain a task record pointing to a subagent transcript, files, labels, and token data. That session must still be condensed. Only a session with no transcript, no touched files, and no task records is empty.

The ownership split is:

- If an ended session has files waiting for a user commit, session-end condensation leaves it alone. Once Git creates the commit, `PostCommit` reads the `Entire-Checkpoint` trailer and condenses the session against that commit.
- If an ended session has no pending files but has transcript or task content, the session-end path eagerly condenses it.
- If eager condensation is interrupted, `entire doctor` recovers it. `PostCommit` must not attach the interrupted session to the next unrelated commit. Doctor remains human-triggered unless it runs with `--force`.
- If the session has no meaningful content, condensation skips it instead of writing a metadata-only checkpoint.

I understand `CondensationAttemptID` as write-ahead identity. Entire reserves the checkpoint ID before the potentially interruptible write. If the process dies after writing the checkpoint but before saving the completed state, recovery reuses the same ID instead of creating another checkpoint over the same transcript range.

There are two synchronization concerns. The state lock prevents two processes from modifying the same session file concurrently. Separately, the reservation and condensation paths must make the same eligibility decision. Sharing `prepareEagerCondensation` prevents one path from reserving an ID while the other skips the session and leaves a stale reservation behind.

Condensation and `PostCommit` write the fast regex-redacted blobs. OPF is the network-backed privacy pass and stays in the pre-push rewrite, where Entire processes the pending checkpoint blobs in one batch. If an OPF run is selected and fails, the push path withholds those checkpoint refs rather than marking them as OPF-processed.

Please correct me if any of these lifecycle, ownership, or recovery boundaries are off.

cmd/entire/cli/strategy/manual_commit_condensation.go
MuskanPaliwal: Restored this above `skillEvents` in `buildCondensationWriteOptions`. The extraction moved the write-options construction and dropped the comment along the way. This is still where post-commit keeps regex-only redaction and leaves OPF to pre-push, so the guardrail belongs here.

cmd/entire/cli/strategy/manual_commit_condensation.go
MuskanPaliwal: Moved the checkpoint ID reservation into `markSessionEnded`, which already holds the session-state lock. Empty sessions are also marked complete there, while sessions with files are left for PostCommit.

`CondenseAndMarkFullyCondensed` can now skip those cases without another lock acquisition. The fallback mutation remains for late subagent captures that add task content after SessionEnd. I added a contention test that holds the state lock and confirms the file-bearing path does not wait for it.

cmd/entire/cli/strategy/manual_commit_condensation.go
MuskanPaliwal: Pulled the shared FilesTouched, empty-content, and shadow-branch checks into `prepareEagerCondensation`. The `StepCount` and task-record check also lives in `hasEagerCondensationContent`.

We still rerun the preflight inside the final mutation because session state can change between marking the session ended and condensing it.

cmd/entire/cli/lifecycle.go
MuskanPaliwal: Fixed the lifecycle comment. The old text said PostCommit retries an interrupted no-file ENDED session, but PostCommit does not condense that state without `FilesTouched`. Doctor is the recovery path.

The updated comment also covers both formats: current versions reuse the ID reserved before the write, and Doctor reconciles states created before reservations existed. `TestSessionEndCondensation_AgentCommitLeavesInterruptedSessionForDoctor` covers that case.

MuskanPaliwal:
The PR is now ready for a review @Soph. Thanks!

cmd/entire/cli/strategy/manual_commit_condensation.go
Soph: Restored at :720-721. Thanks.

cmd/entire/cli/strategy/manual_commit_condensation.go
Soph: Fixed in 914a8bb02. `PrepareSessionEndCondensation` reserves inside `markSessionEnded`'s existing mutation, and `CondenseAndMarkFullyCondensed` picks the ID up with a lock-free `loadSessionState`, falling back to a mutation only when there is no reservation yet.

`TestCondenseAndMarkFullyCondensed_FilesWaitingForCommitDoesNotWaitForStateLock` actually holds the flock and asserts the call returns — a better proof than I asked for. Also good that the no-content branch still returns before `ensureCondensationAttemptID`, so a no-op stop doesn't pay the checkpoints-config load.

cmd/entire/cli/strategy/manual_commit_condensation.go
Soph: Fixed in 914a8bb02 — extracted as `prepareEagerCondensation` and called by both mutations, with `hasEagerCondensationContent` collapsing the StepCount/task-content pair.

Checked that the write-ahead ordering survived the refactor: the reservation still lands in an earlier save, and the condense mutation re-reads `state.PendingCondensationID()` under the lock and bails with `ErrMutationSkip` on empty, so it can't write ahead of a persisted reservation.

Soph (approved):
Looks good!  Thanks again.

PR #2044, opened 2026-08-18
fix(search): route each repo to its canonical placement only (ENT-1672)

evisdren:
https://entire.io/gh/entireio/cli/trails/1086

## Why

`groupReposByCell` fanned out to **every** placement of a repo — home and mirrors alike. Its comment claimed this matches the BFF; the BFF does the opposite. Mirror namespaces are replicated copies of the same content, so the extra legs return duplicate and stale rows, and CLI results diverge from the web.

Measured on `entirehq/peregrine` (3 ready placements: us-east-2 canonical, eu-central-1, ap-southeast-2), query `repo id remap`:

- web / leaf canonical namespace: **6 rows**
- CLI: **100 rows** — provably the union of all three namespaces; 94 rows came from mirror namespaces (mostly ap-southeast-2, whose query-serve currently admits rows down to rerank 0.22 where us-east-2 floors at ~0.45 — that config skew is a separate issue)
- EU namespace is a strict subset of US (replica + index lag): mirrors hold no unique content

## What

Each repo now routes to exactly **one** placement — its **home placement**: the elected processing primary (`primaries.processing` when it names a placement in `placements[]`), else the canonical row-ID convention (row-ID match, else `placements[0]`). One rule for every request shape (`routedRepoPlacement`, the CLI twin of the BFF's `searchPlacement`):

- **Semantic search, scoped**: routes by the home placement and pins its ULID — query-serve resolves pins from its all-accessible-placements byID set (matches the BFF's filtered path, entirehq/entire.io#3782).
- **Semantic search, broad (`--all-repos`)**: same rule; safe because query-serve homes pin-less broad scope by the same election since ENT-1776 (entirehq/entire-search#188). Matches the BFF's broad flip (entirehq/entire.io#3801).
- **Code search, both modes**: same rule; peregrine resolves pins from its all-ids set and homes pin-less broad scope by the election (entirehq/peregrine#214). **Deliberately ahead of the web here**: the BFF's `code-search.ts` still routes canonical until its code-read chain (`/symbols`, `/usages`, file viewer) flips with it in a follow-up PR. No observable difference until then — see sequencing note below.

Shared rules:

- Selection never keys on the `Mirror` flag (real `/repos` rows mark every placement `Mirror:true`, primary included).
- A home placement that is explicitly not ready (`processing`/`failed`/`suspended`) is skipped and reported — an elected-but-unready primary is NOT substituted with a ready mirror (it may predate the clone; the BFF's `searchPlacement` makes the same pick for its caller to skip). Semantic search surfaces the skip as a warning on scoped queries; code search reports it via `skipped_repos`; broad searches also report when the skips left nothing to query. An empty or unknown placement status cannot reach the pick at all — the generated client validates status against the closed enum at decode, failing the whole index read loudly.
- Broad searches were already single-copy in *results* (the leaves narrow unfiltered queries to their own cell's home placements); this change also stops sending wasted legs to cells that host only mirrors.

## Sequencing

Behavior is byte-identical today: every current election matches the row-ID convention (that's how the leaves deployed as a no-op, prod-verified in all 4 cells per entirehq/entire.io#3801's merge gate). Core electing a divergent primary is gated on the BFF's code-read-path flip — peregrine's divergence-warn log is the tripwire. Order of operations: this PR → BFF #3801 (broad semantic) → BFF code-search + code-read chain flip → core may elect divergent primaries.

## Validation

- Unit tests: election preferred over convention, unknown-election fallback, canonical-not-first fallback, unready-home skip + warning, single-home grouping, no-placements guard, placement-ID normalization, skip-report gating (pinned/broad), skipped_repos in both output shapes.
- `go test ./cmd/entire/cli/... ./internal/...` clean; `mise run lint` clean; `mise run test:ci` (unit + integration + canary) clean.
- End-to-end (dev binary against prod): scoped semantic, broad semantic, scoped code, and broad code all verified; the parity case above returns **6 rows** matching the web (was 100). 20-query paired web/CLI parity matrix: scoped went from **1/10** identical result lists to **10/10**; broad top-10 identical (remaining diff is the known limit-semantics skew, not placement scope).

## Notes

- Rebased on main, which resolves single-repo cell targets via the same `primaries.processing` election (`resolveProcessingPlacement`, trails/experts) — this PR is the multi-repo fan-out counterpart. Deliberate difference: the fan-out fails soft (convention fallback, skip-with-report) where the single-repo resolver fails loud, per each path's documented philosophy.
- Stopgap on the road to ENT-1672's server-side coordinator (single fan-out+merge implementation). The invariant this relies on — mirror namespaces are replicated copies with no unique indexed content — should hold by construction; if mirror-unique indexing is ever introduced, search scope needs revisiting.
- The ap-southeast-2 rerank-floor/config skew observed during validation is server-side and unaffected by this PR; filing separately.


evisdren:
Adversarial review (performance / correctness / stability) + blast-radius audit done; fixes pushed in 6c625c060:

**Blast radius:** `groupReposByCell` has exactly two production callers, both under one command — `entire search` (semantic one-shot + TUI, and `--code` + TUI code tab). Every other repo-resolving command (`trail`, `explain`, `experts`, `api --to cell`, `recap`, `activity`, `repo mirror/clone`, checkpoint push/pull, git-remote-entire) uses the mirrors API or `/mirrors/placements` and is untouched. The web BFF's code-search also routes via `canonicalPlacement`, so this aligns code search with web too.

**Findings fixed:**
- ID-less canonical placement was the one remaining silent-loss path (dropped without a report); now rejected as unroutable and reported via `skipped`.
- The skipped warning was inaccurate for unfiltered searches (no repo param is sent, so skips only shrink the cell set); warning is now scoped-only.
- `--code` swallowed skips entirely — a mid-clone repo looked like "no matches". `SkippedRepos` now surfaces in text output and `--json`.
- Stale all-placements comments updated; merge dedupes retained as guards for overlapping cell scopes (crosslinked sessions, home+explicit-cell overlap).
- goconst CI failure fixed (whole-file lint; test literals hoisted to consts).

**Known accepted edges:** (1) unfiltered searches no longer query cells that host only mirrors of the caller's repos — those cells only ever returned their own canonical content, so nothing indexed is lost; under a truncated repo index the fan-out can be narrower than before, and the existing truncation warning covers it. (2) The CLI still has three placement-selection rules (`canonicalRepoPlacement` for search, first-active for recap/activity/explain, refuse-on-multi for trail/experts) — unifying them belongs to ENT-1672's coordinator, not this stopgap.

evisdren:
Adversarial review of the BFF twin (entirehq/entire.io#3782) found a coverage hole that applies here too: routing UNFILTERED (`--all-repos`) fan-out by the elected processing primary silently loses divergently-elected repos. Both leaf services narrow a pin-less broad search to the placements canonical (row id) in their own cell — so when `primaries.processing` names a different cell than the row id, the elected cell excludes the repo from broad scope and the canonical cell never gets a leg. `primaryRepoPlacement` is only safe on the SCOPED path (explicit `repo=` pins resolve from the leaves' all-accessible byID sets).

The BFF now splits: scoped/pinned → election, broad → canonical convention, unify once the leaves narrow broad scope by the election. `groupReposByCell` here needs the same split (callers know scoped vs broad) before core starts electing primaries that differ from the convention. Harmless today — every current row's election matches the convention.

evisdren:
Implemented the split flagged above, and rebased on main (32f1e6e28):

- `groupReposByCell` now takes a `placementRouting` mode: `routeByElection` (processing primary, canonical-convention fallback) vs `routeCanonical` (row-ID match else first, election ignored).
- **Scoped semantic** → `routeByElection` (pins the placement ULID, so the elected cell resolves it).
- **Broad semantic** → `routeCanonical` (pin-less; leaves narrow broad scope to their own canonical placements, so the election would search a divergently elected repo nowhere).
- **Code search, both modes** → `routeCanonical`, matching the BFF's deliberate choice in code-search.ts (placement-replicated content; the web's follow-up reads resolve canonical).

Mirrors entirehq/entire.io#3782 exactly, including the elected-but-unready skip-with-report (no fallback to a ready mirror). New test `TestGroupReposByCell_CanonicalModeIgnoresElection` is the twin of the BFF's code-search election test. `mise run test:ci` (unit + integration + canary) and lint clean after the rebase.

evisdren:
Rebased on main and added the ENT-1776 broad flip: **broad semantic fan-out now routes by the election** (`routeByElection`), collapsing the split rule this PR previously introduced for safety. The gate that forced the split is satisfied — both search leaves narrow broad scope by the election and are deployed to all prod cells ([entire-search#188](https://github.com/entirehq/entire-search/pull/188), [peregrine#214](https://github.com/entirehq/peregrine/pull/214), byte-identical no-op verified per cell), and the BFF's matching flip is entirehq/entire.io#3801.

Code search stays `routeCanonical` deliberately — the whole code-read chain (hits, /symbols, /usages, file viewer) resolves the canonical placement; that flip is tracked separately on ENT-1776.

Shipping this version of the CLI matters for the rollout: released CLIs persist on user machines, so every version released with canonical broad routing grows the population that would mis-route once core elects divergent primaries.

evisdren:
Code search flipped to the processing primary in b9bced4c8 — one home rule for every request shape, so the `placementRouting` split is deleted and `routedRepoPlacement` is now exactly the BFF's `searchPlacement` (election, else canonical convention).

Safe because peregrine#214 already resolves pins from all-ids and homes pin-less broad scope by the election. Deliberately ahead of the web: the BFF's `code-search.ts` stays canonical until its code-read chain (`/symbols`, `/usages`, file viewer) flips in a follow-up PR — behavior is identical until then (every current election matches the convention), and core electing divergent primaries remains gated on that web flip (peregrine's divergence warn is the tripwire).

evisdren:
Adversarial review (correctness / performance / stability / simplicity / standards, findings verified against the generated coreapi client) + simplify pass applied in 72bb2faec:

**Behavior fix**
- Broad (`--all-repos`) searches no longer swallow the skip report when the skips left *nothing* to query — previously a single-repo account with its repo mid-clone got a bare "No results" with no explanation. Skip reporting is now one shared helper (`reportableSkippedRepos`): pinned skips always report; broad skips report only when zero cells remain queryable.

**Verified-and-corrected comment claims**
- The "unknown/empty status is treated as not ready with a report" claim was unreachable: `decodeListReposResponse` validates placement status against the closed enum, so such a value fails the whole index read upstream. Dropped the dead fail-open arm — the readiness check is now exactly the BFF's `status !== "ready"` — and the docs state the real failure mode.
- Standardized on "processing placement" (matching `cell_target.go`/CLAUDE.md), consolidated the routing-rule prose into `routedRepoPlacement`'s doc, and documented the deliberate readiness divergence from `resolveProcessingPlacement` (search skips mid-processing placements; trails tolerate them).

**Simplification**
- Replaced the mutate-by-closure `pick` with `placementByID` + a straight-line fallback chain that reads like the documented selection order.
- Collapsed ~15 repeated placement literals behind a `readyPlacement` fixture; added tests for the previously untested skip-reporting layer (gate + `skipped_repos` in both output shapes).

`mise run test:ci` and lint clean; dev-binary smoke check unchanged (peregrine parity case still 6 rows, single-leg scoped code search).

Soph (approved):

PR #2045, opened 2026-08-19
fix(textutil): strip Cursor's injected <timestamp> block from prompts

peyton-alt:
https://entire.io/gh/entireio/cli/trails/1087

## Summary

Imported Cursor sessions were showing titles like `<timestamp>Tuesday, Aug 18, 2026, 2:37 PM …</timestamp>` in the web UI. Cursor prepends a harness-rendered `<timestamp>...</timestamp>` element to stored user messages, and that tag was missing from `StripIDEContextTags`'s vocabulary, so it leaked straight through into the title.

`StripIDEContextTags` is the single shared prompt cleaner — every prompt-extraction path (live hooks and `entire import` alike) reaches it through `transcript.ExtractUserContent`. Adding one regex to its `systemTagRegexes` table fixes live and imported prompts in one place.

## Design

- The whole `<timestamp>...</timestamp>` block is removed (its body is harness-rendered text), unlike `<user_query>` whose tags are stripped but content is kept — that content is the actual user ask.
- The regex only matches a **single-line body containing no `<`**: `<timestamp(\s[^>]*)?>[^<\n]*</timestamp\s*>`. This is a deliberate boundary, not an oversight:
  - Multi-line spans survive (e.g. a user asking to "rename the `<timestamp>` element ... and update `</timestamp>` closers").
  - A body containing `<` survives, so `<timestamps>` wrapper tags aren't mangled and unrelated tag content isn't clipped.
  - **Accepted tradeoff**: a genuine single-line `<timestamp>short body</timestamp>` element still gets stripped wherever it appears — including inside pasted XML or log text a user quotes verbatim. Covered by an explicit test case so the behavior is pinned rather than incidental.
- `full.jsonl` (the raw stored transcript) stays authoritative and unstripped; this only affects derived surfaces (titles, compact transcripts, commit-message summarization inputs) for every agent, not just Cursor.

## Testing

- `cmd/entire/cli/textutil/ide_tags_test.go`: new cases cover the happy path, attributes on the tag, the multi-line-preserved boundary, the `<timestamps>` wrapper non-mangling, and the accepted single-line-in-pasted-XML tradeoff.
- `cmd/entire/cli/agentimport/cursor_test.go`: pins that the import pipeline strips the tag too.
- `go test ./cmd/entire/cli/textutil/ ./cmd/entire/cli/agentimport/ -count=1` and full `mise run test` (9254 tests) both green; `mise run fmt && mise run lint` clean.

## Related

Sibling entire-api PR (server-side previews for already-pushed data + AI naming for imports): https://github.com/entirehq/entire-api/pull/695

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 844fca59dc32cc50b84d00cf9dd5d836aecdd5ac. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


suhaanthayyil (dismissed):

suhaanthayyil (approved):
LGTM. Anchored `\A` regex correctly scopes the strip to Cursor's injected leading `<timestamp>` without touching user-pasted timestamp content elsewhere in the message. Good catch on the now-redundant double-strip in `cursor/transcript.go` and `factoryaidroid/lifecycle.go` — removing those was necessary once the strip became position-anchored (no longer idempotent). Test cases cover the tradeoffs called out in the description (multi-line preserved, pasted XML preserved, nested tags preserved).

PR #2052, opened 2026-08-19
fix(codex): use repository hook root in linked worktrees

### LINKED ISSUE #2020: Codex sessions are not captured in linked worktrees because hooks are installed in an ignored worktree-local location
## The Problem

When Entire is enabled for Codex from a linked worktree, it writes the hook configuration to `<linked-worktree>/.codex/hooks.json`. The file exists, and Entire may report the hooks as installed, but Codex does not read it. Codex therefore reports **Installed 0 / Active 0**, sends no lifecycle events to Entire, and the session is never captured.

This appears to be a compatibility gap introduced by [openai/codex#21969 — “Use root repo hooks in linked worktrees”](<https://github.com/openai/codex/pull/21969>). Codex now treats the root repository as the shared source of truth for hooks and hook trust across linked worktrees. Entire still installs and checks Codex hooks relative to each checkout’s `WorktreeRoot`, so the two tools disagree about where `.codex/hooks.json` belongs.

I reproduced this against current Entire `main` at `bad8572a` and Codex 0.147-era behavior.

## What I observed

In a linked checkout, `entire enable` creates a worktree-local `.codex/hooks.json`, but Codex `hooks/list` returns no hooks. Placing the unchanged file at the root location Codex considers authoritative makes the same linked checkout discover four Entire hooks.

The two observed states were:

```text
Worktree-local hooks file:
Installed: 0
Active:    0
```

```text
Hooks file at Codex’s authoritative root:
Installed: 4
Active:    0
```

The second result is expected before approval: Codex has found the hooks, but they are not trusted yet. After approval, Active can become 4.

This distinction is important because the original failure is not a trust problem. **Installed 0 / Active 0 means there is nothing to approve.** Codex did not discover the configuration in the first place.

## Reproduction in a conventional linked worktree

A temporary repository and linked checkout can be created with:

```sh
tmp_dir=$(mktemp -d)

git init -b main "$tmp_dir/root"
git -C "$tmp_dir/root" config user.name "Entire Repro"
git -C "$tmp_dir/root" config user.email "entire-repro@example.invalid"
touch "$tmp_dir/root/README.md"
git -C "$tmp_dir/root" add README.md
git -C "$tmp_dir/root" commit -m "Initial commit"

git -C "$tmp_dir/root" worktree add -b feature "$tmp_dir/linked"
```

From `"$tmp_dir/linked"`, run `entire enable` and select Codex. Entire creates `"$tmp_dir/linked/.codex/hooks.json"` as expected. Start Codex from that linked checkout and inspect `/hooks`, or call `hooks/list` through the Codex app server. Codex reports zero installed and zero active hooks.

As a diagnostic positive control, place the unchanged configuration in the original checkout:

```sh
mkdir -p "$tmp_dir/root/.codex"
cp "$tmp_dir/linked/.codex/hooks.json" "$tmp_dir/root/.codex/hooks.json"
```

After restarting Codex in the linked checkout, it discovers four hooks sourced from `"$tmp_dir/root/.codex/hooks.json"`. They initially appear untrusted, so the result is four installed and zero active until the user approves them.

Copying the file is only a way to demonstrate the lookup behavior. It is not the proposed product fix.

## Reproduction in a bare worktree layout

The same problem occurs in repositories organized as a bare common Git directory with sibling checkout directories:

```text
layout/
├── .bare/
├── main/
└── feature/
```

A minimal layout can be created with:

```sh
tmp_dir=$(mktemp -d)

git init -b main "$tmp_dir/seed"
git -C "$tmp_dir/seed" config user.name "Entire Repro"
git -C "$tmp_dir/seed" config user.email "entire-repro@example.invalid"
touch "$tmp_dir/seed/README.md"
git -C "$tmp_dir/seed" add README.md
git -C "$tmp_dir/seed" commit -m "Initial commit"

mkdir -p "$tmp_dir/layout"
git clone --bare "$tmp_dir/seed" "$tmp_dir/layout/.bare"
git --git-dir="$tmp_dir/layout/.bare" worktree add "$tmp_dir/layout/main" main
git --git-dir="$tmp_dir/layout/.bare" worktree add \
  -b feature "$tmp_dir/layout/feature" main
```

Enable Entire for Codex in both `layout/main` and `layout/feature`. Entire creates a `.codex/hooks.json` file inside each checkout, but Codex still discovers no hooks from either one.

When the same configuration is placed at `layout/.codex/hooks.json`, both checkouts discover the four Entire hooks, again initially awaiting approval.

The confusing part of this layout is that the checkout named `main` is also a linked worktree. Enabling Entire in both `main` and `feature` therefore produces two valid-looking files in locations Codex ignores.

## Why Entire enablement does not help

Entire enablement and Codex hook discovery are separate layers. Entire’s settings determine whether Entire processes a hook after receiving it. Codex’s `.codex/hooks.json` determines whether Codex sends the hook in the first place.

Both checkouts can be Entire-enabled, but that does not help when Codex never reads either hooks file. The failure also happens before Entire dispatch runs, so it is unrelated to concurrent-session handling or hook attribution inside Entire.

## Where the paths diverge

[openai/codex#21969](<https://github.com/openai/codex/pull/21969>) changed linked worktrees to use root-repository hooks and shared hook trust. Entire’s current implementation still constructs the Codex hooks path from `paths.WorktreeRoot` in [`InstallHooks`, `UninstallHooks`, and `AreHooksInstalled`](<https://github.com/entireio/cli/blob/bad8572a28885fc88528e8a0201f9067cfc6915b/cmd/entire/cli/agent/codex/hooks.go>). The same assumption is used by [`HookTrustGaps` and `MissingEntireHooks`](<https://github.com/entireio/cli/blob/bad8572a28885fc88528e8a0201f9067cfc6915b/cmd/entire/cli/agent/codex/trust.go>), as well as [`doctor`](<https://github.com/entireio/cli/blob/bad8572a28885fc88528e8a0201f9067cfc6915b/cmd/entire/cli/doctor.go>).

Using `WorktreeRoot` is correct for a normal checkout. In a linked checkout, however, the worktree root and Codex’s authoritative hook root are different directories. This also means Entire’s status and doctor output can inspect a file that exists locally but is irrelevant to Codex.

The precise Git-pointer traversal is inferred from Codex’s implementation and PR. The user-visible discovery behavior was verified directly with `hooks/list`.

## Suggested fix

Entire should have one Codex-specific resolver for the authoritative hooks location and use it everywhere the hooks file is read or written. Installation, reinstallation, removal, installed and active reporting, trust-gap detection, missing-hook checks, and doctor output should all agree on the same path.

For normal checkouts, the resolver should preserve the current project-local behavior. For conventional linked worktrees, it should resolve to the root checkout. For bare layouts, it should resolve to the layout root. It should also handle relative `gitdir` and common-directory forms safely.

Submodules need explicit care here. A resolver that simply takes the parent of the Git common directory can produce a location inside `.git/modules`. Entire must not write `.codex/hooks.json` into Git’s internal storage or mistake an ordinary or linked submodule for a top-level linked worktree. If Codex’s behavior for a linked-submodule shape cannot be reproduced safely, returning a clear diagnostic would be preferable to guessing.

## Migration behavior

Some users will already have Entire-managed hooks in a legacy worktree-local file. Reinstall or repair should merge the Entire entries into the authoritative hooks file while preserving user-defined hooks and unrelated configuration. Once the authoritative file has been updated successfully, Entire should remove only its own managed entries from the old worktree-local file. If that file still contains user configuration, it should remain in place.

The migration should be idempotent and safe when several worktrees share the destination. It may also require one new Codex approval because trust belongs to the authoritative configuration. Entire should expose that state clearly rather than trying to copy or synthesize trusted hashes.

Since all linked checkouts share the authoritative file, uninstalling Codex hooks from one checkout will affect the repository as a whole. The uninstall output should make that scope clear.

## Regression coverage

The focused regression suite should exercise a normal checkout, a conventional linked worktree, a bare layout, and relative `gitdir` or common-directory forms. It should also cover ordinary and linked submodules to prove that the resolver never writes into Git internals.

The behavior worth testing includes installation, reinstallation, uninstall, migration from the legacy location, preservation of user-defined hooks, doctor and status output, and the distinction between installed and active/trusted hooks. Since the destination is shared, concurrent updates from separate worktrees would also be worth covering.

An optional Codex `hooks/list` smoke test could provide useful integration coverage if it can run without invoking a model.

## Related work

This differs from entireio/cli#1197, which covered Codex hook discovery when starting from a nested monorepo directory. This failure is specific to Git linked worktrees and Codex’s root-repository hook semantics.

It is also not the same issue addressed by [4b9b9636e](<https://github.com/entireio/cli/pull/1729/commits/4b9b9636e959bb500a61771b03e7416c28eddffe>) , despite its title, `fix(codex): write hook config to the location codex expects`. That commit handled the Codex `config.toml` feature flag for repositories under Codex’s reserved `agents` directory. It explicitly kept `hooks.json` at the normal project-local path and did not add linked-worktree or Git common-directory resolution. Entire has since stopped managing that feature flag because Codex hooks are enabled by default.

PR [#1958](<https://github.com/entireio/cli/pull/1958>) is related but does not appear to fix the location issue. At the currently reviewed head (`2fe0fbb`), it adds `SubagentStart` and `SubagentStop`, while hook installation and discovery checks still derive `.codex/hooks.json` from `WorktreeRoot`. If that PR lands first, the expected discovered count will become six rather than four, but the location problem will remain.

This issue is not proposing global user hooks, copying files to the layout root as the documented product solution, changes to PR entireio/cli#1958, or changes to concurrent-session behavior. It also should not require paid or real-agent E2E tests; the failure can be verified through Codex hook discovery without running a model.

MuskanPaliwal:
Fixes #2020.

Codex 0.149.0 can load project hooks from a different checkout than the one in which it is running. In a conventional linked worktree, its app-server `hooks/list` response reports the primary checkout's `.codex/hooks.json`; a divergent file present only in the linked worktree is not loaded.

This PR makes that effective location visible without giving Entire ownership of another checkout. Discovery and mutation are deliberately separate: Entire may inspect the file Codex actually loads, but installs, removes, and cleans hooks only in the exact checkout where the command was invoked.

## Discovery and ownership

- `WorktreeHooksPath` identifies `<current-checkout>/.codex/hooks.json`. It is the only mutation target used by install, uninstall, presence, freshness, deselection, and clean flows.
- `DiscoveredHooksPath` identifies the project hooks file Codex is expected to load. It is used only for read-only configuration and trust diagnostics.
- `ResolveHookDiscovery` consumes the shared `gitrepo.ResolveGitLayout` classification rather than maintaining a Codex-specific `.git`, `commondir`, or worktree-marker parser.
- Conventional linked worktrees discover the primary checkout's hook file. Normal checkouts, ordinary submodules, and separate-Git-directory checkouts discover their own hook file.
- `.bare/worktrees`, linked submodules, contradictory metadata, and other behavior not pinned against Codex are reported as unresolved. Entire does not claim an unsupported layout or guess a writable fallback.

`HookDiscovery` carries no write target, migration path, cleanup path, or lock path. Its `RepositoryWide` bit describes the reach of the configuration Codex discovers; it does not authorize repository-wide mutation.

## Checkout-local mutation

`entire enable`, `entire agent remove codex`, agent deselection, and `entire clean` mutate only the current checkout's `.codex/hooks.json`. They do not create another checkout's `.codex` directory, migrate a worktree-local file into the primary checkout, remove a discovered file owned elsewhere, or perform repository-wide cleanup.

Concurrent updates are serialized only with `<current-checkout>/.codex/hooks.json.lock`. There is no `entire-codex-hooks.lock` in the Git common directory and no synchronization across worktrees. An old common-directory lock, if present from a previous build, is neither used nor cleaned from another checkout.

Current-checkout writes are hardened before the final atomic replacement:

- the logical target must be exactly `<current-checkout>/.codex/hooks.json`;
- redirected `.codex` directories, final-file symlinks, non-regular files, unrelated destinations, and Git-metadata redirection are rejected;
- existing user hooks, unknown hook events, and unrelated top-level fields are preserved;
- uninstall removes only Entire-managed entries;
- an existing file keeps its permission bits, while a new file uses `0600`.

## Bounded read-only inspection

The local and discovered files are inspected independently as absent, user-only, Entire-managed, invalid, or unresolved. Inspection is contained to the exact canonical `.codex` directory, requires a stable regular file, caps `hooks.json` at 1 MiB, reads through a limit with one overflow byte, and checks file identity before and after the read. It never creates a directory, rewrites configuration, removes entries, or acquires a mutation lock.

The structured inspection also reports whether the core Entire events are present, which current events are missing, whether managed commands and timeouts match this CLI, whether the current checkout has the project layer Codex requires, whether local trust records were readable, and which declared hooks lack an approval record. Entire does not compute, copy, or synthesize Codex trust hashes.

## Diagnostics and remedy

- `entire doctor` reports the current-worktree path and Codex-discovered path, invalid or unresolved discovery, a missing project layer, managed hook drift, and trust gaps.
- `entire status` keeps current-checkout installation/freshness semantics separate from effective Codex discovery and exposes the mismatch under the existing `codex_hooks` JSON object without changing the warning wording established in this PR.
- SessionStart appends a short, bounded, read-only discovery or trust warning when applicable.

When hooks exist only in a linked worktree, Entire does not repair the primary checkout automatically. The user-owned remedy is to commit `.codex/hooks.json` and apply that commit to the primary checkout (typically by cherry-pick), or run `entire enable` from the primary checkout. Because linked worktrees share Git objects, the commit is immediately available to cherry-pick; no push or fetch is required.

## Pinned Codex integration contract

`TestCodexAppServerHooksList_LinkedWorktreeUsesPrimaryCheckout` is a no-model integration test pinned to `@openai/codex@0.149.0`. It:

1. creates a normal repository and conventional linked worktree with distinct hook markers;
2. starts `codex app-server` in the linked worktree with an isolated `CODEX_HOME`;
3. sends `initialize`, `initialized`, and `hooks/list` JSONL messages over stdio;
4. requires the primary marker and primary `sourcePath`;
5. rejects the linked-only marker; and
6. requires empty `warnings` and `errors`.

The test makes no model request and needs no OpenAI credentials. Local integration runs skip when the exact Codex binary is unavailable. CI installs version 0.149.0 and sets `ENTIRE_TEST_REQUIRE_CODEX_APP_SERVER=1`, so a missing binary or version drift fails rather than silently skipping.

## Verification status

The implementation includes focused unit coverage for Git-layout classification, path-role separation, current-checkout-only install/remove/clean behavior, exact-worktree locking, destination containment, symlink refusal, bounded reads, mode preservation, doctor/status/SessionStart diagnostics, trust inspection, and JSON output, plus the pinned app-server integration contract above.

Full formatting, lint, unit, integration, canary, duplication, and `mise run check` verification is intentionally deferred to implementation-order item 9.


peyton-alt:
Hey @MuskanPaliwal, thanks again  for your contribution, reviewed, this is really solid. A few things before merge:

1. Project-layer premise — ProjectLayerExists assumes a linked checkout needs its own local .codex dir for Codex to discover the shared hooks. That claim gates AreHooksInstalled and a doctor state, and it isn't in the issue. Can you point at the codex-rs source for it, or share a hooks/list run from a linked checkout without a local .codex dir?
2. Trust-key paths — trust matching prefixes on the symlink-resolved hooks.json path. If Codex writes its trust keys with the un-resolved path (macOS /var vs /private/var, symlinked checkouts), we'd report false trust gaps. Can you confirm which form Codex writes and pin it with a test?
3. Lock wait is unbounded — acquireHooksLock retries until the context cancels, so a wedged holder hangs entire enable silently. Please bound the wait and return a clear error instead.
4. Merge conflicts

Happy to merge once these are addressed. Thanks again!

MuskanPaliwal:
Hey @peyton-alt , thanks for the review. I pushed the follow-up changes in [7020927d6](https://github.com/entireio/cli/pull/2052/commits/7020927d66a1edbe5faa607271c7c8adf25ff87b).

**Project-layer behavior:** I tested `hooks/list` from a conventional linked worktree using the latest stable Codex, `0.148.0`. With no local `.codex` directory, Codex returned `hooks: []`. After creating an empty local `.codex` directory, it found the hook from the primary checkout’s `.codex/hooks.json` and reported it as `trustStatus: "untrusted"`. Everything else was kept unchanged. This confirms that the local directory enables project-layer discovery, while the hook configuration remains root-authoritative and approval remains separate. I have screenshots of the version, filesystem states, and raw responses.

Screenshots: 

Screenshot 2026-08-20 at 12 21 07 PM

Screenshot 2026-08-20 at 12 20 58 PM






**Code changes:** Trust matching now parses the complete `<path>:<event>:<group>:<handler>` key, validates the indexes, and compares canonicalized paths. The regression test covers a logical nested-symlink path against the physical authoritative path. Lock acquisition now times out after five seconds with a clear error, while caller cancellation still returns `context.Canceled`.

I also fixed three related cases found while rechecking this path: migration no longer risks deleting an aliased shared hook file, unknown and user-defined JSON fields are preserved, and uninstall can find removable shared hooks when the linked checkout lacks its local project layer.

Current `main` is merged, and the seven-hook set from #1958 remains intact, including `SubagentStart` and `SubagentStop`. The merged head passes `mise run check`, including race-enabled unit and integration tests, all 56 Vogon canaries, and all 4 Roger-Roger canaries.

cmd/entire/cli/agent/codex/hook_root.go
```diff
@@ -0,0 +1,255 @@
...
+	if commonDir != expectedCommonDir {
+		return HookLocation{}, fmt.Errorf("linked worktree common directory %q does not match Git directory %q", commonDir, expectedCommonDir)
+	}
+	if !looksLikeGitDir(commonDir) {
+		return HookLocation{}, fmt.Errorf("linked worktree common directory %q is not a Git directory", commonDir)
+	}
+	location.LockPath = filepath.Join(commonDir, "entire-codex-hooks.lock")
+	location.LegacyHooksPath = filepath.Join(worktreeRoot, ".codex", HooksFileName)
+
+	authoritativeRoot, err := canonicalPath(filepath.Dir(commonDir))
```
peyton-alt: This doesn't check that the parent of the common dir is actually a checkout, and `isInsideGitMetadata` below won't catch it (a bare repo named `repo.git` matches neither `.git`/`.bare`, and the walk starts above it).

```sh
git clone --bare <src> ~/repo.git
git -C ~/repo.git worktree add ~/work/wt
# from ~/work/wt: commondir=../.. -> commonDir=~/repo.git -> Dir() -> $HOME
# HooksPath = ~/.codex/hooks.json, nil error
```

That's `CODEX_HOME`, so `agent add codex` merges one repo's hooks into every Codex session machine-wide, and `agent remove codex` can delete the file. Same cause puts hooks in `~/src/.codex` for a bare repo in `~/src`, or beside the git dir with `--separate-git-dir`.

Could we refuse when the derived root is `CODEX_HOME`/`$HOME`, with a sentinel like the submodule one? `git worktree list --porcelain` also marks the bare main entry if you want a positive signal. (A "must contain `.git`" check won't work — the `.bare` layout root has none.)

peyton-alt: This doesn't check that the parent of the common dir is actually a checkout, and `isInsideGitMetadata` below won't catch it (a bare repo named `repo.git` matches neither `.git`/`.bare`, and the walk starts above it).

```sh
git clone --bare <src> ~/repo.git
git -C ~/repo.git worktree add ~/work/wt
# from ~/work/wt: commondir=../.. -> commonDir=~/repo.git -> Dir() -> $HOME
# HooksPath = ~/.codex/hooks.json, nil error
```

That's `CODEX_HOME`, so `agent add codex` merges one repo's hooks into every Codex session machine-wide, and `agent remove codex` can delete the file. Same cause puts hooks in `~/src/.codex` for a bare repo in `~/src`, or beside the git dir with `--separate-git-dir`.

Could we refuse when the derived root is `CODEX_HOME`/`$HOME`, with a sentinel like the submodule one? `git worktree list --porcelain` also marks the bare main entry if you want a positive signal. (A "must contain `.git`" check won't work — the `.bare` layout root has none.)

cmd/entire/cli/agent/codex/hook_root.go
```diff
@@ -0,0 +1,255 @@
...
+	projectDir := filepath.Dir(l.HooksPath)
+	if l.LegacyHooksPath != "" {
+		projectDir = filepath.Dir(l.LegacyHooksPath)
+	}
+	info, err := os.Stat(projectDir)
+	return err == nil && info.IsDir()
+}
+
+// ResolveHookLocation resolves Codex's repository-authoritative hooks file.
+func ResolveHookLocation(ctx context.Context) (HookLocation, error) {
```
peyton-alt: Worth documenting that on `ErrLinkedSubmoduleHooksUnsupported` this returns a *partially populated* `HookLocation` — `UninstallHooks` depends on `LockPath` surviving there, while every other error path returns the zero value. Someone "normalizing" that return to `HookLocation{}` compiles, passes most tests, and turns the submodule uninstall into `flock.New("")`. A data-carrying sentinel (the shape `V1DivergedError` uses in `strategy/`) would make it compiler-visible instead.

cmd/entire/cli/agent/codex/hooks.go
```diff
@@ -47,216 +52,454 @@ var managedHooks = []managedHook{
...
+	{event: "SubagentStart", label: "subagent_start", verb: HookNameSubagentStart, timeout: defaultHookTimeoutSec, wrap: agent.WrapProductionSilentHookCommandForOS},
+	{event: "SubagentStop", label: "subagent_stop", verb: HookNameSubagentStop, timeout: defaultHookTimeoutSec, wrap: agent.WrapProductionSilentHookCommandForOS},
 }
 
-// InstallHooks installs Codex hooks in .codex/hooks.json.
+// InstallHooks installs Entire hooks in Codex's repository-authoritative
+// .codex/hooks.json.
 func (c *CodexAgent) InstallHooks(ctx context.Context, force bool) (int, error) {
-	repoRoot, err := paths.WorktreeRoot(ctx)
+	location, err := ResolveHookLocation(ctx)
```
peyton-alt: This propagates the submodule sentinel through `applyAgentChanges` into `errors.Join`, so `entire enable` exits non-zero from a linked submodule where it succeeded on main — other agents install fine but the command still fails. Since it's a permanent limitation rather than a transient error (and `UninstallHooks`/`CheckHookConfig` both special-case it with `errors.Is`), can we warn and skip Codex here instead?

cmd/entire/cli/agent/codex/hooks.go
```diff
@@ -47,216 +52,454 @@ var managedHooks = []managedHook{
...
-		var groups []MatcherGroup
-		if err := parseHookType(rawHooks, h.event, &groups); err != nil {
+	if count > 0 {
+		if err := writeHooksDocument(location.HooksPath, destination); err != nil {
 			return 0, err
 		}
-		if force {
-			groups = removeEntireHooks(groups)
+	}
+	if legacy != nil && legacy.exists {
```
peyton-alt: The second-worktree leg of the migration looks untested: A migrates, B still has a legacy file, so install from B writes nothing to the authority (count 0) but must still clean B's legacy file here, `CheckHookConfig` must return Outdated on current-authority-plus-legacy, and doctor must print LEGACY COPY FOUND. Every existing migration test does the authority write and the legacy clean in one call, and LEGACY COPY FOUND is the only doctor state with no coverage. One test — install in primary, seed a legacy Entire file in the linked worktree, install from the linked worktree — pins all three.

cmd/entire/cli/agent/codex/hooks.go
```diff
@@ -47,216 +52,454 @@ var managedHooks = []managedHook{
...
+	location, err := ResolveHookLocation(ctx)
+	if err != nil {
+		if errors.Is(err, ErrLinkedSubmoduleHooksUnsupported) && HasWorktreeLocalEntireHooks(ctx) {
+			return agent.HooksOutdated
+		}
+		return agent.HooksAbsent
+	}
+	inspection := inspectHookConfigAt(ctx, location.HooksPath)
+	switch inspection.State {
+	case HookFileInvalid:
```
peyton-alt: Invalid → `HooksOutdated`, combined with `OutdatedHookAgents` now iterating `agent.List()`, flags Codex in repos that never enabled it: a hand-written or newer-schema hooks.json we can't parse puts codex in `hooks_outdated`, so `checkHookDrift` prints "Run `entire enable --force`" while `checkCodexHookTrust` prints "Fix or restore the file before running `entire enable`". Contradictory, and the first can't work — `installManagedHooks` hits the same parse error. Could we leave invalid files to `checkCodexHookTrust`, or gate this on whether Entire wrote the file?

cmd/entire/cli/agent/codex/hooks.go
```diff
@@ -47,216 +52,454 @@ var managedHooks = []managedHook{
...
-func (c *CodexAgent) AreHooksInstalled(ctx context.Context) bool {
-	repoRoot, err := paths.WorktreeRoot(ctx)
-	if err != nil {
-		repoRoot = "."
-	}
-
-	hooksPath := filepath.Join(repoRoot, ".codex", HooksFileName)
-	data, err := os.ReadFile(hooksPath) //nolint:gosec // path constructed from repo root
-	if err != nil {
+func fileExists(path string) bool {
```
peyton-alt: This collapses every stat error to false, including EACCES, and `uninstallHooksFiles` uses it as a pre-gate. With an unreadable hooks.json: `CheckHookConfig` → Invalid → Outdated → `hooksPresentOrOutdated` true → `runRemoveAgent` proceeds → `!hasFile` → return nil → we print "Removed Codex hooks." Nothing was removed and the hooks keep firing. Without the pre-gate the loop's `readHooksDocument` would have failed loudly with the path. Suggest dropping it (the loop already handles absence via `document.exists`), or treating only `os.ErrNotExist` as false.

cmd/entire/cli/agent/codex/trust.go
```diff
@@ -1,15 +1,17 @@
...
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 )
 
-// HookTrustGaps returns the snake_case event labels declared in
-// <repoRoot>/.codex/hooks.json that don't have a matching
+// HookTrustGaps returns the snake_case event labels declared in Codex's
+// authoritative hooks.json that don't have a matching
```
peyton-alt: `:0:0` below is no longer how matching works: `codexHasTrustedEvent` parses the full `<path>:<event>:<group>:<handler>` key, accepts any indices, and compares both sides after `canonicalPath`. The inline "match any handler index" note that used to qualify this went away with the symlink fix, so this doc is now the only description of matching. The "Returns nil when" list is also missing two new cases — resolver failure, and a missing `.codex` project layer (`InspectHookTrust` short-circuits on both).

cmd/entire/cli/agent/codex/trust.go
```diff
@@ -96,40 +118,50 @@ func declaredCodexEvents(hooksJSONPath string) ([]string, bool) {
...
-// Returns nil when hooks.json is missing or unreadable — those cases
-// are "Codex isn't enabled here", which is a different problem.
-func MissingEntireHooks(repoRoot string) []string {
-	hooksJSONPath := filepath.Join(repoRoot, ".codex", "hooks.json")
-	data, err := os.ReadFile(hooksJSONPath) //nolint:gosec // path constructed from caller-controlled repo root
-	if err != nil {
-		return nil
+// Returns nil when the authoritative file is absent, user-only, or invalid.
+// Callers that need to distinguish those states should use InspectHookConfig.
+func MissingEntireHooks(ctx context.Context) []string {
```
peyton-alt: No production callers left — doctor reads `InspectHookConfig(...).Missing` directly now, and `missingEntireHooks` below is only reachable from tests (and hardcodes `context.Background()`, so it bypasses the real `UseWindowsProductionHooks(ctx)`). Worth deleting both and having the tests call `inspectHookConfigAt`, unless you're keeping this as public API.

cmd/entire/cli/agent/codex/hooks_lock_test.go
```diff
@@ -0,0 +1,40 @@
+package codex
```
peyton-alt: This covers the helper, and `TestInstallHooks_RepositoryLockDoesNotPolluteWorktree` proves the lock lands in the common dir — but nothing pins that `InstallHooks` holds the lock *across* its read-modify-write. Moving `readHooksDocument` above `acquireHooksLock` would pass the whole suite while reintroducing the lost update this PR exists to prevent (A's install erasing what B just wrote). A test racing a few goroutines through install/uninstall on one authoritative file, asserting valid JSON and a surviving user hook, would pin it — flock conflicts across FDs in-process, so it works as a unit test.

cmd/entire/cli/config.go
```diff
@@ -160,15 +160,14 @@ func InstalledAgentDisplayNames(ctx context.Context) []string {
...
-// opted in on this machine.
+// Every freshness implementation is asked directly so it can report a stale
+// artifact that no longer qualifies as an active installation. For generated-
+// file agents (Pi, OpenCode), the committed file *is* the installation, so a
+// repo that ships one gets drift warnings even where nobody ran
+// `entire agent add`.
 func OutdatedHookAgents(ctx context.Context) []types.AgentName {
 	var outdated []types.AgentName
-	for _, name := range GetAgentsWithHooksInstalled(ctx) {
+	for _, name := range agent.List() {
```
peyton-alt, line 169: The doc above still opens with "returns installed agents", which this line is precisely no longer true of — and `TestCheckCodexHookTrust_LinkedWorktreeReportsLegacyFile` pins the opposite (codex absent from `GetAgentsWithHooksInstalled`, present here). Worth dropping "installed" from that first sentence. The widening itself is safe today: I checked claudecode/opencode/pi and all three return `HooksAbsent` rather than `HooksOutdated` when nothing is installed, so this only moves Codex.

cmd/entire/cli/doctor.go
```diff
@@ -622,37 +619,52 @@ func checkHookDrift(cmd *cobra.Command) {
...
-// resolve the worktree root. Warn-only.
+// checkCodexHookTrust reports whether Codex can discover the authoritative
+// hooks file, whether its Entire-managed event set is current, and whether the
+// local Codex config has approval records for every declared hook. All checks
+// are structural; Entire never computes or copies Codex trust hashes.
 func checkCodexHookTrust(cmd *cobra.Command) {
-	repoRoot, err := paths.WorktreeRoot(cmd.Context())
+	location, err := codex.ResolveHookLocation(cmd.Context())
 	if err != nil {
+		if errors.Is(err, codex.ErrLinkedSubmoduleHooksUnsupported) && codex.HasWorktreeLocalEntireHooks(cmd.Context()) {
```
peyton-alt: Only the sentinel gets a diagnostic — every other resolver failure (`.git` with no `gitdir:`, unreadable/empty `commondir`, the commondir-mismatch and `looksLikeGitDir` rejections, EACCES in `canonicalPath`) hits the bare `return` below with no output. Meanwhile `AreHooksInstalled` → false and `CheckHookConfig` → `HooksAbsent`, so Codex vanishes from `entire status`, the review/investigate pickers and doctor, and `agent remove codex` refuses with "not installed" — while Codex may still be firing hooks, since it resolves its own root. Worth an `UNRESOLVED — <err>` branch when there's evidence Codex is in play (a `.codex` dir at the worktree root, or `HasWorktreeLocalEntireHooks`).

cmd/entire/cli/setup.go
```diff
@@ -1680,6 +1685,36 @@ func setupAgentHooks(ctx context.Context, ag agent.Agent, forceHooks bool) (int,
...
+
+func sharedHooksRemovalSuffix(ctx context.Context, ag agent.Agent) string {
+	shared, ok := ag.(agent.RepositorySharedHooks)
+	if !ok || !shared.HooksSharedAcrossWorktrees(ctx) {
+		return ""
+	}
+	return " from the repository-wide configuration; this affects all linked worktrees"
+}
+
+func writeSharedHooksNote(ctx context.Context, w io.Writer, ag agent.Agent) {
```
peyton-alt: `.codex/hooks.json` is tracked and not gitignored in this repo, so installing from a linked worktree leaves an unexplained `modified:` in the primary checkout — a file the user never touched, in a worktree they're not looking at (and removal can leave it deleted there). This note is the only thing that could connect that diff to Entire, and it doesn't say where the file is. Can we print the resolved `HooksPath` here and in `sharedHooksRemovalSuffix`?

cmd/entire/cli/agent/testutil/git.go
```diff
@@ -0,0 +1,36 @@
+package testutil
```
peyton-alt: All five of these delegate verbatim to `cmd/entire/cli/testutil`, and there's no import cycle forcing the indirection — `go list -deps ./cmd/entire/cli/testutil` doesn't reach `agent/codex`. Can the codex tests import `cli/testutil` directly and drop this package?

peyton-alt:
@MuskanPaliwal  thanks for the edits! I left a batch of inline comments. Only the hook_root.go (the derived hook root can land on CODEX_HOME or outside the repo entirely)one blocks, the rest are small.

MuskanPaliwal:
> @MuskanPaliwal thanks for the edits! I left a batch of inline comments. Only the hook_root.go (the derived hook root can land on CODEX_HOME or outside the repo entirely)one blocks, the rest are small.

on it.

cmd/entire/cli/agent/codex/hook_root.go
MuskanPaliwal: Fixed. We now derive the shared Codex hooks path only when Git’s common directory is named `.git` or `.bare`, since its parent is the project root in those layouts. That assumption is false for standard bare repositories and separate Git directories, so Entire skips Codex hook installation there instead of risking a write outside the repository. We also reject paths that would collide with `$HOME` or `CODEX_HOME`, with regression tests for each case.

cmd/entire/cli/agent/codex/hook_root.go
MuskanPaliwal: Changed this to an `UnsupportedHookLocationError`. `ResolveHookLocation` now returns a zero `HookLocation` on failure, while the typed error carries the safe location information needed to clean up a legacy file.

cmd/entire/cli/agent/codex/hooks.go
MuskanPaliwal: Setup now treats this as a skipped Codex installation instead of failing `entire enable`. It prints the reason, continues configuring the other agents, and does not claim that Codex was added. I added a linked-submodule setup test for this behavior.

cmd/entire/cli/agent/codex/hooks.go
MuskanPaliwal: Added a test that installs the shared hooks from the primary checkout, creates a legacy copy in a linked worktree, and checks both status and doctor output. Running installation from the linked worktree removes the legacy copy without duplicating the shared hooks.

cmd/entire/cli/agent/codex/hooks.go
MuskanPaliwal: Fixed. Invalid hook files now report `HooksAbsent`, so they no longer appear under `hooks_outdated` or suggest running `enable --force`. Doctor still reports the file as `INVALID`.

cmd/entire/cli/agent/codex/hooks.go
MuskanPaliwal: Fixed. The uninstall check treats only `os.ErrNotExist` as an absent file. Other `stat` and read errors now include the affected path and return to the caller. I added coverage for an unreadable hooks file.

cmd/entire/cli/agent/codex/trust.go
MuskanPaliwal: Updated the documentation to cover the full `<path>:<event>:<group>:<handler>` key, any valid group and handler index, canonical path comparison, resolver failures, and missing project layers.

cmd/entire/cli/agent/codex/trust.go
MuskanPaliwal: Removed `MissingEntireHooks` and `missingEntireHooks`. The tests now call `inspectHookConfigAt` directly with an explicit context.

cmd/entire/cli/agent/codex/hooks_lock_test.go
MuskanPaliwal: Added a concurrency test that races installs and uninstalls against user-hook updates using the same lock. The test checks that the resulting file is valid JSON and that none of the user hooks are lost.

cmd/entire/cli/config.go
MuskanPaliwal: Removed “installed” from the opening sentence so the comment matches the current behavior.

cmd/entire/cli/doctor.go
MuskanPaliwal: Doctor now reports `Codex hooks: UNRESOLVED` with the resolver error when a local `.codex` project layer shows that Codex is in use. Unsupported derived roots still get their own diagnostic.

cmd/entire/cli/setup.go
MuskanPaliwal: Installation and removal messages now include the resolved shared `hooks.json` path and state that the change applies to all linked worktrees.

cmd/entire/cli/agent/testutil/git.go
MuskanPaliwal: Removed the pass-through Git helper file. Codex tests now import the canonical CLI test helpers directly, with a narrow architecture-test allowance for that package.

MuskanPaliwal:
hey @peyton-alt , I have addressed all the review comments. Thanks for the review. Kindly re-review it. Thanks!

cmd/entire/cli/agent/codex/hook_root.go
```diff
@@ -0,0 +1,313 @@
...
+			return false
+		}
+		codexHome = filepath.Join(home, ".codex")
+	}
+	canonicalCodexHome, err := canonicalPath(codexHome)
+	if err != nil {
+		return false
+	}
+	canonicalProjectDir, err := canonicalPath(filepath.Join(hookRoot, ".codex"))
+	return err == nil && canonicalProjectDir == canonicalCodexHome
```
peyton-alt, line 251: This prevents an exact `CODEX_HOME` collision, but it does not require the resolved `.codex` directory to remain inside the authoritative checkout.

I reproduced `enable` from a linked worktree writing through the primary checkout `.codex` symlink into an unrelated directory.

Could we validate destination containment while preserving the intentional same-file alias where a linked worktree `.codex` resolves to the authoritative `.codex`?

cmd/entire/cli/agent/codex/hooks.go
```diff
@@ -47,216 +52,450 @@ var managedHooks = []managedHook{
...
+	}
+	if err := os.MkdirAll(filepath.Dir(path), 0o750); err != nil {
+		return fmt.Errorf("create .codex directory: %w", err)
+	}
+	output, err := jsonutil.MarshalIndentWithNewline(document.topLevel, "", "  ")
 	if err != nil {
 		return fmt.Errorf("failed to marshal hooks.json: %w", err)
 	}
-	if err := os.WriteFile(hooksPath, output, 0o600); err != nil {
+	if err := jsonutil.WriteFileAtomic(path, output, 0o600); err != nil {
```
peyton-alt: Rename-based atomic replacement breaks a symlinked `hooks.json`.

With `.codex/hooks.json -> ../managed/hooks.json`, `entire enable --force` replaces the symlink with a regular file and detaches the managed target.

Could we preserve atomicity and locking while following only repository-contained, validated symlink targets? The global-writer stack `WriteFileAtomicFollowingSymlinks` helper may help after containment validation is added.

cmd/entire/cli/setup.go
```diff
@@ -605,23 +605,10 @@ func applyAgentChanges(ctx context.Context, w io.Writer, selectedAgentNames []st
...
-			successfullyReinstalledAgents = append(successfullyReinstalledAgents, ag)
-		}
-	}
+	successfullyAddedAgents, setupErrs := setupAgentHookSet(ctx, w, addedAgents, opts.ForceHooks)
+	errs = append(errs, setupErrs...)
+	successfullyReinstalledAgents, setupErrs := setupAgentHookSet(ctx, w, reinstalledAgents, opts.ForceHooks)
+	errs = append(errs, setupErrs...)
 
 	var uninstalledAgents []agent.Agent
 	for _, ag := range removedAgents {
```
peyton-alt, line 617: This removal loop can only see agents in `installedNames`, which `runManageAgents` builds with `GetAgentsWithHooksInstalled`.

Codex returns false for valid partial/outdated state and when the current linked checkout lacks its local project layer, even though shared hooks may still be active in sibling worktrees. Codex then disappears from the picker and cannot be deselected here.

Could we include valid removable/outdated Codex state in preselection and removal while continuing to exclude invalid arbitrary hook files?

cmd/entire/cli/setup.go
```diff
@@ -1658,26 +1650,110 @@ func uninstallDeselectedAgentHooks(ctx context.Context, w io.Writer, selectedAge
...
 	hookAgent, ok := agent.AsHookSupport(ag)
 	if !ok {
-		return 0, fmt.Errorf("agent %s does not support hooks", ag.Name())
+		return hookSetupResult{}, fmt.Errorf("agent %s does not support hooks", ag.Name())
 	}
 
 	count, err := hookAgent.InstallHooks(ctx, forceHooks)
 	if err != nil {
-		return 0, fmt.Errorf("failed to install %s hooks: %w", ag.Name(), err)
+		var skipped agent.HookInstallationSkipError
```
peyton-alt: Skipping is correct when another selected agent remains functional, but this also succeeds when Codex is the only selected agent.

I reproduced `enable --agent codex` printing `Ready`, exiting 0, and leaving `status --json` with `enabled: true` and `agents: []`.

Could we fail when skipping Codex leaves zero hook coverage? This preserves the earlier multi-agent skip behavior.

cmd/entire/cli/agent/codex/hook_root.go
peyton-alt: The basename guard closes the `repo.git` case, but a separate Git directory can legitimately be named `.git`:

```text
git init --separate-git-dir /tmp/storage/.git /tmp/primary
git -C /tmp/primary worktree add /tmp/linked
```

Enabling from `/tmp/linked` writes `/tmp/storage/.codex/hooks.json`, outside both checkouts. Could we positively validate the worktree registration and `gitdir` backlink instead of treating the basename as proof of ownership?

cmd/entire/cli/agent/codex/hooks.go
peyton-alt: The valid second-worktree migration is covered now, but malformed ignored legacy state still blocks the authoritative installation.

An empty `linked/.codex/hooks.json` makes `enable` exit before creating the primary hooks file. Could the authoritative destination be installed first, with any legacy cleanup failure reported separately?

cmd/entire/cli/setup.go
peyton-alt: The helper output is fixed, but the primary interactive enable path does not call it, and deselection/disable reports the shared scope only after removal.

Could we show the resolved path and cross-worktree effect before confirmation or mutation?

peyton-alt:
@MuskanPaliwal Thanks again! I re-reviewed this and left a few more inline comments.

cmd/entire/cli/agent/codex/hook_root.go
MuskanPaliwal: The ownership problem is fixed. Your exact layout still resolves to `<storage>`, because Codex treats it as the trusted project root and reads hooks there. I chose to match that behavior and show the destination before writing. If Entire’s policy is that hooks may only be written inside a visible checkout, this layout should instead return an explicit unsupported-location error. A local fallback would not work.

cmd/entire/cli/agent/codex/hooks.go
MuskanPaliwal: Changed the installation order. Entire now writes the authoritative hooks file before cleaning the legacy copy.

If the legacy file is malformed, the working authoritative installation remains in place. Entire leaves the malformed file untouched and reports that installation succeeded but cleanup failed. The test uses this exact case.

cmd/entire/cli/setup.go
MuskanPaliwal: The interactive picker now shows the resolved shared `hooks.json` path. Enable, deselection, direct removal, and disable also print the path and cross-worktree effect before changing anything.

The tests check that the notice appears before the installation or removal result.

cmd/entire/cli/agent/codex/hook_root.go
MuskanPaliwal: Codex hook paths now go through one destination resolver. It resolves the checkout, `.codex` directory, and final `hooks.json` target, then rejects any target outside the checkout.

The intentional same-directory alias still works. When the linked and authoritative `.codex` paths resolve to the same directory, legacy cleanup is skipped.

cmd/entire/cli/agent/codex/hooks.go
MuskanPaliwal: Contained `hooks.json` symlinks are now preserved. After containment validation, atomic writes target the resolved file instead of replacing the symlink entry.

Uninstall follows the same rule. If removing Entire’s hooks leaves no other properties, it writes `{}` to the target rather than deleting the symlink. Outside-checkout targets remain rejected.

cmd/entire/cli/setup.go
MuskanPaliwal: Agent management now includes Codex when its hooks are installed or when its valid Entire-owned configuration is outdated. That covers partial hooks and shared hooks missing the current worktree’s project layer.

Invalid JSON and user-only files still count as absent, so the picker will not offer to remove configuration Entire does not own.

cmd/entire/cli/setup.go
MuskanPaliwal: Enable now fails if every selected agent is skipped. It does so before writing `.entire` state or marking the repository enabled.

Multi-agent setup still continues when another selected agent installs working hooks. Tests cover both explicit `enable --agent codex` and interactive enable.

cmd/entire/cli/agent/codex/hooks.go
MuskanPaliwal: One correction to my earlier reply: after matching Codex 0.149.0’s resolver, linked submodules no longer need to be skipped. Their shared-root ownership check fails, so both Codex and Entire use the submodule worktree’s local `.codex/hooks.json`.

Skipping remains for unsafe final destinations such as `$HOME` and `CODEX_HOME`. If a skip leaves no selected agent with hook coverage, enable fails instead of reporting success.

MuskanPaliwal:
hey @peyton-alt, the pr is ready for review. thanks!

cmd/entire/cli/setup.go
peyton-alt: Thanks — the agent manager path is fixed. One related fresh-setup path still calls `uninstallDeselectedAgentHooks`, which seeds from `GetAgentsWithHooksInstalled`. In a new linked checkout with shared Codex hooks but no local `.codex` layer, Codex is Outdated rather than Installed, so deselecting it during interactive enable still leaves the repository-wide hooks active. Could this function use `removableAgentHookNames` too, with a regression test for the fresh setup path?

peyton-alt:
Hey! @MuskanPaliwal I reviewed again and left one more in-line comment. Everything else looks good!

cmd/entire/cli/setup.go
MuskanPaliwal: Good catch — fixed in 76f67ec5a. `uninstallDeselectedAgentHooks` now seeds from `removableAgentHookNames` (same as the agent-manager path) so Outdated-but-not-Installed repository-wide hooks are still torn down when deselected. Added `TestUninstallDeselectedAgentHooks_CodexOutdatedLinkedWorktreeWithoutProjectLayer`, which reproduces the linked-worktree-without-a-local-.codex-layer scenario and fails without the fix.

MuskanPaliwal:
Hey @peyton-alt, kindly take a look. Thanks!

MuskanPaliwal:
Hi @peyton-alt let me know if anything else is pending here :))

Soph:
Hey @MuskanPaliwal thanks for doing this. I had a look at this today, and I would like to ask if we can reduce the scope of this PR and maybe split it. What I like to keep is resolver + inspection + warnings but pull all the write parts out for now. I think the main issue here is surfacing that something is wrong but trying to fix it for the user might be a bit too much.

Like I if the user adds this to a worktree but then commits he can just bring that commit to his root worktree and things work. Writing across worktrees - with the potential of him missing the change in the root tree completely - is to much work from my point of view for something we just need to surface more properly.

On top of the split I have these findings / asks:

**1. Fourth implementation of worktree traversal.** `gitrepo.resolveDotGitPath` (`gitrepo/repository.go:155`) + `resolveCommonGitPath` (`:183`) already parse `gitdir:` and `commondir`; `paths.parseWorktreeID` already handles `.git/worktrees/`, `.bare/worktrees/` and the submodule `/modules/` carve-out; `session.GetGitCommonDir` (`session/state.go:1021`) is memoized. The problem isn't line count — `parseWorktreeID` treats `.bare/worktrees/` as a linked worktree while `ownsCommonDir` makes bare fall back to worktree-local. Two answers in one repo, and `dupl` won't flag it. Reusing `gitrepo`'s primitives also gets memoization, which matters since `ResolveHookLocation` is now on the `SessionStart` path via `HookTrustGaps`.

**2. Cross-checkout write needs a gate.** Measured: install from a linked worktree leaves `M .codex/hooks.json` in the primary checkout and resets its mode 0644 → 0600 (`WriteFileAtomic` chmods unconditionally, `hooks.go:460`; the `os.WriteFile` it replaced didn't). Worse in the removal direction — `entire clean` in a scratch worktree strips capture repo-wide, and `writeSharedHooksRemovalNote` prints *after* the confirm prompt in `runUninstall`. Suggest: scope text into the prompt, write behind a flag.

**3. Containment stops at the checkout.** `hook_path.go:67`/`:95` accept any target under `checkoutRoot`, and `.git/` qualifies. Measured with committed symlinks: `.codex -> .git` wrote `.git/hooks.json`; `.codex/hooks.json -> ../package.json` rewrote `package.json` at 0600. Both commit and clone cleanly. Still better than `main` (bare `os.WriteFile` followed symlinks anywhere) — just short. Fix is already in the PR: run the resolved destination through `isInsideGitMetadata` (`hook_root.go:386`) and require the project dir to be exactly `<checkoutRoot>/.codex`.

**4. Nothing in CI catches this regressing.** No `e2e/` or `integration_test/` file touched, and no e2e test runs in a linked worktree — every "worktree" mention in `e2e/tests/` is a prompt saying *"Do not use worktrees."* The unit tests assert the resolver against our beliefs about Codex; the only anchor to Codex's real behavior is the comment at `hook_root_test.go:115`. Can we get one integration test driving `codex app-server`'s `hooks/list` in a linked worktree?

Minor: `ErrLinkedSubmoduleHooksUnsupported` in the description doesn't exist (linked submodules fall back to worktree-local, no error), and the bare-layout description contradicts `TestResolveHookLocation_BareRepositoryWorktreeUsesWorktreeLocalHooks` — #2020 saw the opposite on 0.147-era vs 0.149 in the comment, worth pinning down. Also `hooks.go:341` / `trust.go:96` read the repo-controlled hooks file unbounded while `hook_root.go` caps git metadata at 64KiB.

MuskanPaliwal:
Hey @Soph, yes that makes sense. let me split the pr and get back to you :))

MuskanPaliwal:
Hey @Soph, thanks for the detailed review. I reduced the scope as requested and pushed the follow-up changes in `5690551ce` (on top of `8bd2757e4`).

The resolver remains read-only. Install, uninstall, and clean only modify the checkout where the command runs; when that differs from the file Codex reads, setup, removal, doctor, status, and SessionStart warn instead of writing across checkouts.

I also clarified the warnings so they name the exact files involved: the current worktree's `.codex/hooks.json` and Codex's discovered `.codex/hooks.json`. They now explicitly say that hooks in the current-worktree file will not run when the paths differ. For a normal checkout, the user can run `entire enable` from the discovered checkout. For a `.bare` layout, the discovered project root is not a worktree, so the generated `.codex/hooks.json` change must be applied or merged there instead; `entire enable` should not be run from that root.

I added a doctor regression covering that `.bare` warning, kept the bounded read-only diagnostics, and retained the pinned Codex `0.149.0` app-server integration coverage. The tests drive `hooks/list` in both a conventional linked worktree and a `.bare` worktree, asserting the actual source path and rejecting the linked-worktree marker. CI verifies the installed Codex version and requires those tests.

Validation: focused Codex/doctor tests pass, `git diff --check` passes, and `mise run lint` passes with 0 issues. The local full CLI suite is blocked by this sandbox's inability to bind httptest localhost ports, and the local app-server tests skip when the pinned Codex binary is unavailable; CI is configured to fail rather than skip in that case. I did not run paid-agent E2E tests.

Screenshot 2026-08-26 at 8 28 51 PM

MuskanPaliwal:
On top of this, I’m working on a broader consolidation of the Git traversal logic in a follow-up PR.

.github/workflows/ci.yml
```diff
@@ -31,8 +31,13 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
       - uses: jdx/mise-action@f10502fc09dadecfefb962fff68ce77213930204 # v4
+      - name: Install pinned Codex CLI
+        run: npm install --global @openai/codex@0.149.0
+      - name: Verify pinned Codex CLI
+        run: test "$(codex --version | awk '{print $NF}')" = "0.149.0"
       - name: Integration Tests
         env:
+          ENTIRE_TEST_REQUIRE_CODEX_APP_SERVER: "1"
```
Soph: This is nice. But I think it would be better to make this a dedicated new CI job just separating this.
Also just install latest so we notice when something changes here.

Soph:
Ran this against `f5b1be33` in an isolated repo with a primary checkout plus a linked worktree. The state machine and all three surfaces work; two changes.

**1. There is no healthy state for a linked worktree.**

Repro: `entire enable --agent codex` from the linked worktree, then commit `.codex/hooks.json` and merge it into the primary - the user does exactly the right thing and hooks genuinely fire:

```
Codex hooks: CURRENT-WORKTREE FILE NOT DISCOVERED
  Current-worktree hooks:  .../wt/.codex/hooks.json
  Codex-discovered hooks:  .../primary/.codex/hooks.json
  Codex is active from the discovered file; changes to this worktree's file do not affect it.
  Apply/merge the generated .codex/hooks.json change into the discovered project root.
```

`entire status`: `! Codex ignores this worktree's hooks file · run 'entire doctor'`

The working configuration and the broken one differ only in *which* warning you get, and the remedy points at a merge that already happened.

`codexHookStateWorktreePathNotDiscovered` is the correct steady state for every linked worktree, not a fault. Codex resolves hooks to the root checkout deliberately - `root_checkout_hooks_folder_for_dir` in `codex-rs/config/src/loader/mod.rs`, added by openai/codex#21969 ("Use root repo hooks in linked worktrees", 2026-05-13). That PR rejected the per-worktree alternative on review because divergent worktree hook definitions would fight over one `trusted_hash`, so it's stable and not getting reverted.

Suggest splitting the state at `doctor.go:750`: when the discovered file exists, is Entire-managed, and is current, report success -

```
✓ Codex hooks: ACTIVE (via root checkout)
    .../primary/.codex/hooks.json
```

Keep the warning for missing / not-Entire-managed; stale is already `codexHookStateOutdated`. Same split for the `status` one-liner at `codex_hook_diagnostics.go:152`.

**2. Trim the remedy to the one thing that applies.**

`.codex/hooks.json` is a tracked file, so delivery is git, not a filesystem write - the same way `.github/workflows/*.yml` only take effect once merged to the default branch.

Both remedy blocks currently print two hedged lines unconditionally:

```
If that root is a Git checkout, run `entire enable` from that checkout.
In a .bare layout, the discovered project root is not a worktree; apply the file there instead.
```

Every linked-worktree user sees the `.bare` line whether or not it applies, and the "If that root is a Git checkout" line hedges over a distinction the code already makes - `commonGitDirIsBare` (`hook_root.go:159`) and `hook_root.go:100` both detect it. Suggest dropping both lines from `writeCodexPrimaryCheckoutRemedy` (`doctor.go:850`) and from `warnCodexHooksAfterSetup` (`setup.go:1698-1703`), leaving:

```
  Codex will read the discovered file above, not the current-worktree file above.
  .codex/hooks.json is tracked - commit it and make sure the root worktree has it
  (merge to the default branch, or check that branch out there).
```

Anyone running a `.bare` layout already knows their project root isn't a checkout and will draw the right conclusion from "make sure the root worktree has it"; everyone else just loses a line about a layout they don't use. If it's ever worth saying explicitly, gate it on `commonGitDirIsBare` rather than hedging in prose.

Same trim in `warnCodexHooksAfterRemoval` (`setup.go:1706-1719`).

**Verification:** post-merge repro prints `✓ Codex hooks: ACTIVE (via root checkout)` with no `!` in `status`; `enable` and `doctor` each drop two lines.

Soph (commented):
and thanks for continuing! :)

MuskanPaliwal:
Hey @Soph , kindly re-review. Thanks :)

Soph (approved):

Soph (approved):
Thanks a lot. Looks good!

PR #2053, opened 2026-08-19
refactor(`jsonutil`): centralize streaming atomic replacement

MuskanPaliwal:
Depends on #1877. Until that PR merges, review the stacked range `3a39273b..7c62491a5`.

This change moves OpenCode’s staged transcript publication into `jsonutil.WriteFileAtomicStream`. The previous implementation kept temporary-file creation, validation, rename retries, and recovery inside OpenCode even though most of that lifecycle is general atomic-write behavior. The shared implementation keeps publication ordered as produce, sync, close, validate, chmod, rename, and best-effort directory sync.

## Core Changes

`WriteFileAtomicStream` now owns the staging file from creation through publication. Producer and validator errors are returned unchanged. Any failure before publication leaves the existing destination untouched and makes a best-effort attempt to remove the staging file. If a completed and validated file cannot be renamed into place, the helper returns a `PublishError` and transfers ownership of the retained staging path to the caller so the output can be recovered.

Windows replacement contention is handled inside `jsonutil` rather than OpenCode. Only access-denied and sharing-violation failures are retried, with five total attempts and cancellation-aware waits. The follow-up review found a cancellation window between retries: cancellation could occur after a wait, but the next rename could still replace the destination. The retry loop now checks the context before every rename attempt, retaining the validated staging file if cancellation wins.

OpenCode continues to translate publication failures into its existing actionable “export saved at …” message. Its package-specific staging, rename, and contention helpers have been removed because those responsibilities now belong to the shared atomic-write implementation.

## API Notes

This adds the public `jsonutil.WriteFileAtomicStream` helper and the public `jsonutil.PublishError` ownership contract. A caller receiving `PublishError` owns the validated file at `StagedPath` and must either recover or remove it.

The existing `jsonutil.WriteFileAtomic` function uses the same private engine but keeps its previous behavior. Rename failures do not expose `PublishError`, and its staging files are still cleaned up instead of being retained.

## Compatibility / Release Notes

OpenCode command invocation, timeout and error classification, mock export mode, transcript validation, attach behavior, transcript protocol, and declared capabilities are unchanged. Failed, partial, empty, invalid, and cancelled exports cannot replace an existing valid transcript.

No unrelated temporary-file or rename-based writers are migrated in this change because they may have different locking, permission, confinement, executable, or cross-filesystem requirements.

This branch is stacked on #1877 and should be rebased onto `main` after that PR merges.

## Testing

- [x] `mise run check`
- [x] Formatting and lint passed with zero issues
- [x] Race-enabled unit and integration suites passed
- [x] Vogon deterministic canary passed: 56/56
- [x] Roger-Roger deterministic canary passed: 4/4
- [x] Focused race tests passed for `jsonutil` and OpenCode
- [x] Windows cross-compilation and vet passed for the affected packages
- [ ] Real-agent E2E tests were not run because they make paid external calls and were not requested


MuskanPaliwal:
Hi @Soph , now that PR number #1877 is merged, I would highly appreciate your review on this PR. This was stacked on #1877 .

MuskanPaliwal:
Merged `main` and resolved the conflicts in cda21142d.

The overlap was in OpenCode's export code and tests, and the shared atomic writer. `main` had added filesystem confinement to the OpenCode staging code that this PR moves into `jsonutil`. It had also removed fsync from ordinary atomic writes.

I kept the confinement by adding `WriteFileAtomicStreamIn`. It works through the supplied `os.Root` and holds the parent directory open throughout publication. OpenCode still supplies the export and JSON validation callbacks; `jsonutil` owns staging, syncing, replacement retries, and cleanup. The duplicate OpenCode helpers stay removed.

The failure behavior stays intact too. Failed or invalid exports leave the existing transcript untouched. If replacement fails after validation, `PublishError` retains the completed export so it can be recovered.

Ordinary atomic writes keep `main`'s no-fsync behavior. Streaming transcript exports retain syncing because the cached export may be the session's only local copy.

Added tests for directory replacement during export, unsafe paths, cleanup, and recovery. Lint, race-enabled unit/integration tests, both deterministic canaries, and Windows cross-compilation passed. One plugin-startup timeout passed three isolated runs and the full-suite rerun.

@Soph , does this fit the intended direction of the filesystem changes on `main`? I'd appreciate a review of the confined streaming API and the decision to retain fsync for transcript exports.

Issue #2058, opened 2026-08-19
Subagent data does not survive condensation — durable storage for subagent transcripts


Soph:
Tracking issue for the one open piece of subagent tracking. Plan doc lives at `subagent-durability-plan.md` (untracked).

> **Update 2026-08-20 — the design below is implemented in #2032, not here.** That branch
> (`feat/subagent-stop-capture`, trail 1076) started 2026-08-18 as background-subagent capture
> and its last ~9 commits built exactly this: `session.TaskRecords` as the mid-turn pointer
> ledger, `WriteOptions.Tasks []TaskPayload` materialized into
> `tasks/<tool-use-id>/{task.json, agent-<id>.jsonl}` inside the session checkpoint, with
> `TranscriptUnavailableReason` for the transcript-gone case. It cites this issue by number and
> carries its own plan at `docs/superpowers/plans/2026-08-19-subagent-durable-records.md`.
> I filed this issue without spotting that branch — my mistake.
>
> **What is left after #2032 is the OPF gate and some doc debt.** Everything else below is
> either done or covered there; ticked items say which.

**Status 2026-08-19:** the per-agent ephemeral bugs are merged (7 PRs, through #1958).
Durable storage is the one open piece — this issue tracks it.

**Problem, one line:** condensation builds one `checkpoint.WriteOptions` per session with no
`IsTask`/`ToolUseID` (`manual_commit_condensation.go:398`), so `persistent.go`'s whole task-write
path is unreachable — no producer anywhere sets `IsTask: true`. Subagent detail dies at commit,
for every agent. Measured: 0 of the 150 newest `entire/checkpoints/v1` commits contain `tasks/`;
80-checkpoint sample (CC 58, Codex 10, OpenCode 2, Pi 1) had `subagent_tokens` non-null in 0.

## Decision: capture the pointer mid-turn, defer the payload to condensation

- `SubagentEnd` → small session-state entry only: correlation ID, agent type, description,
  **declared transcript path**, files touched, tokens. No shadow-tree write.
- Condensation → read the declared paths, sanitize → externalize images → redact, store.
- Path gone by then → keep the pointer record, metadata says transcript unavailable.

Why: takes per-subagent tree surgery + CAS off the mid-turn hook path; one producer instead of
two mirrored ones (that duplication is what made unreachable code look durable); no misleading
intermediate state. Works on both backends, so refs-only stops being required — still preferred
for lazy fetch of large transcripts.

[...]

PR #2059, opened 2026-08-19
remove the unreachable persistent task-checkpoint writer

Soph:
[https://entire.io/gh/entireio/cli/trails/1096](<https://entire.io/gh/entireio/cli/trails/1096>)

`persistent.go` carried a full task-checkpoint writer — `tasks/<tool-use-id>/` with per-task `checkpoint.json`, incremental `NNN-<id>.json` files, and a redacted subagent transcript — behind `if opts.IsTask && opts.ToolUseID != ""`.

**Nothing ever reached it.** No producer in the tree sets `IsTask`: condensation builds one `WriteOptions` per session, and neither `attach` nor `agentimport` sets it. The `api/checkpoint` write union has no task variant at all, so a task write could not even be expressed. Measured on this repo's `entire/checkpoints/v1`: 0 of the 150 newest commits contain any `tasks/` path.

Deleting the three functions alone would have left the scaffolding that made this look like working storage, so the always-empty parts go with them:

* the `IsTask` guard in `applySessionWrite`, and its `taskMetadataPath` return value
* `buildCommitMessage`'s `taskMetadataPath` parameter — so a committed checkpoint no longer advertises an `Entire-Metadata-Task` trailer it never populated
* `incrementalCheckpointData` / `taskCheckpointData`
* the `WriteOptions` fields only that writer read: `AgentID`, `CheckpointUUID`, `SubagentTranscriptPath`, and the four `Incremental*` fields

`IsTask`/`ToolUseID` **stay**: they feed stored metadata that live readers consume (`explain`, `checkpoint list`), and the ephemeral store genuinely sets `IsTaskCheckpoint` on the pending path. They now carry a comment saying no producer sets them and pointing at entireio/cli#2058, so the seam isn't mistaken for storage again.

Untouched: the shadow branch's live `tasks/<tool-use-id>/` writer in `ephemeral.go`, and the trailer formatter it uses.

## Coverage

`TestWriteCommitted_SubagentTranscript_JSONLFallback` exercised only the dead path and goes with it. Its live twin `TestWriteTemporaryTask_SubagentTranscript_RedactsSecrets` covers the same invalid-JSONL redaction fallback on the ephemeral store, so the behaviour stays covered.

## Doc and comment fixes

Things this made visible:

* `manual_commit_condensation.go` claimed re-reading subagent transcripts "finds nothing on this path once the agent has cleaned the transcripts up." Agents do not clean them up — Claude Code keeps 180 `agent-*.jsonl` (121 older than two days), Codex 164 rollouts (all older). Cost is the real and sufficient reason for `subagentsDir=""`, so the false sentence is replaced by the measured fact.
* `ref-checkpoint-backend.md` listed pre-push OPF among the lifecycle paths that follow the primary, contradicting its own limitations section 200 lines later; and described a checkpoint ref's tree as containing `tasks/…`.
* `sessions-and-checkpoints.md` now states outright that a committed checkpoint holds one record per session and none per subagent.

## Verification

`mise run fmt && mise run lint && mise run test:ci` — 0 lint issues, no failures, 56 vogon + 4 roger-roger canary tests pass.

Refs entireio/cli#2058. This is the cleanup half of that issue; the storage design (payload location, OPF gate) is still open there.

---

> \[!NOTE\] [Cursor Bugbot](<https://cursor.com/bugbot>) is generating a summary for commit 8dae672bc2b428bd70df99844a1cb4d65be4691a. Configure [here](<https://www.cursor.com/dashboard/bugbot>).


Soph:
Closing — superseded by #2032, which I didn't know about when I opened this.

#2032's last ~9 commits (`5cdde2e3a` onward) build the durable subagent record: `session.TaskRecords` as the mid-turn pointer ledger, `WriteOptions.Tasks []TaskPayload` materialized into `tasks/<tool-use-id>/{task.json, agent-<id>.jsonl}` inside the session checkpoint. Along the way it deletes the same unreachable writer, the same `WriteOptions` fields, and the same `TestWriteCommitted_SubagentTranscript_JSONLFallback` test this PR did — and goes further, also dropping `IsTask`/`ToolUseID`, which I kept here as a seam.

Two of this PR's doc edits would also become wrong once #2032 lands: it states in `sessions-and-checkpoints.md` and `ref-checkpoint-backend.md` that a committed checkpoint has no `tasks/` subtree, which is exactly what #2032 changes.

What survives, split out:

- #2083 — the one piece #2032 doesn't touch: `ref-checkpoint-backend.md` listing pre-push OPF among the paths that follow the checkpoint primary, contradicting its own limitations section.
- The false comment in `manual_commit_condensation.go` ("finds nothing on this path once the agent has cleaned the transcripts up" — agents don't clean them up) belongs in #2032, which already edits that file. Raising it there rather than conflicting over it.

PR #2075, opened 2026-08-19
feat(agent): add Goose coding agent integration

Ar9av:
Adds Goose as a built-in agent. Goose is the Agentic AI Foundation's Apache-2.0 coding agent ([goose-docs.ai](https://goose-docs.ai/)).

```
entire enable --agent goose
```

enable

## Why this agent

I scored the Agentic Tools Almanac's catalogue against the four things Entire needs from an agent: a turn-start hook, a turn-end hook, a native transcript on disk, and resume by session id. No tool in that catalogue documents all four.

Goose came closest. The Almanac's own `_shared/source-audit.md` rates it `official-only` and records a live-verified hook run on v1.44.0. It fires all four lifecycle events, puts `session_id` on hook stdin, and ships export, import and resume as first-class verbs. The Linux Foundation governs it, so the rug-pull risk is lower than the proprietary options.

Two runners-up and why they lost:

- **Qwen Code** hands hooks both `session_id` and `transcript_path`, the best payload in the survey, but the audit rates it `mixed` and no resume-by-id invocation is documented.
- **OpenHands** only supports hooks at project scope (`.openhands/hooks.json`), which breaks enable-once-per-machine.

I ruled out Amazon Q (AWS marks it unmaintained), Continue CLI (the Almanac live-tested it and hooks never fire in headless mode; the repo is read-only after the Cursor acquisition), Amp (no subprocess hooks at all), Cline (turn-end is documented as "coming soon"), Kiro (no session-end event), and Auggie (no turn-start event).

## Existing workflows stay untouched

Every Go change in this PR adds lines. `git show HEAD -- '*.go' | grep '^-'` returns nothing, so no existing branch, switch case, or function body changed behaviour. Outside the new package the diff is 40 added lines across 7 files:

| File | Change |
| --- | --- |
| `agent/registry.go` | 2 name/type constants |
| `hooks_cmd.go` | 1 blank import |
| `agent/resume_command.go` | 1 new `case` |
| `summarize/summarize.go` | 1 new `case` plus a builder function |
| `explain.go` | 2 new `case` arms |
| `strategy/manual_commit_condensation.go` | 1 new `case` |
| `activity_cmd.go` | 1 map entry |

Nothing routes through Goose unless a user runs `entire enable --agent goose`. The four `AgentType` switches I touched are the ones with no compile-time exhaustiveness check, where a missing case degrades to `Unknown` in silence, so I wired all four rather than leaving a latent gap.

## Design: Goose keeps sessions in SQLite

`goose info` reports the store at `~/.local/share/goose/sessions/sessions.db`. The binary carries in-place `ALTER TABLE` migrations for at least four columns, so reading that schema would couple us to something the vendor rewrites under us.

The integration follows the OpenCode pattern instead. It reads through `goose session export --format json` and restores through `goose session import`, and never opens the database. `--format json` is not optional: export defaults to markdown, which is lossy and cannot be imported back.

hooks

## What the docs got wrong

I installed goose v1.46.0 and round-tripped a transcript through an isolated `XDG_DATA_HOME` to capture the export schema. Three findings that guessing would have missed:

1. **The message array is `conversation`, not `messages`.** OpenCode's analogous export uses `messages`. Copying that shape yields an empty transcript with no error.
2. **Cache token fields disagree between layers.** The export spells them `cache_read_input_tokens` and `cache_write_input_tokens`; the SQLite columns behind them are `cache_read_tokens` and `cache_write_tokens`.
3. **Tool names arrive in more than one spelling.** Goose namespaces tools as `<extension>__<tool>` at runtime, but the Almanac records a live-verified case where a hook received the bare name, so a matcher on the documented spelling never fires. File extraction matches on suffix and accepts the bare, namespaced, and import-inherited spellings.

`AGENT.md` in the new package records the protocol with per-claim provenance, tagging each fact `[verified]` (observed from the binary) or `[documented]` (read from docs, not executed).

Chunking and reassembly operate on a raw field map rather than the modelled struct, so top-level keys the package does not describe survive a round trip instead of vanishing on every checkpoint. A test asserts this against the real export.

## Verification

`gofmt`, `golangci-lint` 2.11.3, `go test -race ./cmd/...`, the integration suite, and the Vogon E2E canary all pass.

Beyond that I ran the agent end to end against a built binary. Firing `session-start`, `turn-start` and `turn-end` produced a checkpoint carrying the right commit trailer, the extracted file list, and a transcript that reads back out through `checkpoint explain`. Token totals match the export exactly: input 124.6k, cache read 112.8k, cache write 11.8k, output 394.

status

explain

The 40+ unit tests include a genuine v1.46.0 export committed as `testdata/real_export_v1_46_0.json`, which guards the field names above. `security_contract_test.go` pins the path-component guard on `ResolveSessionFile` and on both session-id entry points.

## Three things to push back on

1. **`CalculateTokenUsage` ignores its `fromOffset` argument.** Goose reports usage as session-level totals with no per-message block to scope by, so the method returns the session total whatever offset it gets. The method comment says so and a test pins it, but it deviates from the interface contract and a reviewer may want a different shape.
2. **The hook payload is `[documented]`, not `[verified]`.** Firing a hook needs a configured model provider (`goose doctor` reports none), so I could not execute one here. The payload shape and native tool names come from docs plus the Almanac's live test on v1.44.0. No `e2e/` harness ships for the same reason.
3. **I set `IsPreview() = true` and left `agent-help` classification alone.** CLAUDE.md calls those product judgments, so I took the safe default and am flagging both rather than deciding them.

## Screenshots branch

The images above live on the `assets-goose` branch of my fork, kept out of this diff on purpose.


Ar9av:
Closing in favour of a single PR covering four agents (Goose plus three more). The Goose work here is unchanged and will be included there.

PR #2076, opened 2026-08-19
feat(agent): add Goose, Qwen Code and OpenHands integrations

Ar9av:
Adds three agents: **Goose** (Agentic AI Foundation), **Qwen Code** (Alibaba), and **OpenHands** (All Hands AI).

```
entire enable --agent goose
entire enable --agent qwen-code
entire enable --agent openhands
```

enable

## How these three were chosen

I scored the Agentic Tools Almanac's 34 tools against the four things Entire needs from an agent: a turn-start hook, a turn-end hook, a native transcript on disk, and resume by session id. No tool in that catalogue documents all four, so I installed each finalist and probed it rather than building from docs.

Several bigger names are absent, each for a specific reason:

| Tool | Vendor | Why not |
| --- | --- | --- |
| Cline | Cline Bot | `TaskComplete` is documented as "coming soon". The turn-end hook does not exist |
| Amp | Amp (ex-Sourcegraph) | No subprocess hooks at all. In-process TypeScript plugins only, so there is nothing for a Go binary to attach to |
| Kiro | Amazon | No session-end event, no transcript, no resume |
| Amazon Q | AWS | AWS marks it unmaintained and points users at Kiro |
| Continue CLI | Continue Dev | The Almanac live-tested it: hooks never fire in headless mode. Repo went read-only after the Cursor acquisition |
| Junie | JetBrains | Hook stdin carries no session id and there is no env var, so Entire cannot key session state without racing on concurrent sessions |
| Grok Build | xAI | Ships only a `curl \| bash` installer, so I could not verify it |

The PyPI package named `devin-cli` self-describes as "Unofficial CLI for Devin AI", so it is not Cognition's and I did not build against it.

## Each agent stores its conversation differently

That difference drives most of the code:

| Agent | Storage | How Entire reads it |
| --- | --- | --- |
| Goose | SQLite (`sessions.db`) | `goose session export --format json` / `goose session import`, the OpenCode pattern |
| Qwen Code | JSONL per session | Directly. The hook hands over `transcript_path`, so no export step exists |
| OpenHands | One JSON file **per event** | Serializes the directory to JSONL and reconstructs it on write |

hooks

## The one design decision worth arguing about

OpenHands stores `event-00000-<uuid>.json`, `event-00001-<uuid>.json`, … and ships no export command. Entire's `SessionRef` is a single path, so this serializes that directory to JSONL.

`agent-integration-checklist.md` warns against inventing intermediate formats, and this is one. My argument for it: the mapping is lossless and mechanically reversible, because the filename is fully determined by `(index, event.id)` through OpenHands' own `EVENT_FILE_PATTERN`. `TestEventDirRoundTrip` pins byte-identical filenames on the way back. No field is added, renamed or dropped; only the container changes. `base_state.json` stays untouched, being run state rather than conversation.

If you would rather Entire grew a directory-shaped `SessionRef` instead, say so and I will split that out. I took the smaller change and documented the trade in `agent/openhands/AGENT.md`.

## What the documentation got wrong

Installing each agent and reading its shipped source turned up six things that contradict the docs. Each would have been a silent bug:

1. **Goose's export names the array `conversation`, not `messages`.** OpenCode's analogous export uses `messages`. Copying that yields an empty transcript with no error.
2. **Goose spells cache tokens differently in each layer:** `cache_read_input_tokens` in the export, `cache_read_tokens` in the SQLite column.
3. **A Qwen `tool_result` line carries `message.role: "user"`.** Filtering prompts by role reports every tool result as a user prompt, so `ExtractPrompts` keys on `provenance: "real_user"`.
4. **OpenHands does have a global hook scope.** `config.py` searches `~/.openhands/hooks.json` as well as the project file; the Almanac states no user scope exists.
5. **OpenHands sets `extra="forbid"` on its hook config**, so any unrecognised top-level key makes it reject the whole file. Goose's integration identifies its own file with a `_comment` marker; doing that here would break OpenHands, so ownership is detected from the hook command instead.
6. **`openhands --resume <id>` exists**, which the Almanac lists as UNKNOWN. The id has two spellings: the directory uses undashed hex32, `--resume` takes the dashed UUID.

Each `AGENT.md` labels every claim `[verified]` (observed from the binary or a real run) or `[documented]` (read, not executed).

## Verification

Both new fixtures are genuine agent output. Neither Qwen nor OpenHands will complete a turn without a paid provider, so I ran each against a local OpenAI-compatible mock and captured what it wrote to disk. Structure is real; the model's prose and its chosen arguments are not, and the AGENT.md files say so.

`gofmt`, `golangci-lint` 2.11.3, `go test -race ./cmd/...`, the integration suite, and the Vogon E2E canary all pass.

All three were also driven end to end against the built binary: enable, fire the hooks, commit, then read the checkpoint back. Goose's token totals match its export exactly (input 124.6k, cache read 112.8k, cache write 11.8k, output 394).

status

qwen

OpenHands, read back through `checkpoint explain` with its tool call intact:

openhands

explain

## Blast radius

Existing agents are untouched. Outside the three new packages the diff adds 8 lines to `registry.go`, 3 blank imports, 3 resume cases, 7 `AgentType` switch arms, and 3 entries to a display map. Nothing routes through a new agent unless someone runs `entire enable --agent <name>`.

The four `AgentType` switches have no compile-time exhaustiveness check, so a missing case degrades to `Unknown` in silence. I wired all four for each agent rather than the minimum. The resume-command registry test caught a missing Qwen case while I was writing this, which is the failure mode those tests exist for.

`dashedConversationID` in `agent/resume_command.go` duplicates `resumeID` from `agent/openhands` rather than importing it, because that package imports `agent` and the import would cycle. The registry test pins the two to agree.

## Three things to push back on

1. **Goose's `CalculateTokenUsage` ignores its `fromOffset`.** Goose reports session-level totals with no per-message usage to scope by, so it returns the total whatever offset it gets. Qwen and OpenHands do not have this problem. Documented on the method and pinned by a test, but it deviates from the interface contract.
2. **OpenHands reports no token usage at all.** Its events carry no usage block, so the agent does not implement `TokenCalculator`. Reporting nothing beats inventing a number, but it means OpenHands checkpoints show no tokens.
3. **All three are `IsPreview() = true` with no `agent-help` classification change.** CLAUDE.md calls those product judgments, so I took the safe default and am flagging rather than deciding.

## Screenshots branch

The images live on the `assets-goose` branch of my fork, kept out of this diff on purpose.


Ar9av:
Superseded by a PR that drops OpenHands. Its hooks load but never execute in headless mode, verified locally with a plain marker command, so I could not stand behind it. The replacement also corrects the Goose and Qwen tool-name lists, which were inferred from docs and were wrong.

PR #2079, opened 2026-08-20
feat(trail): write a trail body through the new trail-body endpoint with etag-based CAS

khaong:
https://entire.io/gh/entireio/cli/trails/1107

## Summary

Implements ENT-1741: `entire trail update --body` now reads the description's ETag and sends it as `If-Match`, so a body write is rejected (412) rather than silently clobbering a description that changed since it was read, instead of always sending `Overwrite: true` unconditionally.

- `TrailBodyDocument.ETag` decodes the new `bodyDocument.etag` field on a trail read.
- `sendTrailBody` dispatches between three modes: `--overwrite` (unconditional, explicit), an available etag (`If-Match`, no `Overwrite`), or neither (graceful fallback to `Overwrite: true` — covers a server predating the etag field, a trail with no body document yet, or a failed best-effort etag read).
- New `--overwrite` bool flag on `trail update`, named for what it destroys per the ticket ("`--overwrite`, not `--force`").
- 412 → "changed since read, re-run or pass --overwrite"; 409 → "not empty, pass --overwrite". Neither auto-retries with overwrite.
- The interactive form path gets its etag for free from the existing detail read; the non-interactive `--body` path does one extra best-effort GET purely for the etag (skipped entirely under `--overwrite`), and now warns on stderr if that read fails rather than silently downgrading the conflict check.

## Live-server verification

Confirmed against a real trail on `gh/entireio/cli` (not just test fixtures) that `bodyDocument.etag` is a real, populated field: trail 1076 returned `etag = W/"2026-08-18T23:29:25.468469Z"`; trail 1075 returned an empty etag alongside an empty snapshot, exercising the graceful-degradation path. The entire-api schema also lists `etag` as required on `TrailBodyDocumentJSON`.

## Known tradeoff (by design, not a bug)

The non-interactive `--body` path fetches an etag purely to attempt CAS, even though the caller supplied replacement text without having read the current document — this etag doesn't protect against a race the caller could have avoided by reading first, only against one that happens to occur in the fetch-to-write window. This is the ticket's explicit design (Linear ENT-1741), not something this PR reconsiders; flagging it here since a reviewer raised it independently.

That fetch-to-write window is also the only one closable today: `trail show --json` emits no etag field, so a read-modify-write caller (read the description, compose an edit, write back later) has no way to pass back what they read — there is no `--if-match` escape hatch. Widening CAS to cover that case (etag on `trail show --json`, an `--if-match` flag) is out of scope for ENT-1741; filed as a follow-up finding on the trail so it isn't lost.

## Out of scope (per ticket)

`--body-file`/stdin, other `trail` subcommands, presence/live-session detection.

## Review process

Ran `slop-reviewer` + 4 `/pr-review-toolkit:review-pr` dimensions (code, tests, silent-failures, comments) in parallel before opening this PR. Slop rating: **small**. Fixed all accepted findings in a second commit: a stale doc comment overclaiming write-etag-chaining, a silently-swallowed refetch failure (now warns), a guard that could mis-pair an etag with unseen content in a corner of the interactive path, plus two added tests and a switch-statement simplification.

## Test plan

- [x] `mise run fmt && mise run lint` clean
- [x] `go test -race -run Trail ./cmd/entire/cli/...` — all pass
- [x] `mise run test:ci` (unit + integration + e2e canary) — pass (one flaky failure in an unrelated pre-existing test, `TestHooksGitCmd_DiscoverExternalAgents_WhenEnabled`-class cross-test pollution in the large parallel `cmd/entire/cli` package; reproduced 3 different unrelated failures across repeated full-package runs, confirmed unrelated to this diff by 2 clean reruns)
- [x] New tests cover all 3 dispatch modes, both error mappings, the etag-fetch-only-when-needed logic, `--overwrite` skipping the fetch, and the refetch-failure warning path

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

---

> [!NOTE]
> **Medium Risk**
> Changes how trail descriptions are written (conflict detection vs silent overwrite) and adds backward-compatible fallback paths; scope is limited to the trail CLI body PUT path with broad test coverage.
> 
> **Overview**
> **Trail description writes** no longer always send `Overwrite: true`. The CLI decodes **`bodyDocument.etag`** on trail reads and threads it through **`fetchTrailDescription`** / **`resolveTrailUpdateBody`** so **`sendTrailBody`** can send **`If-Match`** when an etag is available and **`--overwrite`** is not set.
> 
> **`entire trail update`** gains **`--overwrite`** for unconditional replacement. Non-interactive **`--body`** does a best-effort detail GET for the etag (skipped under **`--overwrite`**); if that fetch fails, stderr warns and the write still proceeds via the **`Overwrite`** fallback. **412** and **409** responses get actionable hints (re-run / **`--overwrite`**). Interactive update seeds the form from the existing detail read, so it gets the etag without an extra fetch.
> 
> Tests cover dispatch modes, error mapping, etag threading, and fallback behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0dc4af34ddb446a855ba8a719867ad8960be5dd1. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/api/trail_types.go
+++ b/cmd/entire/cli/api/trail_types.go
@@ -133,8 +133,8 @@ type TrailUpdateResponse struct {
 // The route also accepts contentJson (ProseMirror JSON, written as-is) in place
 // of markdown; the CLI only ever writes Markdown, so contentJson is not
 // modeled here. The route also accepts an If-Match header for optimistic
-// concurrency, populated from a prior read's or write's TrailBodyDocument.ETag
-// — see sendTrailBody for the dispatch between If-Match and Overwrite.
+// concurrency, populated from a prior read of TrailBodyDocument.ETag — see
+// sendTrailBody for the dispatch between If-Match and Overwrite.
 type TrailBodyRequest struct {
 	Markdown  string `json:"markdown"`
 	Overwrite bool   `json:"overwrite,omitempty"`
--- a/cmd/entire/cli/api/trail_types.go
+++ b/cmd/entire/cli/api/trail_types.go
@@ -50,9 +50,12 @@ type TrailResource struct {
 // is the rendered plain text displayed by the CLI. The document is also what a
 // body write returns (see TrailBodyRequest), so both directions decode into this
 // type; the fields the CLI does not use (id, documentKey, schemaVersion,
-// contentJson, updatedAt) are simply left out of it.
+// contentJson, updatedAt) are simply left out of it. ETag is populated on a
+// read as well as on a write response, and is what makes If-Match viable on
+// the next write (see sendTrailBody).
 type TrailBodyDocument struct {
 	TextSnapshot string `json:"textSnapshot"`
+	ETag         string `json:"etag,omitempty"`
 }
 
 // ToMetadata converts a TrailResource to display metadata.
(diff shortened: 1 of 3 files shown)


khaong:
bugbot run

Soph (approved):

Issue #2063, opened 2026-08-19
Subagent transcripts skip image externalization before size cap / redaction


suhaanthayyil:
## Summary
Main session transcript path: sanitize → externalize images → redact.
Subagent path (`prepareSubagentTranscript`): sanitize only, then hard-drop if `> MaxChunkSize` (50MB), then redact.

## Impact
- Image-heavy Claude/Codex subagent transcripts can be dropped whole.
- Codex `input_image` payloads reach redaction still as base64 and can be corrupted (see companion redact issue).

## Expected
When image externalization is enabled, run the agent image codec on subagent transcripts before the size check, and store returned assets under the task checkpoint tree.

PR #2103, opened 2026-08-23
fix(strategy): externalize task images before applying transcript cap

SparshM8:
This PR fixes issue #2063 where large task transcripts containing base64 images were being capped before externalization, leading to lost transcripts even when externalization was enabled.

Changes:
- Reordered the condensation strategy to externalize images BEFORE applying the transcript cap.
- Added a targeted boundary test to verify that images up to the 50MB cap are externalized and transcripts are preserved.
- Verified that transcripts exceeding the cap after externalization are still safely handled.

Native test passed: TestCondenseSession_TaskRecordImageExternalizationBoundary

### DIFF
--- a/cmd/entire/cli/strategy/manual_commit_condensation.go
+++ b/cmd/entire/cli/strategy/manual_commit_condensation.go
@@ -240,16 +240,22 @@ func prepareTaskTranscriptForStorage(
 	raw []byte,
 ) (redacted redact.RedactedBytes, assets []cpkg.TranscriptAsset, tooLarge bool, err error) {
 	sanitized := agent.SanitizeTranscriptForStorage(ag, raw)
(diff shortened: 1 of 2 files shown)


SparshM8:
Hi @entireio/cli-maintainers,

I've raised PR #2103 to address the transcript cap issue (#2063). I've included a targeted boundary test that verifies the fix by simulating large base64 images that exceed the cap before externalization but stay well within it after.

Could you please approve the CI workflows to run? I'm happy to address any feedback or adjust the aggregate asset policy if needed.

Thanks!

SparshM8:
I've refined the documentation to accurately reflect the pipeline order: `sanitize -> externalize -> cap -> redact`. I also optimized the boundary test to avoid large memory allocations while still verifying the 50MiB cap behavior correctly.

PR #2123, opened 2026-08-25
fix(checkpoint): avoid false empty result after truncated session scan

### LINKED ISSUE #2089: checkpoint explain --session reports "No checkpoints found" when the scan limit truncates (false negative; misleads the shipped search skill)
### Summary

`entire checkpoint explain --session <id>` applies a default scan limit, but when that limit truncates the scan it reports **zero checkpoints in definitive language** rather than reporting truncation. The output simultaneously claims the history is empty and admits it may be hidden:

```
  branch       <branch>
  session      <session-id>
  checkpoints  0

No checkpoints found on this branch.
Checkpoints will appear here after you save changes during an agent session.
note: checkpoint list reached its scan limit; older checkpoints may be hidden. Run 'entire checkpoint explain --json --limit <N>' to see more.
```

The same command with an explicit limit returns the checkpoints immediately:

```
entire checkpoint explain --session <id> --json --limit 2000
→ [ { "checkpoint_id": "…", "session_id": "…", "agent": "Claude Code", … }, … ]
```

So the data is present and reachable. Only the default-limit path misreports it.

### Why this matters more than a cosmetic wording issue

The headline (`checkpoints 0` + `No checkpoints found`) and the footnote (`older checkpoints may be hidden`) contradict each other, and the wrong half is the prominent one. A human may read down to the note. **An agent generally does not.**

This bites the search skill that Entire itself installs (`--search-skill`, `ENTIRE-MANAGED SEARCH SKILL v1`). Observed live: a subagent following the shipped v1 prompt concluded

> "the accessible session transcripts in Entire don't contain the diagnostic discussion … the decision-making appears to have been done post-hoc in commit messages rather than during an interactive session transcript that Entire captured."

That was false. The session file held **306** matches for the search term, and `--limit 2000` returned the checkpoints at once. The agent reported absence of history as a finding, with no signal to the user that the lookup had been truncated.

Step 5 of the shipped v1 prompt compounds it:

> "If nothing looks right, rerun a narrower `entire search --json --compact` instead of explaining many hits or switching tools."

When the cause of "nothing looks like it exists" is a silent truncation, *narrowing* is exactly the wrong recovery — it makes the false negative more confident. The prompt has no instruction to raise the limit before concluding absence.

### Environment

* Entire CLI **0.10.2** (go1.26.6, darwin/arm64)
* Checkpoint backend: `git-refs`
* Agents: Claude Code + Codex
* ~900 checkpoints in the repo, created by the history import that `entire enable` offers (`Imported 921 turn(s) from 29 session(s)`)
* Private GitHub repo; all checkpoint refs confirmed pushed to origin

Possibly relevant: the imported checkpoints carry `"is_logs_only": true`, and some `message` fields are terminal scrollback rather than prompts. I did not test whether the same truncation reports zero on natively-captured (non-imported) checkpoints.

### Reproduction

1. Enable Entire on a repo with substantial agent history, accepting the history import at `enable` (enough to exceed the default scan limit — ~900 checkpoints here).
2. Pick a session id known to contain matching content.
3. `entire checkpoint explain --session <id>` → `checkpoints 0`, `No checkpoints found on this branch.`
4. `entire checkpoint explain --session <id> --json --limit 2000` → checkpoints returned.

### Suggested fixes

Any one of these would remove the false negative; the first two seem most valuable:

1. **Do not print `No checkpoints found` when the scan was truncated.** If the limit was hit and zero matched, that is an *inconclusive* result, not an empty one. Say so in the headline — e.g. `checkpoints 0 of ≥N scanned (scan limit reached — pass --limit to widen)`.
2. **Make the `--json` shape carry it.** A `"truncated": true` / `"scan_limit_reached": true` field would let agents detect this programmatically instead of relying on prose in a note they don't read.
3. **Auto-widen once** when a `--session`-scoped lookup returns zero and the limit was hit, since a session-scoped query has a naturally bounded result set.
4. **Update the shipped search skill** to require an explicit high-limit lookup before an agent may state that history has no match. (Patched locally, but it is an `ENTIRE-MANAGED` file, so a `--force` reinstall or a v2 skill will silently overwrite the fix.)

### Related

Adjacent but distinct from #1195 — there, `entire search` returns 0 while `checkpoint explain` sees the data. Here it is `checkpoint explain` itself under-reporting. I also saw search return only `type: "commit"` hits while matching transcripts existed, which may be the same surface as #1195.

SparshM8:
## Summary

Fixes #2089.

`checkpoint explain --session <id>` can report `checkpoints 0` and `No checkpoints found on this branch.` when the underlying checkpoint scan reaches its limit and the requested session is not present in the returned window. That output is a false negative: older checkpoints may exist but were not examined.

## Root cause

`getBranchCheckpoints` already returns a truncation signal, but the branch-view path discarded that state before applying the session filter. When filtering removed every checkpoint from the bounded result, the formatter could not distinguish “the branch has no checkpoints” from “no matching checkpoint was found in an incomplete scan.”

## Changes

* Preserve the scan-truncation state through the branch formatter.
* Report that no matching checkpoint was found **in the scanned results** when a session-filtered scan is truncated.
* Keep the existing empty-history message unchanged for genuinely non-truncated empty results.
* Add a native regression test covering the truncated, zero-match-after-filter case.

This change does not alter checkpoint collection limits, JSON output, or the behavior of an unfiltered empty branch.

## Validation

* `go test ./cmd/entire/cli -run 'TestFormatBranchCheckpoints|TestGetBranchCheckpoints_TruncationSignal|TestRunExplainBranchWithFilter'`
* `go test ./cmd/entire/cli`
* `git diff --check`

Both targeted and complete `cmd/entire/cli` package tests pass locally.

### DIFF
--- a/cmd/entire/cli/explain.go
+++ b/cmd/entire/cli/explain.go
@@ -2935,7 +2935,7 @@ func runExplainBranchWithFilter(ctx context.Context, w, errW io.Writer, noPager 
 	}
 
 	// Format output
-	output := formatBranchCheckpoints(w, branchName, points, sessionFilter)
+	output := formatBranchCheckpointsWithTruncation(w, branchName, points, sessionFilter, truncated)
 
 	outputExplainContent(w, output, noPager)
 
@@ -3117,6 +3117,12 @@ const (
 // Groups commits by checkpoint ID and shows the prompt for each checkpoint.
 // If sessionFilter is non-empty, only shows checkpoints matching that session ID (or prefix).
 func formatBranchCheckpoints(w io.Writer, branchName string, points []strategy.RewindPoint, sessionFilter string) string {
+	return formatBranchCheckpointsWithTruncation(w, branchName, points, sessionFilter, false)
+}
+
+// formatBranchCheckpointsWithTruncation formats a branch view while preserving
+// whether the underlying checkpoint scan was truncated before filtering.
+func formatBranchCheckpointsWithTruncation(w io.Writer, branchName string, points []strategy.RewindPoint, sessionFilter string, truncated bool) string {
 	var sb strings.Builder
 	styles := newStatusStyles(w)
 
@@ -3149,8 +3155,13 @@ func formatBranchCheckpoints(w io.Writer, branchName string, points []strategy.R
 	sb.WriteString("\n")
 
 	if len(groups) == 0 {
-		sb.WriteString("No checkpoints found on this branch.\n")
-		sb.WriteString("Checkpoints will appear here after you save changes during an agent session.\n")
+		if truncated && sessionFilter != "" {
+			sb.WriteString("No checkpoints matching this session were found in the scanned results.\n")
+			sb.WriteString("The checkpoint scan was truncated; older checkpoints may be hidden.\n")
+		} else {
+			sb.WriteString("No checkpoints found on this branch.\n")
+			sb.WriteString("Checkpoints will appear here after you save changes during an agent session.\n")
+		}
 		return sb.String()
 	}
 
--- a/cmd/entire/cli/explain_test.go
+++ b/cmd/entire/cli/explain_test.go
@@ -4367,6 +4367,24 @@ func TestFormatBranchCheckpoints_SessionFilter(t *testing.T) {
 		}
 	})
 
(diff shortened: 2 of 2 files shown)


SparshM8:
Closing as non-meaningful modest correctness fix. Focus shifting to high-impact engineering work.

PR #2131, opened 2026-08-26
perf(checkpoint): batch persistent session writes with CAS

MuskanPaliwal:
## Problem

`process_sessions` still persists each session separately. For every session, it reloads the checkpoint, rebuilds the session index and summary, creates a commit, and advances the checkpoint ref.

Issue #2091 measures that work at roughly 73–103 ms per session. The repeated checkpoint load and rebuild make condensation slower as more sessions share a checkpoint.

Checkpoint-level batching needs two things first:

- Ref updates must detect concurrent writes from hooks, worktrees, CLI processes, and other Git clients.
- Both persistent Git layouts must use the same batch algorithm.

## Solution

This PR adds native compare-and-swap publication and a shared `BatchSessions` write request.

It lays the storage and concurrency groundwork for #2091 item 2. It does not yet change PostCommit to group and submit sessions in batches. That work is listed below.

## Native compare-and-swap publication

`CompareAndSwapRef` publishes refs through native Git:

```text
git update-ref <ref> <new> <expected-old>
```

Missing-ref creation passes the correct all-zero object ID in both SHA-1 and SHA-256 repositories.

`RunRefTransaction` retries a logical mutation when another writer wins the race. Each attempt reads the current tip, rebuilds from that tip, and publishes with CAS. Conflicts use bounded, cancellation-aware jitter. The transaction stops after 16 attempts.

A mutation that reports `changed=false` is an idempotent no-op and does not call `git update-ref`.

The CAS conversion covers persistent v1 metadata refs and per-checkpoint refs, including:

- initial ref creation;
- v1 migration;
- attribution, summary, and transcript backfills;
- reconciliation and safe local ref advancement;
- orphan cleanup and expected-old deletion;
- shadow-branch migration;
- rewind publication;
- OPF publication.

OPF uses one CAS attempt by design. If its expected ref has moved, it fails closed and withholds the push. It does not rebuild sanitized output from newly appeared content.

Go-git object construction now takes a process-local lock keyed by `*git.Repository`. This protects go-git state shared by concurrent writers in one process. Native Git CAS handles publication conflicts between processes.

## Checkpoint-level batch writes

`BatchSessions` is now part of the sealed `WriteRequest` union. One request contains a checkpoint ID and one or more reserved session writes.

Validation happens before object construction:

- The checkpoint ID must be present and valid.
- Every session must reference that checkpoint.
- Session IDs must be valid and unique.
- Sessions are sorted by `SessionID`.
- Task tool-use IDs and agent IDs must be valid.
- Duplicate task IDs within or across incoming sessions are rejected.
- Sessions without timestamps share one UTC timestamp, truncated to seconds.
- Author fallback, commit-message inputs, and prepared tree hashes are fixed before CAS retries.

`GitStore` and `gitRefsStore` share the same preparation and checkpoint-tree builder. Their wrappers only locate or splice the checkpoint subtree and choose the ref to publish.

For each CAS attempt, the builder:

- loads the existing checkpoint once;
- reads each existing session’s metadata once;
- reuses the numeric index when updating a session;
- preserves untouched session subtree hashes;
- appends new sessions in canonical order;
- merges the task subtree once and preserves untouched tasks;
- reduces the complete final session set into the root summary;
- preserves combined attribution and sticky review, investigation, and import flags;
- aggregates counts, tokens, files, strategy, branch, and commit SHA deterministically;
- creates one checkpoint commit for that backend attempt.

Existing `Session` and `ReservedSession` requests pass through the same builder with one session. Their caller contract does not change.

Combined attribution stays in its existing backfill transaction. The batch write does not fold it into the checkpoint commit.

## Storage and compatibility

The durable checkpoint format does not change:

```text
metadata.json
0/metadata.json
0/full.jsonl
0/transcript.jsonl
1/...
tasks/<tool-use-id>/...
```

Older clients already understand checkpoints with multiple numbered session directories. A batch produces the same tree, task subtree, summary, and metadata types that repeated single-session writes produce.

Compatibility was checked in separate dimensions:

- New clients continue to read existing checkpoint data.
- Old clients can read checkpoints written by `BatchSessions`.
- Existing callers still use `Writer.Write(context.Context, WriteRequest)`.
- Single-session requests keep their existing behavior.
- Malformed or ambiguous requests now fail before publication.
- Concurrent upgraded writers protect each other with native CAS.
- If an old blind writer publishes first, the CAS writer detects the change and rebuilds.
- If an old blind writer publishes second, it can still overwrite the newer write. One upgraded participant cannot make a pre-CAS writer safe.

Rejecting invalid input and replacing last-writer-wins behavior with conflict detection are intentional safety changes, not storage incompatibilities.

## Routing and partial success

Persistent dispatch, kind routing, fanout, and the filesystem test store support `BatchSessions`.

One Git backend makes every session in the batch visible through one ref update. Fanout writes the primary first. Mirror failures remain logged and best-effort, so they do not turn a successful primary write into an error.

Routing follows the backend encoded by the checkpoint ID if configuration changes after reservation. If a readable migrated copy exists, routing updates that copy too. The read-preferred backend is probed before the first physical write.

There is no transaction across physical backends. If the second backend fails, the first may already contain the batch. The writer returns the error. Retrying the same canonical upsert repairs the lagging copy.

A future lifecycle caller must therefore treat an error as “some backend may already contain this batch,” not as proof that nothing was written.

The filesystem store replaces one JSON document atomically under a process-local mutex. It is a test backend and does not provide cross-process serialization.

## Complexity

The deterministic visitation test uses existing checkpoints with 100, 200, and 500 sessions. With no tasks, it records:

```text
root entries visited          N + 1
session entries visited       N
existing metadata reads       N
task entries visited          0
final-session contributions   N
total                         4N + 1
```

These counters measure visits to checkpoint content, not every CPU operation.

One CAS attempt against one Git backend performs one checkpoint load and build. It does not rescan the checkpoint for each incoming session. Request preparation happens once before the retry loop. A CAS conflict causes one more load and build from the winning tip.

## Validation

`mise run check` passed on the committed tree. This included:

- formatting;
- golangci-lint with zero issues;
- the aggregate race-enabled unit and integration suite;
- Vogon, 56 of 56;
- roger-roger, 4 of 4.

Focused race-enabled tests also passed for:

- batch validation, ordering, summaries, timestamps, trailers, task collisions, and both Git layouts;
- concurrent metadata-ref and per-checkpoint-ref writes;
- SHA-1 and SHA-256 CAS;
- transaction retries and cancellation;
- initialization and migration;
- dispatch, kind routing, fanout, and filesystem storage;
- reconciliation, cleanup, ref advancement, rewind, and OPF publication;
- persistent-ref writer coverage.

The advisory duplication scan found nothing in `persistent_batch.go`. Its production persistent-backend findings come from older read-wrapper patterns. The only related new finding is duplicated routing-test setup.

No paid real-agent E2E tests were run.

## Left for the next PR

PostCommit still writes sessions one at a time. Completing #2091 item 2 requires the lifecycle layer to:

1. Group materialized `process_sessions` results by checkpoint ID.
2. Build one `BatchSessions` request per checkpoint.
3. Persist enough intent to recover after a process crash.
4. Submit one batch write per checkpoint.
5. Retry safely when one physical backend may already contain the batch.
6. Keep combined attribution as a separate transaction.
7. Remove completed intent only after every required primary write succeeds.
8. Define when stale intent can be removed without losing recoverable work.
9. Test interruption before publication, partial backend success, retry, and cleanup.
10. Re-run the commit-hook performance tests and confirm that checkpoint loading and building happen once per checkpoint.

Until that work lands, single-session writes are safer and use the shared builder, but the full `process_sessions` latency reduction is not active.

Task-v2 storage, attribution folding, unrelated CLI cleanup, and the existing commit-hook performance-test work remain out of scope.

## Review status

Stages 1 and 2 were reviewed independently. No Critical, High, or Medium correctness defect was found.

The remaining hardening ideas are non-blocking:

- Check cancellation again after acquiring the repository object lock.
- Replace the lexical persistent-ref writer inventory with an AST-based check.
- Narrow native Git’s `cannot lock ref` error classification if callers need more precise operational errors.
- Revisit process-lifetime lock-map retention if this code moves into a long-running process that repeatedly opens repositories.


MuskanPaliwal:
Hey @Soph ,  kindly review the pr whenever you get time to. Thanks :))

Soph:
Sorry, I just merged #1926 which caused a lot of conflicts for you!

Soph:
Before you dig into the conflicts - how are you planning to handle the overlap with #1926? Worth agreeing the shape first so you're not spending an evening reconciling something we'd rather just delete.

MuskanPaliwal:
Hey @Soph , Thanks for flagging this. Before I resolve the overlap, this is the shape I’m planning.

#2131 already has the first two parts of the batching work:

- A shared ref transaction API that reads the current tip, rebuilds against it, and publishes with native Git CAS.
- A `BatchSessions` storage request and shared builder for both `git-branch` and `git-refs`.

The storage-level batching is implemented. It can load one checkpoint, upsert several sessions, recompute the summary once, build one commit, and publish it once. Single-session writes also use that builder.

What #2131 does not do yet is use `BatchSessions` from `PostCommit`. The `process_sessions` loop still calls `CondenseSession` once per session, and each call writes one `ReservedSession`. That lifecycle change is planned for Stage 3.

The direct overlap with #1926 is the persistent-ref transaction code. Both PRs implement native CAS, retry, and rebuild behavior around many of the same writers. #1926 also adds some behavior that #2131 currently lacks:

- A cross-process filesystem lock for each persistent ref.
- Context-aware waiting for the lock.
- Holding the lock across the full read, build, CAS, and retry cycle.
- Cleaning up a losing commit only after a confirmed CAS conflict.
- The concurrent two-Stop-hook integration regression.

My plan is to keep #2131’s transaction API rather than carry `persistent_ref_update.go` forward as a second implementation. I would move #1926’s guarantees into `RunRefTransaction` and the multi-ref transaction path.

The result should work like this:

- The transaction acquires #1926’s per-ref filesystem lock before reading the tip.
- It holds the lock while building and publishing the commit.
- Native Git CAS remains the final publication check.
- If an external writer bypasses Entire’s lock and moves the ref, the transaction rereads the winning tip and rebuilds.
- Multi-ref transactions acquire all required ref locks in a fixed order before using the atomic multi-ref CAS path.
- Losing commits are removed only after confirmed CAS conflicts. Hard errors preserve them because Git may already have published the ref.

The process-local mutex in #2131 and the filesystem lock from #1926 have different jobs. The mutex protects go-git object construction from goroutines in one CLI process. The filesystem lock coordinates separate Entire processes. Git CAS protects the final ref update, including against tools that do not use Entire’s lock.

I would also keep the intent of #1926’s tests. Cooperating Entire writers should serialize on the filesystem lock. A simulated external Git writer should bypass the lock and exercise the CAS retry path. The concurrent Stop-hook integration regression should remain as end-to-end coverage.

Stage 3 would keep this locking. It would group prepared sessions by checkpoint ID and submit one `BatchSessions` write, so the persistent-ref lock is acquired once for the checkpoint batch instead of once per session. The remaining Stage 3 work is in the lifecycle: preparing sessions without immediately writing them, durable recovery intent, state finalization after publication, partial-backend retries, attribution, telemetry, and safe shadow-branch cleanup.

So the merge rule I’m proposing is: keep #2131’s transaction and batching structure, but carry #1926’s locking, cleanup, cancellation, and concurrency guarantees into that structure.

PR #2132, opened 2026-08-26
fix(cli): show all agents in picker

### LINKED ISSUE #2126: Agent picker hides the last options: OpenCode and Pi need scrolling in `entire enable`
`entire enable` shows only 6 of the 8 agents. OpenCode and Pi are cut off and reachable only by scrolling.

This is not terminal height. Reproduced under a pty at 12, 20, 24, 40 and 60 rows: exactly 6 options every time.

## Cause

huh v2.0.3 sizes the viewport from the options block alone, then subtracts the title and description lines from that height (`updateViewportSize`, `field_multiselect.go:495`):

```go
height := m.height          // 0, we never set it
if height <= 0 {
    height = lipgloss.Height(v)   // options view only, 8 lines for 8 agents
}
m.viewport.SetHeight(max(minHeight, height) - yoffset)  // yoffset = title(1) + description(1)
```

Result: 8 options minus 2 header lines = 6 visible rows. One option is lost per header line. `agent.List()` sorts alphabetically, so OpenCode and Pi are the two that fall off.

## Fix

Pass an explicit height that includes the header lines, which cancels the subtraction. The group still clamps the field to the terminal on resize, so short terminals keep scrolling as before (verified at 8 rows: 5 shown, scrolls).

Applied to both agent pickers in `cmd/entire/cli/setup.go` (`promptAgentSelection` for `enable`, and the `entire agent` manage picker).

## Wider blast radius

All 46 `huh.NewSelect` / `NewMultiSelect` call sites in `cmd/` lack an explicit height. 4 have a description and lose 2 options, 42 have a title only and lose 1. So the resume picker, review pickers, dispatch wizard, mirror wizard and doctor are all silently hiding their last entry.

Open question: sweep the remaining call sites individually, or add a wrapper in `uiform` so new pickers cannot reintroduce it.

MuskanPaliwal:
Fixes #2126.

## Problem

`entire enable` offers eight agents, but huh v2.0.3 initially displays only six even when the terminal has enough rows. Its unset `MultiSelect` height starts with the rendered option height and then subtracts the title and description height, leaving OpenCode and Pi below the initial viewport.

The underlying sizing issue is upstream in `huh` and remains present on its current `main` branch. This PR applies a
 focused workaround in Entire; a library-level fix will be handled separately upstream.

## What changed

- Add `uiform.SingleLineMultiSelectHeight` for `MultiSelect` fields with one-line options and normal field headers. Multiline options and unusually tall headers intentionally remain outside its contract.
- Apply it to the `promptAgentSelection` and `runManageAgents` agent pickers.
- Move the mirror wizard’s equivalent workaround into `uiform`, migrating `pickRepos` and `pickRegions` without changing their intended behavior.

This does not sweep other picker call sites or broaden the claim to `Select`, which uses different unset-height logic. Accessibility mode is unchanged because huh’s accessible prompt path does not use the rendered viewport height.

## Validation

- Rendering regression test using all eight agent labels:
  - At 24 rows, every option is initially visible.
  - At 8 rows, Pi is initially hidden and appears after downward navigation.
- Manual PTY verification at 24x100 and 8x100.
- `mise run check` passed on the identical implementation before rebase.
- After rebasing onto current `main`, `mise run lint` and `go test ./cmd/entire/cli/uiform` passed.
- `git diff --check` passed.

https://github.com/user-attachments/assets/dbde5b0a-300a-4ff4-8a22-1ecd5ccf49de

Screenshot 2026-08-26 at 1 36 11 PM

### DIFF
--- a/cmd/entire/cli/repo_mirror_create_wizard.go
+++ b/cmd/entire/cli/repo_mirror_create_wizard.go
@@ -18,6 +18,7 @@ import (
 	"github.com/entireio/cli/cmd/entire/cli/api"
 	"github.com/entireio/cli/cmd/entire/cli/auth"
 	"github.com/entireio/cli/cmd/entire/cli/interactive"
+	"github.com/entireio/cli/cmd/entire/cli/uiform"
 	"github.com/entireio/cli/internal/coreapi"
 )
 
@@ -139,18 +140,6 @@ func selectableAvailableRepos(avail []coreapi.AvailableMirror) []coreapi.Availab
 		return out[i].Repo < out[j].Repo
 	})
 	return out
-}
-
-// multiSelectHeight returns an explicit huh multi-select Height that keeps every
-// option visible. huh auto-sizes an unset height to (rendered option lines −
-// title/description rows), which collapses to ~1 visible row for short lists
-// (e.g. 3 regions vs. a long repo list) — the cause of the region picker
-// appearing clamped to one option. We set it to the option count plus slack for
-// the title + (possibly wrapped) description so the whole list shows; huh still
-// scrolls if the list outgrows the terminal.
-func multiSelectHeight(n int) int {
-	const headerSlack = 3 // title (1) + description (1–2 when wrapped)
-	return n + headerSlack
 }
 
 // clusterChoices maps regions to multi-select options (value = bare host),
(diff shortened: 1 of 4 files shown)


MuskanPaliwal:
Hey @Soph , would appreciate your review on this. Thanks!

Soph (approved):

PR #2133, opened 2026-08-26
Resolve cluster auth against the core's cluster registry

jagregory:
https://entire.io/gh/entireio/cli/trails/1147

## What

The git remote helper (`git clone entire://<host>/…`) and the cluster-addressed `entire repo mirror …` commands decided which login could authenticate a cluster by fetching **that cluster's own** `/.well-known/entire-cluster.json` and checking the active context against the `core_urls` it advertised.

That is a self-reported source: the host under scrutiny nominates the login servers a client should trust it with. Both paths now resolve the target against the **core's cluster registry** (`GET /api/v1/clusters`, matched on `publicUrl`) — the same authoritative source cell routing already uses (`cell_target.go`'s `resolveRepoCellTarget` → `matchClusterByHost`), so cluster identity comes from one place.

The acting identity is unchanged: the active context (or `ENTIRE_TOKEN`), whose core is the core consulted. Resolution is **authoritative, not best-effort** — an unlisted host, or a registry that can't be consulted, fails with an error naming the core it asked and pointing at `entire auth use`. No fallback to per-host discovery.

The `ENTIRE_TOKEN` path still derives its core from the token's own (unverified) `aud`, but that now buys nothing: the core `aud` names must **also** list the cluster host the user typed, so a token pointing at an attacker core reaches a registry that doesn't list the real cluster and the clone fails before any credential is attached.

## Caching

Confirmed pairings are cached in `cluster_registry.json` (cache dir, 24h TTL, keyed on core origin + cluster host), mirroring the `cluster_cores.json` cache this replaces — a warm clone/fetch/push pays no control-plane round trip, and a known cluster keeps working through a brief core outage.

**Only successes are cached.** A host the core doesn't list, and a registry that couldn't be consulted, are re-checked every time: the refusal is the security-relevant answer, and a newly onboarded cluster resolves on the very next command.

## Behaviour change (deliberate)

Acting on a cluster fronted by a **federation other than the active login** now errors instead of silently switching cores. Cross-region within a federation is unaffected — the registry is federation-global. The error points at `entire auth use <context>`.

## Also

- The host matcher and the bare-host guard (`MatchClusterByHost`, `HostFromPublicURL`, `ValidateClusterHost`) move to `internal/coreapi` so `cli`, `coreapi` and `git-remote-entire` share one definition; the `cli` copies delegate.
- `auth.ResolveControlPlaneTargetForCluster` is deleted with its sole consumer.
- `internal/entireclient/clusterdiscovery`'s cluster entry points (`ResolveContextForCluster`, `ResolveClusterAuth`, `ResolveClusterCores`) now have no production callers — left in place since the package's API-host half is still live; deletable follow-up.
- The remote helper binary grows 9.9M → 10.7M from linking the generated coreapi client (no lighter way to call `ListClusters`).
- `docs/architecture/upstream-host-resolution.md` updated.

## Verification

- `mise run lint` — clean
- `mise run test` — 15 failures, all pre-existing `TestExplainCmd_*` (verified failing on the untouched tree)
- `mise run test:integration` — 3 pre-existing `TestPluginRemoteInstall_*` failures (local `git tag` env issue)
- `mise run test:e2e:canary` — passing
- New tests: registered/case-folded host resolves, absent host fails wrapping `ErrClusterNotRegistered` with nothing attached, registry-unavailable fails closed, attacker `aud` gains nothing, warm cache does one round trip across three ops, cached cluster survives a core outage, misses/errors never cached, cache scoped per core, `NewForCluster` covered end-to-end against an httptest control plane.

---

> [!NOTE]
> **High Risk**
> Changes how git credentials and cluster-addressed control-plane calls establish trust (registry vs self-reported well-known), with security and cross-federation user-visible behavior impacts.
> 
> **Overview**
> **Cluster-addressed git and CLI paths no longer trust the target host’s `/.well-known/entire-cluster.json` to pick a login server.** They still use the **active context** (or `ENTIRE_TOKEN`) and dial **that identity’s core**, but first require `GET /api/v1/clusters` on that core to list the typed cluster host (`coreapi.VerifyClusterRegistered`)—the same catalog cell routing already uses.
> 
> **`git-remote-entire`** applies the same gate before handing out credentials; **`ENTIRE_TOKEN`** is checked against the token `aud` core’s registry, not the cluster’s advertised cores. **`NewForCluster`** / **`runCoreForCluster`** (mirror create/remove, collaborators list) dial the active client and gate on the registry instead of **`ResolveControlPlaneTargetForCluster`** and per-host discovery.
> 
> **Shared cluster-host parsing and validation** (`MatchClusterByHost`, `HostFromPublicURL`, `ValidateClusterHost`) move into **`internal/coreapi`** so CLI, remote helper, and control-plane client stay aligned. Successful (core, host) checks are cached in **`cluster_registry.json`** (24h, successes only); misses and registry errors are always re-checked.
> 
> **Deliberate behavior change:** acting on a cluster fronted by another federation **errors** with the core consulted and an **`entire auth use`** hint, rather than silently switching cores. Architecture docs are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26fc1b957173ee507403734be42ccbc33e6a3c7a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/auth/control_plane.go
+++ b/cmd/entire/cli/auth/control_plane.go
@@ -2,26 +2,12 @@ package auth
 
 import (
 	"context"
-	"errors"
 	"fmt"
-	"net/http"
 	"strings"
-	"time"
 
-	"github.com/entireio/cli/internal/entireclient/clusterdiscovery"
 	"github.com/entireio/cli/internal/entireclient/contexts"
 	"github.com/entireio/cli/internal/entireclient/userdirs"
 )
-
-// controlPlaneClusterDiscoveryTimeout bounds the one
-// /.well-known/entire-cluster.json GET a cluster-addressed control-plane
-// command makes to learn which core fronts the cluster. Short so an absent or
-// slow endpoint fails the command promptly.
-const controlPlaneClusterDiscoveryTimeout = 8 * time.Second
-
-// resolveContextForCluster is the discovery seam, swapped in tests so they
-// don't reach the network. Mirrors clusterdiscovery.ResolveContextForCluster.
-var resolveContextForCluster resolveContextFunc = clusterdiscovery.ResolveContextForCluster
 
 // ControlPlaneTarget is the resolved login server a control-plane request
 // (org/repo/project/grant) should dial, plus the bearer source for it.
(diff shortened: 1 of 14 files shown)


toothbrush:
Superseded by https://github.com/entireio/cli/pull/2248, closing.

PR #2135, opened 2026-08-26
refactor(external): load agent plugins concurrently and keep their failures

gtrrz-victor:
https://entire.io/gh/entireio/cli/trails/1149

Stacked on #2134 (`refactor-discover-and-register-external-agents`). Implements the plan in `docs/external-agents/pi-version.md`, which that PR adds. **Review after it merges** — the diff here is only the two discovery files.

## The problem

`discoverAndRegister` — shared by `DiscoverAndRegister` and `DiscoverAndRegisterAlways` — executed every `entire-agent-*` binary on `$PATH` **serially** under one shared 10s budget, checking for cancellation only *between* binaries. So one slow plugin consumed the whole budget and every binary after it was dropped, reported as a timeout it had no part in.

Every failure was `logging.Debug`'d and then discarded, so a plugin that is installed but unloadable looked exactly like one that was never installed. The forgot-`chmod +x` case failed completely silently.

Both matter because `...Always` runs on interactive, user-waiting paths, up to three times per process (`setup.go:891`, `:402`, `:496` — independently reachable entry points).

## What changed

A cheap `$PATH` scan collects candidates; one goroutine per candidate runs `info` under its own budget. A scan costs ~one budget instead of N, and a stalled plugin can't hide the ones behind it. Results are applied in `$PATH` order, so registration stays deterministic despite concurrent loading.

Outcomes go into two registries. Ready agents still land in the `agent` registry, which must keep meaning "agents you can actually use" — so broken binaries are held separately with their reason, reachable via `external.Get`, `ReadyAgents()`, `BrokenAgents()`. That is the groundwork for listing external agent state.

Errors are classified. `exec` reports a context-killed child only as `signal: killed`, so `ErrInfoTimeout` and `ErrNotExecutable` exist to let callers tell "too slow" and "forgot chmod" from a genuine protocol failure without matching on strings.

`DiscoverAndRegisterNamedAlways` shares the same load path and budget, and still returns `nil` for a binary that isn't installed — its `--summarize-provider` caller relies on that to fall through.

## Two decisions worth reviewing

**A registered agent still wins a name collision, and the shadowing binary is still never executed.** Overriding was considered and rejected: `agent.Register` overwrites process-wide and the gated scan runs inside `hooks_cmd.go` / `hooks_git_cmd.go`, so dropping `entire-agent-claude-code` into any writable `$PATH` directory would take over the agent that reads transcripts and writes checkpoints, on every hook. The same mechanism shadows `vogon` and would break the e2e canary in a way that looks like a CLI bug. The collision is now logged at `Warn` instead of vanishing, which fixes the original invisibility complaint without the takeover.

**The budget stays at 10s** — the same total the serial scan already allowed for a whole scan. Worst-case wait is therefore unchanged, while the typical case collapses from N budgets to one. The original spec asked for 300ms; that was measured and rejected:

| Case | Time |
| --- | --- |
| First exec of a freshly written binary | 320–394ms |
| Each further newly written binary | 140–170ms |
| Re-exec once warm | 7–12ms |

A newly installed binary pays code signing and cold page-in on first use, and at 5s a cold exec on a machine running the full parallel `-race` suite was *still* killed mid-flight. Undersizing this turns "the user just installed a plugin" into a broken-agent entry that then fixes itself on the next run — the worst kind of intermittent. It is a `var`, so it stays easy to tune.

## Testing

Tests no longer race the production budget: they raise it when a binary must load, and lower it when one must time out, so no assertion depends on machine load. That was found the hard way — the happy-path tests passed standalone and failed under a saturated `test:ci`.

New coverage: concurrency (four stalled binaries must not stop a fast one, asserted against the serial floor), timeout and non-executable classification, `Get` across ready/broken/unknown, a cancelled caller recording candidates without executing them, and idempotency across repeated calls (marker-file exec counts).

Deleted the mocked `os.Stat`/`exec.LookPath` failure tests and the hand-rolled `stalledPropagationCtx`, which existed to prove a context-shadowing bug this structure removes by construction.

`mise run check` passes: fmt, lint (0 issues), and `test:ci` including both e2e canaries — `roger-roger` is itself a real external agent binary, so it exercises this path end to end.

---

> [!NOTE]
> **Medium Risk**
> Changes how external agents are discovered and registered on hook/setup paths, including anti-hijack rules for name collisions; misbehavior could affect which agent handles transcripts/checkpoints, though built-ins are protected from PATH override.
> 
> **Overview**
> External agent discovery no longer walks `$PATH` serially under one shared timeout. It **globs candidates first**, runs each binary’s `info` **in parallel** with a **per-plugin** `infoTimeout` (still 10s), and applies results in **`$PATH` order** so registration stays deterministic.
> 
> **Failures are retained instead of debug-only skips.** Ready plugins still register in `agent`; unloadable binaries go into a separate **broken** registry with `BrokenAgent` + `Get`, `ReadyAgents()`, and `BrokenAgents()`. Typed errors **`ErrInfoTimeout`** and **`ErrNotExecutable`** distinguish slow kills from missing chmod. Non-executable plugins and bad `info` output are recorded; missing binaries on named discovery still return **nil**.
> 
> **Safety and repeat calls:** an external binary **cannot override** an already-registered agent (shadow binary is not executed; collision is **`Warn`**). Repeated discovery **skips re-exec** for names already judged ready/broken. A cancelled caller records candidates as broken with cancellation and does not run plugins.
> 
> Tests were expanded for concurrency, timeouts, idempotency, and CI-stable budgets via tunable `infoTimeout`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a10a19515a9895d3b57d49ecd56419ed665542ea. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/agent/external/discovery.go
+++ b/cmd/entire/cli/agent/external/discovery.go
@@ -9,7 +9,9 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/entireio/cli/cmd/entire/cli/agent"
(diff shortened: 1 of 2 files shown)


gtrrz-victor:
Closing in favor of #2136. The central-registry implementation completes the user-facing diagnostics in agent list and explicit agent resolution. The winning branch now also includes the PATH fallback, failed named-probe memoization, cold-start timeout, and CLI error-reporting fixes identified during review.

PR #2142, opened 2026-08-27
Expose original trail branches in JSON

pfleidi:
https://entire.io/gh/entireio/cli/trails/1155

## Why

Completed or closed trails can be unlinked from their branch. The API preserves the former branch as `originalBranch`, but `entire trail list --json` and `entire trail show --json` converted responses through `trail.Metadata` and dropped that value. Consumers such as entire-worktree could no longer associate an unlinked trail with its existing worktree.

Local verification also exposed two repository hygiene problems: the Codex hook integration fixture compared the macOS `/var` alias with a canonical `/private/var` path, and Entire-managed worktrees appeared as untracked repository content.

## What changed

Trail JSON now includes an always-present `original_branch` string copied directly from the API while `branch` continues to mean the currently linked branch. Human-readable trail output is unchanged.

The Codex hook fixture canonicalizes its temporary root before comparing source paths, and `.entire/worktrees/` is ignored alongside the other worktree directories.

## Usage example

An unlinked trail now retains both meanings explicitly:

```json
{
  "branch": "",
  "original_branch": "feature/native"
}
```

Linked trails continue to report the current branch through `branch` and expose the raw original-branch state separately.

The CLI extends `trail.Metadata` instead of encoding `TrailResource` directly. Direct encoding would change existing snake_case field names, expose unrelated API-only fields, and bypass the URL and description enrichment shared by list and show.

`original_branch` does not use `omitempty`, giving list and show consumers one stable object shape even when the API supplies no original branch.

## Technical tradeoffs

This is an additive JSON contract. Consumers that want a branch for worktree association should prefer `branch` and use `original_branch` only when `branch` is empty; changing the existing `branch` fallback would erase the current linked-versus-unlinked distinction.

## Reviewer notes

The downstream entire-worktree consumer still needs to adopt `original_branch` for completed, unlinked trails.

---

> [!NOTE]
> **Low Risk**
> Additive JSON field and test/hygiene fixes; no auth or write-path changes, with low risk of breaking consumers that ignore unknown keys.
> 
> **Overview**
> **Trail `list --json` and `show --json` now surface `original_branch`** alongside `branch`, mapping API `originalBranch` through `trail.Metadata` and `ToMetadata()` so unlinked trails still expose the former branch name while `branch` stays the current link (often empty). The field is always emitted (no `omitempty`) for a stable JSON shape; **human-readable trail output is unchanged** and does not show the original branch.
> 
> Also **ignores `.entire/worktrees/`** in git and **canonicalizes temp dirs with `filepath.EvalSymlinks`** in a Codex bare-worktree integration test so path comparisons work on macOS (`/var` vs `/private/var`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c86e1fb3a0d7261cf93bbf6adb0a56eeaa511ac1. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pfleidi:
Bugbot run

computermode (approved):

PR #2152, opened 2026-08-27
Add Windows PowerShell installer

ashtom:
https://entire.io/gh/entireio/cli/trails/1164

## Summary

- add a Windows installer compatible with Windows PowerShell 5.1 and PowerShell 7
- resolve stable/nightly GitHub releases, verify SHA-256 checksums, and install `entire.exe` plus `git-remote-entire.exe`
- prefer Scoop for stable installs when available, including bucket setup and install-versus-update behavior
- document the PowerShell install commands and point Git Bash/MSYS users to the native installer
- exercise installer parsing under both `powershell` and `pwsh` in the Windows CI job

## Verification

- `mise run lint` — passes with 0 issues after formatting and immediately before push
- Azure Windows VM, Windows PowerShell 5.1.26100.33158:
  - stable and nightly verified-download installs
  - checksum verification and both binaries validated
  - Scoop first-install and repeat-update delegation validated with an isolated shim
- Azure Windows VM, portable PowerShell 7.6.5:
  - help path and stable verified-download install
  - `entire.exe` execution and `git-remote-entire.exe` presence
  - Scoop first-install and repeat-update delegation
- Azure test runtimes, binaries, fake Scoop state, and PATH changes were removed after verification

## Known local test failures

`mise run check` completed formatting and lint successfully, but its race-enabled Go test phase hit existing/unrelated failures on this macOS checkout:

- versioncheck tests classify their mocked unknown `/usr/local` executable path as Homebrew on this machine
- `TestCodexAppServerHooksList_BareWorktreeUsesLayoutRoot` compares `/var/...` with the canonical `/private/var/...` path
- parallel external-agent discovery failures from the full run passed when rerun in isolation, as did `TestAttach_DiscoversExternalAgents`

This change does not modify Go production or test code.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to install scripts, docs, and CI smoke checks; no CLI runtime or auth logic is modified.
> 
> **Overview**
> Adds **`scripts/install.ps1`**, a Windows-native installer for PowerShell 5.1 and 7, with **stable** and **nightly** channels. Stable installs **delegate to Scoop** when it is on PATH (bucket add, install vs update); otherwise the script downloads the release zip, **verifies SHA-256** against `checksums.txt`, and installs **`entire.exe`** and **`git-remote-entire.exe`** under `%USERPROFILE%\.local\bin`, optionally updating the user **PATH** and surfacing PATH conflicts like `install.sh`.
> 
> **README** now leads with the `irm … | iex` flow, documents Scoop vs direct install behavior, and lists `install.ps1` in release channels. **`install.sh`**’s MSYS/Git Bash error text points users at the PowerShell installer instead of Scoop-only wording.
> 
> **CI** runs `.\scripts\install.ps1 -Help` under both **`powershell`** (5.x) and **`pwsh`** (7+) in the Windows job before existing Go tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8453cc9e35a89ba8ed7d2da959f7b1cfad741818. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,20 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
+      - name: Windows PowerShell 5 installer syntax
+        shell: powershell
+        run: |
+          if ($PSVersionTable.PSVersion.Major -ne 5) {
+            throw "Expected Windows PowerShell 5.x, got $($PSVersionTable.PSVersion)"
+          }
+          .\scripts\install.ps1 -Help
+      - name: PowerShell 7 installer syntax
+        shell: pwsh
+        run: |
+          if ($PSVersionTable.PSVersion.Major -lt 7) {
+            throw "Expected PowerShell 7 or later, got $($PSVersionTable.PSVersion)"
+          }
+          .\scripts\install.ps1 -Help
       - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
           go-version-file: go.mod
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,22 +77,8 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
-      - name: Windows PowerShell 5 installer tests
-        shell: powershell
-        run: |
-          if ($PSVersionTable.PSVersion.Major -ne 5) {
-            throw "Expected Windows PowerShell 5.x, got $($PSVersionTable.PSVersion)"
-          }
-          .\scripts\install.ps1 -Help
-          .\scripts\test\run.ps1
-      - name: PowerShell 7 installer tests
-        shell: pwsh
-        run: |
-          if ($PSVersionTable.PSVersion.Major -lt 7) {
-            throw "Expected PowerShell 7 or later, got $($PSVersionTable.PSVersion)"
-          }
-          .\scripts\install.ps1 -Help
-          .\scripts\test\run.ps1
+      # install.ps1 and its tests run in install-ps1-e2e.yml, which is gated
+      # to the files that can affect them.
       - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
           go-version-file: go.mod
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,8 +77,6 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
-      # install.ps1 and its tests run in install-ps1-e2e.yml, which is gated
-      # to the files that can affect them.
       - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
           go-version-file: go.mod
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,20 +77,22 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
-      - name: Windows PowerShell 5 installer syntax
+      - name: Windows PowerShell 5 installer tests
         shell: powershell
         run: |
           if ($PSVersionTable.PSVersion.Major -ne 5) {
             throw "Expected Windows PowerShell 5.x, got $($PSVersionTable.PSVersion)"
           }
           .\scripts\install.ps1 -Help
-      - name: PowerShell 7 installer syntax
+          .\scripts\test\run.ps1
+      - name: PowerShell 7 installer tests
         shell: pwsh
         run: |
           if ($PSVersionTable.PSVersion.Major -lt 7) {
             throw "Expected PowerShell 7 or later, got $($PSVersionTable.PSVersion)"
           }
           .\scripts\install.ps1 -Help
+          .\scripts\test\run.ps1
       - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
           go-version-file: go.mod
--- a/README.md
+++ b/README.md
@@ -69,7 +69,20 @@ curl -fsSL https://entire.io/install.sh | bash                          # stable
 
 ### Windows
 
-Install with Scoop:
+Install with Windows PowerShell 5.1 or later:
+
+```powershell
+irm https://raw.githubusercontent.com/entireio/cli/main/scripts/install.ps1 -UseBasicParsing | iex
+# & ([scriptblock]::Create((irm https://raw.githubusercontent.com/entireio/cli/main/scripts/install.ps1 -UseBasicParsing))) -Channel nightly
+```
+
+For stable releases, the PowerShell installer uses Scoop when it is available,
+adding the Entire bucket if needed. Without Scoop — and for nightly releases —
+it verifies the release checksum and installs both `entire.exe` and
+`git-remote-entire.exe` to `%USERPROFILE%\.local\bin`, adding that directory to
+your user `PATH` if needed.
+
+Or install the stable release with Scoop:
 
 ```powershell
 scoop bucket add entire https://github.com/entireio/scoop-bucket.git
@@ -128,6 +141,8 @@ How to use each channel:
 - Homebrew nightly: `brew install --cask entire@nightly`
 - `install.sh` stable: `curl -fsSL https://entire.io/install.sh | bash`
 - `install.sh` nightly: `curl -fsSL https://entire.io/install.sh | bash -s -- --channel nightly`
+- `install.ps1` stable (uses Scoop when available): `irm https://raw.githubusercontent.com/entireio/cli/main/scripts/install.ps1 -UseBasicParsing | iex`
+- `install.ps1` nightly: `& ([scriptblock]::Create((irm https://raw.githubusercontent.com/entireio/cli/main/scripts/install.ps1 -UseBasicParsing))) -Channel nightly`
 - Scoop: currently supports `stable` only via `scoop install entire/entire`
 
 ## Typical Workflow
--- a/README.md
+++ b/README.md
@@ -72,8 +72,8 @@ curl -fsSL https://entire.io/install.sh | bash                          # stable
 Install with Windows PowerShell 5.1 or later:
 
 ```powershell
-irm https://raw.githubusercontent.com/entireio/cli/main/scripts/install.ps1 -UseBasicParsing | iex
-# & ([scriptblock]::Create((irm https://raw.githubusercontent.com/entireio/cli/main/scripts/install.ps1 -UseBasicParsing))) -Channel nightly
+irm https://entire.io/install.ps1 -UseBasicParsing | iex
+# & ([scriptblock]::Create((irm https://entire.io/install.ps1 -UseBasicParsing))) -Channel nightly
 ```
 
 For stable releases, the PowerShell installer uses Scoop when it is available,
@@ -141,8 +141,8 @@ How to use each channel:
 - Homebrew nightly: `brew install --cask entire@nightly`
 - `install.sh` stable: `curl -fsSL https://entire.io/install.sh | bash`
 - `install.sh` nightly: `curl -fsSL https://entire.io/install.sh | bash -s -- --channel nightly`
-- `install.ps1` stable (uses Scoop when available): `irm https://raw.githubusercontent.com/entireio/cli/main/scripts/install.ps1 -UseBasicParsing | iex`
-- `install.ps1` nightly: `& ([scriptblock]::Create((irm https://raw.githubusercontent.com/entireio/cli/main/scripts/install.ps1 -UseBasicParsing))) -Channel nightly`
+- `install.ps1` stable (uses Scoop when available): `irm https://entire.io/install.ps1 -UseBasicParsing | iex`
+- `install.ps1` nightly: `& ([scriptblock]::Create((irm https://entire.io/install.ps1 -UseBasicParsing))) -Channel nightly`
 - Scoop: currently supports `stable` only via `scoop install entire/entire`
 
 ## Typical Workflow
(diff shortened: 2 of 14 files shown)


scripts/install.ps1
```diff
@@ -0,0 +1,389 @@
...
+
+        $headers = @{
+            "Accept"     = "application/vnd.github+json"
+            "User-Agent" = "entire-install.ps1"
+        }
+        if (-not [string]::IsNullOrWhiteSpace($env:GITHUB_TOKEN)) {
+            $headers["Authorization"] = "Bearer $($env:GITHUB_TOKEN)"
+        }
+
+        Invoke-RestMethod -Uri $Uri -Headers $headers -UseBasicParsing
```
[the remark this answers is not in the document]
gtrrz-victor: \ is accepted and silently ignored in PS 7+ (it became the default behavior). It's required for PS 5.1 to skip the IE DOM parser. Keeping it. Added a comment explaining why.

README.md
```diff
@@ -64,7 +64,20 @@ curl -fsSL https://entire.io/install.sh | bash                          # stable
 
 ### Windows
 
-Install with Scoop:
+Install with Windows PowerShell 5.1 or later:
+
+```powershell
+irm https://raw.githubusercontent.com/entireio/cli/main/scripts/install.ps1 -UseBasicParsing | iex
+# & ([scriptblock]::Create((irm https://raw.githubusercontent.com/entireio/cli/main/scripts/install.ps1 -UseBasicParsing))) -Channel nightly
+```
```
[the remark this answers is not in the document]
gtrrz-victor: Same as above — `-UseBasicParsing` is harmless on PS 7+ and required for PS 5.1. The README snippets are correct.

scripts/install.sh
```diff
@@ -61,7 +61,11 @@ detect_os() {
...
             ;;
         mingw*|msys*|cygwin*)
-            error "install.sh does not support Windows, you can install the Entire CLI using scoop:
+            error "install.sh does not support Windows. Run this from PowerShell 5.1 or later:
+
+    irm https://raw.githubusercontent.com/entireio/cli/main/scripts/install.ps1 -UseBasicParsing | iex
+
+  Or install the Entire CLI using Scoop:
 
     scoop install entire/entire
```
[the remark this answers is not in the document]
gtrrz-victor: Same — \ works fine in PS 7+, it's a no-op there. Keeping for PS 5.1 compat.

scripts/install.ps1
```diff
@@ -0,0 +1,389 @@
...
+                Write-Host "! WARNING: PATH conflict detected" -ForegroundColor Yellow
+                Write-Host "!"
+                Write-Host "! Installed to: $installPath"
+                Write-Host "! But 'entire' resolves to: $($pathCommand.Source)"
+                Write-Host "!"
+                Write-Host "! Remove the old installation or adjust PATH to prioritize:"
+                Write-Host "!   $resolvedInstallDir"
+                Write-Host ""
+                throw "Installation completed, but PATH needs adjustment."
+            }
```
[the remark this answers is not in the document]
gtrrz-victor: Fixed in 18f7d5c19 — conflict detection now checks all `entire` binaries on PATH, not just the first. When the installed one takes priority over a Scoop shim, it warns without erroring so the nightly install completes.

scripts/install.ps1
```diff
@@ -0,0 +1,471 @@
...
+                foreach ($cmd in $conflicting) {
+                    Write-Host "! Also found:   $($cmd.Source)"
+                }
+                if (-not $firstIsOurs) {
+                    Write-Host "!"
+                    Write-Host "! 'entire' currently resolves to: $($first.Source)"
+                    Write-Host "! Remove the old installation or adjust PATH to prioritize:"
+                    Write-Host "!   $resolvedInstallDir"
+                    Write-Host ""
+                    throw "Installation completed, but PATH needs adjustment."
```
Soph: `Add-ToUserPath` is at line 446 — after this `throw`. On the documented nightly path (Scoop stable installed, fresh `~\.local\bin`) the new directory isn't on `$env:Path` yet, so line 414 finds only the Scoop shim, `$firstIsOurs` is false, and we throw here before PATH is ever updated. Re-running reproduces it identically forever, because nothing ever adds the directory.

Moving `Add-ToUserPath` above the conflict check fixes it cleanly: it prepends to both the user PATH and `$env:Path`, so the re-evaluated first hit is our binary and the check correctly downgrades to warn-only. With `-NoPathUpdate` the throw is defensible, but then the message should say so. Same trap fires for any pre-existing `entire.exe` earlier on PATH, not just Scoop's.

scripts/install.ps1
```diff
@@ -0,0 +1,471 @@
...
+        $actualChecksum = (Get-FileHash -LiteralPath $ArchivePath -Algorithm SHA256).Hash
+        if (-not [string]::Equals($actualChecksum, $expectedChecksum, [StringComparison]::OrdinalIgnoreCase)) {
+            throw "Checksum verification failed. Expected: $expectedChecksum, actual: $actualChecksum"
+        }
+    }
+
+    function Get-NormalizedPath {
+        param([string] $Path)
+
+        $fullPath = [IO.Path]::GetFullPath($Path)
```
Soph: `[IO.Path]::GetFullPath` resolves relative paths against `[Environment]::CurrentDirectory`, which Windows PowerShell 5.1 does not keep in sync with `Set-Location` — so `-InstallDir .\bin` after a `cd` silently installs somewhere else. `~\tools` is worse: `~` isn't expanded, so you get a literal `~` directory.

`$ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath($Path)` is the PowerShell-correct call. The default is absolute, so this only bites an explicit relative or tilde `-InstallDir`.

scripts/install.ps1
```diff
@@ -0,0 +1,471 @@
...
+            throw "Scoop failed to install Entire CLI."
+        }
+
+        Write-Success "Entire CLI installed with Scoop"
+    }
+
+    function Get-PlatformArchitecture {
+        # PROCESSOR_ARCHITEW6432 identifies the native OS architecture when
+        # this script is running in a 32-bit PowerShell process.
+        $architecture = $env:PROCESSOR_ARCHITEW6432
```
Soph: These env vars describe the *process*, not the OS. The case to check is x64 PowerShell on ARM64 Windows, where they can report AMD64 and hand an ARM64 machine the amd64 build. The Verification section lists only x64 Azure VMs, so the arm64 branch has no evidence behind it yet.

`[Runtime.InteropServices.RuntimeInformation]::OSArchitecture` asks the OS directly (needs .NET 4.7.1+, which every ARM64 Windows has). I can't confirm the env-var behaviour without an ARM64 device — worth a run on real hardware rather than taking my word for it.

scripts/install.ps1
```diff
@@ -0,0 +1,471 @@
...
+                    throw "Installation completed, but PATH needs adjustment."
+                }
+                Write-Host "!"
+                Write-Host "! The installed version takes priority, but consider removing"
+                Write-Host "! the other installation to avoid confusion."
+                Write-Host ""
+            }
+
+            if ($SkipPathUpdate) {
+                if ($pathCommands.Count -eq 0) {
```
Soph: `$pathCommands.Count -eq 0` answers "is any `entire` on PATH", not "is the install dir on PATH". Concretely: if the directory is in the user registry PATH but not in this process's `$env:Path`, you print "Added `$resolvedInstallDir` to your user PATH" without having added anything, plus a needless restart hint — `Add-ToUserPath` had just updated `$env:Path` in-process.

`Test-PathContains -PathValue $env:Path -Directory $resolvedInstallDir` is the actual predicate.

README.md
```diff
@@ -64,7 +64,20 @@ curl -fsSL https://entire.io/install.sh | bash                          # stable
 
 ### Windows
 
-Install with Scoop:
+Install with Windows PowerShell 5.1 or later:
+
+```powershell
+irm https://raw.githubusercontent.com/entireio/cli/main/scripts/install.ps1 -UseBasicParsing | iex
```
Soph: Verified live just now: `entire.io/install.sh` → 307 → raw main (so the `.sh` side is already tip-of-main — no release-gate difference), but `entire.io/install.ps1` → 307 → `/login?next=/install.ps1` → 200 with an HTML login page. The redirect simply doesn't exist yet, which is presumably why this uses the raw URL.

Two reasons to wire it up and use `https://entire.io/install.ps1` instead: it makes `scripts/install.ps1` and the default branch name part of the public contract in three places, one of them the `install.sh` error text that ships to users (install.sh:66); and today that URL returns 200 with HTML, so anyone who guesses the `.sh` naming convention pipes an HTML login page into `iex`.

.github/workflows/ci.yml
```diff
@@ -77,6 +77,20 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
+      - name: Windows PowerShell 5 installer syntax
+        shell: powershell
+        run: |
+          if ($PSVersionTable.PSVersion.Major -ne 5) {
+            throw "Expected Windows PowerShell 5.x, got $($PSVersionTable.PSVersion)"
+          }
+          .\scripts\install.ps1 -Help
```
Soph: `-Help` returns at the first statement of `Install-Entire`, so both steps are parse-and-plumbing checks only — no arch detection, checksum matching, path normalization, or PATH logic.

`Assert-Checksum`, `Test-PathContains`, `Get-NormalizedPath` and `Get-PlatformArchitecture` are pure and trivially Pester-testable, but they live inside the `& { }` block that (correctly) keeps `irm | iex` from leaking into the caller's session. That's a real tradeoff rather than an oversight — but worth deciding deliberately, since the ordering bug on install.ps1:432 is exactly what a test would have caught.

Soph (commented):
Read the script end to end; no Windows or `pwsh` on this machine, so this is code reading plus live checks of the published install URLs — I could not execute it.

On the existing bot findings: @gtrrz-victor is right and Copilot is wrong on all three `-UseBasicParsing` comments — the parameter was retained on `Invoke-WebRequest`/`Invoke-RestMethod` in PS 6+ as an accepted no-op, so nothing throws. Cursor's finding was real, but 18f7d5c19 fixed the ranking rather than the ordering, so it still reproduces; that's the one blocker below.

Worth keeping: the checksum handling is stricter than `install.sh`'s (anchored match with `[regex]::Escape` plus a hash-shape check, vs. the unanchored `grep -iE "${archive_name}$"` on install.sh:229), and `Get-FileHash` is always present so there's no "no checksum tool found, skipping verification" escape hatch. `GITHUB_TOKEN` goes only to `api.github.com` and not to the release download, which avoids leaking it to the `objects.githubusercontent.com` redirect. TLS 1.2 via numeric `-bor 3072`, strict mode 2.0, `-LiteralPath` throughout, and the `exit`-vs-`throw` split for file-vs-pipe are all correct.

One nit not worth its own thread: the commented-out nightly line at README.md:71 sits inside a copy-paste fence and reads like a leftover — it's already given uncommented in the channels list at line 141. And neither web call passes `-TimeoutSec`, so a hung connection hangs the installer indefinitely.

scripts/install.ps1
gtrrz-victor: Fixed in cd793c493.

`Add-ToUserPath` now runs after the copy and before `Get-Command entire`. It prepends the install dir (or moves it to the front if it was already later on PATH), then the conflict check re-evaluates. Nightly next to Scoop warns about the remaining shim instead of throwing. `-NoPathUpdate` still throws, and the message says PATH was left unchanged because of the flag.

scripts/install.ps1
gtrrz-victor: Fixed in 6983bbb98.

`Get-NormalizedPath` now runs the path through `GetUnresolvedProviderPathFromPSPath` before `GetFullPath`, so `-InstallDir .\bin` follows `Set-Location` on Windows PowerShell 5.1 and `~\tools` expands to the home directory. Verified on the machine.

scripts/install.ps1
gtrrz-victor: Tested on my utm but.. 
```
Error: The property 'OSArchitecture' cannot be found on this object. Verify that the property exists.
PS ...> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      5.1.26100.9168
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.26100.9168
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1

```

Finding a proper fix

gtrrz-victor: Seems like `(Get-ItemProperty 'HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Environment').PROCESSOR_ARCHITECTURE ` is the right approach.

gtrrz-victor: Fixed in dae6c112a.

`RuntimeInformation.OSArchitecture` needs .NET 4.7.1+ and is missing on stock Windows PowerShell 5.1, so that approach failed on a Windows 10 VM. The installer now reads `PROCESSOR_ARCHITECTURE` from `HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Environment`, which is the native OS value and is not rewritten for an emulated process. Verified on Windows PowerShell 5.1 (x64 VM) and on ARM64.

.github/workflows/ci.yml
gtrrz-victor: Leaving CI as `.\scripts\install.ps1 -Help` on 5.1 and 7 on purpose.

The helpers sit inside `& { }` so `irm | iex` does not leak functions, `Set-StrictMode`, or `$ErrorActionPreference` into the caller's session. That wrap is load-bearing for the documented one-liner.

Pester on those helpers needs them importable. Pulling them into a second file is the usual module layout, but then `irm …/install.ps1 | iex` only fetches one payload — `$PSScriptRoot` is empty and the helpers file is not on disk. Inlining both copies would drift. Defining the functions at script scope (no child scope) keeps a single file testable by dot-sourcing, and after `iex` the user keeps `Get-NormalizedPath`, StrictMode, and the rest in that shell.

So the choice is: one-file `iex` with isolation, or unit tests, not both without a bootstrap/second fetch. Same tradeoff as other `irm | iex` CLIs. Scoop (GoReleaser) remains the stable Windows path; this script is the nightly / no-Scoop hatch. `-Help` is the parse-and-plumbing check we are keeping.

scripts/install.ps1
gtrrz-victor: Fixed in cd793c493 (same change as the PATH-order fix).

We no longer use `$pathCommands.Count -eq 0`. `-NoPathUpdate` uses `Test-PathContains` on `$env:Path`. Otherwise we print "Added … to your user PATH" only when `Add-ToUserPath` actually wrote the user registry PATH (`$userPathChanged`), including move-to-front.

README.md
gtrrz-victor: Docs now use `https://entire.io/install.ps1` (README one-liners and the `install.sh` Windows error). Fixed in b8ec5c004. The site redirect is a follow-up so that URL stops serving the login page.

Soph (approved):

PR #2154, opened 2026-08-28
Refuse to run when `.entire` is not a real directory

pjbgf:
https://entire.io/gh/entireio/cli/trails/1166

.entire holds session metadata, transcripts, and the redaction settings that decide what may be committed. Nothing checked that the path was a directory Entire owns, so a regular file or a symlink in its place was read and written through: settings loaded from the far end, the log sink created there, and `enable` happy to MkdirAll straight through it.

paths.ValidateEntireDirAt is the single implementation. The stat is Lstat, not Stat, so a symlink is rejected even when it points at a perfectly good directory -- that is the case a Stat check waves through. Absent stays fine, since Entire may not be enabled here yet. A stat error other than "not exist" also fails: it is not evidence the invariant is violated, but it is not evidence that it holds either, and the caller's next move is to write there.

Guarded is the default. The root PersistentPreRunE runs the check above both settings.IsSetUpAny and ensureLogger, because each of those already touches the path. A command opts out with exemptFromEntireDirCheck at its AddCommand line, so the exempt set reads as one list, and every entry needs a reason in entireDirCheckExemptions or the guard test fails.

Exemption does not mean writing through the path anyway: an exempt command in a broken repo runs, but is told the path is unsafe, which is what keeps ensureLogger from creating .entire/logs inside whatever the symlink points at. doctor is exempt so that it can run on a broken repo, and reports the problem in the group's persistent pre-run -- ahead of its own PreRunE, which loads redaction settings through .entire.

help and agent-help stay guarded: someone asking what they can do in this repo is told the repo is broken rather than handed a working command list. `entire <command> --help` is unaffected, because cobra returns flag.ErrHelp before any PersistentPreRunE.

---

> [!NOTE]
> **Medium Risk**
> Touches every guarded CLI entry point and settings/logging paths; misclassification of git or exemption drift could block legitimate use or leave a hole, though behavior is heavily tested and intentionally fail-closed.
> 
> **Overview**
> **Blocks reads and writes through a non-directory `.entire`** (file, symlink—even to a valid dir, FIFO, etc.) so session metadata, settings, and logs cannot be redirected to a path Entire does not own. Validation lives in `paths.ValidateEntireDirAt` / `RequireEntireDir` using **`Lstat`**, with three typed errors (`ErrEntireDirNotDirectory`, `ErrEntireDirUnreadable`, `ErrRepositoryUnresolved`) and matching user remedies.
> 
> **Enforcement is layered:** the root `PersistentPreRunE` runs the guard before `settings.IsSetUpAny` and `ensureLogger`; exempt commands (`exemptFromEntireDirCheck` + audited allowlist) may still run but get `safe == false` so logging does not create `.entire/logs` through a bad path. **`LoadEntireSettings`** and **`newLogger`** repeat the check for plugin/telemetry paths the pre-run misses. **`doctor`** diagnoses via its own `PersistentPreRunE` and stops before touching settings/logs.
> 
> **`WorktreeRoot` now fails closed** on ambiguous git failures: only a positive “not a git repository” skips validation; dubious ownership, missing git, empty output, etc. surface as `ErrRepositoryUnresolved` (locale pinned to C for classification). Unit and integration tests cover symlink repos, exemptions, and no writes through the symlink target.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1194781d85a24b2233461f465c6a0539f83a7b2b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -472,6 +472,19 @@ mean "write through it anyway": `checkEntireDirBeforeRun` returns
 `ensureLogger` from creating `.entire/logs` through the symlink. `newLogger`
 repeats the check for the callers that build a logger outside the pre-run.
 
+The pre-run is not the only enforcement point. `LoadEntireSettings` repeats the
+check, because the pre-run does not cover everything: external plugins are
+dispatched from `main.go` before cobra runs at all, and exempt commands still
+reach settings through the post-run telemetry path. Settings are read *from* the
+directory in question, so loading them is the one operation those callers have
+in common — the duplicated `Lstat` on the ordinary path buys the guarantee that
+the check happens at least once on the unusual ones.
+
+Outside a git repository there is no worktree root and so nothing to validate,
+and the check is skipped rather than failing. Commands that need a repository
+report its absence themselves, with a message about the repository rather than
+about `.entire`.
+
 Every exemption needs an entry in `entireDirCheckExemptions`
 (`entiredir_guard_test.go`) giving the reason; `TestEntireDirCheckExemptions`
 fails both on an unlisted exemption and on a stale entry, so an exemption added
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -439,6 +439,56 @@ return fmt.Errorf("unknown strategy: %s", name)
 - `root.go` - Sets `SilenceErrors: true` on root command
 - `main.go` - Checks for `SilentError` before printing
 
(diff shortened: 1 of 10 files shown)


pjbgf:
@cursor review

pjbgf:
@cursor review

nodo (approved):

PR #2162, opened 2026-08-28
fix(redact): preserve Grok encrypted reasoning payloads

### LINKED ISSUE #2157: Redaction corrupts Grok reasoning blocks: encrypted_content is not covered by the *signature skip rule
## Summary

Grok Build stores its reasoning-block signatures in a field named `encrypted_content`. That name does not end in `signature`, so `shouldSkipJSONLField` does not exempt it, the entropy scanner redacts it, and restoring the transcript back into Grok's session directory produces a session that Grok refuses to replay.

This is the same class of bug as the Oh My Pi `thinkingSignature` fix, just with a field name the current suffix rule doesn't reach.

## Root cause

`redact/redact.go:1033`:

```go
// Skip signature fields: cryptographic attestations, not secrets. Covers
// "signature" (Claude Code) and provider variants like "thinkingSignature"
// (Oh My Pi). ...
if strings.HasSuffix(lower, "signature") {
    return true
}
```

Grok writes reasoning lines in `~/.grok/sessions/<encoded-cwd>/<session-id>/chat_history.jsonl` like this:

```json
{"type":"reasoning","id":"r1","model_id":"grok-4.6",
 "summary":[{"type":"summary_text","text":"..."}],
 "encrypted_content":"8k0KExCUBzdOBM4cHXfgVQFwExIWdb6DC4qJyl1/I5E9cLXoEZ7MtARRP9vN...",
 "status":"completed"}
```

The value is high-entropy base64 (Shannon 5.78, well over the 4.5 threshold), so the entropy scanner rewrites it. Measured on a real 3-turn session: **13 `REDACTED` insertions**, producing values like `"REDACTED/REDACTED"` and partially-mangled strings such as `"fLH/AJ5QPs6B+XQG8dGFNJQPHyMsA1MVJwqm/REDACTED/REDA"`.

## Impact

`grok --resume <session-id>` on a restored session fails outright:

```
Invalid params: "This session's conversation history is incompatible with the current model. Please start a new session."
```

Resuming an Entire-tracked Grok session on another machine (or after any restore) is impossible. As with the previous signature bugs, the corruption is irreversible: a transcript redacted before the fix cannot be repaired, and the user has to start a fresh session.

## Reproduction

Requires the Grok adapter from entireio/external-agents#41 plus the fixes stacked on it (without those, the session is not discoverable at all and you never reach the replay error).

```bash
# 1. Scratch repo with Entire + the grok external agent enabled
cd /tmp && mkdir grok-redact && cd grok-redact && git init
printf '{"enabled":true,"telemetry":false,"external_agents":true}' > .entire/settings.json
entire enable --agent grok --telemetry=false

# 2. Real Grok session (--trust is required or project hooks are silently skipped)
grok --trust --always-approve -p "Create a file hello.txt containing exactly: hello world"
grok --always-approve -c -p "Now create goodbye.txt containing exactly: goodbye world"

# 3. Keep a pristine copy, then wipe the native session to simulate another machine
SESS=~/.grok/sessions/%2Ftmp%2Fgrok-redact/<session-id>
cp "$SESS/chat_history.jsonl" /tmp/pristine.jsonl
rm -rf "$SESS"

# 4. Restore through Entire
git add -A && git commit -m "grok output"
entire session resume master

# 5. Observe the damage and the failure
grep -c REDACTED "$SESS/chat_history.jsonl"     # -> 13
grok --always-approve --resume <session-id> -p "Which files did you create?"
# -> Invalid params: "This session's conversation history is incompatible with the current model."

# 6. Swap the pristine transcript back into the same directory: it resumes correctly
cp /tmp/pristine.jsonl "$SESS/chat_history.jsonl"
grok --always-approve --resume <session-id> -p "Which files did you create?"
# -> correctly lists hello.txt and goodbye.txt
```

Step 6 isolates redaction as the sole cause: same session directory, same `summary.json`, only the transcript bytes differ.

## Suggested fix

Exempt `encrypted_content` in `shouldSkipJSONLField`. A targeted addition alongside the existing suffix rule:

```go
// Skip opaque provider reasoning payloads. Grok Build stores its reasoning
// signature as "encrypted_content", which the *signature rule does not cover.
if lower == "encrypted_content" {
    return true
}
```

Worth considering more broadly: the suffix rule has now been widened twice for the same failure mode (`signature` -> `*signature` -> this). A small allowlist of known reasoning-payload field names, or a rule keyed on the enclosing object's `type` (`thinking`, `reasoning`), would stop the next provider from hitting it. Happy to take direction on which shape you prefer.

A regression test alongside the existing `shouldSkipJSONLField` cases in `redact/redact_test.go` would cover it.

## Context

Found while testing the Grok external agent in entireio/external-agents#41 against real Grok 4.6 (`@xai-official/grok` 1.0.5). Every other part of the resume path is fixed in the branch stacked on that PR; this is the one remaining blocker and it can only be fixed here.

MuskanPaliwal:
Fixes #2157

## The bug

Grok Build stores a replay-critical reasoning attestation in `encrypted_content`. The value is opaque, high-entropy data that must remain byte-for-byte intact.

Entire's JSONL redactor did not recognize this field as provider protocol data. The entropy scanner therefore replaced parts of the value with `REDACTED`, producing a transcript that Grok rejected when the session was restored.

## The fix

Exclude the exact, case-sensitive `encrypted_content` key from JSONL scanning.

The boundary is deliberately narrow. It does not skip entire reasoning objects or match arbitrary encrypted-looking fields. Other fields in the same object remain subject to redaction.

Direct field-classification coverage locks that exact allowlist entry in place, while the regression test exercises the real `JSONLContent` seam.

The regression test sends the same high-entropy value through `JSONLContent` twice:

- `encrypted_content` must remain unchanged.
- Ordinary `content` must still be redacted.

The sharded-redaction fixture now keeps its synthetic high-entropy value under `content`, so it continues exercising the entropy layer after `encrypted_content` becomes exempt.

Codex also uses this key, but its agent-specific sanitizer removes Codex's non-portable ciphertext before storage. Focused sanitizer, checkpoint, and integration tests confirm that behavior remains intact.

## Verification

- Reproduced the corruption with the focused regression test before the fix.
- Confirmed the same regression test passes after the fix.
- Ran focused Grok redaction and Codex sanitization tests.
- Ran `mise run check`, including formatting, lint, race-enabled unit and integration tests, and deterministic E2E canaries.
- Ran the final `mise run lint`: 0 issues.

Real-agent E2E tests were not run.


MuskanPaliwal:
Hi @gtrrz-victor , would appreciate your review here. Thanks :))

gtrrz-victor:
First, thanks for the effort you put into this, but I'm going to close it, since we're removing that field from Grok transcripts. We'll no longer need to skip redaction over it. 
If we do need to apply this redaction skip later on, I'll bring your changes back.

PR #2163, opened 2026-08-28
Keep the per-user config dir private to its owner

pjbgf:
https://entire.io/gh/entireio/cli/trails/1171

`~/.config/entire` was created `0755` where it should be `0700`.

os.MkdirAll is a no-op on a path that already exists, so the directory's mode belongs to whichever caller created it first. Three callers create this one. Two ask for 0700; the third asked for 0755, and because it runs from the root command's PersistentPostRun it fires after essentially any command, so on a fresh machine it is almost always the one that wins. The two 0700 callers could never repair the result.

Route all three through userdirs.EnsurePrivateDir, which creates at 0700 and clears group and other bits from a directory that already exists. That second half is what reaches installs already in this state; correcting the creation mode alone would only help users who have not yet run the CLI. A mode stricter than 0700 is left as the user set it, and the tightening is skipped on Windows, where the permission bits are synthetic and Chmod only toggles the read-only flag.

---

> [!NOTE]
> **Medium Risk**
> Touches credential directory creation and may chmod existing config dirs on Unix; behavior is intentionally limited for custom token-store paths but still affects real user home config layouts.
> 
> **Overview**
> Fixes a case where **`~/.config/entire`** could stay **world/group-traversable (0755)** because the version check often created the directory first with loose permissions, and later **`MkdirAll(0700)`** calls could not fix an existing directory.
> 
> Adds **`userdirs.EnsurePrivateDir`**, which creates directories at **0700** and, on Unix, **strips group/other permission bits** from directories that already exist (while preserving stricter owner modes). **Version check**, **contexts.json**, and the **default file token store** now use this helper instead of ad-hoc **`MkdirAll`**.
> 
> The file token store only tightens directories Entire chose (**`ownsDir`**): paths from **`ENTIRE_TOKEN_STORE_PATH`** are still created with **`MkdirAll(0700)`** without **`chmod`** on the parent, so CI mounts and user-chosen locations are not modified. Tests cover **`EnsurePrivateDir`** behavior and owned vs unowned token-store paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c53af1200a8bfd41905138009b712a05f6fe3e6d. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/versioncheck/versioncheck.go
+++ b/cmd/entire/cli/versioncheck/versioncheck.go
@@ -107,9 +107,16 @@ func globalConfigDirPath() string {
 }
 
 // ensureGlobalConfigDir creates the global config directory if it doesn't exist.
+//
+// It goes through userdirs.EnsurePrivateDir rather than MkdirAll because this
+// directory is shared with contexts.json and the file token store, both of
+// which hold bearer tokens. The version check runs from the root command's
+// PersistentPostRun, so on a fresh machine it is almost always what creates
+// the directory — before the first login ever runs. Creating it world-readable
+// here left it that way permanently, since the credential stores' own
+// MkdirAll(0700) cannot change the mode of a directory that already exists.
 func ensureGlobalConfigDir() error {
-	//nolint:gosec // ~/.config/entire is user home directory, 0o755 is appropriate
-	if err := os.MkdirAll(globalConfigDirPath(), 0o755); err != nil {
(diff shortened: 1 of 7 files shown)


pjbgf:
@cursor review

pjbgf:
@cursor review

pfleidi (approved):

PR #2167, opened 2026-08-29
fix(checkpoint): never store a shadow checkpoint with a transcript chunk hole

suhaanthayyil:
https://entire.io/gh/entireio/cli/trails/1175

## Problem

A transcript larger than `agent.MaxChunkSize` (50 MB) is stored as `full.jsonl`, `full.jsonl.001`, `full.jsonl.002` … and every consumer reassembles it from whatever chunk files are present, in index order, with **no gap marker** (`agent.SortChunkFiles` + `agent.ReassembleTranscript`, and the same `base + ".%03d"` convention elsewhere).

`addTaskMetadataToTree` logged a warning and `continue`d when a chunk's blob write failed:

```go
blobHash, blobErr := CreateBlobFromContent(s.repo, chunk)
if blobErr != nil {
    logging.Warn(ctx, "failed to create blob for transcript chunk", ...)
    continue
}
```

so the checkpoint was committed with that chunk missing. `full.jsonl` + `full.jsonl.002` reassembles into a syntactically clean transcript that is silently short by up to 50 MB — no error, no marker, nothing a reader can detect.

This does not stay in the shadow branch. Condensation extracts the session transcript **out of the shadow tree** (`extractOrCreateSessionData` → `prepareTranscriptForStorage`), so the truncation is copied into the committed checkpoint on `entire/checkpoints/v1` and into everything downstream of it.

Both transcript-chunk loops in the committed store already return on a blob failure (`persistent.go:1063`, `persistent.go:2038`), as does the shadow store's own session-metadata walker (`ephemeral.go:998`). This loop was the only one that swallowed it.

## Fix

- A failed chunk blob write now fails the checkpoint write. A partial chunk set is unusable data, and failing leaves the previous shadow commit intact rather than replacing it with a corrupt one.
- The transcript **read** failure is now logged. It stays non-fatal (the agent may not have flushed its transcript yet), but until now an unreadable transcript produced a transcript-free checkpoint with no trace of why — the `if readErr == nil` had no `else` at all.

## Tests

`TestAddTaskMetadataToTree_PartialTranscriptChunkSetFails` fails on `main` (the write succeeds and the transcript is silently absent) and passes here. `TestAddTaskMetadataToTree_WritesEveryTranscriptChunk` pins the positive half.

`createBlobFromContent` is a test seam over `CreateBlobFromContent` so a single blob write can be failed, mirroring the existing `chunkTranscript` seam in `persistent.go`.

`mise run lint` clean; `go test ./cmd/entire/cli/checkpoint/...` green (677 tests).

### DIFF
--- a/cmd/entire/cli/checkpoint/ephemeral.go
+++ b/cmd/entire/cli/checkpoint/ephemeral.go
@@ -43,6 +43,12 @@ const (
 	WorktreeIDHashLength = 6
 )
 
+// createBlobFromContent is an indirection over CreateBlobFromContent so tests
+// can force a single blob write to fail and assert that the shadow write
+// refuses to commit a partial transcript chunk set. Production code paths
+// always use the real function. Mirrors chunkTranscript in persistent.go.
+var createBlobFromContent = CreateBlobFromContent
+
 // HashWorktreeID returns a short hash of the worktree identifier.
 // Used to create unique shadow branch names per worktree.
 func HashWorktreeID(worktreeID string) string {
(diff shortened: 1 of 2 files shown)


Beehtrix:
@suhaanthayyil and this one https://github.com/entireio/cli/pull/2177

PR #2168, opened 2026-08-29
fix(trailers): stop untrusted commit subjects from forging Entire trailers

suhaanthayyil:
https://entire.io/gh/entireio/cli/trails/1176

## Problem

A shadow checkpoint's commit subject is built from text Entire does not author:

- `strategy.FormatIncrementalMessage(step.TodoContent, ...)` — the `content` of a TodoWrite item the model wrote, verbatim apart from a 60-rune truncation.
- `generateCommitMessage` → `cleanPromptForCommit` — the session's first prompt, verbatim apart from prefix trimming and a 72-rune truncation.

Neither strips newlines. That text is steerable by anything the agent read into its context, including a file in the repository it is working on.

`FormatShadowCommit` / `FormatShadowTaskCommit` write that subject, then a blank line, then the real trailers. The parsers in this package scanned the **whole message, unanchored, and took the first match** — and the subject comes first. So one newline in the subject was enough:

```
Refactored the parser
Entire-Metadata: ../../../../etc      <- injected
Entire-Session: attacker-session      <- injected

Entire-Metadata: .entire/metadata/<real>
Entire-Session: <real>
Entire-Strategy: manual-commit
```
```
ParseSession(msg)  -> "attacker-session"
ParseMetadata(msg) -> "../../../../etc"
```

This is not cosmetic. The forged values drive real decisions:

- `ephemeralStore.ListCheckpoints` (`checkpoint/ephemeral.go:544`) filters commits by `ParseSession`, so a forged session ID re-attributes a checkpoint to another session, or hides it from its own.
- `ReadEphemeral` (`checkpoint/ephemeral.go:206-207`) and `ListCheckpoints` (`:572`) return `ParseMetadata`'s value as the checkpoint's metadata directory.
- Manual-commit rewind (`strategy/manual_commit_rewind.go:343,497,560`) resolves *which session's shadow branch to reset* from `ParseSession`.

## Fix

Both halves are needed — I checked, and flattening alone does not close it. Collapsing the subject to one line still left `Entire-Session:` sitting mid-line, and the unanchored pattern matched it there, so the collapsed subject kept winning.

1. **Flatten** the subject and the trailer values onto single lines in both shadow formatters. Every call site already passes a single-line subject, so this is lossless for legitimate input. Values get the same treatment because a newline in one splices an extra trailer line into the block, and `ValidateSessionID` is a denylist that does not reject a newline.
2. **Anchor** the four trailer patterns to the start of a line (`(?m)^`). A git trailer is a line, not a substring; a key appearing inside prose was never a trailer.

## Not covered here

`FormatCheckpoint` embeds the user's (or agent's) real commit message, which is legitimately multi-line, so it cannot be flattened. Anchoring hardens it against mid-line mentions but a line-start `Entire-Checkpoint:` typed into a commit body still parses. Closing that needs scoping the reads to the final trailer block (as `HasOPFApplied` already does via `finalTrailerBlock`), which changes `ParseAllCheckpoints` semantics for squash merges — a separate call.

## Tests

Three new tests in `trailer_forgery_test.go`, each failing on `main` and passing here. They assert the parsers return the real values and that the only lines *beginning* with a trailer key are the ones the formatter wrote.

The existing `"multiple trailers returns first"` case is unchanged (both trailers are in the real block).

`mise run lint` clean. `go test ./cmd/entire/cli/trailers/... ./cmd/entire/cli/checkpoint/...` green (733). `./cmd/entire/cli/strategy/...` 1065 pass; the 4 `git worktree add` failures there reproduce with these changes stashed and are environmental.

### DIFF
--- a/cmd/entire/cli/trailers/trailer_forgery_test.go
+++ b/cmd/entire/cli/trailers/trailer_forgery_test.go
@@ -108,3 +108,51 @@ func TestFormatShadowCommit_TrailerValuesCannotSpliceLines(t *testing.T) {
 		t.Errorf("message has %d trailer-started lines, want %d:\n%s", got, want, msg)
 	}
 }
+
+// TestParse_IndentedSquashTrailersStillParse pins the boundary of the
+// start-of-line anchor. `git merge --squash` nests every original commit
+// message inside "Squashed commit of the following:" and indents it by four
+// spaces, so a genuine trailer reaches the parsers indented. An anchor that
+// demanded column zero silently stopped resolving checkpoints on squash-merged
+// branches — the e2e TestResumeSquashMergeMultipleCheckpoints caught it — which
+// is why the patterns skip leading horizontal whitespace.
+func TestParse_IndentedSquashTrailersStillParse(t *testing.T) {
+	t.Parallel()
+
+	const session = "2026-01-20-8f76b0e8-b8f1-4a87-9186-848bdd83d62e"
+	msg := "Squashed commit of the following:\n" +
+		"\n" +
+		"commit 1111111111111111111111111111111111111111\n" +
+		"Author: A <a@example.com>\n" +
+		"Date:   Fri Aug 29 00:00:00 2026 +0000\n" +
+		"\n" +
+		"    Add red doc\n" +
+		"\n" +
+		"    Entire-Checkpoint: abc123def456\n" +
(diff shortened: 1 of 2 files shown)


Beehtrix:
@suhaanthayyil brought these in for you here: https://github.com/entireio/cli/pull/2176 :)

PR #2175, opened 2026-08-30
Update ephemeral_test.go

Beehtrix:
These tests don’t call t.Parallel(), but most other tests in this package do (e.g. cmd/entire/cli/checkpoint/objectsigner_test.go:13, cmd/entire/cli/checkpoint/persistent_reader_test.go:16). Adding it here will keep the suite consistent and improve overall test runtime.


Soph:

PR #2176, opened 2026-08-30
fix(trailers): patch single-line bypass in shadow commit subjects

Beehtrix:
Edge case: a single-line bypass. 

Because the regexes are anchored to the line-start, a hostile single-line subject that begins exactly with a trailer key (e.g., `Entire-Session: attacker`) is still evaluated as line 1 and successfully parsed as a trailer. 

This layers onto your forgery fix by:
- Updating `flattenSubject()` to check if the flattened subject starts with `Entire-`. If it does, it prepends `(subject) ` to safely neutralize the line-start anchor.
- Adding a regression test in `trailer_forgery_test.go` to explicitly pin this single-line bypass behavior.


Soph:

PR #2177, opened 2026-08-30
Update ephemeral_transcript_chunk_test.go

Beehtrix:
The new failure test doesn’t actually exercise the “chunk hole” scenario described in the PR: the transcript written here is far below agent.MaxChunkSize (so it produces only one chunk), and the override makes every chunk write fail (so it never simulates “some chunks written, one missing”). This can pass even if we regress on the real multi-chunk truncation case.

Consider forcing multi-chunk input and failing exactly one later chunk write (e.g., the second call) so the test covers the silent-truncation risk the change is meant to prevent.


Soph:

PR #2179, opened 2026-08-30
fix(checkpoint): surface load errors for missing checkpoints

Beehtrix:
This addresses `ReadTaskRecords` swallowing errors when a checkpoint ID is invalid or unavailable. 

**Changes:**
* **`cmd/entire/cli/checkpoint_api_reader.go`**: Updated `ReadTaskRecords` to validate the checkpoint via `loadDetail` and surface resulting errors instead of unconditionally returning `nil, nil`.
* **`cmd/entire/cli/checkpoint/persistent.go`**: Updated `GitStore.ReadTaskRecords` to return the actual underlying `err` rather than hardcoding `ErrCheckpointNotFound`.

### DIFF
--- a/cmd/entire/cli/checkpoint/persistent.go
+++ b/cmd/entire/cli/checkpoint/persistent.go
@@ -1573,7 +1573,7 @@ func (s *GitStore) ReadTaskRecords(ctx context.Context, checkpointID id.Checkpoi
 	}
 	checkpointTree, err := s.getCheckpointFetchingTree(ctx, checkpointID)
 	if err != nil {
-		return nil, ErrCheckpointNotFound
+		return nil, err
 	}
 	return readTaskRecordsFromCheckpointTree(ctx, checkpointTree)
 }
--- a/cmd/entire/cli/checkpoint_api_reader.go
+++ b/cmd/entire/cli/checkpoint_api_reader.go
@@ -288,7 +288,11 @@ func (r *apiCheckpointReader) ReadSessionContent(ctx context.Context, checkpoint
 // ReadTaskRecords reports no task records: the API exposes no per-subagent
 // task record endpoint, so a checkpoint read this way has no attributable
 // subagent work.
-func (r *apiCheckpointReader) ReadTaskRecords(context.Context, id.CheckpointID) ([]checkpoint.StoredTaskRecord, error) {
+func (r *apiCheckpointReader) ReadTaskRecords(ctx context.Context, checkpointID id.CheckpointID) ([]checkpoint.StoredTaskRecord, error) {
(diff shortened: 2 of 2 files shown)


Soph:
Closing this one, with the reasoning kept - the `persistent.go` observation is real and worth revisiting.

**Why it can't land as written.** It breaks two existing tests:

```
--- FAIL: TestReadTaskRecords/git-branch/unknown_checkpoint_is_ErrCheckpointNotFound
--- FAIL: TestKindRoutingStore_ReadTaskRecords/absent_everywhere_is_ErrCheckpointNotFound
    expected: "checkpoint not found"
    in chain: "sessions branch not found: reference not found"
```

Confirmed by reverting the single `persistent.go` line - both pass again.

The premise is half right. `getCheckpointFetchingTree` does return `ErrCheckpointNotFound` on a tree-walk miss (`persistent.go:2298`), but one level up `getSessionsBranchTrees` returns `fmt.Errorf("sessions branch not found: %w", ...)` (`persistent.go:2351,2353`) when the metadata branch doesn't exist. That isn't an `ErrCheckpointNotFound` chain, and a repo with nothing condensed yet is the everyday case. The contract is documented in two places the PR doesn't update: `persistent.go:1568` and `api/checkpoint/interfaces.go:24-31`.

The masking is also deliberate rather than an oversight - the same pattern sits at `persistent.go:1466` with a comment saying so at `:1457-1458`, and `GitStore.Read` (`:1424`) is more lenient still. Changing one of three leaves the family inconsistent.

**Drop the `checkpoint_api_reader.go` half regardless.** `apiCheckpointReader` is constructed only at `explain_repo.go:48` (the `--repo` path), and the sole production caller of `ReadTaskRecords` (`checkpoint_tokens_load.go:89`) can never hold one - so the change is a runtime no-op that adds a `GET /repos/{id}/checkpoints/{id}` per call the moment a caller appears. It also doesn't reach the consistency it aims for: `loadDetail` returns bespoke 404/403 prose (`:357,360`) that doesn't wrap `ErrCheckpointNotFound` either.

**If you want to revisit it**, the shape that works is `errors.Is` gating rather than replacing the sentinel:

```go
if errors.Is(err, ErrCheckpointNotFound) { return nil, err }
return nil, fmt.Errorf("resolve checkpoint tree: %w", err)
```

applied to `getSessionTree` and `Read` too, with both doc comments updated and a test. Or make `getSessionsBranchTrees` wrap `ErrCheckpointNotFound` so the existing contract holds.

Worth noting this targets `tokens-report`, and `ReadTaskRecords` doesn't exist on `main` yet - so it'd need to go to that branch's owner in any case.

PR #2181, opened 2026-08-30
Update dispatch_wizard.go

Beehtrix:
In accessible mode the repo picker is always included, but the validator still requires at least one repo even when the user selects Local mode (where repos are ignored). This makes Local flow unnecessarily blocked/awkward in accessible mode.


Soph:

PR #2182, opened 2026-08-30
Update status.go

Beehtrix:
settings.FilesPresent errors are already contextual; wrapping with fmt.Errorf("%w", err) is a no-op for the message but changes the concrete error value (which can break any direct equality checks). Prefer returning the original error directly.

### DIFF
--- a/cmd/entire/cli/status.go
+++ b/cmd/entire/cli/status.go
@@ -77,7 +77,7 @@ func runStatus(ctx context.Context, w io.Writer, detailed, jsonOutput bool) erro
 	// Check which settings files exist
 	projectExists, localExists, err := settings.FilesPresent(ctx)
 	if err != nil {
-		return fmt.Errorf("%w", err)
+		return err
 	}
 
 	if !projectExists && !localExists {


Soph:
Closing this one - the Go reasoning is right, but the wrap is load-bearing for a linter.

Your reading checks out: `settings.FilesPresent` does already wrap contextually on all three return paths (`settings/settings.go:1649,1653,1657`), and `fmt.Errorf("%w", err)` is message-identical while changing the concrete type.

The reason the line looks redundant is `wrapcheck`, enabled at `.golangci.yaml:75`. `cmd/entire/cli/settings` isn't in its `ignore-package-globs`, so removing the wrap fails lint:

```
cmd/entire/cli/status.go:80:10: error returned from external package is unwrapped:
sig: func settings.FilesPresent(ctx context.Context) (project bool, local bool, err error) (wrapcheck)
```

Same idiom for the same reason at `settings/settings.go:1034` and `:1069`.

On the stated risk: nothing in the repo compares this error. `runStatus` has one caller (`status.go:45`), the error flows to `main.go` which only prints it, and `FilesPresent`'s two call sites don't test identity. `errors.Is` and `errors.As` are unaffected by the wrap either way - only a top-level `==` or a bare type assertion on the outermost value would break, and neither exists here.

Not something you could have inferred from the file. Thanks for reading it closely.

PR #2183, opened 2026-08-30
Add Windows PowerShell installer (with PowerShell 7 compatibility fixes)

Beehtrix:
This PR builds upon the initial work by @ashtom to introduce a Windows PowerShell installer, adding necessary fixes to ensure full compatibility with PowerShell 6 and 7.

**Key Changes:**
* **`scripts/install.ps1`**: Added conditional logic to `Invoke-RestMethod` and `Invoke-WebRequest`. The `-UseBasicParsing` flag is now only passed if the host supports it (Windows PowerShell 5.1), preventing runtime errors in newer PowerShell versions where the flag is deprecated.
* **`README.md`**: Updated the one-liner installation commands to use the same conditional `-UseBasicParsing` check so users can copy/paste safely across different PowerShell versions. Cleaned up duplicate instructions in the channel list.
* **`scripts/install.sh`**: Updated the Windows fallback error message to provide the correct cross-compatible PowerShell installation commands.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,20 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
+      - name: Windows PowerShell 5 installer syntax
+        shell: powershell
+        run: |
+          if ($PSVersionTable.PSVersion.Major -ne 5) {
+            throw "Expected Windows PowerShell 5.x, got $($PSVersionTable.PSVersion)"
+          }
+          .\scripts\install.ps1 -Help
+      - name: PowerShell 7 installer syntax
+        shell: pwsh
+        run: |
+          if ($PSVersionTable.PSVersion.Major -lt 7) {
+            throw "Expected PowerShell 7 or later, got $($PSVersionTable.PSVersion)"
+          }
+          .\scripts\install.ps1 -Help
       - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
           go-version-file: go.mod
--- a/README.md
+++ b/README.md
@@ -64,7 +64,20 @@ curl -fsSL https://entire.io/install.sh | bash                          # stable
 
 ### Windows
 
-Install with Scoop:
+Install with Windows PowerShell 5.1 or later:
+
+```powershell
+irm https://raw.githubusercontent.com/entireio/cli/main/scripts/install.ps1 -UseBasicParsing | iex
+# & ([scriptblock]::Create((irm https://raw.githubusercontent.com/entireio/cli/main/scripts/install.ps1 -UseBasicParsing))) -Channel nightly
+```
+
+For stable releases, the PowerShell installer uses Scoop when it is available,
+adding the Entire bucket if needed. Without Scoop — and for nightly releases —
+it verifies the release checksum and installs both `entire.exe` and
+`git-remote-entire.exe` to `%USERPROFILE%\.local\bin`, adding that directory to
+your user `PATH` if needed.
+
+Or install the stable release with Scoop:
 
 ```powershell
 scoop bucket add entire https://github.com/entireio/scoop-bucket.git
@@ -124,6 +137,8 @@ How to use each channel:
 - Homebrew nightly: `brew install --cask entire@nightly`
 - `install.sh` stable: `curl -fsSL https://entire.io/install.sh | bash`
 - `install.sh` nightly: `curl -fsSL https://entire.io/install.sh | bash -s -- --channel nightly`
+- `install.ps1` stable (uses Scoop when available): `irm https://raw.githubusercontent.com/entireio/cli/main/scripts/install.ps1 -UseBasicParsing | iex`
+- `install.ps1` nightly: `& ([scriptblock]::Create((irm https://raw.githubusercontent.com/entireio/cli/main/scripts/install.ps1 -UseBasicParsing))) -Channel nightly`
 - Scoop: currently supports `stable` only via `scoop install entire/entire`
 
 ## Typical Workflow
--- a/README.md
+++ b/README.md
@@ -66,10 +66,15 @@ curl -fsSL https://entire.io/install.sh | bash                          # stable
 
 Install with Windows PowerShell 5.1 or later:
 
-```powershell
-irm https://raw.githubusercontent.com/entireio/cli/main/scripts/install.ps1 -UseBasicParsing | iex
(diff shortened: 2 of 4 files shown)


Soph:
Reviewed `cbcda52d`. The core is solid - checksum verification is present and stricter than `install.sh`, and the artifact names match the real release layout (verified against live v0.10.3: `entire_windows_amd64.zip` exists, sha256 matches `checksums.txt`, and the archive is flat so `Join-Path $extractDir "entire.exe"` resolves). Security review found nothing: no `Set-ExecutionPolicy`, no elevation, no `LocalMachine` scope, no `http://`, no `-SkipCertificateCheck`, no `iex` on fetched content, host pinned to `github.com/entireio/cli`, `-LiteralPath` used throughout, and the version string is semver-validated before URL interpolation - which `install.sh` doesn't do.

Five things to fix. The first is the only one I'd call blocking.

**Not verifiable locally:** this host has no `powershell`/`pwsh` and the repo ships no PowerShell tooling, so every runtime claim below is from reading the code plus Microsoft's documented behaviour, not from execution. Worth someone confirming on a real Windows box before acting on 2.

## 1. `Add-ToUserPath` permanently flattens other tools' PATH entries

`install.ps1`, `Add-ToUserPath`:

```powershell
$userPath = [Environment]::GetEnvironmentVariable("Path", "User")
...
[Environment]::SetEnvironmentVariable("Path", $newUserPath, "User")
```

`GetEnvironmentVariable` returns the **expanded** value and `SetEnvironmentVariable` writes back a `REG_SZ`. Any `%USERPROFILE%\...`-style entry already in the user's PATH - written by scoop, the Python installer, npm, VS Code - is silently converted to a hardcoded literal, and there is no way to undo it. rustup, scoop and pnpm all write PATH through `Microsoft.Win32.Registry` with `RegistryValueKind.ExpandString` specifically to avoid this.

Secondary: no `WM_SETTINGCHANGE` broadcast, so Explorer-launched shells won't pick up the change until restart (the script does handle the current process correctly).

The rest of the PATH handling is right - **User** scope only, and `Test-PathContains` makes re-runs genuinely idempotent.

## 2. The PowerShell 7 guard is unreachable

Per Microsoft's `Invoke-RestMethod` docs, `-UseBasicParsing` *"has been deprecated... is included for backwards compatibility only. When used, it has no effect on the operation of the cmdlet."* Deprecated, not removed - the parameter still appears in every 7.x syntax block. So `(Get-Command Invoke-RestMethod).Parameters.ContainsKey("UseBasicParsing")` is `$true` on PS7 exactly as on 5.1, both branches are identical, and the `else` arms never run.

The cost isn't the two lines at `install.ps1:129` and `:179` - it's that the same dead check is pasted into four user-visible places: `README.md:70-73`, `:145`, `:146`, and the new `install.sh` Windows message. `:70` spells it `Get-Command Invoke-RestMethod` while `:145`/`:146` spell it `Get-Command irm`, so there are now two variants of a check that never varies. `e2e-windows.yml:30` already uses the plain `irm ... | iex` shape.

Unconditional `-UseBasicParsing` (harmless everywhere) or dropping the flag removes roughly 40% of this PR's user-facing surface.

## 3. The advertised one-liner has no TLS 1.2 bootstrap

`Install-Entire` sets TLS 1.2 correctly - but that runs *after* the script has been downloaded. On stock Windows PowerShell 5.1 without `SchUseStrongCrypto`, `SecurityProtocol` defaults to SSL3|TLS1.0, and `raw.githubusercontent.com` requires TLS 1.2. So the bootstrap `irm` at `README.md:69-73` fails with *"Could not create SSL/TLS secure channel"* before the script exists. Chocolatey, scoop and rustup all prefix their one-liners with `[Net.ServicePointManager]::SecurityProtocol = 3072` for exactly this reason. The README claims 5.1 support, so the prefix belongs there.

## 4. `-Channel nightly` always hard-errors on a machine that has Scoop

`Install-Entire`: nightly + scoop present warns and proceeds to the direct install. Later, `Get-Command "entire" -CommandType Application` finds scoop's shim, `Test-SamePath` is false, and the script `throw`s *"Installation completed, but PATH needs adjustment."* - **after** both binaries have been copied, and before `Add-ToUserPath` runs.

So the nightly command documented at `README.md:146` reliably fails on the machine most likely to run it: one that already has stable via Scoop. The user is left half-configured looking at a red error.

## 5. The Scoop branch silently discards `-InstallDir` and `-NoPathUpdate`

`Install-Entire` calls `Install-EntireWithScoop; return` before either is consulted, and `Install-EntireWithScoop` takes no such parameters. Someone who passes `-InstallDir D:\tools` gets an install elsewhere with no message. That branch also adds a third-party bucket as a side effect without asking. Minimum: warn when `-InstallDir` was set explicitly and the Scoop path is taken, or let an explicit `-InstallDir` force the direct path.

## Testing

Both new CI steps run `.\scripts\install.ps1 -Help`, which returns early before anything happens - so the download, checksum, extract and PATH flow are all untested, and the `irm | iex` shape the README advertises (which parses as a scriptblock and binds no parameters) isn't exercised either. Running both matrix legs with explicit version assertions is more than the repo had, and the placement in `test-windows` is right; it's a syntax smoke test rather than coverage. Piping the file through `iex` in CI, or a `-WhatIf`-style dry run, would catch real regressions.

## Two things this exposes that aren't your problem

`mise-tasks/lint/shellcheck` runs `find mise-tasks -type f`, so **`scripts/install.sh` has never been shellchecked**, and there is no PowerShell linter at all. That's why `mise run lint` reports 0 issues on the 13-space indent and the trailing-whitespace-only line inside the new `install.sh` message. If installers should meet the same bar as the Go code, this is a good moment to widen that glob and add PSScriptAnalyzer - I'll open that separately.

Also `https://entire.io/install.ps1` doesn't exist yet (307s to `/login`), which is why the raw GitHub URL is baked into four places. Adding the vanity redirect before merge keeps the two installers symmetric and gives us a rollback point; `raw.githubusercontent.com` is blocked on a lot more corporate networks than a vendor domain.

## What's good

- Checksum verification anchored at both ends with a 64-hex guard and `*` binary-marker handling - stricter than `install.sh`'s `grep -iE "...$"`.
- The release version is semver-validated before it reaches a URL. `install.sh` doesn't do this.
- `PROCESSOR_ARCHITEW6432` handling for 32-bit PowerShell on ARM64/x64 - most installers get that wrong.
- The `foreach` over `Invoke-RestMethod` array output, with an accurate comment - a real 5.1 gotcha, correctly handled.
- Child-scriptblock scoping so `Set-StrictMode` and `$ErrorActionPreference` don't leak into the caller's shell when piped to `iex`.
- `try`/`finally` temp cleanup in a GUID-named directory, and TLS 1.2 via the numeric literal with a comment explaining why the enum name is avoided.

You clearly read `install.sh` and improved on it in several places. Fixes needed are localized: the registry write in 1, drop the dead guard in 2, add the TLS prefix to the README, and resolve the Scoop interaction in 4 and 5.

PR #2184, opened 2026-08-31
Isolate test git invocations from host git config

toothbrush:
https://entire.io/gh/entireio/cli/trails/1185

## Why

`TestGitStatusCallSitesPassNoOptionalLocks` failed on any machine with `color.ui = always` (ANSI escapes broke its output parse), and an audit found 54 more test call sites running git with the host's real config — hooks, colors, `init.defaultBranch`, gpgsign all leak into test repos.

I went and tried to remove as many as possible of our hand-rolled Git env isolation instances.  Hopefully the next person's agent will copy the shared helper pattern.

## What

- New `cmd/entire/cli/testutil/gitenv` package: the single git-isolation implementation (`Isolated`, `IsolateProcess`, `IsolateMain`, `Run`, `EnvHermetic`). Dependency-free, so internal tests of packages testutil imports (gitrepo) can use it too.
- Call sites keep the existing `testutil.GitIsolatedEnv` / `testutil.RunGit` / `testutil.IsolateGitConfigEnv` spelling — testutil delegates to gitenv, so the diff touches only the previously un-isolated call sites, not every caller.
- Every bare `exec.Command("git", …)` in test code now goes through `testutil.RunGit` or sets an isolated env (biggest win: `strategy/hooks_test.go`, −200 lines of exec boilerplate).
- e2e's `TestMain` uses `gitenv.IsolateMain()`, gaining system-config and `GIT_CONFIG_PARAMETERS` isolation it previously lacked.
- `gitrepo/reftable_test.go` drops its three hand-rolled env blocks; being an internal test of a package testutil imports, it calls `gitenv` directly.

## Verification

Full `mise run check` green (86 packages + vogon canary), on a machine with `color.ui = always` where the guard previously failed. Details in PR comment.


toothbrush:
Verification (local, on a machine with `color.ui = always` in `~/.gitconfig` — the config that broke the guard):

- Before: `TestGitStatusCallSitesPassNoOptionalLocks` failed every run with "guard matched no git status invocations at all; the detection pattern has gone stale" (ANSI escapes from the forced color broke the `.go`-suffix parse).
- After: `mise run test:ci` fully green — 86 packages `ok`, zero `FAIL`, vogon canary passed (exercises the changed e2e `TestMain`).
- `mise run fmt` + `mise run lint` clean.
- Audit basis: 140 `exec.Command(…"git"…)` call sites in test code; 95 already isolated, 54 bare (now migrated), plus e2e covered process-wide via `gitenv.IsolateMain()`. `gitrepo/reftable_test.go` kept its own equivalent env until the last commit, which folds it onto `gitenv` too.

toothbrush:
Second pass (2a81cce2e2): 122 more call sites that hand-rolled `exec.Command("git", …)` + `GitIsolatedEnv` around a run-and-fail shape are now `testutil.RunGit` one-liners (−575 lines net; forwarding helpers also lost their dead `ctx` params). Of the original 140 raw git execs in test code, 11 remain, each deliberate: capability probes that `Skipf`, the hermeticity tripwire that asserts failure, error-tolerant readers, e2e's artifact-logging harness, and the vogon binary. Full `mise run check` green again (86 packages + canary).

gtrrz-victor (dismissed):

peyton-alt (approved):

PR #2190, opened 2026-08-31
refactor(gitrepo): consolidate worktree metadata traversal

MuskanPaliwal:
Refs #2052, particularly [the duplicate-traversal finding](https://github.com/entireio/cli/pull/2052#issuecomment-5400538888) and [the follow-up commitment](https://github.com/entireio/cli/pull/2052#issuecomment-5429341093).

PR #2052 exposed that several packages independently interpreted `.git`, `gitdir:`, `commondir`, and linked-worktree identity. Those implementations had already diverged on layouts such as bare containers, linked worktrees, and submodules, while session and checkpoint maintained separate caches for answers derived from the same filesystem metadata.

This consolidates those physical Git facts under the existing `gitrepo` ownership boundary and removes the duplicate traversal and cache layers.

## Why this touches 67 files

This is one ownership change with a wide dependency graph, not 67 separate features.

Before this change, checkpoint, session, strategy, settings, status, review,
investigate, trail, setup/cleanup, and Codex independently interpreted some
combination of `.git`, `gitdir:`, `commondir`, and linked-worktree identity:

```text
Before

checkpoint ─┐
session ────┤
strategy ───┤ each resolves Git metadata independently
settings ───┤ and some cache the result separately
status ─────┤
review ─────┤
Codex ──────┘

After

                    gitrepo.ResolveWorktreeMetadata
                                  │
              ┌───────────────────┼───────────────────┐
           checkpoint          session            strategy...
```

The 67 changed files break down into 37 production files and 30 test files:

| Area | Files | Why they changed |
| --- | ---: | --- |
| Canonical `gitrepo` owner | 7 | Add the resolver, repository-opening boundary, tests, and ownership guard |
| Checkpoint migration | 7 | Remove its common-directory cache and migrate consumers |
| Session migration | 7 | Remove its separate cache and preserve adoption/sweep behavior |
| Strategy and hook paths | 18 | Migrate the hottest and most policy-sensitive consumers |
| Other product consumers | 18 | Migrate settings, status, review, investigate, trail, Codex, setup, and cleanup |
| Legacy `paths` deletion | 2 | Remove the old marker-based worktree-ID parser |
| Cross-cutting proof | 8 | Update fixtures and prove caller policy and subprocess behavior |

Most production changes are small call-site migrations. The largest additions
are tests proving real Git layouts, failure behavior, ownership boundaries, and
subprocess counts. Despite the broad footprint, production Go decreases by 36
lines.

## One owner for physical worktree metadata

`gitrepo.ResolveWorktreeMetadata` now resolves one explicit worktree root into:

- its per-worktree Git directory
- its common Git directory
- its optional linked-worktree ID

The resolver has no CWD lookup, Git subprocess, cache, context, options, environment interpretation, layout taxonomy, or feature-specific policy.

Returned paths remain cleaned absolute lexical paths. Physical filesystem identity is used only to determine whether the Git directory equals the common directory and whether a linked-worktree Git directory is an immediate child of `CommonDir/worktrees`.

Resolution is all-or-error. Malformed worktree metadata cannot silently downgrade into another repository-opening path, while deliberately opening a validated bare repository remains a separate `gitrepo.OpenPath` operation.

## Consolidation and deletion

All former filesystem-metadata consumers now use the canonical resolver while retaining their existing caller-owned error, warning, fallback, trust, and installation policies.

This removes:

- the session common-directory cache and its clearing behavior
- the checkpoint common-directory cache
- duplicate `.git` and `commondir` traversal
- marker-based worktree-ID parsing
- the old `paths` worktree metadata helpers
- migration wrappers that would otherwise preserve fragmented ownership

A source guard rejects new production parsing of `gitdir:`, `commondir`, or independent `.git` filesystem traversal outside `gitrepo`, with two documented policy-only exceptions.

The production Go diff is 557 additions and 593 deletions: net 36 lines removed despite adding the canonical resolver.

## Hook-path subprocess evidence

Operation-level Git tracing records these changes against the historical pre-consolidation counts:

| Operation | Before | After |
| --- | ---: | ---: |
| Session state construction | 2 | 1 |
| Construct and enqueue through checkpoint queue | 1 | 0 |
| `prepare-commit-msg` without an active session | 4 | 2 |
| `entire status` | 4 | 3 |
| Codex hook discovery | 1 | 1 |

These values count Git subprocesses, not whether the operation still occurs. The checkpoint queue retains its locking, append, drain, and removal behavior; it now obtains the common directory through filesystem metadata resolution instead of spawning Git.

The consolidation adds no Git subprocesses and removes the filesystem-metadata `--git-common-dir` queries. The one remaining common-directory query on the prepare-commit path is retained semantic repository-identity behavior rather than metadata traversal.

## Verification

- `mise run check`
  - formatting passed
  - lint passed with 0 issues
  - Vogon canary passed 56/56
  - external deterministic canary passed 4/4
- `git diff upstream/main...HEAD --check`
- Real Git-created fixtures cover ordinary repositories, linked worktrees, `.bare/worktrees`, bare repositories with linked worktrees, separate Git directories, ordinary and linked submodules, relative pointers, symlink aliases, and moved registrations.
- Handmade fixtures cover malformed and broken metadata.
- Tests verify caller-owned failure policies, fail-closed repository opening, common-directory/worktree-ID agreement, and physical alias identity.
- The source guard was positive-control tested by temporarily adding forbidden independent traversal and observing the expected failure.
- No paid real-agent E2E tests were run.

## Deliberately unchanged

This does not change Codex backlink, ownership, trust, warning, diagnostic, or installation policy. It does not add cross-worktree writes, resolver flags, a generic Git facade, or replace genuine Git queries for refs, objects, configuration, or repository semantics.


MuskanPaliwal:
Hey @Soph , I would really appreciate your review here. Thanks!

MuskanPaliwal:
@Soph, I think this PR should land before #2131, even if we assume both branches are clean and ready to merge.

The reason is architectural ordering, not conflict resolution. This PR makes `gitrepo.ResolveWorktreeMetadata` the single owner of `.git`, `gitdir:`, `commondir`, and linked-worktree identity. It also removes the separate checkpoint and session traversal code and their caches.

#2131 changes checkpoint persistence by adding CAS-based ref transactions and batch writers. Parts of that code still need the worktree root or common Git directory for native `git update-ref` calls and shadow-ref handling. If this PR lands first, #2131 has a clear rule to follow: `gitrepo` resolves physical repository metadata, while `checkpoint` owns ref publication and retries. That separation should also prevent the batching work from preserving or reintroducing another metadata traversal path.

We could merge #2131 first, but then this PR would need to absorb and migrate its new transaction code while already changing metadata ownership across the repository. That would mix two separate review questions: who owns Git metadata, and how checkpoint refs are published safely under concurrent writers.

My preferred order is:

1. Merge this PR.
2. Rebase #2131 onto it and migrate any remaining `resolveGitCommonDir` calls to `ResolveWorktreeMetadata`.
3. Rerun the batching and CAS concurrency tests, followed by the hook performance tests.

Do you see a dependency that argues for merging #2131 first?

MuskanPaliwal:
Before I resolve the conflicts with `main`, here’s the approach I’m taking.

`gitrepo.ResolveWorktreeMetadata` will remain the only code that interprets `.git`, `gitdir:`, `commondir`, and linked-worktree identity. `gitdir` will handle confined access to an already resolved Git directory through `os.Root`. It will not have its own resolver or cache.

I’ll remove the remaining `CommonDir`, `CommonDirForWorktree`, session wrapper, and cache APIs. `gitdir` will keep `OpenAt` and `OpenPathIn`, along with a clearly named helper for opening the current worktree’s common directory.

For the overlapping code, I’ll keep the rooted and no-follow I/O from `main`. That includes constructing stores with all their rooted fields, rejecting symlinked queue and redaction-cache files, and acquiring locks through the rooted flock API.

Session locks need both changes. We first resolve and deduplicate common directories by physical identity, then acquire one rooted lock per repository. This avoids taking the same lock twice when callers pass a symlink alias and the real path.

A few `--git-common-dir` calls will remain because they ask Git whether two worktrees belong to the same repository. They are semantic checks, not metadata traversal. I’ll restrict them to an explicit allowlist and remove repository override variables when they operate on a supplied path.

I’ll cover the result with the ownership guard, symlink tests, alias-lock tests, and operation-level Git subprocess counts before running the full check.

Soph:
Hold off on the reconcile, I would rather you did not spend the evening on those 18 files.

Ordering: you are right, and I see nothing arguing for #2131 first. It is not ready anyway with Stage 3 still to write.

Looking at the last days this PR will keep getting conflicts given the scope of it. We can't review fast enough before your work is conflicting again. #2180 landed today and #2220 is already open behind it, another 29 files of `os.Root` work touching five of yours, so I expect this to keep going rather than settle. I see two options:

- I continue your PR and get a better grip on getting colleagues to review in time
- You split up your PR in smaller chunks that we can more easily approve and merge.

If you are up for the split, my suggestion would be:

1. `gitrepo.ResolveWorktreeMetadata`, the repository-opening boundary, its tests and the ownership guard. No consumers migrated, so it is near pure addition and should conflict with nothing.
2. Checkpoint (7 files), then session (7). They carry the duplicate caches, so they pay off first.
3. Strategy and hooks (18).
4. The remaining consumers (18), the `paths` deletion (2), and the proof tests (8).

That is your own table, just landed in order. The ownership guard going in first is what makes the rest safe: it stops new traversal code appearing while the migration is in flight.

Your 16:34 plan carries over unchanged either way and I agree with all of it, including the session-lock dedup by physical identity, which is the part I would have expected to get missed.

MuskanPaliwal:
Hey @Soph , let's opt for the split option. I will start creating smaller prs as you suggested. We will keep the PR open until all the Split parts get delivered. Does it make sense? Thanks :)

PR #2195, opened 2026-08-31
feat(search): report result selections from the CLI search TUI (ENT-2073)

alishakawaguchi:
https://entire.io/gh/entireio/cli/trails/1192

## What was missing

We couldn't tell whether CLI search is any *good* — only that it was alive.

An important correction to the ticket first: **the CLI is not uninstrumented.** An existing search event already reports that a search ran, how long it took, how many results came back, and how it failed. It shipped a few days ago and is landing normally — 64 events from 7 users. It just lands in a **separate PostHog project for CLI telemetry**, not the web app's project, which is where the audit looked. There's no ingestion bug to chase, and four of the five things the ticket asked for already existed.

The real gap was the fifth: whether the user opened any of the results.

## Why it matters

Counts and timings measure liveness, not relevance. A search returning 80 results in three seconds looks equally healthy whether the user found what they wanted in the first row or gave up after the twentieth. The web app has tracked result clicks all along; the CLI had no equivalent, so the click-through half of "is search any good?" was invisible.

## The fix

One new event, emitted when the user opens a result's detail view in the search TUI.

1. Instrumented at the single key handler both the checkpoint tabs and the code tab open through, so both surfaces are covered without per-tab wiring.
2. Reports the result's kind, its position in the list, the size of that list, and which surface. **Position is the point** — a healthy ranking gets picked from the top row far more often than the twentieth, and that isn't derivable from counts.
3. Deduped per position within a result set, so opening and closing the same row repeatedly counts once. Opening several *different* rows still counts each, because having to dig is itself the signal.
4. The result-kind value is clamped to a known list. The search decoder passes unrecognized kinds straight through from the server, so without clamping an arbitrary server-supplied string could have been forwarded into analytics.

No query text, result titles, repo names, or file paths are sent — enums, counts, and one position only. A test asserts on the payload's key set, so a field added later fails rather than quietly leaking. Opt-out gating is identical to the existing search event's, so anyone opted out of one is opted out of both.

## Naming, for the parallel MCP work

Scheme is surface prefix plus what happened, so the MCP surface should emit its own completed/selected pair reusing these property names. Note that cross-surface dashboards need **both** PostHog projects — CLI events don't go to the web app's.

## Verified

All unit tests and lint pass clean. Main merged in with no conflicts. Reviewed the branch diff for regressions and dead code — none found.

Trail review caught one real bug during review, now fixed in a follow-up commit: the dedupe was keyed on the reported surface rather than the active tab, and since two tabs report the same surface, opening the top row of the second tab you visited was silently swallowed. Keyed on the tab instead; the added regression test fails against the old key.

No changelog entry: this repo writes the changelog in one batch per release, not per PR.

---

> [!NOTE]
> **Low Risk**
> Telemetry-only changes with opt-in gating and no search or auth behavior changes; main risk is miscounted analytics if dedupe logic regresses, which the new tests target.
> 
> **Overview**
> Adds **click-through telemetry** for the interactive `entire search` TUI via a new `cli_search_result_selected` event, emitted when the user confirms a row and opens the detail view (checkpoint tabs and Code tab share the same Confirm handler).
> 
> Payloads stay **content-free**: command path, search mode, clamped result type, rank, and result count—gated the same way as existing search outcome telemetry (settings + env opt-out), with emit deferred as a `tea.Cmd` so settings load and detached analytics do not block the Update loop.
> 
> **Deduping** uses separate semantic and code ledgers keyed by `tab:rank`, reset only when that result set refreshes (not on pagination), fixing undercount when Commits/Sessions share mode `checkpoint` and double-report when parallel semantic/code responses land at different times. Unknown API result types map to `other` before analytics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c3e7e3e5a6928f11acb1b3c04f0c6323b4666afc. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/search_selection_telemetry_test.go
+++ b/cmd/entire/cli/search_selection_telemetry_test.go
@@ -1,0 +1,100 @@
+package cli
+
+import (
+	"testing"
+
+	"github.com/entireio/cli/cmd/entire/cli/search"
+	"github.com/entireio/cli/cmd/entire/cli/telemetry"
+)
+
+// The selection ledger (searchModel.selectionsReported) is the dedupe
+// mechanism, so these tests assert on it directly. emitSearchSelection is gated
+// on loaded settings and is a no-op under test, so calling reportSelection here
+// exercises the dedupe without emitting anything.
+
+func TestReportSelection_DedupesRepeatedOpensOfSameRow(t *testing.T) {
+	t.Parallel()
+	var m searchModel
+
+	m = m.reportSelection(telemetry.SearchModeCheckpoint, "commit", 2, 10)
+	if len(m.selectionsReported) != 1 {
+		t.Fatalf("after first selection, ledger has %d entries, want 1", len(m.selectionsReported))
+	}
+
+	// Opening the same row again (enter, esc, enter) is one act of selection.
+	m = m.reportSelection(telemetry.SearchModeCheckpoint, "commit", 2, 10)
+	if len(m.selectionsReported) != 1 {
+		t.Errorf("reopening the same row grew the ledger to %d entries, want 1", len(m.selectionsReported))
+	}
+}
+
+func TestReportSelection_DistinctRanksEachReport(t *testing.T) {
+	t.Parallel()
+	var m searchModel
+
+	// Digging through four results before finding the right one is itself the
+	// relevance signal, so each distinct rank must report.
+	for rank := range 4 {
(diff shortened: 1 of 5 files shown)


alishakawaguchi:
bugbot run

alishakawaguchi:
bugbot run

PR #2199, opened 2026-08-31
Fix stale lock cleanup

### LINKED ISSUE #2197: Persistent-ref lock files accumulate one per checkpoint and nothing reaps them
Follow-up to #1926 (merged as `c7a96436`).

`persistentRefLockPath` creates a lock file per ref under `<git-common-dir>/entire-persistent-ref-locks/` (`cmd/entire/cli/checkpoint/persistent_ref_update.go:63`). Under the git-refs backend that is one file per checkpoint ref, created on first write and never removed. The `MigrateBranchToRefs` path creates one for every migrated checkpoint in a single pass.

`cmd/entire/cli/strategy/cleanup.go` has no reference to that directory, so `entire clean` does not discover or reclaim it.

## Why this is worse than the existing precedent

`shadow_ref.go:80-84` uses the same pattern for `<git-common-dir>/entire-shadow-locks/`, and `cleanup.go` doesn't know about that one either. The difference is bound: shadow lock files are keyed by base commit and worktree, so the set is bounded by branch activity. Persistent-ref locks are keyed by checkpoint ref, so the set grows monotonically with checkpoint count and never shrinks.

## Suggested fix

Add both lock directories to cleanup discovery. A lock file is safe to remove when no process holds it, which `flock` makes cheap to test - try to acquire non-blocking, and delete on success. Reaping both together also closes the pre-existing shadow-locks gap.

Rohitkanithi:
Closes #2197.

## Summary

Adds cleanup support for stale lock files created under the git common dir:

- `entire-shadow-locks`
- `entire-persistent-ref-locks`

`entire clean --all` now discovers those lock files, reports them in dry-run output, and includes them in deletion summaries.

## Review follow-up

This update addresses the lock-safety review feedback:

- Revalidates Unix flock acquisitions against the current lock-file inode before returning from both blocking `Acquire` and bounded `AcquireContext` paths.
- Moves stale lock cleanup to run after shadow branch, session state, and checkpoint cleanup so normal cleanup operations use their locks first.
- Treats actively held lock files as skipped, not failed, so live sessions do not make `entire clean --all` exit non-zero just because a lock is busy.
- Adds an active-session guard for `entire clean --all` unless `--force` or `--dry-run` is used.
- Keeps lock-directory listing best-effort if one lock directory is unreadable.
- Uses platform-specific lock-file removal ordering for Unix and Windows.

## Tests

```bash
GOCACHE=/private/tmp/entire-go-build-cache go test ./cmd/entire/cli/internal/flock ./cmd/entire/cli/strategy
GOCACHE=/private/tmp/entire-go-build-cache go test ./cmd/entire/cli -run TestClean
GOCACHE=/private/tmp/entire-go-build-cache go test ./cmd/entire/...
git diff --check
```

### DIFF
--- a/cmd/entire/cli/clean.go
+++ b/cmd/entire/cli/clean.go
@@ -294,6 +294,23 @@ func runCleanAll(ctx context.Context, cmd *cobra.Command, force, dryRun bool) er
 		unscanned = append(unscanned, "stray agent temp files")
 	}
 
+	if !force && !dryRun {
+		activeSessions, err := activeSessionsInRepository(ctx)
+		if err != nil {
+			fmt.Fprintf(cmd.ErrOrStderr(), "Warning: could not check for active sessions: %v\n", err)
+			fmt.Fprintln(cmd.ErrOrStderr(), "Use --force to override.")
+			return nil
+		}
+		if len(activeSessions) > 0 {
+			fmt.Fprintln(cmd.ErrOrStderr(), "Active sessions detected:")
+			for _, s := range activeSessions {
+				fmt.Fprintf(cmd.ErrOrStderr(), "  %s (phase: %s)\n", s.SessionID, s.Phase)
+			}
+			fmt.Fprintln(cmd.ErrOrStderr(), "Use --force to override or wait for sessions to finish.")
+			return nil
+		}
+	}
+
 	return runCleanAllWithItems(ctx, cmd, force, dryRun, items, tempFiles, orphanTemps, unscanned)
 }
 
@@ -345,7 +362,7 @@ func runCleanAllWithItems(ctx context.Context, cmd *cobra.Command, force, dryRun
 	}
 
 	// Group items by type for display
-	var branches, states, checkpoints, redactCaches []strategy.CleanupItem
+	var branches, states, checkpoints, redactCaches, lockFiles []strategy.CleanupItem
 	for _, item := range items {
 		switch item.Type {
 		case strategy.CleanupTypeShadowBranch:
@@ -356,6 +373,8 @@ func runCleanAllWithItems(ctx context.Context, cmd *cobra.Command, force, dryRun
 			checkpoints = append(checkpoints, item)
 		case strategy.CleanupTypeRedactCache:
 			redactCaches = append(redactCaches, item)
+		case strategy.CleanupTypeLockFile:
+			lockFiles = append(lockFiles, item)
 		}
 	}
 
@@ -368,6 +387,7 @@ func runCleanAllWithItems(ctx context.Context, cmd *cobra.Command, force, dryRun
 		printSection(w, "Session states", cleanupItemIDs(states))
 		printSection(w, "Checkpoint metadata", cleanupItemIDs(checkpoints))
 		printSection(w, "Redaction cache", cleanupItemIDs(redactCaches))
+		printSection(w, "Lock files", cleanupItemIDs(lockFiles))
 		printSection(w, "Temp files", tempFiles)
 		printSection(w, "Stray agent temp files", orphanTemps)
 		printUnscannedNote(w, unscanned)
@@ -409,8 +429,9 @@ func runCleanAllWithItems(ctx context.Context, cmd *cobra.Command, force, dryRun
 	failedTempFiles = append(failedTempFiles, failedOrphans...)
 
 	// Report results
-	totalDeleted := len(result.ShadowBranches) + len(result.SessionStates) + len(result.Checkpoints) + len(deletedTempFiles)
-	totalFailed := len(result.FailedBranches) + len(result.FailedStates) + len(result.FailedCheckpoints) + len(failedTempFiles)
+	totalDeleted := len(result.ShadowBranches) + len(result.SessionStates) + len(result.Checkpoints) + len(result.RedactCaches) + len(result.LockFiles) + len(deletedTempFiles)
+	totalSkipped := len(result.SkippedLockFiles)
+	totalFailed := len(result.FailedBranches) + len(result.FailedStates) + len(result.FailedCheckpoints) + len(result.FailedRedactCache) + len(result.FailedLockFiles) + len(failedTempFiles)
 
 	if totalDeleted > 0 {
 		fmt.Fprintf(w, "✓ Deleted %d %s:\n", totalDeleted, itemWord(totalDeleted))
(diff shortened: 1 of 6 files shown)


Soph:
Thanks for this, and an apology up front: the protocol you implemented is the one I prescribed in #2197 ("try to acquire non-blocking, and delete on success"), and it is not safe. That is my mistake, not yours.

flock binds to the open file description, which points at an inode, not at a name. `deleteCleanupLockFiles` unlinks the file while still holding it (`cleanup.go:712-718`), and after that unlink the name is free: the next `flock.Acquire(path)` creates a fresh inode and locks that, while anyone already blocked on the old inode still gets the old one. Two processes then hold what they both believe is the same lock. The flock package says this in its own doc comment, which I did not re-read before filing the issue: "flock state is held by the file descriptor, not by the inode on disk".

Four contending goroutines plus one reaper doing exactly what this PR does, counting how often two of them are inside the critical section at once:

```
CONTROL (no reaper):  entries=30302  violations=0
WITH #2199 REAPER:    entries=30817  violations=2074
```

I do not think there is a safe version of "delete the lock file" that leaves the acquire side unchanged. What does work is revalidating the inode after the lock is taken: stat the name, compare it against the held fd, and if they differ drop it and lock the current one.

```go
release, err := lockFile(ctx, f)
...
if holdsCurrentInode(f, path) {
	return release, nil
}
release() // the name no longer points at the inode we hold
```

That takes the same harness to 0 violations with throughput unchanged (27,487 entries vs 28,236), and the existing flock tests still pass. It has to cover `AcquireContextIn` as well, and both the blocking `LOCK_EX` and polling `LOCK_NB` paths, since `entire-shadow-locks` is taken through `Acquire`.

So my suggestion is to land the inode revalidation in `flock` first as its own change, and rebase this on top. I am happy to file that issue, or to take that piece myself if you would rather not carry it.

Smaller things, all independent of the above:

- `cleanup.go:748` fails lint: `error returned from external package is unwrapped` (wrapcheck). It is the only lint issue on the branch.
- A lock held by a live session lands in `FailedLockFiles`, and `runCleanAllWithItems` turns any failure into a returned error, so `entire clean --all` exits non-zero whenever a session is running. A held lock is the normal case, not a failure; it wants to be skipped, or reported as skipped.
- `runCleanAll` has no active-session guard at all, unlike `runCleanCurrentHead` (`clean.go:96-113`). Reaping locks makes that gap matter more than it used to.
- The lock block runs first in `DeleteAllCleanupItems`, ahead of shadow branches and session states. Those deletions take the very locks being reaped, so it should run last.
- A read error on either lock directory fails `ListAllItems` outright, so one unreadable directory aborts the whole clean. Everything else there is best-effort.
- `os.Remove` on a file the same process still holds open is a sharing violation on Windows, so this path may fail there every time. I have not verified that. Worth noting the Windows CI job only runs tests matching `(Windows|MSYS)` (`ci.yml:83`), so a normally-named test will not catch it either way.

Rohitkanithi:
Thanks again for the detailed review @Soph. I updated the implementation to address the lock-safety issue:

- added inode revalidation in Unix `flock`, including `Acquire`, `AcquireContext`, and `AcquireContextIn`
- covered both blocking `LOCK_EX` and polling `LOCK_NB` paths
- changed held lock files to be skipped instead of failed
- moved lock-file cleanup to run last
- made lock directory read errors best-effort
- added platform-specific remove ordering for Windows

Soph:
Sorry for the late feedback

Your flock fix is right, and better than what I sketched: both entry points, both the blocking and polling paths, a fresh fd per retry, and the Unix/Windows split covers the case I had flagged as unverified. I re-ran the harness from my first review against it: 0 violations over ~26k critical-section entries, where unpatched main gave 2074.

Would you be up for splitting it? The revalidation fixes a hazard in a primitive every lock in the CLI goes through, so it should land whether or not we ever clean up lock files, and the lock cleanup side still has open questions. I would rather those got settled on their own merits than decide the fate of a fix that is already correct.

Your eight files happen to divide exactly on that line. The two `internal/flock/` files apply to current main with no conflicts; the lock cleanup half holds the only one, `strategy/cleanup.go` against the adjacent var block #2230 landed. One fix needed either way: `flock_unix.go:45` returns `os.OpenFile`'s error unwrapped, the same wrapcheck failure as before, just moved.

Happy to open the flock PR myself with you as the author, so you do not spend another evening on our delay. Equally happy for it to be yours if you would rather do it. Just say which.

#2219 also looks like this branch at the same head, so one of the two is worth closing.

Rohitkanithi:
Thanks @Soph, that split makes sense. I opened the focused flock-only PR here:

https://github.com/entireio/cli/pull/2273

It contains just the two `internal/flock/` files and includes the wrapcheck fix for the path-based `os.OpenFile` error. I will keep the lock cleanup changes separate and update/close the combined PRs after the flock fix lands.

I closed https://github.com/entireio/cli/pull/2219

Rohitkanithi:
Thanks, that split makes sense. I opened the focused flock-only PR here:

https://github.com/entireio/cli/pull/2273

It contains just the two `internal/flock/` files and includes the wrapcheck fix for the path-based `os.OpenFile` error. I closed #2219 as the duplicate combined PR and will keep the lock cleanup changes separate.

Rohitkanithi:
Thanks, I cleaned this branch up now that #2273 has landed.

This is just the cleanup side now: held lock files are skipped instead of failed, `clean --all` has the active-session guard, lock cleanup runs after the other cleanup steps, unreadable lock dirs are best-effort, and the remove path is split for Unix vs Windows.

I checked it with:

- `go test ./cmd/entire/cli -run "TestCleanCmd_All_ActiveSessionGuard|TestRunCleanAllWithItems_SkippedLockFilesDoNotFail"`
- `go test ./cmd/entire/cli/internal/flock`
- `go test ./cmd/entire/cli/...`
- `go build ./cmd/entire`
- `git diff --check`

I could not run `mise run lint` locally because `mise` is not installed in this shell.

Rohitkanithi:
Thanks @Soph for merging the PR, Here

I cleaned this branch up now that #2273 has landed. This PR is only the cleanup side now.

For the cleanup approach, I kept it conservative:

- only clean the two known lock directories
- list unreadable lock dirs as best-effort skips instead of failing the whole clean
- try to take each lock with an immediate timeout before deleting it
- treat held locks as skipped, not failed
- run lock cleanup last, after the other cleanup work has had a chance to use those locks
- keep the Windows remove path separate, since removing an open file there can fail

That seemed like the safest version of reaping stale lock files now that the lower-level lock revalidation has landed. Does this direction look right to you? If you had a different shape in mind for the cleanup behavior, I can adjust it accordingly.

PR #2200, opened 2026-08-31
fix(checkpoint): skip ref updates when a checkpoint is already migrated

### LINKED ISSUE #2196: MigrateBranchToRefs re-runs take a flock and a git subprocess per already-imported checkpoint
Follow-up to #1926 (merged as `c7a96436`), which routed persistent ref updates through a cross-process flock plus native `git update-ref` CAS. Correct fix; this is a performance regression it introduced on the migration path.

## What happens now

The real migration path enters `updatePersistentRef` unconditionally at `cmd/entire/cli/checkpoint/migrate.go:108`. The in-transaction idempotency re-check at `:130` (`treeInRefHistory`) correctly detects an already-imported checkpoint and returns a no-op CAS - but only *after* the flock has been acquired and `git update-ref <ref> H H` has been forked. Before #1926 that case returned before touching a ref at all.

So an idempotent `MigrateBranchToRefs` re-run on a repo with thousands of migrated checkpoints now costs thousands of extra lock acquisitions and git subprocesses, all to conclude there was nothing to do.

The dry-run branch is unaffected - `:94-100` still short-circuits on `alreadyImported` and writes nothing, as its comment says.

## Suggested fix

Restore a `treeInRefHistory` fast path *outside* the transaction, before `updatePersistentRef` is called. This is additive and needs no new correctness reasoning: the in-transaction re-check at `:130` stays, so a checkpoint that becomes already-imported between the outer check and the CAS is still handled safely. The outer check is purely an optimisation for the common case.

## Verifying a fix

Run a migration to completion, then run it again and assert the second pass acquires no persistent-ref locks and spawns no `git update-ref`. `result.Skipped` should be unchanged.

MuskanPaliwal:
Closes #2196

#1926 made persistent ref writes safe when multiple Entire processes write at once. It also moved the migration's already-imported check inside `updatePersistentRef`.

As a result, a second migration pass takes the per-ref flock and runs `git update-ref H H` for every checkpoint, even though there is nothing to update. That adds thousands of lock acquisitions and Git processes in a large repository.

This moves the first history check back outside `updatePersistentRef`. When the normalized tree already exists in the ref's first-parent history, migration re-enqueues the ref, counts it as skipped, and returns.

The check inside `updatePersistentRef` stays. If another writer imports the checkpoint after the first check, the locked code reads the new tip and skips it safely. Re-enqueueing also stays because a previous run may have written the ref and failed before recording it in the push queue.

## Verification

The new regression test migrates a checkpoint once, then runs migration again with either the persistent-ref flock or Git's ref lock held. Both reruns succeed, report one skipped checkpoint, and leave `Migrated` empty.

On the old code, the flock case times out and the Git lock case fails inside `git update-ref`.

Git Trace2 provides a separate check. Before the fix, the second pass recorded two `update-ref H H` calls. After the fix, it recorded none.

`mise run check` passes, including lint, race-enabled unit and integration tests, and the cost-free E2E canaries.


MuskanPaliwal:
Hey @Soph , would appreciate your review here. Thanks :)

Soph:
Approach is right, and the two lock-held tests are a nice way to prove the lock is never taken.

The unlocked read is safe because `treeInRefHistory` is monotonic under ref advancement: `setRef` appends, so a tree already on the first-parent chain stays on it. The read can only go stale in the "not imported yet" direction, which the closure's re-read at `migrate.go:137` catches.

One ask. That re-read is now the correctness backstop for the TOCTOU this introduces, and it lost its only coverage:

```
origin/main   migrate.go:130.52,133.5   count 1
this PR       migrate.go:137.52,140.5   count 0
```

Full `checkpoint` package, `go test -coverprofile -count=1`. Package coverage moves 74.7% to 74.6%, so CI will not flag it.

On main that guard doubled as the ordinary idempotency check, so any plain re-run reached it. Here every already-imported re-run short-circuits at `:106`, and the only way in is the real race: the outside read says not imported, another writer imports it, the closure's re-read sees it. Writing that test is genuinely awkward, since the concurrent importer needs the same lock the migration is waiting on, so I am not going to prescribe a shape. But I would rather not merge a race backstop that nothing exercises.

MuskanPaliwal:
Hi @Soph , Fixed in [0e2565d](https://github.com/entireio/cli/pull/2200/commits/0e2565d9cb0f0904d31487e7965fa71b080b283a). Added deterministic coverage for the TOCTOU path. The test simulates another importer winning after the unlocked check, then verifies the locked re-read skips the duplicate import and preserves the winning ref.

Soph (approved):

Soph:
Thanks again! 🙇

PR #2211, opened 2026-09-01
fix(checkpoint): serialize OPF ref updates

### LINKED ISSUE #2204: OPF writers still CAS checkpoint refs via go-git, racing #1926's native update-ref
#1926 routed persistent checkpoint ref updates through a cross-process flock plus native `git update-ref <ref> <new> <old>`, which closed #1917 for the checkpoint store's own writers.

Two writers in `strategy` still update those same refs outside that lock, using go-git's `CheckAndSetReference`:

- `cmd/entire/cli/strategy/manual_commit_opf_rewrite.go:815` - CAS on `refs/heads/entire/checkpoints/v1`
- `cmd/entire/cli/strategy/manual_commit_opf_refs.go:169` - CAS on the per-checkpoint refs

Both compare-and-swap, so this is not the unconditional-update defect from #1917. The problem is that go-git's `CheckAndSetReference` does not interoperate with native git's `.lock` files - the reason documented at `cmd/entire/cli/checkpoint/shadow_ref.go:53`. So a pre-push OPF rewrite and a post-commit condensation can now race on the same ref through two mutually invisible CAS mechanisms.

The author of #1926 called this out in that PR's description as remaining scope.

Needs its own reproduction before a fix - #1917's does not cover it, since both sides here verify an expected old value.

MuskanPaliwal:
Fixes #2204

A native Git checkpoint writer and an OPF rewrite could both read the same ref value and then both report a successful update. The OPF paths used go-git's compare-and-swap, which does not participate in native Git's reference-locking protocol. A native transaction could therefore commit after the OPF rewrite and silently discard it.

## The fix

Introduce `gitrepo.CompareAndSwapRef`, a shared direct-ref compare-and-swap built on a native `git update-ref --stdin` transaction. The transaction prepares the update first, acquiring Git's ref lock, verifies that the ref is not symbolic while that lock is held, and then commits.

The native helper now backs:

- shadow-branch updates
- all persistent-ref writers using `updatePersistentRef`, including the Stop hook
- both OPF rewrite paths: the shared v1 ref and per-checkpoint refs

The OPF scan remains outside the ref transaction. Its final CAS uses the value observed before the scan, so a concurrent update is rejected instead of overwritten.

`CASPersistentRef` deliberately does not acquire Entire's persistent-writer flock. Its expected value predates the call, and native Git is the correctness boundary. Avoiding the flock also prevents deadline-free pre-push operations from waiting indefinitely. Transient native lock contention is retried with a bounded budget; a genuine stale-value conflict is terminal.

## Transaction and failure behavior

The separate `git symbolic-ref` invocation is deliberate. Git does not allow `symref-verify` and an update of the same ref in one transaction because it treats them as multiple updates to that ref. Preparing the native update first keeps the ref locked while the symbolic-ref probe runs, closing the conversion race.

CAS error classification recognizes only Git's exact stale-value diagnostics. Corrupt refs, unreadable refs, namespace conflicts, and other permanent failures retain their underlying diagnosis rather than being mislabeled and retried. The v1 rewrite reports `V1RefMovedError` only after a successful reread confirms that the ref actually changed.

Cancellation closes the transaction's stdin so Git can abort cleanly. A three-second `WaitDelay` bounds shutdown if a hook or inherited pipe prevents the process from exiting.

## Performance

Routing the shared write path through the transaction increases a measured ref update from 13.4 ms to 22.0 ms—an 8.6 ms cost per update. That is under 1% of the roughly 3-second p50 Stop-hook latency.

The extra symbolic-ref process is required for the direct-ref guarantee described above; it is not an accidental round trip.

## Verification

Regression tests use real native Git transactions and cover both files and reftable repositories. They exercise:

- both OPF lost-update races
- symbolic-ref rejection and concurrent symbolic conversion
- native lock contention and bounded retry behavior
- cancellation and lock release
- malformed, corrupt, unreadable, and namespace-conflicting refs
- preservation of the original failure when a v1 ref cannot be reread

The focused race tests passed ten consecutive runs.

`mise run check` passed after the review fixes, including formatting, lint, race-enabled unit and integration tests, and the Vogon and Roger-Roger canary suites.


MuskanPaliwal:
Pushed a follow-up in [15427c7b1](https://github.com/entireio/cli/pull/2211/commits/15427c7b100bd55b2c5e8ed5e92402e463c21aae) for the review findings. The ref CAS now lives in one place, uses `--no-deref`, and tells a stale ref apart from a held lock. Persistent writers retry when the ref disappears, and flock waits now stop on cancellation even without a deadline. I also added real loose-ref and reftable contention tests, plus coverage through the public OPF paths.

`mise run check` passes, including race tests, integration tests, and the deterministic canaries. Ready for another look.

Soph:
#2131 and #2211 both convert the two OPF call sites to native CAS, through two different APIs. `git merge-tree` on the current heads conflicts in six files, including a modify/delete on `persistent_ref_update.go`, which #2131 removes and this PR extends.

What do you want to carry forward?

MuskanPaliwal:
Hey @Soph , My bad, I missed the overlap between my two PRs. I want #2131’s transaction and batching design to win in the long run. Since #2204  is a correctness bug, should we merge #2211  now, then reconcile #2131 ? I’d keep #2131 ’s transaction API, move over the locking behavior, CAS fixes, and OPF tests, and drop `CASPersistentRef` and `persistent_ref_update.go`.

Or would you rather I move those pieces into #2131  now and close #2211 ? I outlined the design in [my latest #2131 comment](https://github.com/entireio/cli/pull/2131#issuecomment-5489979780), and I’d appreciate a review before I start moving code.

Soph:
Agreed on the order: land this one, reconcile #2131 on top. #2204 is a lost update and #2131 is a 3,900 line refactor with Stage 3 still to write, so I would rather not couple them. Your reconcile plan is the right shape, and #2131's OPF hunks become no-ops once this lands.

Re-reviewed at `f342d4fbd`. Both earlier findings are fixed, verified rather than taken from the description. One thing to fix before merge.

`ErrRefLocked` is split out because lock contention does not prove the ref moved, and then treated as terminal anyway. `atomicSetV1Ref` maps only `ErrRefCASConflict`; everything else falls through to `set v1 ref: %w`, which `manual_commit_push.go:164-170` returns, so the user's `git push` fails. `manual_commit_opf_refs.go:169` has the same shape via `warnOPFCheckpointRefsWithheld`. Every other CAS caller in this call graph does retry it: `persistent_ref_update.go:118-127` and both shadow loops build `ErrShadowRefBusy` from both sentinels and retry 16 times.

This is reachable from Entire alone on reftable repos, because reftable's ref lock is repo-global rather than per-ref. A prepared transaction on a shadow branch blocks a CAS on the unrelated v1 ref (git 2.50.1):

```
--- shadow-branch txn prepared; now CAS the UNRELATED v1 ref ---
fatal: update_ref failed for ref 'refs/heads/entire/checkpoints/v1': cannot lock references
```

Both sides there are our own native `update-ref`: a TurnEnd shadow write against the pre-push OPF rewrite. The shadow writer retries and survives; the OPF writer aborts the push and discards the rewrite. Loose-ref repos are much narrower, and `packed-refs.lock` does not widen it, since a held `packed-refs.lock` blocks deletions only and updates always land loose.

The fix is to give the two OPF CAS sites the same bounded retry the persistent path has, on `ErrRefLocked` only. `ErrRefCASConflict` should stay terminal there, since a moved ref really does invalidate the rewrite.

MuskanPaliwal:
Hey @Soph , kindly re-review. Thanks :))

Soph:
Ran a security review over this. The main question, whether the new classification can let un-redacted content reach the remote, came back clean: git's actual stderr was measured for eight failure modes against both the files and reftable backends, every one still yields a non-nil error, and every caller either aborts the push or withholds the flush with the refs left queued. Nothing fails open. Two other things came out of it.

**The v1 CAS now waits unboundedly on the user's push path.** `atomicSetV1Ref` took no flock before this PR; it now goes through `CASPersistentRef` into `withPersistentRefFlock`, and the hook's root context is `context.WithCancel(context.Background())` (`main.go:29`), so there is no deadline to bound the wait. `GitStore.updatePrimaryRef` holds that same lock across its entire build closure, condensation included, so a turn-end hook running concurrently stalls `git push` for as long as condensation takes. Measured: with the lock held from another fd, `atomicSetV1Ref` was still blocked at 3s and returned in 3.017s the instant it was released. Ctrl-C does interrupt it, which the old kernel-blocking path would not have, so this is not a wedge. Still, a bounded acquire that fails with a clear message reads better here than an indefinite wait, given what is on the other side of it is the user's push.

**`--no-deref` destroys a symbolic ref, and the test pins that as intended.** `git update-ref --no-deref` compares `<oldvalue>` against the resolved hash but writes the symref itself, so the CAS succeeds and the symbolic ref is replaced by a hash reference; against `HEAD` that detaches it. The go-git call this replaced returned `reference has changed concurrently` and left the symref intact, so this direction fails open where the old one failed closed. `TestCASPersistentRef_DoesNotDereferenceSymbolicRef` asserts `require.NoError` and then `require.Equal(t, plumbing.HashReference, updated.Type())`, which locks in the replacement, while the name reads as protection against it. No current caller passes a symbolic ref, so this is latent rather than exploitable. I would still rather `CompareAndSwapRef` rejected a symbolic target than have a test asserting it is handled.

MuskanPaliwal:
Thanks @Soph  for digging into this. Fixed both issues in [`462fd2d2c`](https://github.com/entireio/cli/pull/2211/commits/462fd2d2c44fc22b0c0d09405fe4677302d78772).

`CASPersistentRef` now waits up to two seconds for Entire's writer lock, then returns a clear timeout. The timeout only covers lock acquisition. Once acquired, the CAS uses the caller's original context.

Symbolic refs are now rejected while Git holds the ref lock. The code prepares an `update-ref` transaction, checks the ref type, then aborts for a symbolic ref or commits for a direct ref. A competing Git process cannot change the ref type between the check and the write. Ref names are also validated before they enter the stdin protocol.

I tested this with files and reftable repositories. Symbolic refs and their targets stay unchanged, a concurrent `git symbolic-ref` fails while the transaction is prepared, and a held Entire lock times out around two seconds without moving the ref. I also exercised eight failure cases on each backend; all returned errors and left the refs unchanged.

`mise run check` passes, including the integration race suite and canaries.

MuskanPaliwal:
Hi @Soph, let me know if this pr is good to go. Thanks :))

cmd/entire/cli/gitrepo/ref_cas.go
```diff
@@ -0,0 +1,234 @@
...
+) (*refCASTransaction, error) {
+	if err := refName.Validate(); err != nil {
+		return nil, fmt.Errorf("validate ref for compare-and-swap: %w", err)
+	}
+	oldValue := strings.Repeat("0", newHash.HexSize())
+	if expectedHash != plumbing.ZeroHash {
+		oldValue = expectedHash.String()
+	}
+
+	cmd := exec.CommandContext(ctx, "git", "update-ref", "--stdin")
```
Soph, line 84: **Blocking: cancellation here persistently wedges the ref.**

`exec.CommandContext`'s default cancel is `Process.Kill()` (SIGKILL), and no `Cancel`/`WaitDelay` is set. `CompareAndSwapRef` holds the transaction *prepared* — git owns `.git/refs/<name>.lock` — across the `git symbolic-ref` fork before committing. The root ctx cancels on SIGINT/SIGTERM (`main.go:44-66`), and the pre-push hook runs under it, so Ctrl-C during `git push` lands in that window.

Reproduced on git 2.50.1:

| cancel | `.lock` left | later plain `git update-ref` | txn |
| --- | --- | --- | --- |
| SIGKILL (current) | **yes** | **fails: cannot lock ref … File exists** | aborted |
| SIGTERM | no | succeeds | aborted |
| close stdin | no | succeeds | aborted |

Git has no stale ref-lock recovery — its own message says "remove the file manually to continue". The OPF path takes the lock once per ref, so a migration backfill has N windows.

Two things that may help: **SIGTERM is sufficient** (git's tempfile signal cleanup removes the lock), so `procutil.TerminateOnCancel(cmd)` fixes it in one line with an existing helper. And no variant silently commits — I checked the ref value after each, all three abort — so this is purely lock hygiene, not a spurious write. Closing stdin also works but exits 0, which the Go side would then need to not read as success.

cmd/entire/cli/checkpoint/persistent_ref_update.go
```diff
@@ -19,6 +20,10 @@ import (
 
 type persistentRefBuilder func() (newHash, expectedHash plumbing.Hash, err error)
 
+// CAS runs on the push path, so it must not wait through another writer's
+// potentially expensive checkpoint build while holding up the user's push.
+const persistentRefCASLockWait = 2 * time.Second
```
Soph: **Blocking: this doesn't delay the push, it aborts it.**

`DeadlineExceeded` → `atomicSetV1Ref` (not `ErrRefCASConflict`, so it falls through to `fmt.Errorf("set v1 ref: %w")`) → `RewriteUnpushedV1WithOPF` → `manual_commit_push.go:164` `return rewriteErr` → non-zero pre-push exit → the user's `git push` fails.

Nothing classifies it and nothing retries it: `retryPersistentRefLockContention` sits inside the flock, so it can't retry a failed acquire. And the comment above names "another writer's potentially expensive checkpoint build" as the expected holder — so it fires precisely in the scenario it was written for, where previously the push waited and succeeded.

Scope: git-branch/v1 only. The refs backend's fail-closed path withholds and returns nil, so it degrades correctly. Also worth noting the 2s bounds only the acquire — the CAS retries after it are unbounded, so it doesn't actually cap the hook's contribution to push latency.

Either drop the bound (a delayed push beats a dead one), or classify `DeadlineExceeded` so v1 withholds instead of aborting.

cmd/entire/cli/internal/flock/flock_unix.go
```diff
@@ -66,10 +62,8 @@ func AcquireContextIn(ctx context.Context, root *os.Root, name string) (release
 // lockFile holds the locking logic shared by the path- and root-based entry
 // points, which differ only in how the file was opened.
 func lockFile(ctx context.Context, f *os.File) (release func(), err error) {
-	// Fast path: no deadline -> block in the kernel exactly like the historical
-	// Acquire. This preserves behavior (and efficiency) for callers that must
-	// wait as long as it takes, such as turn-end checkpoint condensation.
-	if _, hasDeadline := ctx.Deadline(); !hasDeadline {
+	// Fast path for contexts that can never finish, such as Background.
+	if ctx.Done() == nil {
```
Soph: This changes the gate from deadline-gated to cancelability-gated, which is the opposite of what the comment being deleted promised: it said a deadline-free ctx "takes the same blocking kernel path as Acquire, **so existing callers keep their exact behavior**", and named turn-end condensation as a must-wait caller.

Every CLI context is cancelable (root ctx, `main.go:44`), so in practice only `context.Background()` now reaches the kernel path. Turn-end condensation now does 25ms `LOCK_NB` polling.

The part I'd raise loudest is fairness rather than the syscalls: `LOCK_EX` gives you a kernel queue position, poll-and-retry gives you none, so under the N-agents-one-repo workflow a poller can be starved indefinitely. This is a shared primitive affecting every caller, and no commit message mentions the change — worth either scoping it to the callers that need it or making it an explicit, documented decision.

Soph (commented):
Native CAS as the correctness boundary with the scan outside the lock is the right shape, and the lost-update tests are convincing. Three findings, two of which I'd fix before merge because they make the failure mode worse than the bug being fixed.

Minor, not blocking: `gitPlumbingEnv()` is `append(os.Environ(), …)`, so it inherits `GIT_DIR`/`GIT_WORK_TREE`, which outrank `cmd.Dir`. CLAUDE.md asks for `gitrepo.EnvWithoutRepoOverrides()` on any git subprocess that can run in a hook and targets via `cmd.Dir` — `CompareAndSwapRef` does both. Resolves to the same repo today, and the pattern predates this PR in `reftable.go`, so it's latent. `LC_ALL=C`/`LANG=C` are pinned, so the stderr matching is locale-safe.

MuskanPaliwal:
Thanks @Soph for being patient with me and for the detailed review. I fixed the latest findings in  [82b12d2](https://github.com/entireio/cli/pull/2211/commits/82b12d2e32f47bccc254194ea5e0dc3aecb38bc9). Cancellation now aborts prepared ref transactions without leaving stale locks, OPF CAS waits instead of failing the user’s push, and Git commands ignore inherited repo overrides. I also restored kernel lock fairness and added loose-ref and reftable regression tests. `mise run check` passes.

MuskanPaliwal:
Hey @Soph , highly appreciate your feedback here. The PR looks a lot cleaner now. Thanks :))

Soph:
bugbot run

cmd/entire/cli/checkpoint/persistent_ref_update.go
```diff
@@ -81,16 +70,72 @@ func withPersistentRefFlock(ctx context.Context, commonDir string, refName plumb
...
+	repo *git.Repository,
+	refName plumbing.ReferenceName,
+	newHash, expectedHash plumbing.Hash,
+) error {
+	repoRoot, commonDir, err := repositoryDirs(ctx, repo)
+	if err != nil {
+		return fmt.Errorf("resolve repository directories: %w", err)
+	}
 	return withPersistentRefFlock(ctx, commonDir, refName, func() error {
+		return retryPersistentRefLockContention(ctx, refName, func() error {
```
Soph: **This replaces my last finding with an unbounded, uninterruptible wait on the same path — and the flock it waits for is not what makes the fix work.**

The CLI root context (`main.go:29`) is cancelable but carries no deadline, so `lockFile` takes the blocking `syscall.Flock(LOCK_EX)` branch, which never consults ctx. The holder is `updatePersistentRef`, which keeps this same lock across its whole build-and-CAS loop — up to `shadowRefMaxRetries` = 16 checkpoint-tree rebuilds — and the lock file lives in the git **common** dir, so a sibling worktree contends on the same v1 ref name.

Scenario: condensation takes the v1 flock in worktree 1 and starts an expensive build (redaction dominates there). A `git push` in worktree 2 reaches `atomicSetV1Ref` -> `CASPersistentRef` and blocks in the kernel. Ctrl-C prints `Interrupting… press Ctrl-C again to force quit` and does nothing; only a second Ctrl-C escapes. No message explains the stall.

Confirmed: holding the flock and calling `CASPersistentRef` with a `context.WithCancel(context.Background())` — exactly main.go's shape — then cancelling gives `still blocked 3s after cancel`.

**And the flock earns almost nothing.** The condensation path already used native `git update-ref`, so the #2204 lost update came purely from OPF's `CheckAndSetReference`; routing OPF through native CAS is the entire fix. This flock only covers the window between its own acquire and the CAS, but `expectedOld` was read at the start of the rewrite, long before the lock — so a condensation landing mid-rewrite yields a conflict either way. Removing the flock from `CASPersistentRef` leaves all three OPF regression tests passing (`TestOPFV1RewriteDoesNotLoseConcurrentCheckpointUpdate`, `TestOPFCheckpointRefRewriteDoesNotLoseConcurrentCheckpointUpdate`, `TestAtomicSetV1Ref_ReportsActualCASConflictAsMoved`); the only failure is `TestCASPersistentRef_WaitsForHeldFlockWithoutCallerDeadline`, which exists to describe the flock itself.

Suggest dropping it here and relying on native CAS plus `retryPersistentRefLockContention` (16 attempts x ~10-25ms, so bounded well under a second). If it is kept for retry-churn reasons it needs a bound **and** a soft classification, so a timeout leaves the refs queued rather than aborting the user's push, plus a stderr line so the wait is not silent.

cmd/entire/cli/gitrepo/ref_cas.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	return "", false, fmt.Errorf("probe symbolic ref %s: %s: %w", refName, detail, err)
+}
+
+// isRefCASConflict reports that git rejected the expected old value, including
+// the cases where a required ref disappeared or a create-only ref appeared.
+func isRefCASConflict(stderr []byte) bool {
+	msg := strings.ToLower(string(stderr))
+	return strings.Contains(msg, "but expected") ||
+		strings.Contains(msg, "unable to resolve reference") ||
```
Soph: **This substring also matches three permanent, non-concurrency failures, and one route is a diagnosis regression against main.**

Reproduced on git 2.50.1, all classifying as `conflict=true`:

| planted condition | git stderr |
| --- | --- |
| corrupt loose ref | `cannot lock ref 'refs/probe/x': unable to resolve reference 'refs/probe/x': reference broken` |
| unreadable ref (mode 000) | `… unable to resolve reference 'refs/probe/y': Permission denied` |
| name is a directory | `… unable to resolve reference 'refs/probe/z'` |

Via `casUpdateRef` these become `ErrShadowRefBusy`, so `updatePersistentRef` rebuilds the checkpoint tree and retries **16 times** on the turn-end hook before failing with "shadow branch ref busy" — wrong cause, 16x the work. Via `atomicSetV1Ref` they become `V1RefMovedError`, so a corrupt or unreadable v1 ref aborts the push with a concurrency story. That second one is a regression against main, where `storage.ErrReferenceHasChanged` would not fire for these and the real cause surfaced.

`TestCASPersistentRef_DoesNotClassifyRefNamespaceConflictAsContention` covers the D/F case in the other direction (creating `parent/child` when `parent` exists yields `'…' exists; cannot create '…'` and classifies fine). The direction that breaks is the untested one.

The genuine "deleted concurrently" case has no trailing reason (verified: `unable to resolve reference 'refs/probe/gone'` and nothing more), so requiring the absence of a tail separates the first two rows. For the third, gate `V1RefMovedError` in `atomicSetV1Ref` on the re-read actually differing from `expectedOld` — it already does that re-read to fill `Actual`.

cmd/entire/cli/gitrepo/ref_cas.go
```diff
@@ -0,0 +1,245 @@
...
+		return nil, fmt.Errorf("open git update-ref stdout: %w", err)
+	}
+	tx := &refCASTransaction{
+		ctx:     ctx,
+		refName: refName,
+		cmd:     cmd,
+		stdin:   stdin,
+		stdout:  bufio.NewScanner(stdout),
+	}
+	cmd.Cancel = func() error {
```
Soph, line 103: `WaitDelay` is left at zero. Go's docs recommend setting it precisely for a custom `Cancel` that does not kill, because otherwise `Wait` blocks indefinitely on both the process and the stderr copier. `cmd.Stderr` is a `bytes.Buffer`, so os/exec allocates a pipe plus a copier goroutine that `Wait` joins — any grandchild inheriting that pipe holds `Wait` open even after git exits.

A blocking `reference-transaction` hook (git runs it at prepared/committed/aborted) reproduces the same wedge inside the child: git never exits on EOF, `tx.wait()` never returns, and the hook hangs with the transaction prepared and the `.lock` held — with the root context's cancel unable to break it.

Plausible rather than confirmed; I reasoned it from os/exec semantics and git's hook points without staging a blocking hook. A few seconds of `cmd.WaitDelay` covers it.

cmd/entire/cli/checkpoint/persistent_ref_update.go
```diff
@@ -35,28 +35,17 @@ func repositoryDirs(ctx context.Context, repo *git.Repository) (worktreeRoot, co
...
-	oldValue := strings.Repeat("0", newHash.HexSize())
-	if expectedHash != plumbing.ZeroHash {
-		oldValue = expectedHash.String()
-	}
-
-	cmd := exec.CommandContext(ctx, "git", "update-ref", refName.String(), newValue, oldValue)
-	cmd.Dir = repoRoot
-	cmd.Env = append(os.Environ(), "LC_ALL=C", "LANG=C")
-	output, err := cmd.CombinedOutput()
+	err := gitrepo.CompareAndSwapRef(ctx, repoRoot, refName, newHash, expectedHash)
```
Soph, line 26: Worth adding to the PR description: `casUpdateRef` is the shared entry point, so this reroutes more than the two OPF writers the description names — `casUpdateShadowBranchRef` (`checkpoint/shadow_ref.go`) and every `updatePersistentRef` caller go through it too, which puts it on the per-checkpoint shadow-branch write in the Stop hook.

Benchmarked at 60 iterations on a 400-commit repo: the old single `git update-ref <ref> <new> <old>` fork is 13.4ms/op, the new transaction plus the `symbolic-ref` probe is 22.0ms/op — +8.6ms per ref update, and `updatePersistentRef` can retry up to 16 times. Against the ~3s p50 hook floor that is under 1%, so not a blocker, just an unmentioned scope expansion onto a path this repo watches.

I checked the obvious way to drop the extra fork and it does not work: `symref-verify <ref>` alongside `update <ref> …` in one transaction fails with `fatal: prepare: multiple updates for ref '…' not allowed` on both backends. So the separate fork is a legitimate workaround, not an oversight.

Soph (commented):
Re-reviewed at fc410717c. All three findings from my last pass are fixed, and so is the `gitPlumbingEnv` note — the flock change is fully reverted (the diff against main for `internal/flock/` is now empty), the 2s bound is gone, and the SIGKILL path now closes stdin instead. I verified the new `cmd.Cancel` behaves as intended on git 2.50.1: with an explicit `start` in the stream, stdin EOF aborts at STARTED and PREPARED, leaving the ref unchanged, exit 0, and no leftover `.lock`. Deleting that block does fail `TestPreparedRefCASCancellationReleasesGitLock`.

One correction to my earlier review: I questioned whether holding the transaction prepared across the `symbolic-ref` fork was necessary. It is, and more so than the code says. `option no-deref` + `update <symref> <new> <old>` **succeeds** and converts the symref to a hash ref — git verifies the old value against the dereferenced target while writing the raw one — so without the probe the CAS compares against the wrong ref and destroys the symref. I also confirmed a prepared transaction blocks a concurrent `git symbolic-ref`. That ordering is load-bearing; worth a comment saying so, since it looks removable.

Four comments inline. The first is the one I would want changed before merge — it is the same shape as my previous two: the remedy fails worse than the bug.

Two minor ones not worth inlining. `finish` (ref_cas.go) reads `commit: ok` and then calls `wait()`; a cancel in that gap makes `wait` discard a nil `cmd.Wait()` and return `ctx.Err()`, so a write that succeeded is reported as a failure. Self-healing (local v1 already points at the rewritten tip and the next push stops at the trailer watermark) but it is a lies-about-the-outcome path in code whose job is not lying about ref outcomes. And `CompareAndSwapRef`'s `errors.Join(symbolic, tx.abort())` means a lock error from the abort makes `errors.Is(err, ErrRefLocked)` true as well, so a symbolic ref could be retried 16 times — check `ErrRefSymbolic` first, or don't join.

Things I checked and found correct, so they don't need defending: the injection defence (go-git's `Validate` rejects backslash and whitespace per component, so no C-style escape or extra line can reach git's quoted-arg parser — the newline test covers a family), error-classification wording against real git on both backends, transaction bookkeeping (no leaks, no double-wait, no pipe-fill deadlock, scanner limit fails into abort), the sha256-correct zero OID, and that both regression tests are genuinely red when the writers go back to `CheckAndSetReference`. `-race` clean, `mise run lint` 0 issues, no flakiness across repeated runs.

MuskanPaliwal:
I went into this thinking an extra lock would make the OPF update safer, but this review showed the opposite. Git's CAS already protects the write, and the extra lock only gave this path another way to hang. That was a useful correction to my mental model.

I've pushed the fixes and updated the description. One thing I'm still wondering about is whether native Git CAS should stay a small helper for direct refs, or eventually become the one path for every ref mutation we own.

@soph, could you take another look when you have a chance?

MuskanPaliwal:
Hi @Soph , a gentle reminder for this pr. Thanks :))

MuskanPaliwal:
Resolved the conflicts with `main` in [eba084710](https://github.com/entireio/cli/commit/eba0847108dfd5541f013ca30c2e7b04222edbc8).

I kept main's Git metadata consolidation. Checkpoint helpers now use `gitrepo.ResolveWorktreeMetadata` through `repository_metadata.go`; the old common-dir parser stays removed. Checkpoint writers also keep main's symlink-safe lock handling through `gitdir.OpenAt`, `MkdirAllNoSymlink`, and `flock.AcquireContextIn`.

OPF still bypasses the Entire flock. It already has an expected hash, and native Git checks that hash atomically. Lock contention retries with the same expected hash. A stale-value conflict stops without rebuilding, so we don't overwrite concurrent work. I adapted the tests to main's APIs and kept the existing CAS race and safety coverage.

I also checked Soph's earlier review fixes against the merge and found no reintroductions. That check did uncover remaining edge cases, fixed separately in [03effe5af](https://github.com/entireio/cli/commit/03effe5af):

- A directory at a ref path now returns an error instead of causing 16 pointless rebuilds.
- Late cancellation no longer reports failure after Git acknowledges `commit: ok`.
- Symbolic-ref rejection stops retries even if transaction cleanup also returns a lock error.

The added regressions cover files and reftable backends, plus normal, linked, and bare repositories where applicable.

`mise run check` passed, including race-enabled unit and integration tests, 56 Vogon canaries, and 4 Roger canaries. Pre-push lint passed too.

Soph (approved):
Verified at 03effe5af — everything from both rounds is addressed, and I re-ran the checks rather than reading the diff.

**The flock removal does not regress the fix.** `CASPersistentRef` is now `retryPersistentRefLockContention` + native CAS with no flock, and both lost-update regressions still pass, along with `TestAtomicSetV1Ref_ReportsActualCASConflictAsMoved` and the whole OPF rewrite suite. That was the one I wanted changed and it is the right shape: native CAS was always the correctness boundary, so the lock was cost without cover.

**The misclassification fix is stronger than what I suggested, and I was wrong about why it works.** I re-planted my three conditions and they now classify correctly (corrupt loose ref, unreadable ref, directory-at-ref-path all non-conflict; genuine concurrent delete still a conflict). But I had assumed the suffix check alone did it, and it does not — I dumped git's raw stderr and the directory case is **byte-identical** to genuine absence, so no string match could separate them. `refIsAbsent` is what does, by requiring positive confirmation of absence, and it correctly `Lstat`s through the filesystem storage first because go-git falls back to packed refs after a loose-ref read error. That residual would have survived my suggested fix. Nice catch.

**`WaitDelay`** is set to 3s with a comment naming the hazard, and `wait()` now short-circuits on `tx.committed` so a successful commit can no longer surface as a cancellation. Making `canceled` atomic is a correctness fix beyond what I raised — `cmd.Cancel` runs on another goroutine. The shared-`casUpdateRef` scope and the 13.4ms → 22.0ms cost are in the description now.

Gate: `gitrepo`, `checkpoint`, `flock` and `strategy` all pass under `-race`; `mise run lint` 0 issues.

One non-blocking leftover, the nit from last time: `CompareAndSwapRef` still returns `errors.Join(symbolic, tx.abort())`, so a lock error from the abort makes `errors.Is(err, ErrRefLocked)` true too and a symbolic ref could be retried 16 times. It needs an abort of a transaction we own to hit lock contention, so it is close to unreachable — worth a follow-up at most.

Approving.

PR #2213, opened 2026-09-01
fix(e2e-test): Isolate auth locks in spawned test binaries

### LINKED ISSUE #2201: auth-go lock dir isolation doesn't cover spawned binaries
Follow-up to #2173, which stopped auth-go's process lock landing in the developer's real user cache under `go test`.

That fix is in-process only. `tokenManagerLockDir` (`cmd/entire/cli/auth/refresh.go:156`) returns `testdirs.Dir("authlock")`, and `testdirs.Dir` returns `ok=false` in a subprocess (`internal/testdirs/testdirs.go:26-29`). The integration and e2e TestMains set `ENTIRE_CONFIG_DIR` and `XDG_CACHE_HOME` but not `HOME`, and on macOS `os.UserCacheDir()` ignores `XDG_CACHE_HOME` - so a spawned `entire` that reaches a token refresh still writes into the real `~/Library/Caches/auth-go`.

`userdirs` and `tokenstore` each pair `testdirs` with an env override (`ENTIRE_CONFIG_DIR`, `ENTIRE_TOKEN_STORE_PATH`). This one implements only half that pattern.

Suggested: check an `ENTIRE_AUTH_LOCK_DIR` env var ahead of `testdirs.Dir`, and set it in the integration and e2e TestMains alongside the existing isolation vars.

Note the practical exposure is unproven - no spawned binary was observed reaching the lock before failing at the network - but the mechanism is confirmed.

MuskanPaliwal:
Closes #2201

The existing auth-go lock isolation only protected code running inside a `go test` executable. Integration and E2E tests spawn a separately built `entire` binary, where `testing.Testing()` is false. That binary therefore fell back to the auth-go user-cache lock directory. On macOS, `os.UserCacheDir()` ignores `XDG_CACHE_HOME`, leaving the developer real `~/Library/Caches/auth-go` exposed.

## The fix

Add `ENTIRE_AUTH_LOCK_DIR` as an explicit override ahead of the existing in-process test fallback. The CLI package, integration, and E2E TestMains now point it at their test-owned directories so spawned binaries and hooks inherit the isolation.

When the variable is unset, production behavior is unchanged. Auth-go continues using its normal user-cache default.

The repository spawned-binary testing guidance now includes the new isolation variable.

## Verification

The focused regression test failed before the resolver change and passed afterward:

```text
TestTokenManagerLockDir_UsesExplicitOverride
before: tokenManagerLockDir() returned the testdirs fallback
after:  tokenManagerLockDir() returned the explicit override
```

A manual spawned-process check built the CLI with `go build`, seeded an expired login and refresh token, and ran `entire auth token` against a local `/oauth/token` endpoint twice:

- The positive control, with no override, created the auth-go lock under a fake macOS `HOME/Library/Caches/auth-go`.
- The fixed run created the same hashed lock under `ENTIRE_AUTH_LOCK_DIR`.
- The fixed run left its fake user-cache auth-go directory absent.
- The local endpoint received both refresh requests, proving both runs exercised the real refresh and lock path.

Automated checks passed:

```text
go test ./cmd/entire/cli/auth -count=1
mise run check
```

The resolver override and no-override branches also passed together as the blast-radius guard.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -299,10 +299,10 @@ using `entire` for real while tests run.
   `tokenstore.UseFileBackendForTesting(...)`.
 - **Spawned binaries are NOT covered**: `testing.Testing()` is false in a
   subprocess. The integration and e2e TestMains set `ENTIRE_CONFIG_DIR`,
-  `XDG_CACHE_HOME`, `ENTIRE_TOKEN_STORE=file`, `ENTIRE_TOKEN_STORE_PATH`, and
-  `ENTIRE_TEST_AUTH_STORE_FILE` process-wide so every spawned `entire` (and
-  every agent-invoked hook) inherits isolation. Any new harness that spawns
-  the real binary must do the same.
+  `XDG_CACHE_HOME`, `ENTIRE_AUTH_LOCK_DIR`, `ENTIRE_TOKEN_STORE=file`,
+  `ENTIRE_TOKEN_STORE_PATH`, and `ENTIRE_TEST_AUTH_STORE_FILE` process-wide so
+  every spawned `entire` (and every agent-invoked hook) inherits isolation.
+  Any new harness that spawns the real binary must do the same.
 - **Legacy auth store**: `auth.NewStore()` talks straight to the zalando
   keyring; packages whose tests can reach it need `keyring.MockInit()` in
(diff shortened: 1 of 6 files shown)


MuskanPaliwal:
Hey @Soph , would appreciate your review here. Thanks :)

PR #2219, opened 2026-09-01
Fix stale lock cleanup

### LINKED ISSUE #2197: Persistent-ref lock files accumulate one per checkpoint and nothing reaps them

Rohitkanithi:
Closes #2197.

## Summary

Adds cleanup support for stale lock files created under the git common dir:

- `entire-shadow-locks`
- `entire-persistent-ref-locks`

`entire clean --all` now discovers those lock files, reports them in dry-run output, and includes them in deletion summaries.

## Review follow-up

This update addresses the lock-safety review feedback:

- Revalidates Unix flock acquisitions against the current lock-file inode before returning from both blocking `Acquire` and bounded `AcquireContext` paths.
- Moves stale lock cleanup to run after shadow branch, session state, and checkpoint cleanup so normal cleanup operations use their locks first.
- Treats actively held lock files as skipped, not failed, so live sessions do not make `entire clean --all` exit non-zero just because a lock is busy.
- Adds an active-session guard for `entire clean --all` unless `--force` or `--dry-run` is used.
- Keeps lock-directory listing best-effort if one lock directory is unreadable.
- Uses platform-specific lock-file removal ordering for Unix and Windows.

## Tests

```bash
GOCACHE=/private/tmp/entire-go-build-cache go test ./cmd/entire/cli/internal/flock ./cmd/entire/cli/strategy
GOCACHE=/private/tmp/entire-go-build-cache go test ./cmd/entire/cli -run TestClean
GOCACHE=/private/tmp/entire-go-build-cache go test ./cmd/entire/...
git diff --check
```

### DIFF
--- a/cmd/entire/cli/clean.go
+++ b/cmd/entire/cli/clean.go
@@ -345,7 +345,7 @@ func runCleanAllWithItems(ctx context.Context, cmd *cobra.Command, force, dryRun
 	}
 
 	// Group items by type for display
-	var branches, states, checkpoints, redactCaches []strategy.CleanupItem
+	var branches, states, checkpoints, redactCaches, lockFiles []strategy.CleanupItem
 	for _, item := range items {
 		switch item.Type {
 		case strategy.CleanupTypeShadowBranch:
@@ -356,6 +356,8 @@ func runCleanAllWithItems(ctx context.Context, cmd *cobra.Command, force, dryRun
 			checkpoints = append(checkpoints, item)
 		case strategy.CleanupTypeRedactCache:
 			redactCaches = append(redactCaches, item)
+		case strategy.CleanupTypeLockFile:
+			lockFiles = append(lockFiles, item)
 		}
 	}
 
@@ -368,6 +370,7 @@ func runCleanAllWithItems(ctx context.Context, cmd *cobra.Command, force, dryRun
 		printSection(w, "Session states", cleanupItemIDs(states))
 		printSection(w, "Checkpoint metadata", cleanupItemIDs(checkpoints))
 		printSection(w, "Redaction cache", cleanupItemIDs(redactCaches))
+		printSection(w, "Lock files", cleanupItemIDs(lockFiles))
 		printSection(w, "Temp files", tempFiles)
 		printSection(w, "Stray agent temp files", orphanTemps)
 		printUnscannedNote(w, unscanned)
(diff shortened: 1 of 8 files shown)


Rohitkanithi:
Thanks again for the detailed review. I updated the implementation to address the lock-safety issue:

- added inode revalidation in Unix `flock`, including `Acquire`, `AcquireContext`, and `AcquireContextIn`
- covered both blocking `LOCK_EX` and polling `LOCK_NB` paths
- changed held lock files to be skipped instead of failed
- moved lock-file cleanup to run last
- made lock directory read errors best-effort
- added platform-specific remove ordering for Windows

I also reran:

```bash
go test ./cmd/entire/cli/internal/flock ./cmd/entire/cli/strategy
go test ./cmd/entire/cli -run 'TestClean'
go test ./cmd/entire/...
go build ./cmd/entire
git diff --check origin/main...HEAD
```

The PR is mergeable now and waiting on workflow approval / code owner review.

Rohitkanithi:
Closing this duplicate combined PR. The flock-only fix has been split out into #2273, and the remaining lock cleanup work will stay on the original discussion PR #2199.

PR #2220, opened 2026-09-01
Expand on os.Root coverage

pjbgf:
https://entire.io/gh/entireio/cli/trails/1202

Follow-up to #2180. A second pass over the migrated tree found five places still
doing filesystem work outside an anchor, and the diagnostic that would surface the
condition they now refuse. This brings them onto the same footing as the rest of
the tree.

## Pi's extension directory

`agent/pi/hooks.go` was the last registered agent whose hook-config surface had
not moved. It joined `.pi/extensions/entire/index.ts` onto the repo root and
handed the result to `os.ReadFile` / `os.MkdirAll` / `os.WriteFile` /
`os.RemoveAll`, each with a `//nolint:gosec` about where the path was *built*
rather than where it *resolves* — the justification `agent.HookConfigFile`'s doc
comment already calls out as the one that says nothing useful. A working tree
arrives by clone, so `.pi` is a component the repository supplies, and every one
of those four calls resolved through it.

Routed through `agent.OpenHookConfig`, which the other eight agents already use:
the base is the worktree root, `.pi/extensions/entire/index.ts` is a name inside
it, and directories are created with `MkdirAllNoSymlink` so a symlinked compon
is refused by name instead of followed. `.opencode/plugins/entire.ts` is the
structurally identical case — embedded TS file, worktree-resident, marker-chec
— and was already on this path; pi is now the same shape rather than the exception
beside it. Worth noting for the next integration: pi is auto-detected
(`DetectPresence` stats `.pi`), so it reaches these operations without the user
naming it, which is the argument for routing a new agent through the anchor by
default rather than trusting review to notice.

Two supporting pieces:

- **`osroot.RemoveAllNoSymlinks`.** `os.Root.RemoveAll` refuses a descent that
  would leave the root and unlinks a symlinked leaf rather than following it, but
  it says nothing about the components *above* the leaf — which is where `.pi`
  sits. This pins the parent first, then removes a single named component. Same
  relationship `OpenChild` has to a bare `Root.OpenRoot`.
- **`HookConfigFile.RemoveDir`.** Pi discovers extensions *by directory*, so
  removing only the file leaves an empty `.pi/extensions/entire` for pi to find
  with nothing to load. Deliberately not general: every other agent writes into a
  directory the agent itself owns, where `Remove` of the file is the correct
  uninstall and taking the parent would delete the user's own config with it.
  Refuses the case where the directory to remove would be the worktree root.

`agent.GeneratedHookFileState` — the path-based drift check that predates the
rooted `HookConfigFile.GeneratedState` — had pi as its only caller and is
deleted, so there is no unrooted variant left to pick up.

## Four sites beside it

- **`vercel.json`** (`setup.go`, `vercelconfig`). A working-tree file read wit
  `os.Stat` / `os.ReadFile` on a joined path. Now read through the worktree an
  as `LoadIn(root, name)`, with a 1 MiB cap: the file arrives with the checkout, so
  its size is not ours to choose, and everything downstream of the parse collapses
  to a single bool. In-repo symlinks are still followed on purpose — pointing
  `vercel.json` at a monorepo's shared config is a real setup, and this file i
  user's rather than Entire's — while one leaving the worktree is refused, which is
  the property the anchor exists to give.
- **`worktreedir.Name`**. Given the `VolumeName` pairing `SessionStore.Name` got in
  57b275ac6, which `ValidateSessionID` and gitrepo's alternates checks already
  `os.Root` rejects a drive-relative name one layer down (`filepathlite.IsLocal`),
  so this changes no behaviour today; it is closed because `Name` exists to an
  "is this inside the worktree?" independently of its caller, and
  `readWorktreeFileSafely` feeds it paths that arrive from the API. It was the
  remaining `IsAbs` in a containment role without the pairing.
- **The persistent-ref lock** (`checkpoint/persistent_ref_update.go`). The las
  `.git`-resident lock still opening by path rather than through the common dir's
  root. Now mirrors `shadowBranchLock`: `gitdir.OpenAt` + `MkdirAllNoSymlink` +
  `AcquireContextIn`. Git will not check a path out into `.git`, so this is
  consistency and defence in depth — but `openLockFileIn` exists for exactly t
  and one caller sitting outside it is how the next one gets written that way
- **`readCapped`** (`runner_gather.go`). Bounds the read, not just the result.
  These are working-tree files named by convention (`README.md`, `CLAUDE.md`), so
  their size is not ours to assume, and the caller has already said how much of one
  it will use. Reading a gigabyte to keep its first 4 KB is a cost with no ret

## `doctor` now reports symlinked agent directories

The condition these refusals produce was invisible after the fact. `enable` fa
loudly on it, but once a repo is enabled `HookConfigFile.Exists()` deliberately
reports a symlinked parent as *absent* — so `entire status` showed hooks missi
without saying why, and `entire clean` skipped the directory on the stated grounds
that doctor reports a symlinked agent directory. Nothing did.

`checkAgentDirSymlinks` runs beside `checkEntireDirSymlinks` and names the
outermost link, once per finding. Candidates come from the registry and the
scaffold templates rather than a list in the function. Unlike `.entire`, these
trees are largely the user's, so it examines only the components Entire itself
creates and writes through: `.claude/skills/my-own -> ../../shared/skills/my-o
is a real setup, and reporting it would train people to skip this section.

## Follow-up

`agentSymlinkCheckPaths` derives from `AllProtectedDirs()` plus the skill
templates, not from the hook-config relPaths, so the two agents whose config nests
below their top-level directory have their Entire-created intermediate directories
unchecked — a symlink at `.pi/extensions` or `.opencode/plugins` produces no
doctor output. The refusals cover both; this is reporting only.

---

> [!NOTE]
> **Medium Risk**
> Touches hook install/uninstall and recursive delete paths where symlink mistakes could write or destroy files outside the repo; changes are defensive with new doctor reporting and tests, but they affect every agent’s enable flow.
> 
> **Overview**
> This PR finishes **os.Root / worktree anchoring** for agent hook configuration and closes gaps where symlinked repo paths could still be followed or deleted outside the worktree.
> 
> **Pi** is migrated off raw `os.ReadFile` / `MkdirAll` / `WriteFile` / `RemoveAll` onto `agent.HookConfigFile` like the other agents. Uninstall uses new **`HookConfigFile.RemoveDir`** (backed by **`osroot.RemoveAllNoSymlinks`**) so the whole `.pi/extensions/entire` tree is removed without following a symlinked `.pi`. Path-based **`GeneratedHookFileState`** is removed; drift checks go through **`HookConfigFile.GeneratedState`**.
> 
> **`agent.HookConfigLocator`** (`HookConfigRelPath`) centralizes each agent’s config path for **`AllHookConfigRelPaths()`**, refactors all `OpenHookConfig` call sites, and powers a **build guard** ensuring every agent that opens a hook config declares its path.
> 
> **`entire doctor`** gains **`checkAgentDirSymlinks`**, listing symlinked or unreadable components on paths Entire creates (hook config paths from the locator plus skill scaffold parents), so “hooks missing” on `status` gets an explicit cause.
> 
> Additional hardening: **persistent ref flock** via `gitdir` + `flock.AcquireContextIn`; **`vercelconfig.LoadIn`** and setup’s Vercel probe through the worktree with a read cap; **`readCapped`** bounded reads; **`worktreedir.Name`** treats Windows drive-relative paths like other containment checks. Docs in **`CLAUDE.md`** are expanded for symlinked hook files/dirs and Pi auto-detection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 98bdf18e1e25b02ce9f5f32112eec79bd6503236. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -891,7 +891,12 @@ comments at each site say which case applies:
   symlink that *escapes* a root but follows one pointing elsewhere *inside* it,
   and an escaping one otherwise fails later with an opaque errno far from the
   cause. `entire doctor` reports what is already there
-  (`checkEntireDirSymlinks`). `.entire` **itself is refused too**: `entiredir`
+  (`checkEntireDirSymlinks` for `.entire`, `checkAgentDirSymlinks` for every path
+  Entire creates or writes on an agent's behalf — derived from
+  `agent.HookConfigLocator` and the scaffold templates, deliberately not from
+  `ProtectedDirs`, which both misses the levels Entire creates under an agent's
+  directory and includes directories it never writes to). `.entire` **itself is
+  refused too**: `entiredir`
   opens it as a checked child of the worktree root (`osroot.SharedChild`), which
   `Lstat`s before and `SameFile`s after. An earlier revision allowed it on the
   grounds that `os.OpenRoot` follows a symlinked root and an existing setup
@@ -912,11 +917,18 @@ comments at each site say which case applies:
   through it — and uninstall `RemoveAll` through it — without the user ever
   naming pi. Its uninstall is also the one caller of
   `HookConfigFile.RemoveDir`, because pi discovers extensions by directory, so
-  removing only the file would leave a half-uninstalled extension behind. The config FILE may still be a symlink — pointing
-  `.claude/settings.json` at a dotfile repo is a real setup — with one behaviour
-  change worth knowing: an **absolute** link there no longer resolves, since
-  `os.Root` refuses absolute symlinks unconditionally. A relative one inside the
-  worktree still works.
+  removing only the file would leave a half-uninstalled extension behind.
+
+  The config FILE is refused too, not just its parents: the merge READ would
+  otherwise pull the link target's contents into
+  what Entire then writes, and the write is a rename, which replaces the link
+  with a regular file rather than following it. Both happen silently, so
+  refusing is the legible version of the same outcome. Pointing
+  `.claude/settings.json` at a dotfile repo is still a real setup — it just has
+  to stay out of the repository (`git rm --cached` plus a .gitignore entry),
+  which is what makes it the developer's rather than the checkout's.
+  `checkAgentDirSymlinks` reports every one of these paths, so the condition is
+  named rather than showing up as hooks that mysteriously will not install.
 
   **A symlinked agent DIRECTORY is refused by every operation, not just the ones
   that create.** `os.Root` blocks a link that escapes the worktree and follows
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -778,7 +778,7 @@ to `os.ReadFile`/`os.WriteFile`/`os.MkdirAll`/`os.ReadDir`/`filepath.Walk`.**
 | `.entire` | `entiredir` | worktree root (`paths.WorktreeRoot`), cwd only when there is provably no repo |
 | git common dir | `gitdir` | `git rev-parse --git-common-dir`, absolutized |
 | the working tree | `worktreedir` | worktree root |
-| an agent's hook config | `agent.HookConfigFile` | worktree root (`.claude/`, `.cursor/`, `.gemini/`, `.github/hooks/`, `.factory/`, `.codex/`, `.opencode/plugin/`) |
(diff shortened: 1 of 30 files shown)


pjbgf:
@cursor review

gtrrz-victor (dismissed):

pjbgf:
@cursor review

gtrrz-victor (approved):

PR #2224, opened 2026-09-02
fix: refuse cross-host redirects in OAuth token-exchange POST

suhaanthayyil:
https://entire.io/gh/entireio/cli/trails/1206

## Summary
- `PostOAuthToken`'s request body carries `subject_token` (the caller's login JWT, or a jurisdiction-scoped exchange token) as a POST form field, not a header. Go's default `http.Client` only strips sensitive *headers* on a cross-host redirect (`net/http`'s `shouldCopyHeaderOnRedirect`) — it copies the request body unconditionally regardless of host change. A malicious or compromised token-exchange endpoint issuing a 307/308 redirect could therefore have the CLI hand its login JWT to a different origin.
- Both call sites that reach `PostOAuthToken` (`auth/cell_data_api.go`'s jurisdiction exchange, `internal/coreapi/cross_juris_transport.go`'s cross-jurisdiction exchange) construct their own `*http.Client` with no `CheckRedirect`, so this affected both.
- Fixed at the shared choke point (`PostOAuthToken` itself) rather than requiring every caller's client construction to opt in — mirrors the `rejectCrossHostRedirect` pattern `cmd/entire/cli/api/client.go` already uses for its own client.

## Verification
- Added `TestPostOAuthToken_RefusesCrossHostRedirect`: a genuine two-`httptest.Server` reproduction (origin redirects to attacker via 307) asserting the attacker server never receives `subject_token`.
- Confirmed this test **fails** against the pre-fix code (redirect silently followed, no error) before confirming it **passes** with the fix — mutation-verified, not just "test exists and is green."
- `mise run check` (fmt + lint + test:ci) passes.

## Test plan
- [x] `mise run check` green

### DIFF
--- a/internal/entireclient/httputil/oauth.go
+++ b/internal/entireclient/httputil/oauth.go
@@ -129,9 +129,12 @@ func PostOAuthToken(ctx context.Context, httpClient *http.Client, coreURL string
 	// copies the request body unconditionally — a POST body is not
 	// protected the way a bearer header is. Guard here, once, for every
 	// caller, rather than requiring each constructed *http.Client to set
-	// its own CheckRedirect. A shallow copy avoids mutating (and
-	// overriding any redirect policy already set on) a client instance
-	// callers may share elsewhere.
+	// its own CheckRedirect. The guard is applied to a shallow copy so a
+	// client instance callers may share elsewhere is not mutated; note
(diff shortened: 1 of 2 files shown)


suhaanthayyil:

pjbgf (approved):

PR #2225, opened 2026-09-02
fix: verify cross-repo checkpoint response identity before rendering

suhaanthayyil:
https://entire.io/gh/entireio/cli/trails/1207

## Summary
- `checkpoint_api_reader.go` decodes an entire-api cell's response identity fields (`repo_full_name`, `checkpointId`) but never checked them against what was actually requested (`r.ownerRepo`, the requested `checkpointID`).
- `loadDetail` cached the response under the *requested* id, and `Read()`/`ReadSessionMetadataAndPrompts()` relabeled whatever came back with the requested `checkpointID` rather than the server's own value — so a wrong-repo/wrong-checkpoint response (server bug, cache-key collision, authz bug) would render silently as belonging to the repo the caller asked about.
- Added `verifyResponseIdentity`, checked before the response is cached or rendered — mirroring `cell_target.go`'s existing self-check pattern (`resolveProcessingPlacement`'s `EqualFold` check) at the routing layer, brought up to the content layer.
- `Read()`/`ReadSessionMetadataAndPrompts()` now source `CheckpointID` from the verified server response, not the request param, so a future regression in the check can't silently paper over a real mismatch by relabeling.

## Verification
- Two-`httptest.Server`-style reproduction test with a mismatched `repo_full_name`/`checkpointId` in the response.
- Confirmed the test fails against the pre-fix code (mismatch silently accepted) before confirming it passes with the fix.
- `go build ./cmd/entire/...` and the package's own test suite (`go test ./cmd/entire/cli/ -run TestAPICheckpointReader`) pass.

## Test plan
- [x] New identity-mismatch reproduction test added and passing

### DIFF
--- a/cmd/entire/cli/checkpoint_api_reader.go
+++ b/cmd/entire/cli/checkpoint_api_reader.go
@@ -134,7 +134,11 @@ func (r *apiCheckpointReader) Read(ctx context.Context, checkpointID id.Checkpoi
 	// second one behind the first one's JSON field would quietly mislead every
 	// consumer of `explain --json`.
 	return &checkpoint.CheckpointSummary{
-		CheckpointID:     checkpointID,
+		// The server's own checkpointId, not the requested param: loadDetail has
+		// already verified the two agree, but sourcing this from info keeps a
+		// future regression in that check from silently papering over a real
+		// mismatch by relabeling foreign data with the id the caller asked for.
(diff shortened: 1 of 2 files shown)


suhaanthayyil:

pjbgf (approved):

PR #2226, opened 2026-09-02
fix: serialize shadow-branch migration against concurrent checkpoint writes

suhaanthayyil:
https://entire.io/gh/entireio/cli/trails/1208

## Summary
- `migrateShadowBranchToBaseCommit` moved a session's shadow branch to a new base-commit-derived name via an unlocked go-git `Reference` read, a non-CAS `SetReference`, and an unconditional `git branch -D` — entirely outside the flock+CAS discipline `writeCheckpoint`/`writeTask` use for the exact same refs (`withShadowBranchFlock` + `casUpdateShadowBranchRef`, `checkpoint/ephemeral.go`).
- Two sessions sharing a worktree's shadow branch is a documented, supported configuration (main agent + Task-tool subagent). A checkpoint write racing this migration could have its work silently orphaned (migration reads a stale hash and migrates that, missing the write) or the write's own CAS could fail against a branch the migration already deleted — checkpoint data loss with no error surfaced to the user.
- Adds `checkpoint.MigrateShadowBranchRef`, which takes both branches' flocks (stable sorted order, avoids deadlocking against a migration running the opposite direction) before touching either ref, CAS-creates the new branch, CAS-deletes the old one. `manual_commit_migration.go` now routes through it instead of the unlocked read/write/delete sequence.

## Verification
- Real-git, real-flock concurrency reproduction test (`TestMigrateShadowBranchRef_SerializesAgainstConcurrentCheckpointWrite`): a writer goroutine holds the shadow flock and advances the branch while migration is blocked waiting on the same lock, then confirms the migrated branch carries the writer's actual latest commit rather than a stale pre-race hash.
- `go build ./cmd/entire/...` and the targeted new tests pass.

## Test plan
- [x] New concurrency reproduction test added and passing
- [x] `go build` passes

### DIFF
--- a/cmd/entire/cli/checkpoint/persistent_ref_update.go
+++ b/cmd/entire/cli/checkpoint/persistent_ref_update.go
@@ -59,6 +59,27 @@ func casUpdateRef(ctx context.Context, repoRoot string, refName plumbing.Referen
 	return fmt.Errorf("git update-ref %s: %s: %w", refName, strings.TrimSpace(out), err)
 }
 
+// casDeleteRef atomically deletes refName through native Git's lock protocol,
+// via `git update-ref -d <ref> <old>`. The delete only succeeds if the ref
+// currently points at expectedHash; otherwise it returns ErrShadowRefBusy so
+// the caller can decide whether that's a benign race (someone else already
+// moved or removed it) or a real conflict.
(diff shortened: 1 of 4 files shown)


suhaanthayyil:

PR #2227, opened 2026-09-02
docs: disclose that pasted images bypass all redaction

suhaanthayyil:
https://entire.io/gh/entireio/cli/trails/1209

## Summary
- Pasted images/screenshots are externalized as raw binary blobs (`checkpoint/persistent.go`'s `writeAssets`) with no redaction call anywhere in the write path, and both OPF rewrite tree walkers explicitly skip the `assets/` subtree. This is a deliberate, defensible design choice — byte-level regex/entropy redaction can't safely touch binary image data without corrupting it — but `docs/security-and-privacy.md` didn't disclose it anywhere a user would actually look for it.
- The only related text was a low-level "JSONL skip rules" footnote framing the `type:image` skip as a false-positive-avoidance optimization for the text scanner, which reads as "we're careful not to over-flag images," not "images are never scanned by anything, ever."

## Change
- New, prominent "What Entire does NOT scan: pasted images and screenshots" section directly under "What Entire redacts automatically."
- Updated the "protecting sensitive data" checklist with an explicit "never paste a screenshot with secrets" bullet.
- Corrected the misleading JSONL-skip-rules footnote to point at the real explanation.

Docs-only change; no code behavior modified (deliberately — remediation depth like OCR-based scanning is a separate product decision, out of scope here).

## Test plan
- [x] Verified the disclosure accurately matches current code behavior (`writeAssets`, `manual_commit_opf_rewrite.go`)
- [ ] Full `mise run check` deferred to a follow-up pass (docs-only change, low risk)

### DIFF
--- a/docs/security-and-privacy.md
+++ b/docs/security-and-privacy.md
@@ -36,6 +36,16 @@ Entire automatically scans transcript and metadata content before writing it to 
 
 Detected secrets are replaced with `REDACTED` before the data is ever written to a git object. Of the six secret-detection passes above, the scanner layer (pass 2) is configurable — see [Choosing secret-scanner engines](#choosing-secret-scanner-engines) below — while the other five are **always on** and cannot be disabled. User-defined rules (inline `custom_redactions` and rule packs) add a seventh secret-detection pass that only runs when configured.
 
+### What Entire does NOT scan: pasted images and screenshots
+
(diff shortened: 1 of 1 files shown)


suhaanthayyil:

pjbgf (approved):

PR #2228, opened 2026-09-02
fix: don't let findInaccessiblePlugin re-find a binary via a relative PATH entry

suhaanthayyil:
https://entire.io/gh/entireio/cli/trails/1210

## Summary
- `exec.LookPath` deliberately refuses (`exec.ErrDot`) to resolve a binary found via a relative `\$PATH` entry — Go's fix for the classic "dot in PATH" RCE class.
- `resolvePlugin`'s fallback for `LookPath` failures, `findInaccessiblePlugin`, didn't distinguish that refusal from a genuine "exists but lacks the executable bit" case: it re-walked the same PATH entries with a plain `os.Stat` and returned `found=true` for the exact file `LookPath` just refused.
- `runPlugin` then execs it via a path containing a separator, which bypasses Go's `exec.Command` ErrDot re-check (that only fires for a bare name with no separator).
- **Net effect:** any relative `\$PATH` entry (direnv `PATH_add`, a project-local `bin/`, `node_modules/.bin`-style patterns — all common, not exotic) plus an untrusted repo containing `<relative-dir>/entire-<name>` gave silent code execution with the user's full privileges on `entire <name>`.

## Fix
`findInaccessiblePlugin` now only scans absolute PATH entries, mirroring `exec.LookPath`'s own refusal for the relative case.

## Verification
- Real reproduction (not a mock): a real executable shebang'd binary planted under a relative `bin/` directory, `PATH` set to include it, `MaybeRunPlugin` invoked directly, and a marker-file assertion proving the planted binary was never executed.
- **Mutation-verified**: temporarily reverted just the one-line fix, reran the new test, confirmed it fails exactly as expected (RCE reproduced — "planted binary via relative PATH entry WAS EXECUTED"), then restored the fix and confirmed the test passes again.
- Full package test suite for the affected area (`TestResolvePlugin*`, `TestRunPlugin*`, `TestMaybeRunPlugin*`) passes.

## Test plan
- [x] Verified test fails without the fix (mutation check) — confirmed actual RCE reproduction
- [x] Targeted package test suite green

### DIFF
--- a/cmd/entire/cli/plugin.go
+++ b/cmd/entire/cli/plugin.go
@@ -126,9 +126,25 @@ func resolvePlugin(rootCmd *cobra.Command, args []string) (binPath string, plugi
 // given name. Only meaningful after exec.LookPath has already failed —
 // indicates the file exists but lacks the executable bit (or the
 // equivalent platform-specific accessibility).
+//
+// Only ABSOLUTE PATH entries are scanned. A relative entry (including bare
+// "." and the empty string, which POSIX treats as the current directory) is
+// skipped outright, even if it points at a real, inaccessible file. That
+// mirrors exec.LookPath's own refusal: when LookPath finds a binary via a
+// relative PATH entry it deliberately returns exec.ErrDot rather than
+// resolving it — the fix for the classic "dot in PATH" RCE class, where a
+// planted binary in an untrusted working directory silently executes with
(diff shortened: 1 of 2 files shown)


suhaanthayyil:

pjbgf (approved):

PR #2230, opened 2026-09-02
fix: require the worktree-suffixed shadow-branch shape for unattended deletion

suhaanthayyil:
https://entire.io/gh/entireio/cli/trails/1212

## Summary
- `CleanupPushedShadowBranches` runs automatically after every successful push, with no confirmation, and trusted `shadowBranchPattern` (`entire/<7+ hex>(-<6 hex>)?`) directly for deletion eligibility.
- The bare, unsuffixed half of that pattern (`entire/<hex>`, no dash) is also a plausible human branch-naming convention — short-SHA branch names are an ordinary convention, and Entire reserves no documented namespace. A branch like `entire/1234567` with no session-state entry (a hand-created branch never has one) was silently, permanently force-deleted (`git update-ref -d`, no merge-check) the next time `entire` pushed anything in that repo.
- Adds `isAutoDeletableShadowBranch`, requiring the worktree-hash suffix that `checkpoint.ShadowBranchNameForCommit` always appends (even for the main worktree — `HashWorktreeID` hashes an empty worktree ID to a real 6-hex value, never an empty one). That shape is the actual, unspoofable-in-practice output of the one function that mints shadow branches — a much stronger ownership signal than bare name matching.
- Wired into both `CleanupPushedShadowBranches` (the eligibility filter) and `DeleteShadowBranchesIfUnchanged` (its own independent guard, defense-in-depth).
- Old-format branches remain listed and deletable through the interactive `entire clean --all` path, where a human sees the branch name and confirms before anything is deleted — only the unattended, no-confirmation path is narrowed here.

## Verification
- Real-git reproduction test (`TestCleanupPushedShadowBranches_HumanBranch_Survives`): plants a human-plausible `entire/abc1234` branch alongside a genuine orphaned shadow branch, confirms only the genuine one is deleted.
- **Mutation-verified in two steps**: reverting only the eligibility-filter gate still passed (the independent `DeleteShadowBranchesIfUnchanged` guard caught it — confirms the defense-in-depth is real, not just written); reverting both gates together reproduced the original bug exactly (2 branches deleted instead of 1, including the human one); restored the fix and confirmed the full targeted test suite (every pre-existing shadow-branch/cleanup test) passes with no regressions.
- `go build ./cmd/entire/...` passes.
- Full `mise run check` deferred to a follow-up pass to keep local CI load down.

## Test plan
- [x] Verified test fails without the fix (mutation check, both gates reverted)
- [x] Full targeted package test suite green, no regressions

### DIFF
--- a/cmd/entire/cli/strategy/cleanup.go
+++ b/cmd/entire/cli/strategy/cleanup.go
@@ -60,18 +60,87 @@ type CleanupResult struct {
 //
 // The pattern requires at least 7 hex characters for the commit, optionally followed
 // by a dash and exactly 6 hex characters for the worktree hash.
+//
+// This pattern is name-shape ONLY -- matching it is not proof Entire created the
+// branch. In particular the "old format" half (bare "entire/<hex>", no worktree
+// suffix) is also a plausible human branch-naming convention (e.g. tracking a
+// short commit SHA), and nothing here is namespace-reserved. Use this broad
+// pattern only for listing/reporting paths that a human confirms before any
+// deletion happens (ListShadowBranches, ListAllItems, `entire clean --all`'s
+// interactive picker). For unattended, no-confirmation deletion, use
+// isAutoDeletableShadowBranch instead -- see its doc comment.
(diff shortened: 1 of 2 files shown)


suhaanthayyil:

pjbgf (approved):

PR #2231, opened 2026-09-02
fix: stop entire investigate fix from following an untrusted file path

suhaanthayyil:
https://entire.io/gh/entireio/cli/trails/1213

## Summary
- `manifest.FindingsDoc` is an absolute path decoded from a JSON manifest file on disk, so `RunFix` reading it directly via `os.ReadFile` let whoever writes the manifest choose which file on the filesystem gets read and fed into the launched fix agent's prompt.
- `entire investigate` already spawns agents with sandbox/approval bypassed and can ingest untrusted seed content (`--issue-link`, a PR diff) — a prompt-injected investigate agent could plant a manifest naming an arbitrary path (SSH keys, another repo's secrets) and have its contents read into the next `entire investigate fix` invocation's prompt.
- Two sibling call sites (`cmd.go`'s `--findings` listing, `show.go`) already route through `ReadRunFindings`, which resolves the manifest's (validated) `RunID` as a name inside the investigation store rather than following the untrusted path — `fix.go` was the one call site that still read `FindingsDoc` directly.
- `RunFix` now does the same: adds a `StateStore` field to `FixDeps`, production wiring opens one with a soft fallback (no findings section rather than a hard failure, matching `RunShow`'s existing behavior), and findings resolution goes through `ReadRunFindings(deps.StateStore, manifest.RunID)`.
- Also updates `FindingsDoc`'s doc comment in `manifest.go`: it is display-only now, never read back directly by `show` or `fix`.

## Verification
- `TestRunFix_IgnoresFindingsDoc_ArbitraryFileRead`: plants a decoy file with a marker string and a manifest whose `FindingsDoc` names it, asserts the marker never reaches the launched agent's prompt.
- **Mutation-verified**: temporarily restored the old direct `os.ReadFile(manifest.FindingsDoc)` behavior, confirmed the test genuinely fails (decoy content appears in the launched prompt exactly as the finding describes), then restored the fix and confirmed the full `investigate` package test suite passes with no regressions.
- `go build`/`go vet` pass. Full `mise run check` deferred to a follow-up pass to keep local CI load down.

## Test plan
- [x] Verified test fails without the fix (mutation check, decoy content genuinely leaked)
- [x] Full `investigate` package test suite green, no regressions


suhaanthayyil:

pjbgf (approved):

PR #2232, opened 2026-09-02
fix: serialize session-state clear against concurrent MutateSessionState

suhaanthayyil:
https://entire.io/gh/entireio/cli/trails/1214

## Summary
- `clearSessionState` (and its exported wrapper `ClearSessionState`, used by `entire doctor`) deleted a session's state file with no lock at all, while every other mutation of the same file goes through `MutateSessionState`'s per-session gate.
- A concurrent, properly-locked write (e.g. a PostToolUse hook for the same session) landing in the gap between a caller's "safe to clear" decision and the actual delete was silently destroyed — the file the write just produced gets deleted out from under it, with nothing surfacing the loss.
- `initializeSession` documents and fixes the identical hazard class elsewhere in this file ("take the gate, then re-check under lock"); this closes the same gap for the clear path.
- `clearSessionState` now acquires `sessionID`'s gate before clearing.
- `CondenseSessionByID`'s `clearAfter` path — the concrete, routine-traffic instance of this race (no shadow branch, condensation decides to clear state only) — now clears via a new `clearSessionStateLocked` helper from inside its own already-locked mutation closure, so the decision and the delete are atomic with respect to other writers rather than racing across an unlocked gap.
- `Reset`/`ResetSession`/doctor's `ClearSessionState` all call the same hardened `clearSessionState`, so they're covered with no additional changes.

## Verification
- Real-goroutine concurrency reproduction (`TestClearSessionState_SerializesAgainstConcurrentMutation`): a writer holds the real gate via `MutateSessionState` and blocks mid-mutation; `clearSessionState` must block until it releases, not run concurrently.
- **Mutation-verified**: temporarily reverted `clearSessionState` to skip gate acquisition (matching the exact pre-fix code path), confirmed the test genuinely fails (clear returns immediately while the writer is still mid-flight), then restored the fix and confirmed the full `strategy` package test suite passes with no regressions.
- `go build` passes. Full `mise run check` deferred to a follow-up pass to keep local CI load down.

## Test plan
- [x] New concurrency reproduction test added and passing
- [x] Full `strategy` package test suite green, no regressions

### DIFF
--- a/cmd/entire/cli/doctor.go
+++ b/cmd/entire/cli/doctor.go
@@ -409,10 +409,42 @@ func promptSessionAction(ss stuckSession) (string, error) {
 	return action, nil
 }
 
+// sessionLockNoticeDelay is how long discardSession waits before telling the
+// user it is blocked on a session's state lock. Short enough that a real stall
+// is announced promptly, long enough that the uncontended case (the common
+// one) stays silent.
+const sessionLockNoticeDelay = time.Second
+
+// clearSessionStateWithProgress clears sessionID's state, announcing the wait
+// on errW if the per-session state lock is not free within notifyAfter.
+//
+// The wait itself is deliberate and must not be shortened here:
+// strategy.ClearSessionState takes the same gate every mutation of that state
+// uses, because deleting the file out from under an in-flight write destroys
+// it. But the acquire is unbounded -- only the TurnStart hook opts into a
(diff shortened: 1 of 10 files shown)


suhaanthayyil:

pjbgf (dismissed):

pjbgf (approved):

PR #2233, opened 2026-09-02
fix: warn on every entire investigate run that agent sandboxing is off

suhaanthayyil:
https://entire.io/gh/entireio/cli/trails/1215

## Summary
- Every agent `entire investigate` spawns runs with sandbox/approval checks unconditionally disabled — Claude Code's `--permission-mode bypassPermissions` and Codex's `--dangerously-bypass-approvals-and-sandbox` are not conditioned on any flag.
- The only existing warning (`confirmUntrustedIssueSeed`) covers just the `--issue-link` case, framed specifically around external GitHub content. A plain `entire investigate "..."` run — including on a freshly cloned, not-yet-reviewed repo, the tool's own documented use case — gave the operator zero signal that the spawned agent could execute arbitrary commands unsandboxed, restrained only by its own prompt text.
- Adds a non-blocking informational notice, printed once per run (fresh, without an issue-link seed, and on `--continue`), naming both bypass flags concretely.

## Why not a confirmation gate like `--issue-link`'s
That mechanism exists for the genuinely higher-risk case (remote-attacker-controlled content driving an unsandboxed agent). Turning `entire investigate`'s ordinary use into a prompted confirmation would just train operators to reflexively accept it — this closes the "totally silent" gap without adding friction to the tool's core function. Happy to revisit the UX call if a maintainer feels this should gate harder.

## Verification
- Real command-execution test (`TestNewCommand_FreshRunWarnsSandboxBypass`, via `investigate.NewCommand` + `cmd.Execute()`) asserting the notice appears on stderr for a plain run with no `--issue-link`.
- **Mutation-verified**: temporarily removed the notice call, confirmed the test fails (stderr genuinely empty, matching the exact pre-fix silence), then restored the fix and confirmed the full `investigate` package test suite passes with no regressions.
- `go build` passes. Full `mise run check` deferred to a follow-up pass to keep local CI load down.

## Test plan
- [x] Full `investigate` package test suite green, no regressions

### DIFF
--- a/cmd/entire/cli/investigate/cmd.go
+++ b/cmd/entire/cli/investigate/cmd.go
@@ -373,6 +373,27 @@ func runInvestigate(ctx context.Context, cmd *cobra.Command, args []string, f ru
 	return runFresh(ctx, cmd, args, f, deps)
 }
 
+// sandboxBypassNotice is printed once per investigate run (fresh or
+// resumed) that reaches agent spawn without going through the stronger,
+// blocking confirmUntrustedIssueSeed gate. Every agent this package spawns
+// runs with sandbox and approval checks unconditionally disabled --
+// Claude Code's --permission-mode bypassPermissions and Codex's
+// --dangerously-bypass-approvals-and-sandbox are not conditioned on
+// --issue-link or any other flag. Only the launched agent's own prompt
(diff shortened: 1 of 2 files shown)


suhaanthayyil:

pjbgf (approved):

PR #2234, opened 2026-09-02
fix: don't let a tool call's own payload spoof the image-object redaction skip

suhaanthayyil:
https://entire.io/gh/entireio/cli/trails/1216

## Summary
- `collectJSONLReplacements`'s `shouldSkipJSONLObject` check (skip an entire subtree when an object's `"type"` field starts with `"image"` or equals `"base64"`) applied unconditionally anywhere in a transcript line's JSON tree — including inside a `tool_use` block's own `"input"` or a `tool_result`'s `"content"`/`"output"`, fields whose shape is controlled by whatever tool declared them. For an MCP tool, that's a third party, not Entire's transcript writer.
- A tool call argument or response deliberately shaped as `{"type":"base64", ...}` hid an arbitrarily large sibling secret behind a single spoofed tag, from every scanning layer (regex, entropy), then shipped unredacted to a pushed branch.
- Threads a `toolPayload` flag through the walk, set once the recursion enters a `tool_use` block's `input` or a `tool_result` block's `content`/`output`, sticky for all descendants. `shouldSkipJSONLObject` is honored only outside `toolPayload`. Genuine externalized image assets (pasted screenshots) are message content blocks, not tool call input/output, so the documented, intentional image-asset skip (`docs/security-and-privacy.md`) is unaffected.

## Deliberately scoped — what this does NOT touch
`shouldSkipJSONLField`'s id/signature-suffix and path-family skips are **not** touched here. Several are load-bearing for real built-in tool calls (Read/Grep's own `file_path`/`path` arguments live inside `tool_use.input`), and narrowing them without full visibility into every supported agent's wire format (Claude Code, Codex, Copilot CLI, Gemini, Factory Droid, Pi each have their own shape) risks reintroducing the false positives they exist to prevent. I verified this against real fixture data (`cmd/entire/cli/transcript/compact/testdata/*.jsonl`) and confirmed e.g. Copilot's own hook payload legitimately nests `sessionId`/`cwd` inside an `input` object unrelated to tool calls. Flagging this as follow-up rather than guessing at a broader fix under time pressure.

## Verification
- Genuine reproduction (`TestCollectJSONLReplacements_ToolPayloadCannotSpoofImageSkip`): a real secret wrapped in a spoofed `type:base64`/`type:image` object inside both a `tool_use.input` and a `tool_result.content`, asserting it is now caught.
- **Mutation-verified**: temporarily reverted the `toolPayload` gate, confirmed the test fails (both evasion shapes reproduce exactly as described, secret silently unredacted), then restored the fix.
- Full `redact` package test suite passes with no changes to existing behavior — including `TestShouldSkipJSONLObject_RedactionBehavior` (the genuine image-skip case) and `TestShouldSkipJSONLField` (the field-name skip cases), both unmodified and still green.
- `go build ./...` passes across the whole module. Full `mise run check` deferred to a follow-up pass to keep local CI load down.

## Test plan
- [x] Verified test fails without the fix (mutation check, both evasion shapes reproduced)
- [x] Full `redact` package test suite green, no regressions to legitimate skip behavior

### DIFF
--- a/redact/redact.go
+++ b/redact/redact.go
@@ -980,26 +980,49 @@ func isSingleJSONValue(dec *json.Decoder) bool {
 // string replacements via the supplied per-leaf redactor. JSONLContent
 // passes String; the OPF-enabled flow passes a closure that combines the
 // regex layers with cached batched OPF spans.
+//
+// The walk also tracks toolPayload: whether the current position is inside
+// a tool call's own schema-controlled data (a "tool_use" block's "input",
+// or a "tool_result" block's "content"/"output"). Field names and shapes in
+// there come from whatever tool declared them — for an MCP tool, that's a
+// third party, not Entire's own transcript writer — so shouldSkipJSONLObject
+// (a whole-subtree skip keyed on a bare "type" value) is honored only
+// outside toolPayload: a tool argument or result deliberately shaped as
+// {"type":"base64", ...} would otherwise hide an arbitrarily large sibling
+// secret behind a single spoofed tag. Genuine externalized image assets
+// (pasted screenshots) are message content blocks, not tool call
+// input/output, so this does not affect that already-documented,
+// intentional skip (see docs/security-and-privacy.md).
+//
+// shouldSkipJSONLField's per-leaf skips (id/signature suffixes, path-family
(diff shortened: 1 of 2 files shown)


suhaanthayyil:

PR #2241, opened 2026-09-02
refactor(gitrepo): centralize metadata resolution (split part 1)

MuskanPaliwal:
Refs #2190.

PR #2190 moved every Git metadata consumer at once. The review exposed two separate jobs: define one owner for the metadata, then migrate its callers.

This is split part 1. It adds the resolver and changes only repository opening. Checkpoint, session, strategy, hooks, and other consumers stay unchanged here.

## Files and ownership

Three production files and four test files change. All paths below are relative to `cmd/entire/cli/gitrepo/`.

| File | Before | After |
| --- | --- | --- |
| `metadata.go` | No file | Resolves the three metadata facts |
| `repository.go` | Opening rediscovered paths | Resolves and passes metadata |
| `reftable.go` | Errors looked like absence | Returns inspection errors |
| `metadata_guard_test.go` | No guard | Blocks new duplicate traversal |
| `metadata_test.go` | No resolver tests | Covers real and broken layouts |
| `repository_test.go` | Tested the old path | Covers fallback and failures |
| `reftable_test.go` | Tested boolean detection | Covers inspection errors |

The call path changes like this:

```text
Before

OpenPath
  ├── resolve .git and commondir while opening
  ├── resolve them again while checking alternates
  └── fall back without one validated metadata result

After

OpenPath
  └── ResolveWorktreeMetadata
        ├── GitDir
        ├── CommonDir
        └── WorktreeID
              │
              ├── repository storage
              ├── alternates inspection
              └── reftable inspection
```

## One resolver for three facts

`gitrepo.ResolveWorktreeMetadata` resolves an explicit worktree root into:

- `GitDir`
- `CommonDir`
- `WorktreeID`

The resolver takes a worktree root. It does not discover one from CWD, read Git environment variables, run Git, cache results, or choose a caller's policy.

Resolution is all-or-error. Malformed or unreadable metadata stops the open.

`gitrepo.OpenPath` now resolves metadata once and passes those facts into repository opening, object-alternates detection, and reftable inspection.

A bare-repository open is tried only when worktree metadata does not exist.

## Symlinks and `..` keep their filesystem meaning

Lexically cleaning a path through a symlink can change where `..` leads. The resolver checks filesystem identity before it returns a cleaned path.

Linked-worktree registrations may use physical aliases. They must still resolve to an immediate child of `CommonDir/worktrees`.

Inspection errors now reach the caller. They no longer mean "feature absent." This applies to reftable inspection and malformed `.git` layouts.

## Fixes a hang that is already released

`OpenPath` on `main` identifies a `.git` entry by reading it: `os.Stat`, then `os.ReadFile`. That works for a directory or a regular file. For three other entry types the read never finishes.

| `.git` is | `main` | This PR |
| --- | --- | --- |
| A FIFO with no writer | Blocks until a writer opens it | Refused on type |
| A symlink to such a FIFO | Blocks until a writer opens it | Refused on type |
| A symlink to `/dev/zero` | Reads forever, memory grows | Refused on type |

The resolver checks the type first. It opens only a directory or a regular file, and refuses anything else by name.

None of these arrive by clone. Git refuses a `.git` path component in a tree at any depth, so each one needs local write access to the worktree. This is a robustness fix, not a vulnerability.

## `gitdir:` strictness now matches Git

The legacy parser accepted pointer files that Git rejects. It cut the prefix at `gitdir:` instead of `gitdir: `, then trimmed whitespace from the remainder. That let `gitdir:target`, ` gitdir: target`, and `gitdir:\ttarget` all resolve. Git requires the single space and trims nothing.
The fixture covers seven `.git` pointer spellings. The new resolver agrees with Git on all seven. The legacy parser agreed on three.
This changes behavior. A `.git` file that worked only because Entire was more permissive than Git now fails, and every other Git tool was already failing it.

## Stop new duplicate traversal

A source guard scans production Go across the repository. It rejects new `.git`, `gitdir:`, and `commondir` traversal outside `gitrepo.ResolveWorktreeMetadata`.
The guard lists the existing callers that later PRs must migrate:

- checkpoint metadata
- session metadata
- strategy and hooks
- remaining product consumers

The exception list records each function and the one `git rev-parse` flag it may use. If that function starts using another metadata flag, the guard fails. The test also fails when a listed exception no longer exists.

A temporary multi-file repository test runs the complete scanner. It proves that the scanner follows a package constant into another file, scans production code under `internal/`, and ignores a parameter named `gitDir`.

Positive controls cover rooted filesystem APIs, package constants, `Readlink`, aliased imports, and legacy resolver calls.

## Verification

- `mise run check`
  - formatting passed
  - lint passed with 0 issues
  - Vogon canary passed 56/56
  - external deterministic canary passed 4/4
- `git diff main --check`
- Real Git fixtures cover ordinary, linked, bare, separate-Git-dir, submodule, symlink-alias, moved-registration, alternates, and reftable layouts.
- On the seven `.git` pointer spellings in the fixture, the new resolver matches Git on 7/7. The legacy parser matched 3/7.
- Broken fixtures verify fail-closed repository opening and typed absence.

No paid real-agent E2E tests were run.

## Deliberately not included

Checkpoint, session, strategy, hooks, and other product consumers still use their existing paths. Separate PRs will migrate them without mixing their error and fallback policies into this resolver change. The last migration will remove the legacy resolvers.

### DIFF
--- a/cmd/entire/cli/gitrepo/metadata.go
+++ b/cmd/entire/cli/gitrepo/metadata.go
@@ -3,11 +3,23 @@ package gitrepo
 import (
 	"errors"
 	"fmt"
+	"io"
 	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
+	"unicode/utf8"
 )
+
+// maxMetadataPointerSize matches git's own refusal threshold for .git pointer
+// files (setup.c caps at 4*PATH_MAX with "too large to be a .git file").
+// Anything larger is not repository metadata, and reading it unbounded turns a
+// sparse pointer file of near-zero disk size into gigabytes of allocation.
+const maxMetadataPointerSize = 4 * 4096
+
+// maxMetadataErrorPathLen bounds how much of a pointer-derived path an error
+// message echoes; a hostile pointer must not become a multi-kilobyte error.
+const maxMetadataErrorPathLen = 256
 
 // ErrWorktreeMetadataNotFound reports that an explicit worktree root has no
 // .git entry. Repository opening uses it to distinguish a possible bare
@@ -84,11 +96,11 @@ func resolveWorktreeGitDir(worktreeRoot string) (string, error) {
 		return "", fmt.Errorf(".git entry at %s is neither a directory nor a regular file", gitEntry)
 	}
 
-	content, err := os.ReadFile(gitEntry) //nolint:gosec // path is anchored at the explicit worktree root.
+	content, err := readMetadataPointerFile(".git", gitEntry)
 	if err != nil {
-		return "", fmt.Errorf("read .git file at %s: %w", gitEntry, err)
+		return "", err
 	}
-	pointer, ok := strings.CutPrefix(strings.TrimRight(string(content), "\r\n"), "gitdir: ")
+	pointer, ok := strings.CutPrefix(strings.TrimRight(content, "\r\n"), "gitdir: ")
 	if !ok {
 		return "", fmt.Errorf("parse .git file at %s: missing gitdir prefix", gitEntry)
 	}
@@ -116,11 +128,11 @@ func resolveWorktreeCommonDir(gitDirPath string) (string, error) {
 		return "", fmt.Errorf("commondir entry at %s is not a regular file", commonFile)
 	}
 
-	content, err := os.ReadFile(commonFile) //nolint:gosec // path is inside the validated per-worktree Git directory.
+	content, err := readMetadataPointerFile("commondir", commonFile)
 	if err != nil {
-		return "", fmt.Errorf("read commondir file at %s: %w", commonFile, err)
+		return "", err
 	}
-	pointer := strings.TrimRight(string(content), "\r\n")
+	pointer := strings.TrimRight(content, "\r\n")
 	if pointer == "" {
 		return "", fmt.Errorf("parse commondir file at %s: empty value", commonFile)
 	}
@@ -161,6 +173,38 @@ func resolveWorktreeID(gitDirPath, commonDir string) (string, error) {
 	return id, nil
 }
 
+// readMetadataPointerFile reads a .git or commondir pointer file with git's own
+// size cap, so a hostile pointer file cannot force an unbounded allocation.
+func readMetadataPointerFile(kind, path string) (string, error) {
+	file, err := os.Open(path) //nolint:gosec // path is anchored at validated repository metadata.
+	if err != nil {
+		return "", fmt.Errorf("read %s file at %s: %w", kind, path, err)
+	}
+	defer func() { _ = file.Close() }()
+
+	content, err := io.ReadAll(io.LimitReader(file, maxMetadataPointerSize+1))
+	if err != nil {
+		return "", fmt.Errorf("read %s file at %s: %w", kind, path, err)
+	}
+	if len(content) > maxMetadataPointerSize {
+		return "", fmt.Errorf("parse %s file at %s: too large to be a %s file", kind, path, kind)
+	}
+	return string(content), nil
+}
+
+// elideMetadataPath bounds a pointer-derived path for error display. Elision
+// backs up to a rune boundary so the message stays valid UTF-8.
+func elideMetadataPath(path string) string {
+	if len(path) <= maxMetadataErrorPathLen {
+		return path
+	}
+	cut := maxMetadataErrorPathLen
+	for cut > 0 && !utf8.RuneStart(path[cut]) {
+		cut--
+	}
+	return fmt.Sprintf("%s… (%d bytes elided)", path[:cut], len(path)-cut)
+}
+
 // Preserve the pointer spelling until validation so a missing component cannot
 // disappear through lexical cleaning before the filesystem sees it.
 func resolveMetadataPath(base, value string) string {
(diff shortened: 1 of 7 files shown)


MuskanPaliwal:
Hey @Soph, this is part 1 of the split from #2190. It introduces `gitrepo.ResolveWorktreeMetadata` as the single owner for resolving `GitDir`, `CommonDir`, and `WorktreeID`, and updates `gitrepo.OpenPath` to use that result. The checkpoint, session, strategy and hooks, and remaining-consumer migrations are intentionally left for separate PRs so each caller’s error and fallback policy can be reviewed independently.

Soph:
Exactly the slice I asked for, and fast. Reviewed at `d3df6e287`: no conflicts, builds, lint clean, tests green. I built the 13 layouts that matter (linked worktree, relative `commondir`, submodule, `--separate-git-dir`, `--shared`, reftable, sha256) and checked them through to blob contents: byte-identical to main, both package invariants hold, subprocesses per open still zero.

The blocker is the guard, not the resolver.

`metadata_guard_test.go` misses the most likely way the next duplicate gets written: hoist the token to a package-level `const` and it walks straight past, because the walker only inspects `fn.Body`. The guard already solves this for `.git` at `:81-82` by collecting string-const identifiers, and `repository.go:21` is itself `const gitDir = ".git"`, so the house style is the bypass. Also unchecked: `--absolute-git-dir`/`--git-dir`/`--show-toplevel` at `:106`, `OpenRoot` in the verb list at `:236`, and everything outside `cmd/entire/cli`. I confirmed each by dropping a probe in and watching it pass.

That matters because part 1 alone takes 3 `gitdir:` parsers to 4 (`metadata.go` sits alongside `repository.go:235`, kept for three Codex call sites). Fine for a strangler fig, and I am not asking you to migrate consumers here, but it makes the guard load-bearing across parts 2 to 4 rather than a nice-to-have.

It also false-positives in the other direction: a function that just takes a `gitDir` parameter and reads an unrelated file gets flagged. `strategy` and `checkpoint` do that constantly, and they are parts 2 and 3.

Two small things. Nothing tests `ResolveWorktreeMetadata("")` and the empty-root guard at `metadata.go:29-31` is the only thing stopping it resolving the process CWD's repo; I deleted the guard and the suite still passed. And worth putting the strictness win in the description: your parser agrees with real `git rev-parse` on 7 of 7 pointer spellings where main agreed on 3 of 7, so this is a correctness fix, not only a refactor.

MuskanPaliwal:
@Soph , I have addressed the feedback in [64f3ba8](https://github.com/entireio/cli/pull/2241/commits/64f3ba8dc164dfde5abcb6c685e113862a9beebe).

Guard exceptions now include both the function and the exact `git rev-parse` flag, so an allowlisted function cannot pick up another metadata flag unnoticed.

I also pulled the repository scan into a testable function and added a multi-file fixture. It catches a constant defined in another file and a forbidden read under `internal/`, while leaving a shadowed `gitDir` parameter alone. `Readlink` is back as a positive control.

I updated the PR description with the parser result too. The new resolver matches Git on all seven pointer spellings. The legacy parser matched three.

The full `mise run check` passes, including both deterministic canaries.

Soph:
**1. The guard false positive is back, and it lands in `strategy`.** `metadata_guard_test.go:585-589` resolves unresolved identifiers by name against package constants, and `strategy/common.go:361` declares `gitDir = ".git"`. Identical code, three packages:

```
control (clean tree):            not flagged
strategy (gitDir const exists):  FLAGGED
checkpoint (no such const):      not flagged
```

So a `gitDir` field on a `strategy` type whose methods live in another file fails the `gitrepo` build, with an error naming neither cause. That is the package part 2 migrates. Type-checking whole packages instead of per file lets the name-based fallback go away. My earlier all-clear was wrong: I probed in `cli`, which has no such const.

**2. `repository.go:151` rejects git directories that git accepts.** `openBareRepository` requires `config.Core.IsBare`; git's own test is HEAD + objects + refs and never reads that key. An ordinary `clone --bare` with `core.bare` unset opens on the merge base and fails here. Only explicit-path callers can reach it, but it is a contract change in a PR that says no consumers change.

**3. Unbounded gitfile read, payload echoed into the error.** `metadata.go:87` and `:119` are bare `os.ReadFile`; `:199` formats the whole pointer into the error. A 64MB sparse `.git`, roughly zero bytes on disk, allocates ~860MB and yields a 128MB error string; 256MB took the review machine out. Real git refuses the same file with `fatal: too large to be a .git file`. This is the one place we are more permissive than git, and this PR is where the parsing gets one owner, so it is the right place to cap it and truncate the pointer.

Two for the description rather than the diff: this fixes a shipped hang (`OpenPath` on main blocks forever on a FIFO `.git`, a symlink to one, or to `/dev/zero`), and your `gitdir:` strictness change matches git where main did not.

MuskanPaliwal:
Thanks @Soph. Fixed all three.

**Guard** [875aff5](https://github.com/entireio/cli/pull/2241/commits/875aff5d36cb66c10d1483cee7e4eeac8f8af595). It type-checks whole packages now, and the name-based fallback is gone. I reproduced your `strategy` case first; the regression test uses that cross-file shape and fails against the old guard. The ledger assertions still pass, so the guard detects everything it did before.

**`openBareRepository`** [1f627f0](https://github.com/entireio/cli/pull/2241/commits/1f627f0428f00e343e16d80e7b2da1d5941e01af). You were right, and it was worse than a contract change. `main` fell through to `PlainOpen`, which opens a bare repo whatever `core.bare` says, so requiring the key was a regression. Only an explicit `core.bare = false` is refused now, which still catches `--separate-git-dir`.

**Pointer reads** [6bcfe65](https://github.com/entireio/cli/pull/2241/commits/6bcfe650ab83ab86366e51acb48b65460b34c290) . One reader caps both files at `4*PATH_MAX`, matching where git itself refuses a `.git` file. A one-byte probe catches the overrun, so an oversized file gets refused rather than truncated into a different valid pointer. Errors elide long paths on a rune boundary.

Both description items are in. I verified the hang against `main` rather than restating it.

One thing to flag on the hang. The check that refuses a non-regular `.git` entry, `metadata.go:95`, has no test. I wrote FIFO coverage for it in this round and then reverted it, since you scoped the hang to the description rather than the diff and I did not want to widen the change under review. The line was already untested when you reviewed the PR, so nothing regressed, but it is the one thing the description claims that the suite does not guard. I can add it in a follow-up.

Separately, a bare reftable repo cannot be opened at all, `unknown extension: refstorage`. Same on `main`, so not from this PR, but I will file it.

`mise run check` is clean: 0 lint, race tests, canaries 56/56 and 4/4.

Soph (approved):
Thanks a lot.

PR #2247, opened 2026-09-02
fix checkpoint explain for native repos

computermode:
https://entire.io/gh/entireio/cli/trails/1226

## Summary

- accept et/project/repo and native entire clone URLs in checkpoint explain --repo
- preserve the forge through repository resolution and route native repos through their project-scoped lookup
- make same-repo detection forge-aware and update command/search guidance

## Testing

- go test ./cmd/entire/cli
- mise run lint

---

> [!NOTE]
> **Medium Risk**
> Changes forge-aware control-plane cell/repo_id resolution for cross-repo checkpoint reads; mis-routing could 404 or hit the wrong region, though behavior is covered by targeted tests.
> 
> **Overview**
> **`entire checkpoint explain --repo`** now works for Entire-native repositories, not only GitHub mirrors.
> 
> The **`--repo`** flag accepts **`et/project/repo`** and **`entire://<host>/et/<project>/<repo>`** in addition to **`owner/name`** / **`gh/owner/name`**. Parsing returns a **forge** plus coordinates; cross-repo reads call **`resolveForgeRepoCellPlacement`** so **`et`** repos use project-scoped native lookup instead of the forge-blind repos index (which could pick a same-named **`/gh/`** mirror). **Same-repo detection** compares forge as well as owner/name, so native and GitHub repos with identical path segments no longer collide. Trail API client wiring uses the same renamed resolver; help text and the search skill document the new **`--repo`** shapes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5bfe91259aca81daad412b08acce7d9403bc19e8. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pfleidi (dismissed):

cmd/entire/cli/explain_repo.go
```diff
@@ -86,59 +85,104 @@ type crossRepoExplainOptions struct {
...
-	_, owner, repo, err = parseMirrorCloneRef(ref)
-	if err != nil {
-		return "", "", fmt.Errorf("invalid --repo %q: expected owner/name or gh/owner/name: %w", value, err)
+}
+
+func explainRepoRef(forge, owner, repo string) string {
+	return forge + "/" + owner + "/" + repo
+}
+
+func explainRepoFullName(forge, owner, repo string) string {
```
Soph: **Blocker: every native `--repo` read fails.**

`explainRepoFullName` returns `et/<project>/<repo>` for native repos, and its comment states this is "the identity spelling returned by entire-api's repo_full_name". It is not. entire-api returns the unprefixed `project/repo` for native rows, on the same endpoint `verifyResponseIdentity` reads:

```
GET /api/v1/repos/01M1DEAAMVDTY2P11ZPG9K18N7/checkpoints/26f98d94b165   (et/entirehq/company-knowledge, eu cell)
→ "repo_full_name": "entirehq/company-knowledge"
```

Same shape for `et/entirehq/marvin` and `et/entirehq/entire-self-serve` on the us cell. So `verifyResponseIdentity` compares `entirehq/company-knowledge` against `et/entirehq/company-knowledge`, misses, falls through to the repo-ULID compare, and misses again. Every native `--repo` read fails, and the message blames the server:

```
$ entire checkpoint explain 26f98d94b165 --repo et/entirehq/company-knowledge
checkpoint identity mismatch: requested checkpoint 26f98d94b165 from et/entirehq/company-knowledge, but the server returned data for repo "entirehq/company-knowledge"; refusing to display possibly-mismatched data (this looks like a server-side bug, please report it)
exit 1
```

Built from 6596b59a. `--repo gh/entireio/cli` renders correctly on the same binary, so the `gh` half of the earlier high finding is genuinely fixed — this is that same defect on the other side.

It survived the tests because every one of them stubs `newCrossRepoReader`, so `explainRepoFullName` never meets a real envelope, and `TestExplainRepoRefKeepsExplicitForgeSeparateFromAPIIdentity` asserts the function returns what it was written to return rather than what the server sends.

Fix: return `owner + "/" + repo` for both forges. Forge disambiguation already happened in `resolveForgeRepoCellPlacement`, which picks the right repo ULID, and the request is keyed on that ULID — so the identity check's only job is "did the server hand me a different repo's data", and the bare pair is the comparand it actually receives. Worth adding the native counterpart to `TestAPICheckpointReader_DisplayRefIsSeparateFromResponseIdentity`, with `repo_full_name: "project/repo"` in the fixture.

cmd/entire/cli/explain_repo.go
```diff
@@ -86,59 +85,104 @@ type crossRepoExplainOptions struct {
...
+		}
+		return nativeCloneForge, strings.ToLower(project), strings.ToLower(repoName), nil
+	case strings.HasPrefix(ref, mirrorCloneForge+"/"):
+		_, owner, repo, mirrorErr := parseMirrorCloneRef(ref)
+		if mirrorErr != nil {
+			return "", "", "", fmt.Errorf("invalid --repo %q: expected %s: %w", value, explainRepoFlagShapes, mirrorErr)
+		}
+		return mirrorCloneForge, owner, repo, nil
+	default:
+		return "", "", "", fmt.Errorf("invalid --repo %q: forge prefix is required; expected %s", value, explainRepoFlagShapes)
```
Soph, line 142: **Breaking change to a stable-released flag.**

`explain --repo acme/widgets` shipped in v0.10.0 (2026-08-10, `346b7fb5b`) and has been in every stable release since. This line removes it — and the bare pair is exactly what the drill-down produces.

`writeSearchCompactJSON` emits `repo` as `org + "/" + repo`, and `compactSearchHit` carries no forge field at all:

```
$ entire search --json --compact "…"
  hit keys: [id, type, repo, branch, author, date, title, score, matchType]
  "repo": "entireio/cli"

$ entire checkpoint explain <id> --repo entireio/cli
invalid --repo "entireio/cli": forge prefix is required; expected gh/owner/name, et/project/repo, …
```

So a search hit hands an agent a string that no longer works, plus nothing from which to build one that does — in the exact workflow this flag exists for, and which `setup_search_skill.go` in this same PR points agents at. `search`'s own `repo:<owner/name>` inline filter and `experts --repo owner/repo` both still take bare pairs, so the three commands now disagree with each other.

The namesquatting argument reads differently here than it does for `repo clone`, where the bare pair is genuinely ambiguous about which forge was meant. CLAUDE.md's own paragraph — unchanged by this PR, directly above the lines it does change — still lists `explain` among the places where the bare pair is safe precisely because lookups cannot cross namespaces. CLAUDE.md:113-117 and the code now contradict each other.

Two ways to close it: keep bare = `gh` (back-compatible, and consistent with search/experts), or keep the requirement and make it recoverable — have the error suggest the forge-qualified readings the way `bareRefSuggestions` does for `repo clone`, and put a forge on the compact hit so the workflow closes. What is here now does neither.

computermode: > keep bare = gh (back-compatible, and consistent with search/experts)

We don't want to default to one forge or the other. I'm investigating this review comment. "have the error suggest the forge-qualified readings the way bareRefSuggestions does for repo clone" <- this sounds like the better path forward.

Soph: yeah sorry I should not have let claude just post this, I think I'm fine to completely remove the none prefix behavior. But then maybe let's ship this in a distinct PR that changes this everywhere? It's not necessary for the explain function for now. And I just want to be sure we are not doing a release in the middle of it and ship some methods doing one thing and others doing another thing 🤔

computermode: I updated it to accept it when /gh/ or /et/ is set, and when there isn't a forge, it returns stating it's invalid and it needs the forge provided:

```
➜  cli git:(fix-native-checkpoint-explain) ✗ entire checkpoint explain 7d7273ff67b8 --repo entirehq/marvin
invalid --repo "entirehq/marvin": forge prefix is required; expected gh/owner/name, et/project/repo, entire://<host>/gh/<owner>/<repo>, or entire://<host>/et/<project>/<repo>
```

cmd/entire/cli/explain_repo.go
computermode: Addressed in the latest commit

cmd/entire/cli/explain_repo.go
computermode: Let me know if that behavior is not desired and I'm happy to give it another look^

Soph (approved):

cmd/entire/cli/explain_repo.go
Soph: I'd love to separate things and fix all other places (like search) in one go, but also fine merging this and I'd do a PR now to get the other parts.

PR #2267, opened 2026-09-03
CLI: remove synchronous mirror creation

pfleidi:
https://entire.io/gh/entireio/cli/trails/1235

## Why

Mirror creation now uses the asynchronous request workflow by default. Keeping the synchronous branch and its kill-switch setting would leave two behavior modes to maintain after the rollout is proven.

## What changed

- Route every `repo mirror create` invocation through the asynchronous mirror-request API.
- Remove the `async_mirror_requests` field from layered settings and delete the synchronous-path tests.
- Describe `--wait-timeout` only in terms of request submission, placement, and clone readiness.

## Usage example

```console
entire repo mirror create github.com/owner/repo aws-us-east-2.entire.io
```

The command now always submits and polls a mirror request. `--wait-timeout` controls the asynchronous workflow only.

The removed setting is not retained as a compatibility tombstone. Settings decoding remains strict, so a file that still contains `async_mirror_requests` fails unknown-field validation and the user must delete that key. Expected external use is near zero because the setting existed only as an emergency kill switch.

## Technical tradeoffs

A permanent tombstone would prevent immediate startup failures for the unlikely existing user, but it would also defer the same breaking removal indefinitely. This PR accepts the narrow compatibility break so the retired setting is removed completely.

## Reviewer notes

- **Merge gate cleared:** COR-1430 shipped in v0.10.5 and is marked Done.
- Changelog and released-version recording remain release-time work and are intentionally outside this implementation.
- Linear: https://linear.app/entirehq/issue/COR-1450

---

> [!NOTE]
> **Medium Risk**
> Removes an emergency sync fallback and breaks settings files that still set `async_mirror_requests`; mirror create is a core onboarding path but behavior is intentionally unified rather than security-critical.
> 
> **Overview**
> **`entire repo mirror create` always uses the mirror-request API** — submit `CreateMirrorRequest`, poll placement, then optionally wait for clone readiness. The synchronous `CreateMirror` branch and the `async_mirror_requests` layered setting are gone.
> 
> **User-facing behavior is simplified:** help text and `--wait-timeout` describe one async deadline (submission, placement, clone). One-shot output is consistently **“Mirror placed at …”** instead of “Registered mirror” / “Mirror exists”. The wizard drops the **empty upstream** status; `--no-wait` paths always show **registered** with an in-progress clone hint.
> 
> **Settings break:** `async_mirror_requests` is removed from `EntireSettings` with no tombstone — configs that still include it will fail strict JSON merge/validation.
> 
> Tests were retargeted to the request flow and synchronous-path coverage was deleted.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8e11938e5d2d52d5748aa6bdb1b07d41412450bf. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pfleidi:
Bugbot run

computermode (approved):

PR #2268, opened 2026-09-03
Harden executable resolution across the CLI

pjbgf:
https://entire.io/gh/entireio/cli/trails/1236

Every $PATH scanner now goes through execx.PathScanDirs(), which returns only absolute entries. Go's own protections do not reach the external-agent scanner: it resolves by filepath.Glob, so a match never passes exec.LookPath (no ErrDot) and it carries separators, so exec.Command's re-check does not apply either -- and discovery executes what it finds. findInaccessiblePlugin carried the same rule separately, which is how the two drifted; one helper keeps them aligned.

external.Agent.run refuses a non-absolute binaryPath before spawning. run sets cmd.Dir to the worktree root and os/exec resolves a relative Path against Dir, so the file registerExternalAgent statted is not necessarily the file that executes. The check sits at the exec, so it also covers the exported New.

external_agents joins openai_privacy_filter.command under the settings trust gate: honored only from a verified-untracked .entire/settings.local.json, a downgrade rather than an error, with the reason surfaced by entire status and by DiscoverAndRegister. It grants execution of every entire-agent-* binary on $PATH, and a committed one-line JSON diff does not read as executable to a reviewer. Every auto-enable therefore writes the key to the local file through enableExternalAgentsLocally: writing it to the project file would produce a setting the user can read back and that never takes effect.

Auditing the six DiscoverAndRegisterAlways call sites -- runSetupFlow, detectOrSelectAgent and runUninstall genuinely need the ungated sweep; --summarize-provider, the non-interactive skill install, and enable's --agent path already knew which agent they wanted and are now named lookups.

pluginParentDir applies its absoluteness rule to XDG_DATA_HOME and LOCALAPPDATA as well, rather than leaving those to osroot and to main.go's PATH restore two layers away from where the invariant is decided.

Regression tests assert behaviour rather than error text: each planted binary writes a marker file, and the tests assert it never appears.

---

> [!NOTE]
> **High Risk**
> Changes authentication-adjacent execution policy: PATH scanning, external agent spawning, and settings that gate arbitrary binary execution—misconfiguration or a bug could break discovery/plugins or leave a path to unintended exec.
> 
> **Overview**
> This PR tightens **who can cause Entire to execute binaries** and **which paths count as on `$PATH`**.
> 
> **Absolute `$PATH` only:** New `execx.PathScanDirs()` centralizes scanning so relative or empty `$PATH` entries are dropped. External-agent discovery and plugin fallback scanning (`findInaccessiblePlugin`) use it instead of splitting `$PATH` directly, closing glob/LookPath gaps where repo-local files could run during hooks.
> 
> **External agent execution:** `external.Agent.run` (and `New`) refuse non-absolute `binaryPath` before spawn, since `cmd.Dir` is the worktree and relative paths would not match what discovery stat’d.
> 
> **`external_agents` as a trust boundary:** Settings load applies `enforceExternalAgentsTrust` (mirroring OPF `command`): `true` is honored only from an untracked `.entire/settings.local.json`. Rejected grants are recorded and surfaced via `entire status` and discovery warnings. Auto-enable paths write through `enableExternalAgentsLocally` with read-back verification so project-file writes or tracked local files cannot silently look “on” while doing nothing.
> 
> **Fewer ungated sweeps:** Summary/configure/enable flows use **named** external discovery where only one agent is needed; full `$PATH` sweeps stay for flows that genuinely need the picker. Non-interactive summary auto-select no longer flips `external_agents` without a human pick.
> 
> **Plugin dirs:** `pluginParentDir` rejects relative `XDG_DATA_HOME` / `LOCALAPPDATA` overrides, not only `ENTIRE_PLUGIN_DIR`.
> 
> Docs (`CLAUDE.md`, security/architecture) and regression tests (marker files proving binaries never run) accompany the behavior changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4a35d4d9797e73816ac14fa539d42e7e815edd9b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1236,6 +1236,21 @@ every directory override it reads — `ENTIRE_PLUGIN_DIR`, `XDG_DATA_HOME`,
 two to `osroot` and to `main.go`'s `PATH` restore. Those backstops hold, but
 they state the invariant two layers from where it is decided.
 
+**The OPF `command` is the deliberate exception, and stays one.**
+`redaction.openai_privacy_filter.command` becomes `argv[0]` of an
+`exec.CommandContext` in `redact/opf.go` with no `cmd.Dir` and no absoluteness
+check, so `"command": "./tools/opf"` resolves against the pre-push process's
+working directory. That is allowed because the field is not repo-supplied: the
+ownership gate above honors it only from an untracked, index-and-HEAD-verified
+`.entire/settings.local.json`, so a relative path there is the developer's own
+choice about their own machine, and a bare name is still covered by
+`exec.Command`'s `ErrDot` re-check. The scanners are the opposite case — they
+resolve names nobody chose, out of a `$PATH` the repository can reach — which
+is why the rule is theirs and not this field's. Do not "finish the job" by
+rejecting a relative OPF `command`: it would break the GUI-git-client setups
+the explicit `command` exists to serve, without closing anything the trust gate
+leaves open.
+
 #### Invoking Commands on Windows - Never Put a Dynamic Value on a cmd.exe Line
 
 **When Entire performs the exec itself, do not go through `cmd.exe`.** Pass the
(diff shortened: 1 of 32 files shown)


pjbgf:
@cursor review

pjbgf:
@cursor review

pjbgf:
@cursor review

Soph (approved):

PR #2273, opened 2026-09-04
flock: revalidate lock file inode after acquire

Rohitkanithi:
## Summary

Adds inode revalidation to the Unix flock implementation after a lock is acquired. If the locked file descriptor no longer matches the current lock-file name, the lock is released and acquisition retries against the current file.

This covers:

- `Acquire`
- `AcquireContext`
- `AcquireIn`
- `AcquireContextIn`
- blocking `LOCK_EX`
- polling `LOCK_NB`

The path-based open error is wrapped so wrapcheck is satisfied.

## Tests

```bash
GOCACHE=/private/tmp/entire-go-build-cache go test ./cmd/entire/cli/internal/flock
git diff --check
```

### DIFF
--- a/cmd/entire/cli/internal/flock/flock_unix.go
+++ b/cmd/entire/cli/internal/flock/flock_unix.go
@@ -13,6 +13,8 @@ import (
 	"os"
 	"syscall"
 	"time"
+
+	"github.com/entireio/cli/cmd/entire/cli/osroot"
 )
 
 // pollInterval is how often the bounded AcquireContext path retries a
@@ -22,8 +24,8 @@ const pollInterval = 25 * time.Millisecond
 // Acquire takes an exclusive advisory lock on path, creating the file if
 // needed. The returned release closes the file, which drops the flock.
 // Callers must invoke release exactly once. The lock file persists between
-// runs — flock state is held by the file descriptor, not by the inode on
-// disk — so the lockfile contents are immaterial.
+// runs - flock state is held by the file descriptor, not by the inode on
+// disk - so the lockfile contents are immaterial.
 //
 // Acquire blocks indefinitely until the lock is available. Use AcquireContext
 // with a deadline to bound the wait.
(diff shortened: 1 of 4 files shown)


Soph (commented):
Thanks! 🙇 Two findings and a few nits


**1. `TestAcquireContextRevalidatesReplacedLockFileInode` can pass on broken code.** If the 50ms handshake at `flock_unix_test.go:87` lands after the waiter's 200ms deadline (`:77`), the waiter times out on the old inode before the unlink happens, the `DeadlineExceeded` assertion at `:105` is satisfied for the wrong reason, and the test passes against unmodified main - I ran it that way to check. `TestAcquireRevalidatesReplacedLockFileInode` stays honest under the same stall because of its two-phase shape: assert no acquire while the new lock is held, release, assert acquire. Giving test 2 that shape fixes it.

**2. The retry loop is unbounded in both branches, and the bounded branch skips its own deadline.** The `continue` at `flock_unix.go:111` jumps back over the ctx check and the `time.After` backoff at `:123`. Driving `lockFile` with a `holdsCurrent` that never converges: 196k open+flock+close cycles in the 3s after a 200ms deadline, and the blocking branch at `:74` never exits at all.

On how reachable that is: I couldn't trigger it through the public API. A competing unlinker converged in 470us, and nothing in the tree unlinks lock files today - `openlock.go` says as much. So it's latent. But the thing that arms it is the lock cleanup half we split out, and `openlock.go` already has the house pattern for the bound (`for range 3`).

Nits:

- The path-based open is wrapped twice: `open flock: open lock file: open /x: ...`, against the root path's `open flock: statat ...`. `openlock.go` carries three `//nolint:wrapcheck` comments saying "wrapping twice would bury the cause" - that's the package's existing answer to the wrapcheck complaint.
- `AcquireIn` / `AcquireContextIn` have no coverage, and that's where nearly every caller is (session_state, pushqueue, shadow_ref, persistent_ref_update, checkpoint_sync_capture, settings, trail_context_cache). They use a different opener and a different revalidator, so the tested half isn't the used half.
- `holdsCurrentRootInode` uses `root.Stat` (`flock_unix.go:140`), which follows symlinks, where this package's discipline is no-follow and `osroot.LstatNoSymlinks` exists. Not reachable today - the opener refuses a planted symlink first - so consistency only.
- Missing `t.Parallel()`; the existing `flock_test.go` has it. And the em-dash to hyphen edit in the `Acquire` doc comment is unrelated to the change.

Rohitkanithi:
Thanks @Soph, addressed these in the latest push.

- Reshaped the replaced-inode tests so the waiter must stay blocked while the current lock file is held, then acquire after it is released.
- Bounded stale-inode retries in both blocking and deadline paths, with context checks/backoff on the deadline path.
- Added coverage for `AcquireIn` and `AcquireContextIn`.
- Switched the root-based revalidation to `osroot.LstatNoSymlinks`.
- Kept the path-based open error raw so `lockFile` wraps it once, added `t.Parallel()`, and undid the unrelated doc-comment churn.

Soph (approved):

dipree (approved):

Soph:
@Rohitkanithi thanks a lot! I found a race in the tests and didn't wanted to put more on your, this already had a few round trips.

PR #2315, opened 2026-09-07
fix: harden filesystem and exec boundaries (hooks dir, summary providers, branch names, user dirs)

pjbgf:
https://entire.io/gh/entireio/cli/trails/1262

Five hardening changes at Entire's remaining un-anchored edges, plus two ratchets
that keep the resulting invariants enforced rather than documented. Each commit
stands alone; they are batched because they came out of one review pass.

## Hardening

**Git hook installation is now rooted.** It was the last tree Entire touched with
bare `os.ReadFile`/`os.WriteFile` on a joined path. Installation now anchors a
root on git's own `--git-path hooks` answer (refusing a symlink at the directory
itself), reads through `osroot.ReadFileNoFollow`, and writes via
`jsonutil.WriteFileAtomicIn`, whose rename replaces a leaf link instead of
following it. A symlinked hook is treated as *foreign*: backed up to
`<hook>.pre-entire` and chained to, exactly as a foreign script is. `doctor`
grows `checkGitHookSymlinks`, since a symlinked hooks directory otherwise just
looks like "hooks absent".

**External summary-provider discovery now respects the `external_agents` grant.**
`discoverSummaryProviderIfMissing` resolves an unregistered
`summary_generation.provider` by running `entire-agent-<name> info`; that lookup
is now gated on `settings.IsExternalAgentsEnabled`, consistent with every other
external-plugin exec. The gate sits after the already-registered early return, so
a committed `"provider": "claude-code"` keeps working with external agents off
Two tests that granted the flag from the committed project file — which the trust
gate has never honoured — now grant it from the untracked local layer.

**Branch names are validated in all three helpers that took one on trust.**
`BranchExistsOnRemote`, `BranchExistsLocally` and `CheckoutBranch` now run
`ValidateBranchName`, replacing `CheckoutBranch`'s leading-dash guard, which
covered less than `git checkout` actually reads (`--`, a pathspec, `@{-1}`).
Validating both existence helpers keeps their answers symmetric — they are cal
side by side on the same name.

**One absolute-override rule now covers every per-user directory.**
`userdirs.RequireAbsoluteOverride` replaces the plugin-local copy and extends
rule to `ENTIRE_CONFIG_DIR` and `XDG_CACHE_HOME`, which redirect the config and
cache trees. `Config()`/`Cache()` keep their string signatures, so the refusal
lands at the roots: `resolveUserRoot`, `contexts.configRoot` and
`discovery.cacheFile.root`. The last two no longer run `filepath.Abs` on the
override first. Rejecting beats falling back to the platform default — for the
config directory that default is the developer's real `~/.config/entire`.

---

> [!NOTE]
> **High Risk**
> Touches git hook install/removal, symlink-following policy, external binary discovery, and checkout behavior—security-sensitive paths where mistakes cause data loss or unintended execution.
> 
> **Overview**
> This PR tightens several filesystem and execution boundaries and adds machinery so those rules stay enforced in CI.
> 
> **Symlinked agent config directories** can be opted into via `allow_symlinked_agent_dirs` in untracked `.entire/settings.local.json`. Settings load applies the same trust gate as `external_agents`, installs a worktree-scoped policy (`SetVouchedSymlinkedDirs` / `OpenAnchoredRoot`), and `HookConfigFile` now separates worktree-relative vs root-relative paths so `RemoveDir` and pi uninstall work when a vouched link is followed. Doctor/status report followed vs refused links and continue scanning beneath vouched dirs.
> 
> **Git hooks** are treated as a rooted tree (install vs removal asymmetry for symlinked hook dirs), with `doctor` reporting symlinked hook directories and managed hook files. **Summary providers**: named external discovery in `discoverSummaryProviderIfMissing` requires the `external_agents` grant; auto-selected externals are not persisted without that grant. **Git checkout/branch helpers** validate branch names consistently and `CheckoutBranch` uses a trailing `--` to avoid silent pathspec restores. **Per-user dir overrides** funnel through `userdirs.RequireAbsoluteOverride` (including plugin index cache).
> 
> **Source guard tests** share `testutil.GitGrepGuard` (`--untracked`, `--no-color`, repo env scrubbing); new/updated guards cover hook-config coverage, transcript-read ratchet, git-status `--no-optional-locks`, and root-base allowlists. **CLAUDE.md** documents these invariants at length.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c35f7606d8396c75d496754e1dfc1bf4ec0a8d8. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -788,7 +788,7 @@ Don't use `fmt.Print*` for operational messages (checkpoint saves, hook invocati
 
 ### The Root Anchors
 
-Entire does filesystem I/O in seven trees, and each has one package that owns a
+Entire does filesystem I/O in eight trees, and each has one package that owns a
 shared `*os.Root` over it. **Never assemble a path into one of these and hand it
 to `os.ReadFile`/`os.WriteFile`/`os.MkdirAll`/`os.ReadDir`/`filepath.Walk`.**
 
@@ -799,6 +799,7 @@ to `os.ReadFile`/`os.WriteFile`/`os.MkdirAll`/`os.ReadDir`/`filepath.Walk`.**
 | the working tree | `worktreedir` | worktree root |
 | an agent's hook config | `agent.HookConfigFile` | worktree root (`.claude/`, `.cursor/`, `.gemini/`, `.github/hooks/`, `.factory/`, `.codex/`, `.opencode/plugins/`, `.pi/extensions/entire/`) |
 | an agent's session store | `agent.SessionStore` | the agent's own `GetSessionDir` |
+| the active git hooks dir | `strategy.hooksRoot` | `git rev-parse --git-path hooks`, absolutized |
 | per-user config / cache | `userdirs.ConfigRoot` / `CacheRoot` | `$ENTIRE_CONFIG_DIR` else `~/.config/entire`; `$XDG_CACHE_HOME/entire` else `~/.cache/entire` |
 | managed plugin tree | `pluginRoot` (`plugin_store.go`) | `pluginParentDir()` — `$ENTIRE_PLUGIN_DIR`, `%LOCALAPPDATA%`, or `$XDG_DATA_HOME` |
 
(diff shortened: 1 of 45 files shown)


pjbgf:
@cursor review

pjbgf:
@cursor review

pjbgf:
@cursor review

pjbgf:
@cursor review

Soph (approved):

PR #2327, opened 2026-09-08
feat(checkpoint): add `entire checkpoint migrate`

toby:
https://entire.io/gh/entireio/cli/trails/1271

> **Stacked on #2335** (`feat(checkpoint): elect the Entire remote for checkpoint sync`). Review that first; this PR's base is its branch and will be retargeted to `main` once it merges. Best reviewed **commit by commit**: the four commits are layered so each builds and is reviewable on its own.

## Why

Once an Entire remote is elected (#2335), new checkpoints flow there, but the ones already pushed to GitHub stay behind. This PR brings that backlog over, verifies it, and cleans up the old remote with consent, then rewards the user with what they have accumulated.

## Commits

1. **`feat(strategy): checkpoint artifact migration engine`** — inventory, hydrate into a tmp namespace and promote via `SafelyAdvanceLocalRef`, requeue, verify at the local hash, chunked delete. Removal addresses the old remote by its single push URL so the dedicated-`checkpoint_remote` rewrite can never apply, and refuses fan-out remotes. `batchPushRefs` chunks refspecs (a backlog publish overran ARG_MAX); `PushQueue.EnqueueAll`; `remote.DeleteRefs` (trail cleanup reuses it); the ls-remote parser moves into the `checkpoint` package. **Review question:** can anything be deleted that was not verified? Never force-pushed?
2. **`feat(checkpoint): migration planner, copy and value summary`** — the pure planner over the repo's state (no remotes, dedicated store, fail-closed setting, choice among remotes, pinned elsewhere, already home, publish only, migrate, convert first) and every user-facing string in one file; the local value block; `formatTokenCount` gains `M`. **Review question:** are the states and the copy right?
3. **`feat(checkpoint): add entire checkpoint migrate`** — the command. Two planning passes (local facts, then ls-remote inventories), pickers whose first option is always no-change, `--to` recorded in `.entire/settings.local.json` only when the election would not already pick it, git-branch repos converted to refs and flipped, publish through the OPF- and policy-gated queue flush, verify, then delete only verified refs and only with consent (`--remove` when non-interactive; `--yes` never implies it). Bare non-interactive runs report and change nothing. Classified user-owned and unlisted in agent-help. **Review question:** is the non-interactive contract safe for agents?
4. **`feat(status): surface the checkpoint migration in status, doctor, enable and docs`** — the network-free status nudge and `checkpoint_sync_migration` JSON field, doctor's destination check, the re-enable pointer, the command name restored in the announcement, hint and topology advice, and the docs.

## Decisions for reviewers

- `checkpoint migrate` is **user-owned and unlisted** in agent-help: it pushes transcripts, deletes remote refs and writes settings.
- The backend flip writes the **project** `.entire/settings.json` (a repo-wide format decision) and prints a commit hint; it is skipped under `ENTIRE_CHECKPOINTS_PRIMARY`.
- The ledger records `done` once data is verified on the destination, and `declined` when cleanup is refused; both silence the status nudge.
- No prompt inside `git push`; the hook announces and points at the command. Offering the migration from `repo mirror use`, which creates the remote, is a natural follow-up.

## Testing

- `mise run lint`: 0 issues. Full unit and integration suites (both backends) and the Vogon canary pass on the stack.
- New: 14 engine tests over two local bares, 30-row planner table, executor tests against a fake engine (verify-missing stops before remove, remove failure keeps the ledger honest, decline records declined, `--yes` never removes, `--dry-run` only lists), status rendering tests, and seven end-to-end `checkpoint migrate` scenarios on both backends.
- Copilot's earlier comment that `for i := range count` does not compile is incorrect (Go 1.22+; this module is Go 1.26) and its suggested patch would not compile; the test passes.


Soph (commented):
Findings from reviewing this before the split, re-checked against `b3fef4e` — all still apply. Tier-side findings are on #2335.

First, a heads-up: **this PR has no CI.** Zero workflow runs on `b3fef4e`, while #2335 got 12/12 in about five minutes. `pull_request`'s default activity types do not include `edited`, so retargeting the base did not trigger one — it needs an empty commit or a close/reopen before the green means anything.

**The headline path fails on first run.** `checkpoint_migrate_plan.go:433` gates `stepRequeue` on `in.LocalCheckpoints > 0`, which is measured before hydration. In any clone that does not already hold the refs — i.e. any fresh clone, since `git clone` does not fetch `refs/entire/checkpoints/*` — it is 0, so nothing is ever queued: hydration fetches the refs, `SafelyAdvanceLocalRef` does not enqueue, and `MigrateBranchToRefs` does not run without a local v1 branch. Ran it against a remote holding 3 refs:

```
✓ Fetched 3 checkpoint refs from origin
✓ Pushed 0 checkpoint refs to dest
✓ Verified 0 checkpoint refs on dest
  ! 3 of 3 checkpoint refs did not arrive on dest. Nothing was removed from origin.
```

A second run succeeds, so it is not data loss — but the primary path fails the first time with an alarming message and a non-zero exit. The gate wants `in.PrimaryIsRefs` alone, or `|| len(plan.Sources) > 0`. It is untested because every integration test seeds a local checkpoint first.

**The closing block asserts the one thing this area is careful about.** `checkpoint_migrate_plan.go:961` prints "whichever remote you push your code to" unconditionally, even though the function takes `destIsEntire` and branches on it two lines earlier. That is only true under the Entire tier; under `config` / `observed` / `default` / `sole` / `first` the single-remote gate makes it the opposite, which is what `entire status`'s gated-push hint exists to say. Reproduced on a plain `default`-tier repo.

**An invalid `--to` / `--from` exits 1 printing nothing.** plan.go:312, 384 and 387 set `ExitError` while leaving `State` and `Reason` empty, so `finishReportOnly` prints `"  " + ""` — a blank line — and returns `NewSilentError`, which main.go suppresses by design. `--json` reports it correctly, so it is the human path only. `TestRunCheckpointMigrate_UnknownToIsAnError` asserts only `require.Error`, which is why it passes.

**With `ENTIRE_CHECKPOINTS_PRIMARY` set, the backend never flips but v1 is still deleted.** `convert` prints a note and returns early on `BackendEnvOverride` without writing settings, so `PrimaryIsRefs` stays false — yet execution continues to publish, verify and delete. Publish and verify only cover per-checkpoint refs, so v1 is never pushed to the destination but is still removed from the old remote. Contrast the error path, which aborts before removal.

The rest, smaller:

- `r.convert` is called unconditionally rather than gated on a plan step, so a git-refs repo with a leftover local v1 branch gets a real conversion that `--dry-run` never listed and the confirmation never mentioned — saw `✓ Converted 0 checkpoints to refs` on exactly such a run.
- `doctor.go`'s `checkCheckpointDestination` tests `entireElected != ""` before `ambiguous()`, so an Entire remote with two pushurls never reports the fan-out. `describeCheckpointDestination` handles that case correctly, and the previous unconditional call did report it.
- The `already_home` branch writes the migration marker before the `--yes` gate, contradicting "it writes nothing unless `--yes` is passed"; it is also recorded as `done` in states where nothing moved, and it is one global flag with no per-remote scope, so a backlog appearing later is permanently un-nudged.
- Declining the picker prints the header and reason twice — `chooseSyncDestination` renders both, then `finishReportOnly` renders both again.
- `convertedAll` can never be false: `Total++` runs once per checkpoint, exactly one of `Migrated` / `Skipped` runs per checkpoint, and any per-checkpoint failure errors out of the walk. The interlock guarding the irreversible v1 deletion is dead code.
- Every remote is ls-remoted twice (the stored inventory drops the hashes hydration needs) and `inventorySyncRemotes` ignores `--from`, so it dials remotes the user excluded. It is also a TOCTOU: the plan the user consents to comes from listing #1, while hydration and removal act on listing #2.
- README's ordered list puts convert before fetch, contradicting both the command's own `Long` and `planSteps`, whose comment says the order is load-bearing.
- `formatTokenCount`: 999,999 renders as `1000k` rather than `1M`, and 2.1e9 as `2100M`.
- `pushQueuedCheckpointRefsReporting`'s doc claims it is shared by both commands; it has one caller, and `strategySyncEngine.Publish` re-implements the same sequence with different strings.
- `legacyElectedRemote` duplicates `strategy.LegacyCheckpointRemote`; `formatDestinationUnchanged` is referenced only from tests; `LegacyRemote` appears in the text nudge but not in `--json`; `status.go:483` still says `entire checkpoint sync`.

Unrelated to your code: Copilot's inline comment on `refs_push_test.go` is a false positive — `for i := range count` has been legal since Go 1.22 and this repo is on 1.26.6.

toby:
Thanks for the thorough pass — everything is addressed in the new top commit `56cdfcae3` (the four layered commits are unchanged apart from a rebase onto #2335's fixes), and the rebase onto the new base gives this PR its first CI runs.

**First run fails in a fresh clone.** Requeue is now planned whenever data moves (dedup is free), not gated on a local count taken before hydration. `TestCheckpointMigrate_FreshCloneMigratesOnFirstRun` seeds origin, deletes every local ref, the v1 branch and the queue, and requires the first `--yes` run to land on both backends.

**Closing block.** "whichever remote you push your code to" prints only when the destination is the Entire remote; otherwise it says pushes to the destination carry checkpoints and pushes elsewhere do not.

**Invalid `--to` / `--from` exits silently.** An `ExitError` with no reason is now returned as an ordinary error that main prints; `TestRunCheckpointMigrate_UnknownToIsAnError` asserts the message and that it is not a `SilentError`.

**`ENTIRE_CHECKPOINTS_PRIMARY` deletes v1 without flipping.** Deleting an old remote's v1 branch now requires both that this run converted it and that git-refs is the store (`storeIsRefs`, set by the flip); under the override the branch is kept and the run says why. `convertedAll` is gone. Covered by `TestRunCheckpointMigrate_EnvPinnedBackendKeepsOldV1Branch` and the git-branch leg of the idempotency integration test, which now expects the branch to survive until the store is git-refs.

**Smaller items, in your order:** conversion is a planned `convert_v1` step (listed by `--dry-run` and the confirmation) whenever a v1 branch will exist locally after hydration; doctor checks `ambiguous()` first so an Entire remote with two pushurls still reports the fan-out; already-home records the ledger only with `--yes` or an interactive run, and otherwise prints how to record it (the single-flag scope is a known limit I would rather live with than make status dial remotes); declining the picker renders the footer only; each remote is listed once — the planner keeps the advertised hashes, execution acts on that listing, and `--from` now excludes remotes from the listing as well as removal (`TestRunCheckpointMigrate_FromNarrowsInventoryAndRemoval`); README lists fetch before convert; `formatTokenCount` rounds before choosing the unit (999,999 → `1M`, 2.1e9 → `2.1B`); the one-caller `pushQueuedCheckpointRefsReporting` refactor is reverted; `legacyElectedRemote` is gone; `formatDestinationUnchanged` is printed when `--to` names the elected remote; `--json` carries `checkpoint_sync_legacy_remote`; the stale comment is fixed.

Copilot's `range` comment: agreed, false positive; no change.

PR #2335, opened 2026-09-09
feat(checkpoint): elect the Entire remote for checkpoint sync

toby:
https://entire.io/gh/entireio/cli/trails/1276

## Why

Repos with several remotes were told to set `checkpoint_remote`, which names a dedicated GitHub checkpoint repository and silently ignores a remote name. The remote picker is `strategy_options.checkpoint_push_remote` in `.entire/settings.local.json`; the topology note and the git-refs multi-URL warning now say so, and tests pin both strings.

Entire's own `entire://` remote is the store built for checkpoint data, and refs pushed to an Entire mirror stay on Entire. When a repo has one, checkpoints should sync there without configuration.

## What

**Entire tier in the election.** A sole `entire://` remote is elected after an explicit setting or a captured election and before origin. Detection reads raw `.git/config` urls and pushurls (never `git remote get-url`, which expands `insteadOf`), and a remote counts only when every URL is `entire://`. Capture cannot displace it.

**Pre-push follows the Entire remote.** Once elected, every push — any remote or raw URL — carries checkpoints to the Entire remote by name, the way the dedicated `checkpoint_remote` URL mode already bypasses the single-remote gate. The git-branch empty-remote defer is skipped under the redirect. The first delivery is announced once on stderr, latched by `entire-checkpoint-sync-entire.json` in the git common dir.

**Copy.** The wrong `checkpoint_remote` advice is replaced with the setting and file that actually pick a remote; with an Entire remote elected the note reports the destination instead of calling it ambiguous. `entire status` shows `(your Entire remote)`.

## Decisions for reviewers

- **Redirect on every push.** After the tier flips the election, a `git push origin` habit would otherwise strand checkpoints locally forever. The dedicated `checkpoint_remote` URL mode already has this shape.
- **Skipped empty-remote defer.** Assumes the Entire server never adopts `entire/checkpoints/v1` as a mirror's default branch.
- **Product order.** An explicit `checkpoint_push_remote` and a captured election still outrank the tier.

## What is not here

The command that brings the backlog off the old remote and cleans it up (`entire checkpoint migrate`, #2327) stacks on this PR. Until it lands, an elected Entire remote receives new checkpoints while older copies stay where they were; nothing is lost in the interim.

## Testing

- `mise run lint`: 0 issues. Unit, full integration (both backends) and the Vogon canary pass.
- Integration tests simulate an `entire://` remote with `url.file://<bare>.insteadOf entire://…`, so raw config says `entire://` while transport reaches a bare repo.
- New: election tier and capture tests, redirect and announcement tests, `remote_topology_test.go` pinning the corrected advice, and five end-to-end pre-push scenarios.


Soph (commented):
Reviewed the split — it is clean: the re-partition cost 3 lines moved between two test files and nothing else, and there are no dangling `checkpoint migrate` references here. Three findings; the first I would block on, given this lands ahead of #2327.

**"nothing is lost in the interim" does not hold for repos without an `origin`.** `CheckpointReadRemotes` returns `[elected, "origin"]` with origin hardcoded, so when the tier displaces a `sole` or `first` election on a differently-named remote, that remote drops out of the read chain entirely. Probed against this branch:

```
SAFE    was "origin"   (default) -> now "ent" (entire), readChain=[ent origin]
BROKEN  was "gh"       (sole)    -> now "ent" (entire), readChain=[ent]
BROKEN  was "upstream" (first)   -> now "ent" (entire), readChain=[ent]
```

The checkpoints are not deleted, but nothing looks for them any more — and #2327, the recovery, is stacked behind this. It is also silent: the announcement dropped the "Earlier checkpoints may still be on X" line, so `LegacyCheckpointRemote` now only reaches a `slog` attribute. Smallest fix I can see is to persist the displaced election in `entire-checkpoint-sync-entire.json` (which the tier already writes) and append it to the read chain — no new command needed.

**The redirect can abort the user's own `git push` on an OPF + git-branch repo.** `manual_commit_push.go:195` passes `ps.pushTarget()` to `RewriteUnpushedV1WithOPF`. The call site is unchanged but its meaning is not: under the redirect that is now the Entire remote, which has no v1 yet, so `resolveRemoteV1Tip` finds nothing and the whole local v1 history counts as unpushed. Above `bootstrapDefaultLimit` (100) that is `BootstrapTooLargeError`, and git-branch fails closed by aborting the user's push. Below 100 it silently OPF-rewrites and CAS-updates the entire history instead.

**The rationale at `checkpoint_sync_remote.go:264-269` is factually wrong** — "which is also what a transport helper is dispatched on before any rewrite". Git applies `insteadOf` / `pushInsteadOf` *before* dispatch; I verified it, a remote whose raw url is `entire://…` plus a rewrite pushes to the rewritten target and the entire helper is never invoked. So a remote can be classified Entire, elected, reported as "your Entire remote", and receive transcripts with the single-remote gate bypassed, while actually pointing elsewhere. Not repo-supplied — `insteadOf` is user-local config — so this is a wrong justification rather than a vulnerability, but it is the sentence the gate bypass rests on, and the tests cannot distinguish the two cases because they use `insteadOf` to fake the remote. Worth noting the two halves already disagree: the election reads raw config, `RemoveCheckpointArtifacts` reads expanded push URLs.

Nit: `EntireRemotes` is exported here with no production caller until #2327.

toby:
Thanks — all four findings and Copilot's four are addressed in `683a34e7b`.

**Displaced remote drops out of the read chain (blocking).** `CheckpointReadRemotesWithElection` now appends the remote the tier displaced (`LegacyCheckpointRemote`: origin, else the sole, else the first non-Entire remote) under the Entire tier, so `gh` / `upstream` stay readable. Pinned by `TestCheckpointReadRemotes_EntireTierKeepsDisplacedRemote` over your three cases. I went with computing it from config rather than persisting the displaced election in the ledger: it answers the same question while the remote exists, and if the remote is gone there is nothing to read from anyway. The announcement names it again — `Earlier checkpoints may still be on "origin"; they stay readable from there.` — while the migration ledger is pending; #2327 turns that into the command pointer.

**Redirect can abort the user's push on OPF + git-branch.** The v1 rewrite is now bounded by the displaced remote's v1 tip when the Entire remote has none (`rewriteUnpushedV1WithOPFBounded`), so commits already published there are not a bootstrap; `TestRewriteUnpushedV1WithOPFBounded_FallbackBoundSkipsPublished` shows the same chain hitting `BootstrapTooLargeError` unbounded and passing bounded. And under the redirect an OPF decision or rewrite failure now withholds the checkpoints with a stderr line and lets the push continue (`skipRedirectedCheckpointPush`) — the user aimed that push elsewhere, so our delivery must not veto it. Direct pushes to the Entire remote keep the fail-closed abort.

**Rationale factually wrong.** Corrected: git applies `insteadOf` before dispatch, so raw classification answers "which remote did the user declare as Entire", not "where do bytes go"; the comment now says that, why it is the intended question (user-local config, never repo-supplied), and that the removal engine deliberately reads the expanded push URL because it addresses a concrete delete target.

**Copilot:** raw hook target redacted in the debug log; `pushSettings.entireTier` / `targetsEntireRemote()` so a direct `git push entire` skips the empty-remote defer too; the topology fan-out text under the tier now describes code only ("checkpoints do not fan out with it") and drops the first-URL marker; the status counter says "with your next git push, to any remote" instead of naming `entire`. `EntireRemotes` is unexported here and re-exported in #2327's command commit.

toby:
You're right, and thanks for catching that — the derivation moves when config does, and `LegacyCheckpointRemote` preferring `origin` unconditionally is exactly the swap you describe. Fixed in `351036b79` by persisting it, the option I should have taken the first time.

**Recorded, not derived.** `EntireSyncState` gains `displaced_remote`, written once at the first push under the tier — in the redirect rather than the announcement, because "which remote did this tier displace" is true whether or not that push lands. `DisplacedCheckpointRemote` returns the record while that remote is still configured, and falls back to the live answer only when nothing is recorded yet or the recorded remote is gone (nothing to read from it under that name either way). The announcement, the OPF rewrite bound, and — on #2327 — the status nudge, doctor and enable all read it, so none of them can be re-pointed by a later `git remote add`.

Your sequence is now a test: sole `gh`, tier takes over and records `gh`, origin added afterwards → `LegacyCheckpointRemote` returns `origin`, `DisplacedCheckpointRemote` returns `gh`, and the chain is `[entire origin gh]`. Also covered: write-once (a later live answer does not overwrite), the fallback when the recorded remote is removed, and an announcement that names `gh` while an origin exists.

**One thing I went further on.** The read chain no longer names a single fallback at all: under the tier it appends *every* non-Entire remote, recorded one first. A single fallback is a pick, and any pick computed from the remote set moves when the set does — the record fixes the specific sequence you found, but the general property I want is "the chain cannot lose a holder". Any non-Entire remote may hold pre-tier checkpoints, reads are best-effort and fail open per candidate, so the cost is a bounded extra probe on read paths under the tier only.

Residual I'm accepting, stated for the record: between adding the Entire remote and the first push under it nothing is recorded, so a config change inside that one-push window still gets the live answer — but the chain-carries-everything change means it cannot drop a holder even then; it only affects which remote gets *named* in prose.

Soph:
One thing left, which I should have raised alongside the last comment rather than after it. The OPF sites are gated on `redirectedToSyncRemote()` while the empty-remote defer uses `targetsEntireRemote()`:

```
173:  if ps.redirectedToSyncRemote() {   // OPF decision failure
206:  if ps.redirectedToSyncRemote() {   // boundFallback
212:  if ps.redirectedToSyncRemote() {   // skipRedirectedCheckpointPush
409:  if ps.targetsEntireRemote() {      // empty-remote defer
```

`redirectedToSyncRemote()` is `syncRemote != ""`, which is only set when the push names a *different* remote. So a direct `git push entire` on an OPF + git-branch repo with more than `bootstrapDefaultLimit` v1 commits gets neither the fallback bound nor the withhold, and takes the original abort path — the case the bounded rewrite exists for. `recordDisplacedCheckpointRemote` runs on that push too, so the recorded name is right there; only the predicate is keeping it out. Widening all three to `targetsEntireRemote()` looks like the whole fix.

toby:
Good catch, and agreed — the asymmetry was accidental, not a decision. Widened in `c59bd4127`; all three sites now key on `targetsEntireRemote()`.

You're right that the bound is the clear bug: the destination is the same Entire remote whether or not the push was redirected there, so it has the same missing v1, and the recorded displaced remote is sitting right there on that push. I confirmed the abort by reverting the predicate against the new tests — `OPF bootstrap would rewrite 3 checkpoint commits (limit 2)` out of the pre-push hook on a direct `git push entire`.

I also widened the **withhold**, which I want to flag because it is a behaviour change rather than a bug fix. My original rationale for it was "the user aimed this push elsewhere", which genuinely does not apply to a direct push — so the honest question is whether a direct push to the checkpoint destination should keep the v1 path's abort. I decided it should not, on the grounds that under the tier checkpoint delivery is attached to *every* push, so it is machinery Entire added to a push the user aimed at their code, and which remote they happened to name is incidental. Making the response to a broken OPF depend on that would be exactly the surprise the tier exists to remove. It is still fail-closed — nothing un-OPF'd ships, the checkpoints stay local and retry — and it is the stance the git-refs backend already takes for a failed OPF gate. Happy to reverse it if you read the fail-closed intent as "abort whenever the push names the destination"; the bound half is not in question either way.

`redirectedToSyncRemote()` had no production caller left, so it is gone; the tests that used it assert `syncRemote` directly. `skipRedirectedCheckpointPush` is now `withholdEntireCheckpointPush`.

Two regression tests drive the real hook against a git-branch repo whose origin holds the v1 history and whose Entire remote does not: a direct push gets the fallback bound rather than `BootstrapTooLargeError`, and an OPF misconfiguration is withheld with the stderr line rather than aborting. Both fail with the predicate reverted to the redirect check, so they bite. Lint clean, full unit and integration suites and the canary pass on both branches; #2327 is rebased on top.

Soph (approved):
Let's do this as a follow up. The whole OPF integration is something I'm questioning more and more if it's not causing more work then is actually useful or used

PR #2321, opened 2026-09-08
Hardening: verify redaction splices, gate agent instruction fields, apply checkpoint ownership to fetches, minimal Copilot tool surface, shared terminal sanitizer

pjbgf:
https://entire.io/gh/entireio/cli/trails/1265

Five independent hardening changes, one commit each.

## redact: verify each JSONL splice landed
`applyJSONReplacements` splices by matching the Go re-encoding of each leaf
against the raw line, and JSON admits more than one encoding of the same
string (raw U+2028, `\/`, `\uXXXX`-escaped ASCII, ensure_ascii-style
non-ASCII) — a producer's encoding choice could leave a flagged leaf
unredacted inside certified `RedactedBytes` with no signal. Redaction now
re-checks every spliced line against the collected originals (honoring key
scoping) and falls back to a structural re-encode of that line (`UseNumber`
so large integers survive; surrounding whitespace reattached). A line that
cannot be rewritten fails the redaction instead of certifying. The
byte-preserving fast path is pinned by test.

## settings: provenance gate for agent instruction fields
`investigate.always_prompt` and every `ReviewConfig.Prompt` (workers + judge,
profiles + legacy map) are appended verbatim to prompts of agents spawned
with approvals disabled, so they get the same trust gate as the OPF `command`
and `external_agents`: honored only from a deep-verified untracked
`.entire/settings.local.json` or clone-local preferences (in `.git/`,
unreachable by clone). Rejection is a downgrade, recorded on
`AgentPromptRejections()`; `entire review`/`entire investigate` print a
one-line stderr notice. A reflection test pins that any future `ReviewConfig`
placement in the schema must be gated.

## checkpoint/remote: ownership check on fetches
`FetchURL` now applies the same `checkpointRemoteIsInherited` check as
`PushURL` (identity set: origin + the caller's read candidate), so an
inherited committed `checkpoint_remote` cannot select the repository local
checkpoint refs are populated from. On rejection, resolution falls back to
the read-candidate chain with `authoritative=false`, keeping ref-absence
classification intact.

**Deliberate reversal:** two tests pinned "reading checkpoints is always
allowed" and are inverted. Consequence: a fork clone of a repo with a
committed `checkpoint_remote` no longer reads from the upstream checkpoint
repo; the escape hatch (as for pushes) is declaring it in
`.entire/settings.local.json`.

## copilotcli: minimal tool surface for text generation
Replaces `--allow-all-tools` with `--deny-tool shell/write/url` (automatic,
[...]

PR #2342, opened 2026-09-09
fix: preserve selected remote when fetching checkpoint refs

peyton-alt:
https://entire.io/gh/entireio/cli/trails/1283

## Summary

Fix checkpoint reads from a selected fork when shared settings still name an inherited dedicated checkpoint repository.

After #2321, discovery correctly finds the checkpoint ID on the fork, but individual-ref fetching drops the selected read candidate when a `checkpoint_remote` key exists. A fresh clone therefore lists incomplete session metadata and cannot retrieve checkpoint detail or transcripts, despite successful delivery to the fork.

Pass the already-selected lead through the existing single-target fetch helpers. This reuses ownership and URL resolution, preserves authority/error classification and timeout behavior, and leaves pushing, discovery, hook probes, and settings unchanged.

## Verification

- `TMPDIR=/private/tmp mise run check`: formatting, lint, race-enabled unit/integration tests, and deterministic canaries passed.
- Pre-push lint: zero issues.
- Ten independent fresh-clone CLI scenarios passed: list, explain-list, session-filtered list, detail, and exact transcript export, with inherited settings and without them.
- The identical integration test on unchanged main fails all five inherited-setting cases with the original metadata/not-found errors; all five controls pass. Each read verifies the checkpoint ref and object are absent first.
- Unit controls cover accepted dedicated/local overrides, invalid settings, failed/empty election, transport failure, non-authoritative absence, and cancellation.
- Independent spec and code-quality reviews passed.

## Scope and caveats

- Six executable production lines; no new resolver, settings schema, ownership policy, or fallback chain.
- No checkpoint migration or user configuration changes.
- New transport-mapping regression fixtures use bash and skip Windows. Tests use isolated local Git repositories, not real GitHub pushes or agent API calls.

---

> [!NOTE]
> **Medium Risk**
> Touches checkpoint ref fetch and local ref installation when dedicated settings are inherited; wrong routing could still mis-hydrate, but behavior is narrowly gated and heavily regression-tested.
> 
> **Overview**
> Fixes checkpoint **per-ref hydration** when shared settings include an inherited `checkpoint_remote` but checkpoints were pushed to an elected fork. Previously, presence of the `checkpoint_remote` key forced `FetchCheckpointRefFrom` onto legacy single-target resolution and **ignored the lead read candidate**, so discovery could find refs on the fork while list/detail/transcript reads failed on fresh clones.
> 
> When settings load cleanly, `checkpoint_remote` is valid, election succeeded, and the first read candidate is non-empty, the fetch now resolves the target via **`checkpointFetchTargetFrom` with that lead** and runs the existing single-target probe/fetch path. Legacy behavior is unchanged for failed election, empty candidates, malformed/unreadable settings, or accepted dedicated overrides.
> 
> Adds unit coverage (forge URL fixtures, authority, no chain retry on fallback failure) and an integration test that exercises five CLI read paths from a fork-only clone with inherited vs local-only settings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aa41e629721df623188df51780ce9c543ebd0cfa. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


peyton-alt:
Addressed the timeout suggestion from the Copilot review in 92baa1dd4: the configured-remote fetch now honors the supplied fetchTimeout. The public wrapper still supplies the same two-minute default. Added a stalled-loopback regression, verified red before the change and green after, and reran the full mise run check (passed, including deterministic canaries) plus final lint. Independent follow-up review found no actionable issues.

Soph (approved):

Issue #2237, opened 2026-09-02
Re-install discards another tool's newer hook and chains to a stale .pre-entire backup


seungpyoson:
### What happened?

When another tool that manages git hooks reinstalls its hook at a name Entire wraps (`commit-msg`, `prepare-commit-msg`, `post-commit`, `pre-push`, `post-rewrite`), the next Entire hook install discards that tool's **new** hook and chains to the **old** copy in `<hook>.pre-entire`. Git then runs the stale copy on every commit. The only signal is one stderr line, which is invisible when the reinstall happens through the agent-hook self-heal path (`EnsureSetup` on user-prompt-submit) rather than an interactive `entire enable`.

Where it comes from (`cmd/entire/cli/strategy/hooks.go`, `InstallGitHook`, at fce049c):

```go
if existingErr == nil && !strings.Contains(string(existing), entireHookMarker) {
    if !backupExists {
        os.Rename(hookPath, backupPath)          // first install: back up the foreign hook
    } else {
        fmt.Fprintf(os.Stderr, "[entire] Warning: replacing %s (backup %s%s already exists from a previous install)\n", ...)
        // the foreign hook now at hookPath is overwritten below and never saved
    }
    backupExists = true
}
```

`TestInstallGitHook_DoesNotOverwriteExistingBackup` pins this: the backup is never refreshed once it exists. That protects the first backup, but it means the file at the hook name, which is the other tool's current hook, is the one that gets lost. The chain then runs whatever the other tool installed at the time of the *first* Entire install, forever.

This is the mirror image of #1349: there, the other tool reclaims the name and Entire's wrapper is lost; here, Entire reclaims the name and the other tool's newer hook is lost, silently replaced by an older version of itself.

### Steps to reproduce

```sh
git init -q -b main repro && cd repro
git commit -q --allow-empty -m init

# another tool installs its hook (v1)
printf '#!/bin/sh\necho "tool A v1 ran" >> "$(git rev-parse --git-dir)/hook-log"\n' > .git/hooks/commit-msg
chmod +x .git/hooks/commit-msg

entire enable --agent claude-code --no-github --local
# -> "[entire] Backed up existing commit-msg to commit-msg.pre-entire"   (correct)

# the other tool reinstalls a newer hook (v2), replacing Entire's wrapper
printf '#!/bin/sh\necho "tool A v2 ran" >> "$(git rev-parse --git-dir)/hook-log"\n' > .git/hooks/commit-msg
[...]

PR #2343, opened 2026-09-09
fix: make checkpoint status reflect hook health

### LINKED ISSUE #2264: entire status reports "Checkpoints sync to: origin" when the git hooks are not installed
### What happened?

`entire status` reports `● Enabled` and `Checkpoints sync to: origin` when Entire's git hooks are not installed at all. `entire doctor` on the same repo correctly reports `Git hooks: NOT INSTALLED`. `status` is the command people actually run, so the one surface a user checks reports health while capture is off.

Reproduced on the released **0.10.4** binary. I let lefthook remove the hooks (`lefthook install -f`, the same thing lefthook's npm postinstall runs, see #2263; #1349 is the config-edit trigger), confirmed Entire's wrapper was gone from `.git/hooks/pre-push`, then ran both commands:

```
$ entire status
● Enabled · branch main
  Agents · Claude Code
  Checkpoints sync to: origin

$ entire doctor
✓ Metadata branches: OK (no remote-tracking metadata ref found)

Git hooks: NOT INSTALLED
  Commits in this repository are not captured as checkpoints.
  Fix: reinstall the managed git hooks (any non-Entire hook is backed up).
  Run `entire doctor --force` to apply it.
✓ Claude Code hook config: OK
```

Nothing is enabled in the sense that matters: no hook will run, so no checkpoint will be captured and none will be pushed. In a repo with pending checkpoints the line goes further and promises "N checkpoints not yet on origin — they sync with your next `git push origin`", which cannot happen.

In the source, `strategy.CheckGitHookState` is what `doctor` uses (`checkGitHooks` in `cmd/entire/cli/doctor.go`). It appears zero times in `cmd/entire/cli/status.go` at v0.10.3, v0.10.4 and current `main`. `status.go` does already render agent hook-config drift (`! <agent> hooks out of date`), so the shape exists; it never consults the git-hook state that `doctor` uses two files away.

### Steps to reproduce

1. Fresh repo with a `lefthook.yml` declaring a `pre-push` command; `git commit`
2. `lefthook install`, then `entire enable`; confirm `.git/hooks/pre-push` contains the `Entire CLI hooks` marker
3. `lefthook install -f` (or delete `.git/hooks/pre-push` by hand; the cause does not matter)
4. `entire status` → `● Enabled … Checkpoints sync to: origin`
5. `entire doctor` → `Git hooks: NOT INSTALLED`

Scripted, re-runnable: `repro-status-blind.sh`. Happy to paste it.

### Entire CLI version

Entire CLI 0.10.4 (Go 1.26.6, darwin/arm64)

### OS and architecture

macOS 26.6.2 (Darwin 25.6.0) arm64

### Agent

Claude Code

### Additional context

**Why this is worth separating from #1349.** #1349 is about hook *ownership*: who wins the file, and how to stop the fight. This is about the *diagnostic* being wrong afterwards, whatever caused it. Any cause has the same effect: a manual `rm`, a `core.hooksPath` change, a fresh clone before `entire enable`, a teammate's tooling. That `doctor` already has this check is the reason I think `status` is worth its own issue: the project has already decided the signal is worth surfacing, and `status` is simply the surface that did not get it.

**Suggested fix.** `status.go` calls `strategy.CheckGitHookState(ctx)` and renders a line for `GitHooksAbsent` / `GitHooksOutdated`, pointing at `entire doctor`. Read-only; the fix stays in `doctor`.

Two judgement calls I would want your steer on before writing it:

1. **Suppress the sync line, or add a warning beside it?** Suppressing removes a false statement; warning keeps the information for someone mid-diagnosis. I lean warning.
2. **`GitHooksAbsent` in a repo that never opted in.** `checkGitHooks` deliberately stays silent unless `settings.IsSetUpAny`, so a diagnostic never nags a repo that did not ask. `status` only prints this block when enabled, so the same guard probably applies for free; worth confirming rather than assuming.

Related: #1349 (how the hooks go missing), #1965 (a different silent-capture failure: hooks present but no session recorded).

peyton-alt:
https://entire.io/gh/entireio/cli/trails/1284

## Problem

Issue #2264 reports that `entire status` says checkpoints sync to `origin` even when Entire git hooks are missing, so capture and push cannot happen. The immediate bug is a false healthy status; the underlying cause in the reproduction is Lefthook overwriting the Entire wrapper during `lefthook install -f` or an npm postinstall.

This PR addresses both layers: it makes status reflect actual hook delivery health, and it installs a durable Lefthook integration so refreshes do not silently remove checkpoint capture. It also emits lifecycle warnings when delivery is unhealthy, so users do not have to remember to run `entire status` to discover a failure. The session-listing and storage-versus-sync changes make the same status snapshot complete and unambiguous.

## Summary

- make native and Lefthook checkpoint-hook installation durable, ownership-aware, and transactional
- surface hook delivery failures proactively and share truthful health diagnostics across status and doctor
- separate storage from sync readiness and keep session reporting complete and read-only

## Verification

- mise run check
- real Codex OAuth E2E: TestLefthookRefreshKeepsAgentCommitCheckpointed
- real Claude Code OAuth E2E: TestLefthookRefreshKeepsAgentCommitCheckpointed
- independent implementation review approved with no remaining actionable findings

Fixes #2264

---

> [!NOTE]
> **Medium Risk**
> Changes how Git hooks are installed, repaired, and reported across commits and push; mistakes could miss checkpoints or mislead sync readiness, but behavior is heavily tested and fails open on warnings.
> 
> **Overview**
> This PR makes Git hook health a first-class part of checkpoint capture and sync reporting, with durable **native** and **Lefthook** integrations that survive manager refreshes instead of waiting for the next agent turn.
> 
> **Hook integration:** `CheckGitHookIntegration` / `EnsureGitHookIntegration` / `RemoveGitHookIntegration` replace the older install-only checks. Lefthook repos get owned `lefthook-local` scripts and config; native hooks remain the bridge when wrappers are missing. `entire doctor`, enable/setup, and `EnsureSetup` all heal through the same coordinator and surface **DEGRADED** / manager-specific reasons.
> 
> **Lifecycle:** Turn start runs hook repair, emits a deduplicated `entire doctor` warning (fingerprinted in session state), and folds it into a single stdout payload for Claude/Codex JSON or a standalone plain-text path for Gemini (deferring context injection when both would collide). Session start can append the same warning to the banner.
> 
> **Status & sessions:** Text and JSON share `buildStatusSnapshot` with `checkpoint_sync_state` (`ready` / `degraded` / `blocked`), `git_hooks` health, and storage backend. Sync is not advertised as healthy when hooks are absent or broken. `entire status` lists every active session read-only (no stale cleanup or finalize-on-read) with richer JSON fields. **pre-push** accepts an optional remote URL argument.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 598bc4f0112de79eb3f64705d9f476e62d5bc347. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/agent/agent.go
+++ b/cmd/entire/cli/agent/agent.go
@@ -409,6 +409,25 @@ type HookResponseWriter interface {
 	WriteHookResponse(message string) error
 }
 
+// StandaloneHookResponseWriter identifies an agent whose user-visible hook
+// response cannot share the same native payload as model context injection.
+// The lifecycle dispatcher emits the warning through WriteHookResponse and
+// retries context injection on a later warning-free turn.
+type StandaloneHookResponseWriter interface {
+	HookResponseWriter
+	RequiresStandaloneHookResponse()
+}
+
+// AsStandaloneHookResponseWriter returns ag when its native protocol requires
+// a standalone user-visible hook response.
+func AsStandaloneHookResponseWriter(ag Agent) (StandaloneHookResponseWriter, bool) {
+	if ag == nil {
+		return nil, false
+	}
+	writer, ok := ag.(StandaloneHookResponseWriter)
+	return writer, ok
+}
+
 // SessionEndBudgeter is implemented by agents whose host enforces a hard
 // wall-clock budget on the session-end hook, because it runs inside the agent's
 // own shutdown sequence rather than between turns.
--- a/cmd/entire/cli/agent/geminicli/lifecycle.go
+++ b/cmd/entire/cli/agent/geminicli/lifecycle.go
@@ -13,10 +13,11 @@ import (
 
 // Compile-time interface assertions for new interfaces.
 var (
-	_ agent.TranscriptAnalyzer = (*GeminiCLIAgent)(nil)
-	_ agent.TokenCalculator    = (*GeminiCLIAgent)(nil)
-	_ agent.HookResponseWriter = (*GeminiCLIAgent)(nil)
-	_ agent.ContextInjector    = (*GeminiCLIAgent)(nil)
+	_ agent.TranscriptAnalyzer           = (*GeminiCLIAgent)(nil)
+	_ agent.TokenCalculator              = (*GeminiCLIAgent)(nil)
+	_ agent.HookResponseWriter           = (*GeminiCLIAgent)(nil)
+	_ agent.StandaloneHookResponseWriter = (*GeminiCLIAgent)(nil)
+	_ agent.ContextInjector              = (*GeminiCLIAgent)(nil)
 )
 
 // WriteHookResponse outputs a hook response message as plain text to stdout.
@@ -37,6 +38,10 @@ func (g *GeminiCLIAgent) WriteHookResponse(message string) error {
 	}
 	return nil
 }
+
+// RequiresStandaloneHookResponse preserves Gemini's plain-text response path;
+// JSON systemMessage is displayed twice by Gemini CLI.
+func (g *GeminiCLIAgent) RequiresStandaloneHookResponse() {}
 
 // InjectionEvent reports that Gemini injects model context at TurnStart (its
 // BeforeAgent hook). Gemini CLI's hook runner merges
--- a/cmd/entire/cli/doctor.go
+++ b/cmd/entire/cli/doctor.go
@@ -622,12 +622,13 @@ func checkGitHooks(cmd *cobra.Command, force bool) error {
 	ctx := cmd.Context()
 	w := cmd.OutOrStdout()
 
-	switch strategy.CheckGitHookState(ctx) {
-	case strategy.GitHooksCurrent:
+	health := strategy.CheckGitHookIntegration(ctx)
+	switch health.State {
+	case strategy.GitHookIntegrationCurrent:
 		fmt.Fprintln(w, "✓ Git hooks: OK")
 		return nil
 
-	case strategy.GitHooksAbsent:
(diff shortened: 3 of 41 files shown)


peyton-alt:
Superseded and closing: the two things this PR did are now split so each is reviewable on its own.

- **#2363** — `fix(status): list every session without mutating state` (+301/−70) — the read-only session-listing half.
- **#2388** — `fix(hooks): register Entire with Lefthook instead of owning its hook files` (+1,413/−60) — the hook half, rewritten against `main` on the mechanism verified against lefthook 2.1.10: Entire registers itself in Lefthook's own config (`extends`) rather than trying to keep a wrapper in `.git/hooks/*`, which Lefthook reclaims at the top of every run.

What this PR's title claimed is worth noting for the record: it says "make checkpoint status reflect hook health", but the bulk of the diff was the durable-hooks work, and it does **not** fix #2237 — the diff touches no `backupExists`/`.pre-entire` line in `hooks.go` and leaves `TestInstallGitHook_DoesNotOverwriteExistingBackup` untouched.

Two defects found while rebuilding that were not in this PR's design, and are fixed in #2388: the integration as designed here made every git hook run Entire **twice** (its own hook, then Lefthook's launcher via the chain call — measured 2× `prepare-commit-msg`/`commit-msg`/`post-commit` per commit), and `LefthookManaged` trusted `lefthook-local.yml` — a file Entire creates itself, so a repo stayed classified Lefthook-managed after Lefthook was removed.

PR #2349, opened 2026-09-10
factoryai-droid: install the cmd.exe hook wrappers on Windows

gtrrz-victor:
https://entire.io/gh/entireio/cli/trails/1289

`factoryai-droid` recorded no checkpoints on Windows. Commits landed normally and not one agent hook fired, so no session ever existed for the commit hook to condense. It passed on Linux and macOS throughout.

## Cause

Entire writes droid's hook command as:

```
sh -c 'if ! command -v entire >/dev/null 2>&1; then exit 0; fi; exec entire hooks factoryai-droid session-start'
```

Droid's **Windows** build hands every hook command to `cmd.exe`; only its macOS/Linux build uses `sh`. cmd.exe reads that string's `>` as its own redirection and `&` as a command separator, so the line is cut apart and `entire` is never executed — with **exit code 0**, which is why nothing reported a problem.

Read out of the shipped droid binaries, and consistent across every version checked:

| droid build | how it runs a hook command |
| --- | --- |
| darwin 0.215.1 | `{command, shell: "sh"}` |
| windows 0.109.1 | `spawn("cmd.exe", ["/c", H])` |
| windows 0.178.0 | `{command, shell: "cmd.exe"}` |
| windows 0.215.1 | `spawn(%ComSpec%, ["/d","/s","/c",H], {windowsVerbatimArguments:true})` |

## Fix

Droid installs the native cmd.exe wrappers whenever the host is Windows, via a new `agent.HookHostIsWindows()`.

Deliberately **not** gated on the existing `agent.UseWindowsProductionHooks`. That probe only establishes that `sh -c 'exit 0'` runs — a command carrying no cmd.exe metacharacters — so a Windows host with Git Bash reports success and keeps the broken wrapper. This is not hypothetical: both the machine used for verification and the GitHub Windows runner have `sh` on PATH, so the probe would have succeeded on each. Codex and Cursor keep the probe, but not because their runners differ — this repo describes those as going through cmd.exe too. Codex keeps it on evidence: it demonstrably passes the Windows nightly with the sh wrapper installed, so whatever its composition does, the wrapper survives it. Cursor has no such evidence, and may share this defect; see Not covered.

Also adds `WrapProductionPlainTextWarningHookCommandForOS`, the missing selector for the one wrapper shape droid uses that had no `ForOS` variant.

Migration needed no new code: `dropStaleEntireHooks` already replaces a managed hook that is not the wanted one, and `IsManagedHookCommand` recognises both Windows shapes. There is a test pinning that, since it is what stops a repo carrying both entries and double-firing.

## Verification

Run [34462110574](https://github.com/entireio/cli/actions/runs/34462110574) — the previously failing test, on a real Windows runner, against real droid, built from this branch's code (`0.0.0-20260910094138-bb2b7c9257cf`): `Total: 1  Passed: 1  Failed: 0`.

At the time, that run needed a throwaway change to `e2e-windows.yml`, which hardcoded claude-code. **That is no longer the case**: the workflow now takes an `agent` input (see the commit list below), so the same run is reproducible from this branch via `workflow_dispatch` with `agent: factoryai-droid`. The recorded run predates the parameterization and was not itself produced through it.

Confirmed independently since, on a real Windows ARM64 machine: a binary built from this branch writes the `cmd.exe` wrappers, and droid runs them. The same `entire enable` from a stale pre-fix binary on the same box writes the sh wrappers — the two shapes side by side, which is also a reminder that `where.exe entire` is worth checking before trusting a behavioural result on a machine with more than one `entire` installed.

Artifact, against the original failure:

| | before | after |
| --- | --- | --- |
| agent hooks in `entire.log` | none | session-start, user-prompt-submit, stop, session-end (×2 each) |
| `prepare-commit-msg` | "no active sessions" | "agent commit trailer added" |
| checkpoint trailers | 0 | 2 |

Additionally, on a separate Windows ARM64 machine: the old wrapper pasted at a cmd prompt exits 0 and never creates `.entire\logs` at all, while the new one reaches hook dispatch and logs a real lifecycle event; and both of droid's shell compositions (0.178.0's and 0.215.1's) were exercised through a byte-exact model of `windowsVerbatimArguments`, since droid nests our stored string inside its own `cmd.exe /d /s /c`.

## The other commits

- **`key the install-test expectations to the host`** — three existing droid tests hardcoded the sh wrapper, so on a Windows host they asserted a command `InstallHooks` no longer writes. CI would not have caught it: the Windows job runs `-run '(Windows|MSYS)'` and none of those names match. Confirmed by same-commit before/after on a Windows box: 13 assertion failures → 0.
- **`nightly-e2e: put factoryai-droid back on the Windows agent list`** — restores the three things the exclusion comment named. Note this workflow installs the *published nightly* and checks out the tag it resolves, so it does not test this branch; droid there is first exercised on the scheduled run after a nightly carrying this fix ships. A dispatch before then would install a pre-fix nightly and fail misleadingly.
- **`pin the droid cmd.exe composition, and correct the gate's doc`** and **`lift the hook-command wrappers out of InstallHooks`** — review responses. The first adds `TestWindowsWrappers_DroidComposition`, which executes the wrappers under a model of droid's own runner rather than Codex's, and corrects `HookHostIsWindows`'s doc: what separates codex and cursor from droid is evidence, not a different runner. The second replaces seven hand-built command literals with `silentHookCommand` / `stopHookCommand`, which also ties the installed verb to the `HookName*` constants the subcommands register.
- **`e2e-windows: take an agent input instead of hardcoding claude-code`** — gives droid a pre-merge Windows check. The bootstrap call, `--agent` and the artifact name parameterize by substitution; the installer does not, so there is a step per agent (droid lands its binary in `%USERPROFILE%\bin`, claude-code in `.local\bin`). The concurrency group gains the agent so a droid run does not cancel a claude-code one. `e2e.yml` forwards it; an unset agent still means claude-code, so the push path is unchanged.
- **Two documentation commits** — `agent-guide.md`'s `InstallHooks` template was teaching the unconditional sh wrapper this PR exists to replace, with no mention of either host gate, and `agent-integration-checklist.md` points every new integration at that template. It now selects per host and gains a "Choosing the hook wrapper" section. The second commit fixes the same template's hand-assembled config path, which bypassed `agent.OpenHookConfig`.

## Existing Windows installs need a re-enable

There is no drift signal for this. Droid implements no `HookFreshness` (claude-code, codex, opencode and pi do), so `status` and `doctor` will not report the stale sh wrapper, and `AreHooksInstalled` still answers true for it. A repo already enabled on Windows keeps its broken hooks until someone re-runs `entire enable`, which is what performs the migration. Giving droid a `HookFreshness` is the durable fix and is not in this PR.

### A committed `.factory/settings.json` is now host-specific

The same missing signal has a second consequence, and this PR is what makes it
visible. Before, one file worked on Unix and silently no-op'd on Windows. Now a
Windows-written config carries `cmd.exe /d /s /c "…"` into a file a Unix
teammate's droid hands to `sh`: `cmd.exe: not found`, exit 127 — and the bare
Stop form is an `sh` syntax error, also non-zero. That is a
SessionStart/UserPromptSubmit hook returning non-zero, which is the one
condition the silent wrapper exists to avoid.

It self-heals: `dropStaleEntireHooks` replaces the other host's form on that
host's next `entire enable`. So the exposure is the window between enables, plus
a ping-pong diff for teams that commit the file.

Codex and cursor carry the same exposure, but only flip on a Windows host
*without* a working sh, which is rare. Droid flips on every Windows host, so it
is the first agent where this is the common case rather than the corner. A
`HookFreshness` covers this and the stale-install case above together — it is
the one signal that would say "this wrapper does not match this host".

## Version scope

Verified end to end on droid 0.178.0 and 0.215.1. 0.109.1 spawned `cmd.exe /c <H>` *without* `windowsVerbatimArguments`, so libuv backslash-escapes the inner quotes of the nested silent wrapper into something cmd.exe does not parse. That version stays broken — equally broken as before, since the sh wrapper never survived its cmd.exe either. Switching every hook to the bare wrapper shape would likely fix it; not taken here because it would trade the shape verified on current droid for one that is not.

## Not covered

The subagent path (`PreToolUse`/`PostToolUse[Task]`) is not exercised by this test.

~~No Windows workflow in the repo builds from source and can target an agent other than claude-code.~~ Closed by the `e2e-windows` commit below: it now takes an `agent` input, so a Windows-specific agent fix has a pre-merge check that builds from source. Not yet exercised through that path — the verification run above predates it.

Cursor may share this defect. Its `InstallHooks` comment says it too spawns hooks through cmd.exe on Windows, in which case its `UseWindowsProductionHooks` probe is wrong for the same reason droid's was — but it is excluded from the Windows matrix (no tmux), so there is no evidence either way. Settling that means reading its runner the way droid's was read; deliberately out of scope here.


cmd/entire/cli/agent/hook_command.go
```diff
@@ -375,6 +375,26 @@ func defaultSHHookWrapperWorks(ctx context.Context, command string) bool {
...
+// Windows whatever else is installed. Factory Droid is one: its Windows build
+// runs each hook command as an argument of cmd.exe, while its macOS/Linux
+// build runs the same string under sh. For such an agent
+// UseWindowsProductionHooks answers the wrong question — its probe only
+// establishes that `sh -c 'exit 0'` runs, and that command carries no cmd.exe
+// metacharacters, so a host with Git Bash reports success while the real sh
+// wrapper, which is full of `>` and `&`, is still cut apart by cmd.exe before
+// any sh sees it.
+//
+// Agents that resolve a shell themselves (Codex, Cursor) keep
```
gtrrz-victor: This claim is contradicted by the repo's own notes about those two runners:

- `hook_command.go:170` — "Codex already runs hook commands through cmd.exe /C"
- `hook_command_exec_windows_test.go:15` — "mirrors Codex's Windows command runner: `cmd.exe /C` followed by the raw, quoted hook command"
- `cursor/hooks.go:125` — "Cursor spawns hook commands through the native OS shell (cmd.exe on Windows)"

By our own description neither resolves a shell itself. And `cmd /C "sh -c 'if ! command -v entire >/dev/null 2>&1; …'"` has its outer quotes stripped (special characters present, so cmd takes the "otherwise" branch of its quote rule), which makes `>/dev/null` a cmd redirection to a nonexistent path — the statement dies before `sh` is ever reached, sh installed or not. That is this PR's argument, applied to the other two agents; `TestUseWindowsProductionHooks`'s `"windows with working sh keeps sh wrappers"` case pins exactly the behaviour the PR argues is wrong.

Suggest cutting the reassurance about Codex/Cursor from this doc comment and saying only what droid needs, then opening a follow-up. Cursor looks like the likelier live bug: its wrapper is the *silent* one and windows-latest has Git Bash on PATH, so it would present as the same "exit 0, nothing happens" shape this PR chased — and its Windows e2e is excluded for an unrelated reason (tmux), so nothing would have caught it.

cmd/entire/cli/agent/factoryaidroid/hooks_test.go
```diff
@@ -754,3 +776,122 @@ func hooksInstalledNow(t *testing.T, ag interface {
...
+	assertFactoryHookExists(t, settings.Hooks.PreCompact, "", want["pre-compact"], "PreCompact")
+}
+
+// TestInstallHooks_WindowsUsesCmdWrappersDespiteWorkingSh pins that droid picks
+// the native cmd.exe wrappers on a Windows host even when a POSIX sh is
+// runnable there. The probe deliberately reports a working sh: droid's Windows
+// build never spawns sh for a hook, so its presence must not keep the sh
+// wrapper — which cmd.exe would cut apart at the first `>`, firing no hook at
+// all. Mutates the shared probe, so no t.Parallel().
+func TestInstallHooks_WindowsUsesCmdWrappersDespiteWorkingSh(t *testing.T) {
```
gtrrz-victor, line 808: These pin *which* wrapper is chosen, which is the right thing to pin here — but the claim the fix actually rests on (that the cmd.exe wrapper survives droid's own composition) is still verified only manually, off-repo.

There is already a harness for exactly that: `hook_command_exec_windows_test.go` executes the wrappers for real, is Windows-only, and `ci.yml:86` runs `-run '(Windows|MSYS)'` on windows-latest. A droid-composition case there — `%ComSpec% /d /s /c <H>` verbatim, plus the `shell: "cmd.exe"` → `/d /s /c "<H>"` form — would turn the byte-exact model you ran on the ARM64 box into a pre-merge check, and would answer the "Not covered" paragraph about there being no source-built Windows workflow. Just name it so it matches the filter.

cmd/entire/cli/agent/factoryaidroid/hooks.go
```diff
@@ -123,14 +123,26 @@ func (f *FactoryAIDroidAgent) InstallHooks(ctx context.Context, force bool) (int
...
+	// the sh wrapper's `>` and `&` as its own redirections and separators, so
+	// the command is cut apart and no `entire hooks` process is ever created —
+	// the session never starts, so a commit produces no checkpoint and nothing
+	// says why.
+	//
+	// Unlike Codex and Cursor this is NOT gated on agent.UseWindowsProductionHooks:
+	// droid never hands a hook to sh on Windows, so whether a working sh exists
+	// there changes nothing. See agent.HookHostIsWindows.
+	useWindowsHooks := agent.HookHostIsWindows()
+	sessionStartCmd := agent.WrapProductionSilentHookCommandForOS("entire hooks factoryai-droid session-start", useWindowsHooks)
```
gtrrz-victor: Worth noting in the comment above: the silent wrapper double-nests cmd.exe under droid. Droid already spawns `cmd.exe /d /s /c <H>`, and `H` is itself `cmd.exe /d /s /c "…"`.

That is fine on 0.178/0.215, but it is the shape that breaks on 0.109.1's `spawn("cmd.exe", ["/c", H])`: non-verbatim, so libuv escapes the inner quotes as `\"`, cmd never strips a backslash, and the inner command name is garbage — exit ≠ 0, no hook, quietly. The *bare* form used for Stop survives that path, because it carries no quotes for libuv to escape.

Either say 0.109.1 is out of scope (it was equally broken before, so there is no regression) or give droid a bare silent variant — `IsManagedHookCommand` already accepts both prefixes, so migration would cost nothing.

cmd/entire/cli/agent/factoryaidroid/hooks_test.go
```diff
@@ -754,3 +776,122 @@ func hooksInstalledNow(t *testing.T, ag interface {
...
+
+	assertDroidWindowsHooks(t, readFactorySettings(t, tempDir))
+}
+
+// TestInstallHooks_WindowsMigratesShWrappers pins that a repo whose hooks were
+// installed from a non-Windows host is migrated to the cmd.exe wrappers by a
+// plain (non-force) reinstall, replacing the sh entries rather than leaving
+// both — two entries would fire the same hook twice. Mutates the shared probe,
+// so no t.Parallel().
+func TestInstallHooks_WindowsMigratesShWrappers(t *testing.T) {
```
gtrrz-victor, line 835: The migration works on the next `InstallHooks`, but nothing prompts an existing install to run one. Droid implements no `HookFreshness`, so `OutdatedHookAgents` skips it and neither `entire status` nor `doctor` reports the stale sh wrapper, while `AreHooksInstalled` (matching via `IsManagedHookCommand`) reports installed. A Windows droid user who upgrades the CLI therefore stays broken until they happen to re-run `entire enable`.

Cursor has the same gap, so a `CheckHookConfig` for droid is fair as a follow-up — but the PR body should say "re-run `entire enable`" under a Migration note, since today it reads as if the fix reaches existing repos on its own.

gtrrz-victor (commented):
The fix itself looks right: droid's Windows build hands the string to `cmd.exe`, so `HookHostIsWindows()` rather than the sh probe is the correct predicate, and the migration path (`dropStaleEntireHooks` + `IsManagedHookCommand` accepting both shapes, pinned by the per-type count assertion) holds up.

Verified locally: `go build ./...`, `go test` on `agent`, `factoryaidroid` and `cli -run 'Droid|Factory|Hook'` all pass; golangci-lint 2.11.3 clean on `./cmd/entire/cli/agent/...`; gofmt clean. Workflow side: the `Smoke test factoryai-droid` step already exists gated on `contains(AGENTS, ' factoryai-droid ')` with `FACTORY_API_KEY` present, and the 06:00-tag → 09:00-e2e cron ordering means the first scheduled run after merge does pick up a nightly carrying the fix.

Four comments inline, none of them blocking the droid change. Two nits not worth their own threads: in `TestInstallHooks_WindowsMigratesShWrappers` the `goos` local is assigned twice and only ever passed as an argument (pass the literals), and `droidWindowsHookCommands`'s stringly-keyed map is built only to be read back seven times — direct wrapper calls in `assertDroidWindowsHooks` would be shorter and typo-proof.

gtrrz-victor (commented):
Second round, on `005b9eac7`..`d99ea12c0`. The code is clean now; one stale sentence left in this description.

**The lint fix is better than a comment trim.** Lifting the seven constructions into `silentHookCommand` / `stopHookCommand` also closes a drift risk that existed before this PR: the installed command's verb and the subcommand `HookNames()` registers were independent strings, and are now the same `HookName*` constants. I checked each substitution against the old literal — `session-start`, `session-end`, `stop`, `user-prompt-submit`, `pre-tool-use`, `post-tool-use`, `pre-compact` — all identical, so the refactor is behaviour-preserving.

The test side is split the right way: `droidHookCommand`/`droidStopHookCommand` delegate (host-dependent, and the form is pinned elsewhere), while `droidWindowsHookCommand`/`droidWindowsStopHookCommand` keep the literal `"entire hooks factoryai-droid " + verb` and call the Windows wrapper outright. That is what keeps the prefix and every verb pinned in ordinary CI rather than only asserted against themselves — worth not "tidying up" later.

Verified on `d99ea12c0`: `golangci-lint run ./cmd/entire/cli/agent/...` at 2.11.3 with no baseline (the exact shape that caught maintidx) reports 0 issues, `gofmt` clean, `GOOS=windows go vet` clean, `agent` + `factoryaidroid` tests pass. CI agrees: `lint`, `test-core`, all three `test-integration` shards, both canaries, `test-codex-hooks` and `test-windows` are green.

**One thing to fix — the Fix section still carries the sentence the code correction removed:**

> Codex and Cursor keep the probe; they resolve a shell themselves, so for them a working sh really does mean the sh wrapper runs.

`005b9eac7` replaced exactly that claim in `HookHostIsWindows`'s doc, and the new Not-covered section says the opposite — that cursor's own comment describes it as spawning through cmd.exe, so its probe may be wrong the same way. As it stands the description contradicts itself, and it is the paragraph a reader reaches first. Suggest matching the code: the probe is kept for Codex because Codex demonstrably passes the Windows nightly with the sh wrapper installed, not because its runner is different.

The two remaining red checks are not this PR. `Analyze workflows` is the CodeQL default-setup-vs-committed-workflow conflict (all four `Analyze (…)` jobs and `CodeQL` itself pass; the upload is what is rejected), and `Entire Gates` lists exactly two reasons — that failing check, and no approval yet. Nothing else outstanding from round one: findings 2, 3 and 4 are closed as pinned, scoped and documented respectively.

cmd/entire/cli/agent/hook_command.go
gtrrz-victor: Fixed in 005b9eac7 — you were right, and this was a claim I reasoned into rather than checked.

I inferred "they resolve a shell themselves" from the fact that Codex and Cursor use the probe, and never read the three places that say otherwise. The doc now states what the code actually supports: what separates them is evidence, not mechanism. Codex passes the Windows nightly with the sh wrapper installed, so whatever its composition does the wrapper survives it; Cursor has no such evidence, being excluded from that matrix for want of tmux, and may share this defect. The comment says to settle it by reading the runner before touching cursor's gate.

d99ea12c0 moved the droid half of this to `silentHookCommand`, so the rationale now sits next to the call it explains.

Not changing cursor here: it needs the same binary-reading evidence gathered for droid, and it has no Windows coverage at all, so a speculative gate swap could break a working path with nothing to catch it. Recorded in the PR body as a suspected defect plus the method.

cmd/entire/cli/agent/factoryaidroid/hooks_test.go
gtrrz-victor: Done in 005b9eac7 — good catch that the infrastructure was already there.

Added `TestWindowsWrappers_DroidComposition` with a `runDroidWrapper` modelling droid's runner rather than Codex's: `/d /s /c`, and the command appended verbatim instead of quoted, so the stored wrapper genuinely nests one cmd.exe inside another. Five cases — silent present / exit-code / absent, plus the plain-text warning both ways, since Stop uses the bare shape and is the one wrapper whose behaviour differs under this runner. Both helpers now share one body.

It matches the `(Windows|MSYS)` filter, and `test-windows` has since passed at this head, so the composition is pinned pre-merge rather than hand-checked.

cmd/entire/cli/agent/factoryaidroid/hooks.go
gtrrz-victor: Scoped explicitly rather than changed — 005b9eac7, then moved to `silentHookCommand`'s doc in d99ea12c0.

Your libuv reasoning matches mine: 0.109.1 spawns without `windowsVerbatimArguments`, so the inner quotes are backslash-escaped into something cmd.exe does not parse.

I took the scope-it-out option over the bare variant deliberately. The nested shape is the one verified end to end on 0.178.0 and 0.215.1 — by hand on a Windows box and by a real passing e2e run. Switching to bare would trade a verified shape for an inferred one, to rescue a version that is already broken today and is no worse under this PR, since the sh wrapper never survived its cmd.exe either. The comment records that, and that bare is the fix if anyone still runs 0.109.1.

cmd/entire/cli/agent/factoryaidroid/hooks_test.go
gtrrz-victor: Confirmed and documented in the PR body.

Checked it: droid implements no `HookFreshness`, while claude-code, codex, opencode and pi do. So `status` and `doctor` stay silent about the stale sh wrapper and `AreHooksInstalled` still answers true for it.

The PR body now says an existing Windows install keeps its broken hooks until someone re-runs `entire enable`, and that giving droid a `HookFreshness` is the durable fix and is not in this PR. Left as a follow-up rather than added here — it is a new capability for the agent, not part of the wrapper fix, and it wants its own drift tests.

Soph (approved):

PR #2352, opened 2026-09-10
cursor: install the cmd.exe hook wrappers on Windows, and tolerate a BOM on hook input

### LINKED ISSUE #1424: Cursor hooks not firing on Windows 11 (Scoop install)

gtrrz-victor:
Makes Cursor hooks fire on Windows, so sessions and checkpoints get recorded there. Fixes #1424.

Two things stopped them: the wrapper Entire installs cannot run under Cursor's PowerShell hook runner, and the payload arrives with a UTF-8 BOM our JSON decode rejects. Both fixed here; the BOM fix is at the shared chokepoint, so it covers every agent.

Trail: https://entire.io/gh/entireio/cli/trails/1291

### DIFF
--- a/cmd/entire/cli/agent/cursor/hooks.go
+++ b/cmd/entire/cli/agent/cursor/hooks.go
@@ -62,6 +62,52 @@ func cursorHookConfig(ctx context.Context) (*agent.HookConfigFile, error) {
 	return agent.OpenHookConfig(worktreeRoot, (&CursorAgent{}).HookConfigRelPath()) //nolint:wrapcheck // agent.HookConfigFile already names the file in its error
 }
 
+// hookCommandPrefix is what every hook command Entire writes for Cursor begins
+// with; the verb is one of the HookName* constants.
+const hookCommandPrefix = "entire hooks cursor "
+
+// silentHookCommand wraps one hook verb in the silent production wrapper for
+// this host.
+//
+// Cursor runs every hook command through **PowerShell** on Windows — not
+// cmd.exe. Read out of the shipped builds (Cursor IDE 3.19.19 win32/x64, and
+// the native cursor-agent CLI windows/x64 2026.09.08-6caf4ff), the spawn is
+// `<pwsh|powershell> [-NoProfile -NonInteractive -ExecutionPolicy Bypass] -c
+// "$OutputEncoding = [System.Text.Encoding]::UTF8; Get-Content -LiteralPath
+// '<tmp>\cursor-hook-payload-*.json' -Raw | & { $input | <command> }"`, with the
+// stored command inserted verbatim. Both Windows runners read the same
+// .cursor/hooks.json, and both compose that identically.
+//
+// So the sh wrapper is not mangled the way droid's is: PowerShell single quotes
+// are literal, and the whole script reaches sh as one argument. It fails for a
+// different reason — `sh` has to resolve in the PATH of the PowerShell child
+// Cursor spawns, and agent.UseWindowsProductionHooks probes for sh in the
+// `entire enable` process instead. Git for Windows installs sh.exe under
+// …\Git\usr\bin and …\Git\bin, neither of which is on the machine PATH (only
+// …\Git\cmd is), while MSYS translates PATH for native children — so running
+// `entire enable` from Git Bash passes the probe and writes a wrapper Cursor
+// cannot run. That is issue #1424's reported environment (Windows 11 + Git
+// Bash), and no stronger probe COMMAND fixes it, because the probe is measuring
+// the wrong process. Hence agent.HookHostIsWindows rather than the probe.
+//
+// The failure is invisible rather than swallowed: a CommandNotFoundException
+// raised inside `& { $input | … }` does not set powershell.exe's exit code, so
+// Cursor sees exit 0 and has no error to report. The same command standalone
+// exits 1. TestWindowsWrappers_CursorComposition asserts this on a real Windows
+// runner — the sh-wrapper subtest is exactly this case.
+//
+// This moves the PATH dependency rather than removing it — the wrapper opens
+// with `where.exe entire`, so it needs `entire` on the child's PATH, which every
+// documented install method satisfies and Git for Windows' sh does not.
+//
+// Not fixed by any wrapper: the cursor-agent CLI picks a bash shell executor
+// whenever MSYSTEM is set (i.e. started from Git Bash) while still composing the
+// PowerShell script above, so hooks cannot run there at all. That is Cursor's to
+// fix, and it does not affect the IDE.
+func silentHookCommand(verb string, useWindows bool) string {
+	return agent.WrapProductionSilentHookCommandForOS(hookCommandPrefix+verb, useWindows)
+}
+
 // InstallHooks installs Cursor hooks in .cursor/hooks.json.
 // If force is true, removes existing Entire hooks before installing.
 // Returns the number of hooks installed.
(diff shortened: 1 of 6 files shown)


cmd/entire/cli/agent/hook_command_exec_windows_test.go
```diff
@@ -12,6 +12,119 @@ import (
...
+// wrapped command's stdin.
+//
+// It cannot share runWrapperCmdLine. That builds a cmd.exe command line via
+// SysProcAttr.CmdLine because cmd.exe does not follow the standard argv
+// unquoting rules; Cursor spawns PowerShell with a real argv, so the ordinary
+// exec.Command path is the faithful model. And Cursor is the only runner that
+// delivers the payload on the wrapped command's stdin, which is what stdinSeen
+// exists to check — a wrapper can run and still starve the hook of its input.
+//
+// PowerShell is resolved the way Cursor resolves it: pwsh, else powershell,
```
gtrrz-victor: This claims a resolution order the test cannot reach. `setWrapperPATH` runs first and scrubs `PATH` down to System32 (plus the stub dir), so by the time these `LookPath` calls run, `pwsh.exe` is unreachable by construction and the shell is always System32's `powershell.exe` — Windows PowerShell 5.1. The third tier is not implemented either: if both lookups fail, `shell` stays the bare string `"powershell.exe"` rather than the System32 path, and `exec.Command` then resolves it from the same scrubbed `PATH`.

Pinning to 5.1 is the right choice — it is where the BOM comes from, it is the weaker shell, and leaving it to `LookPath` would silently pick pwsh on a runner that has it and 5.1 on one that does not, which is a test that means something different depending on where it runs. So the code is right and the comment is wrong. Say that the scrub pins this to Windows PowerShell 5.1 and that the pwsh branch is therefore not exercised here, which is what Not covered already says — as written, a reader of this helper would conclude the opposite.

cmd/entire/cli/agent/cursor/hooks_test.go
```diff
@@ -617,6 +626,14 @@ func TestInstallHooks_DropsLegacyHookAlongsideCurrent(t *testing.T) {
 // InstallHooks writes. A stale committed config is how the pi extension ended up
 // invoking a launcher script that had been deleted.
 func TestCommittedDogfoodHooksIsCurrent(t *testing.T) {
+	// The committed file can only carry one wrapper form, and it carries the sh
+	// one — correct for this repo's own CI, which runs on Linux and macOS. On a
+	// Windows host InstallHooks writes the cmd.exe form, so the comparison would
+	// fail on a difference that is the intended behaviour. Codex has the same
+	// condition for the same reason.
+	if agent.HookHostIsWindows() {
+		t.Skip("committed .cursor/hooks.json holds the sh wrappers; a Windows host installs the cmd.exe ones")
```
gtrrz-victor, line 637: The skip is right, and its consequence is worth one line in the PR body: on a Windows host `entire enable` in THIS repo will now rewrite the committed `.cursor/hooks.json` to the cmd.exe wrappers, so a Windows contributor gets a modified tracked file in `git status` and must not commit it. Under the probe gate that only happened on a box with no reachable `sh`; now it happens on every Windows host, and skipping this test is exactly what removes the thing that would otherwise have told them.

Not an argument against the change — the alternative is a committed file that is wrong for half the population either way. Just name it, since it will be met by whoever next runs `entire enable` on Windows.

cmd/entire/cli/agent/cursor/hooks_test.go
```diff
@@ -58,78 +58,69 @@ func TestInstallHooks_FreshInstall(t *testing.T) {
...
+// TestInstallHooks_WindowsMigratesShWrappers verifies that a repo enabled
+// before this change — carrying sh-wrapped hooks — is migrated by a non-force
+// reinstall on a Windows host, rather than gaining a second entry beside the
+// stale one. Two entries for one hook type double-fire, which is why the count
+// is asserted per type and not just the command.
+//
+// The sh config is seeded by installing as a non-Windows host, which is exactly
+// what wrote it. Mutates the shared probe, so no t.Parallel().
+func TestInstallHooks_WindowsMigratesShWrappers(t *testing.T) {
+	goos := "darwin"
```
gtrrz-victor: `goos` is assigned once and read once, by the call on the next line. This is the same nit I raised on #2349, where droid's `TestInstallHooks_WindowsMigratesShWrappers` — same test name, sibling package, and the PR this one is stacked on — now passes the literals. Worth matching, so the two do not read as deliberately different.

gtrrz-victor (commented):
Reviewed both commits. The investigation behind this is the strongest part and it survived into the text: PowerShell rather than cmd.exe with the composition read per-platform, the probe measuring the wrong process's PATH, the failure being invisible at the exit-code level rather than swallowed, and the PATH claim cited from the Git for Windows on-disk layout plus the absent-`sh` case rather than from a box where scoop happened to supply `sh.exe`. `HookHostIsWindows`'s doc now carries two mechanisms for one predicate with codex's exemption kept on evidence, which is what the droid PR's version should have said in the first place.

The BOM half reads correctly against the contract we agreed: `Peek(1)` gates the three-byte peek, reader errors are not propagated so `"empty hook input"` survives (asserted in 16 places across 6 files, so that mattered), `StdinLooksInteractive` still runs on the raw `stdin` ahead of the wrapping, the loop is justified by both measured producers, and the stripping cost is stated with the `#1398` test's claim scoped to valid payloads in the same breath. `skipUTF8BOM` cannot spin on a partial BOM either — a failed `Peek(3)` returns the reader rather than looping.

Verified on `8ab5086`: `golangci-lint` 2.11.3 with no baseline over `./cmd/entire/cli/agent/...` reports 0 issues, `gofmt` clean, `GOOS=windows go vet` clean, and the `agent` and `cursor` packages pass. (A first run showed `TestParseHookEvent_SessionEnd_CLINoTranscriptPath` failing; that is an artifact of my running from a `git archive` copy with no repository — it passes once the copy is `git init`ed, and the test is untouched here.)

Four comments inline. Nothing blocking; the first is the only one I would want changed before merge.

On the single-PR deviation: agreed it is your user's call, and the cost is the one we named — the BOM commit is the only thing here that helps a Windows Cursor user today, and it is now waiting on #2349's approval and the CodeQL default-setup conflict. Worth saying that in the description too, so whoever clears those knows what is behind them.

cmd/entire/cli/agent/cursor/hooks_test.go
```diff
@@ -142,13 +133,15 @@ func TestInstallHooks_WindowsProbeFlipMigratesCleanly(t *testing.T) {
...
+	// Seed the sh-wrapped config an earlier version installed.
 	if _, err := ag.InstallHooks(context.Background(), false); err != nil {
-		t.Fatalf("first InstallHooks() error = %v", err)
+		t.Fatalf("seed InstallHooks() error = %v", err)
 	}
 
-	// sh stops working; reinstall WITHOUT force.
-	shWorks = false
+	// Same repo, now on a Windows host; reinstall WITHOUT force.
+	agent.SetWindowsHookProbeForTesting("windows", func(context.Context, string) bool {
```
gtrrz-victor: This discards the restore function, which is *correct* — the `t.Cleanup` above captured the real GOOS and probe before the first override, so it restores them whatever happens here — but it is not obvious, and droid's sibling test registers both restores. Either match the sibling or say in a clause why one restore is enough; as it stands it reads like a dropped return value.

cmd/entire/cli/agent/hook_command_exec_windows_test.go
gtrrz-victor: Right, and fixed on the revision before this one — the runner now names Windows PowerShell 5.1 absolutely (windowsPowerShellPath) instead of resolving it, and the comment says the pwsh branch is therefore not exercised. 24d7962b cleans up the last of the old premise: setWrapperPATH still claimed System32 supplies powershell.exe.

cmd/entire/cli/agent/cursor/hooks_test.go
gtrrz-victor: Done — the test passes the literals, same as droid's sibling.

cmd/entire/cli/agent/cursor/hooks_test.go
gtrrz-victor: Done — the restore is captured and registered with t.Cleanup, matching droid's sibling.

cmd/entire/cli/agent/cursor/hooks_test.go
gtrrz-victor: Named, but in the skip comment rather than the PR body (24d7962b/c4120b12f) — the body is deliberately minimal now, and this outlives it there.

It also stops being invisible: c4120b12f gives Cursor a CheckHookConfig, so `entire status` and `entire doctor` report the committed sh config as drift on a Windows host instead of saying nothing. That was finding 01M25NVVVSEA.

cmd/entire/cli/agent/hook_command_exec_windows_test.go
```diff
@@ -12,6 +12,119 @@ import (
...
+	pathEntries := []string{filepath.Join(sysRoot, "System32")}
+	stubDir := ""
+	if entirePresent {
+		stubDir = t.TempDir()
+		if err := os.WriteFile(filepath.Join(stubDir, "entire.bat"), []byte(stubBody), 0o700); err != nil {
+			t.Fatalf("write entire stub: %v", err)
+		}
+		pathEntries = append([]string{stubDir}, pathEntries...)
+	}
+	t.Setenv("PATH", strings.Join(pathEntries, ";"))
```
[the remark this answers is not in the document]
gtrrz-victor: The failure itself was fixed before this: runCursorWrapper no longer calls LookPath after the scrub, it names System32\WindowsPowerShell\v1.0\powershell.exe absolutely, so the test cannot fatal on an unresolved shell.

The premise was still sitting in setWrapperPATH's comment, which claimed System32 supplies powershell.exe. Corrected in 24d7962b.

cmd/entire/cli/agent/cursor/hooks.go
```diff
@@ -62,6 +62,51 @@ func cursorHookConfig(ctx context.Context) (*agent.HookConfigFile, error) {
...
+// This moves the PATH dependency rather than removing it — the wrapper opens
+// with `where.exe entire`, so it needs `entire` on the child's PATH, which every
+// documented install method satisfies and Git for Windows' sh does not.
+//
+// Not fixed by any wrapper: the cursor-agent CLI picks a bash shell executor
+// whenever MSYSTEM is set (i.e. started from Git Bash) while still composing the
+// PowerShell script above, so hooks cannot run there at all. That is Cursor's to
+// fix, and it does not affect the IDE.
+func silentHookCommand(verb string, useWindows bool) string {
+	return agent.WrapProductionSilentHookCommandForOS(hookCommandPrefix+verb, useWindows)
```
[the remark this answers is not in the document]
gtrrz-victor: Correct about cmd.exe, and worth stating precisely: cmd.exe resolves both `where.exe` and the bare `entire` from the current directory before PATH, and /d only disables AutoRun. The POSIX wrapper did not do this.

It is not introduced here, though. WrapWindowsProductionSilentHookCommand is the shared wrapper Codex already installs on Windows today, and droid gains it in #2349 — Cursor is the third caller. Hardening it at this call site would leave the other two exposed, so it belongs in hook_command.go: an absolute %SystemRoot%\System32\where.exe, plus NoDefaultCurrentDirectoryInExePath for the bare `entire`, which needs checking against cmd.exe's actual search order before we lean on it.

Tracked as its own finding on the trail rather than fixed here.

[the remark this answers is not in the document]
gtrrz-victor: Fixed in #2487, on main so Codex, droid (#2349) and Cursor all inherit it — 678abfbd7.

where.exe is now named under %SystemRoot%, and NoDefaultCurrentDirectoryInExePath takes the current directory out of cmd.exe's search for the bare `entire`. Since the docs do not say whether cmd.exe re-reads that variable mid-line, there is a test that plants an entire.bat in the working directory and asserts against a real cmd.exe on the Windows runner that it never runs.

PR #2361, opened 2026-09-10
feat: add OpenCode as a summary provider

peyton-alt:
https://entire.io/gh/entireio/cli/trails/1298

OpenCode is supported for session tracking but cannot currently generate summaries or tailor runners. Add it to the existing summary-provider selection and validation so `entire configure --summarize-provider opencode --summarize-model openai/gpt-5.6-sol` works for `explain --generate` and `runner setup --run`. Omitting the model uses OpenCode's default; Pi and the existing picker are unchanged.

The adapter sends prompts through stdin, parses OpenCode JSON events (including API errors emitted with exit code zero), and runs with a temporary directory and a dedicated agent that denies tools and disables sharing. Existing provider/auth/model configuration is retained, including JSONC inline configuration.

Validation: affected unit tests and independent review passed. Live calls through OpenCode with ChatGPT (`openai/gpt-5.6-sol`) passed both the real summary parser and the actual `runner setup --run` flow in an isolated fixture repository, including a saved runner and preserved placeholders. Pre-push lint passed. The complete unit/integration race suite passed in a space-free verification checkout; canary validation is delegated to CI at the author’s request.

---

> [!NOTE]
> **Medium Risk**
> Spawns external `opencode` with merged user config for summaries; subprocess and config parsing bugs could affect explain/runner flows, but scope is isolated from git hooks and repo mutation.
> 
> **Overview**
> **OpenCode** can now power `explain --generate` and runner tuning the same way as other summary CLIs, selectable via `entire configure --summarize-provider opencode` (optional `--summarize-model`).
> 
> The new adapter implements `TextGenerator` by running `opencode run` in an isolated temp directory with prompts on stdin, parsing JSON `text` events (and failing on embedded API errors even when the process exits 0). It injects a one-off agent config via `OPENCODE_CONFIG_CONTENT` that denies tools and disables sharing while preserving the user’s provider/model settings (including JSONC), and extends the shared isolated CLI helper with optional environment overrides.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8a3ee754d1799e428b0f8d1365ab4a901945b117. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


gtrrz-victor:
## Review

I checked the adapter's assumptions against the installed `opencode` 1.18.30 binary rather than taking them on trust. The mechanics hold up well — three findings, and a proposed change that dissolves one of them plus the dependency question.

<details>
<summary><b>Verified correct against the real binary</b></summary>

- `run --format json --dir --agent --title --model` all exist; stdin is read when stdout isn't a TTY (`process.stdin.isTTY ? void 0 : await Bun.stdin.text()`).
- Event shapes match exactly. `text` is emitted only for **completed** parts (`J.type==="text" && J.time?.end`), so joining parts is right and there are no streaming deltas. The `error` event's `data.message` → `name` fallback in `parseOpenCodeGeneration` mirrors OpenCode's own fallback line for line.
- The generated config is schema-valid. Confirmed validation is live by feeding a bogus permission action — OpenCode rejected it with `Expected PermissionActionConfig ... agent.<name>.permission.*`, which also proves `{"*":"deny"}` is an accepted shape. `share:"disabled"` is one of the three literals; `mode:"primary"` and inline `prompt`/`description` match the Agent struct.
- `OPENCODE_CONFIG_CONTENT` is an **additive last layer**, not a replacement — the global `~/.config/opencode/*.json` still loads (only default-file *creation* is suppressed). So the read-merge-rewrite really does avoid clobbering a user's inline config.
- `OPENCODE_AUTO_SHARE` is parsed as a boolean, so `=false` means off. A presence-check would have silently *enabled* sharing.
- Warnings that could corrupt the JSON stream — "agent not found, falling back", "Configuration is invalid" — all go to **stderr**. And without `--auto`, a permission request is **auto-rejected** (`reply:"reject"`), not left hanging, so omitting `--auto` is correct and there's no non-TTY stall.
- `append(os.Environ(), overrides...)` genuinely gives overrides precedence: `dedupEnvCase` iterates in reverse "to preserve the last occurrence of each key". The new doc comment is accurate.

</details>

---

### 1. README now contradicts the feature

`README.md:618`:

> `opencode` and `factoryai-droid` cannot generate summaries.

And `README.md:612` lists `claude-code`, `codex`, `copilot-cli`, `cursor`, `gemini`, `pi` without `opencode`. Both in-code equivalents were updated (`explain_summary_provider.go:113`, `setup.go:797`) — this is the user-facing doc that still names opencode as unsupported.

### 2. The tests fail on an OpenCode user's machine

`GenerateText` reads `os.Getenv("OPENCODE_CONFIG_CONTENT")`, and the tests are `t.Parallel()` so they inherit the developer's real environment:

```
OPENCODE_CONFIG_CONTENT='{"agent":1}' go test ./cmd/entire/cli/agent/opencode/
--- FAIL: TestOpenCodeGenerateOutput/api_error_after_text
    error = reading OpenCode agent configuration: json: cannot unmarshal number ... want "API key is invalid."
```

Six subtests plus `TestOpenCodeGenerateText`. That's exactly the population likely to have the var set. `t.Setenv` is unavailable under `t.Parallel()`, so the seam that needs breaking is the `os.Getenv` call itself.

### 3. The friendly error path is mostly unreachable

The comment at `generate.go:85` says OpenCode "can emit an API error even when the process exits successfully" — true, but the case I could actually reproduce exits **1**:

```
EXIT=1
stdout: {"type":"error",...,"error":{"name":"UnknownError","data":{"message":"..."}}}
stderr: (empty)
```

On non-zero exit `err != nil`, so `parseOpenCodeGeneration` never runs. `RunIsolatedTextGeneratorCLI` falls back to stdout for its detail, so the user gets the raw JSON event rather than `OpenCode: <message>`. A real fix needs the shared helper to hand stdout back on error (it returns `""` plus a byte count today), so leaving it is reasonable — but the comment shouldn't imply exit-zero is the interesting case.

---

## Proposal: let OpenCode parse its own config

This addresses **#2** and removes the `hujson` dependency. Built and verified; happy to push it as a commit.

**The key fact:** OpenCode already parses the user's JSONC. Entire only needs its own parser because it insists on *rewriting* `OPENCODE_CONFIG_CONTENT`.

Put our agent block in `<tmpdir>/opencode.json` — the project layer, which OpenCode loads from `--dir` — and leave the user's config untouched:

```diff
 raw, stderr, stdoutBytes, err := agent.RunIsolatedTextGeneratorCLI(ctx, a.CommandRunner, openCodeBinary, openCodeBinary, args, prompt,
-    "OPENCODE_CONFIG_CONTENT="+config, "OPENCODE_AUTO_SHARE=false")
+    "OPENCODE_DISABLE_SHARE=true", "OPENCODE_AUTO_SHARE=false")
```

The whole `hujson.Standardize` → unmarshal → merge → remarshal function collapses to one `json.Marshal` of a literal.

**Why `OPENCODE_DISABLE_SHARE` rather than the `share` key.** The one thing the project layer gives up is precedence: `OPENCODE_CONFIG_CONTENT` is applied *after* project config, so a user's inline `{"share":"auto"}` would outrank the file — and the share gate is an OR (`W.data.share!=="auto" && !U.autoShare && !j.share`), so `OPENCODE_AUTO_SHARE=false` alone doesn't cover it. `OPENCODE_DISABLE_SHARE` is `L=process.env.OPENCODE_DISABLE_SHARE==="true"||==="1"`, and every entry point of the share module opens with `if(L)return` (`create` returns an empty record). No config layer can override it — **strictly stronger than the current `share:"disabled"`**, which is just a merged-config value. I kept the `share` key too, with a comment saying it's the documented setting and not the guarantee.

**Two details worth knowing if you take this:**

- The temp dir needs an `osroot` root. `plugin_install_remote.go:167` is the precedent — `os.MkdirTemp` → `osroot.Shared` → `osroot.Forget` before `RemoveAll`, because a memoized root outliving its directory is a handle to an unlinked inode. It matters more here than for plugins: one leaked fd and one dead registry entry *per summary generated*. Needs an `allowedRootBases` entry or `TestRootBasesAreTrusted` fails the build.
- The test inverts: instead of tripping over the ambient env, it asserts about it. `TestOpenCodeGenerateText` now requires the child's `OPENCODE_CONFIG_CONTENT` to equal the parent's exactly (or be absent if unset), and reads `<dir>/opencode.json` from inside the `CommandRunner` closure — the only moment it exists.

**Verification:** `go build ./...`, the opencode/osroot/agent packages and the `cli` summary tests, and `golangci-lint` 2.11.3 all clean. Tests pass with `OPENCODE_CONFIG_CONTENT` unset, set to `{"agent":1}` (which fails six tests on this branch today), and set to JSONC with a comment. A real `opencode` 1.18.30 run with the exact bytes the new code emits resolved our agent from the project file while the user's JSONC passed through and was parsed by OpenCode itself. `go mod tidy` returns `go.mod`/`go.sum` to their pre-PR state.

Net: −85 lines, −1 dependency, a stronger sharing guarantee, and finding #2 disappears rather than getting patched.

**#1 and #3 are independent of this** and still want their own fix.

computermode (approved):

Issue #2263, opened 2026-09-03
Lefthook is classified as not overwriting Entire's hooks, so users get the weaker warning and no permanent fix


2bTwist:
### What happened?

`cmd/entire/cli/strategy/hook_managers.go` tags each detected hook manager with `OverwritesHooks`, and that flag decides which of two very different messages a user sees at `entire enable`:

- `true`: **"Warning: X detected"**, plus the exact lines to add to X's own hook files so Entire survives permanently.
- `false`: **"Note: X detected"**, plus "if X reinstalls hooks, run `entire enable` to restore Entire's hooks". A manual step, after the fact, with no signal for when it is needed.

Husky is `true`. **Lefthook is `false`**, and Husky is the *only* manager flagged `true` (`hook_managers.go:27-46`, unchanged from v0.10.0 through v0.10.4 and current `main`):

```go
checks := []hookManager{
	{"Husky", ".husky/", true},
	{"pre-commit", ".pre-commit-config.yaml", false},
	{"Overcommit", ".overcommit.yml", false},
}
// ... then, in loops:
checks = append(checks, hookManager{"Lefthook", name, false})
checks = append(checks, hookManager{"hk", name, false})
```

Lefthook does overwrite, by the *exact* criterion the `true` branch is written for. That branch's message (`hook_managers.go:80`) reads:

> `%s may overwrite hooks installed by Entire on npm install.`

The `lefthook` npm package ships a `postinstall` that force-installs its hooks:

```js
// lefthook@2.1.10 postinstall.js, line 10
spawnSync(getExePath(), ["install", "-f"], { cwd: process.env.INIT_CWD || process.cwd() });
```

Verified on Entire 0.10.4 with lefthook 2.1.10, fresh repo, positive control before every negative claim:

| step | Entire wrapper in `pre-push` | in `commit-msg` |
|---|---|---|
| control, after `entire enable` | PRESENT | PRESENT |
| `lefthook install` | **ABSENT** | **ABSENT** |
| `entire enable` (restore) | PRESENT | PRESENT |
| `lefthook install -f` (what postinstall runs) | **ABSENT** | **ABSENT** |
| `entire enable` (remedy) | PRESENT | PRESENT |

So every `npm install` in a lefthook repo silently removes Entire's git hooks. That is exactly what Husky is flagged `true` for. Lefthook meets that criterion literally and is flagged `false`.

[...]

Issue #1965, opened 2026-08-12
warn when a coding agent commits in an enabled repo but no session is being recorded


MuskanPaliwal:
### Problem or use case

an enabled repo can silently accept agent-driven commits while recording nothing. entire's git hooks run on every commit, in the right repo, with settings loaded, and when they find no session they log a debug line (`manual_commit_hooks.go:414`) and move on. the user finds out much later, when they go looking for a session that was never recorded. i'd like to propose that this combination becomes a warning instead.

how i got here: i ran a long claude code session against this repo. it created a worktree, wrote a fix, committed twice, opened a pr. days later i wanted to look back at the session:

```
~/W/o/e/c/fix-checkpoint-metadata-bloat> entire checkpoint list
  branch       fix-checkpoint-metadata-bloat
  checkpoints  0
```

nothing on the branch and nothing in the web ui. the mistake was mine: i use a bare-clone worktree layout (a parent folder holding `.bare/` plus one directory per worktree) and i started the agent in the parent folder instead of inside a worktree. agents read their repo-level config from the directory the session starts in, and that's where entire's hooks live (`.claude/settings.json` in my case, and the equivalent checked-in files for the other integrations). the parent folder has none, so no agent hook ever fired. the folder doesn't feel wrong either: its `.git` file points at the bare gitdir, so git commands work, branches resolve, commits succeed.

but the specific mistake matters less than the shape of it. "agent committing, nothing recorded" is reachable many ways:

- the agent session started outside the worktree (my case, and possible with any integrated agent, and from any directory above a normal repo)
- the hooks are registered but the binary they invoke fails at startup (i also had a second, older `entire` on my path that can't parse this repo's current settings schema; every hook invocation through it would die silently)
- the hooks were removed or drifted out of the agent's settings file
- session state was wiped mid-conversation with `entire clean`

no one warning can name the cause, and it doesn't need to. the commit is the anchor: at that moment the git hook already has everything needed to notice. entire is enabled in the worktree. the commit is agent-driven (claude code sets `CLAUDECODE=1` in the environment of every command it runs, verified, and it also sets `CLAUDE_CODE_SESSION_ID`, so the warning could even name the unrecorded session; the `interactive` package already reads `GEMINI_CLI`, `COPILOT_CLI` and `PI_CODING_AGENT` as agent markers for tty detection; markers for the remaining integrations would need to be verified per agent). and no session anywhere in the repo has seen recent activity.

### Desired behavior

```
[...]

Issue #1250, opened 2026-05-22
Add an external Git hooks backend for Rush, Husky, and other hook managers


SnowingFox:
### Problem or use case

Entire CLI currently decides whether Git hooks are installed by resolving Git's active hooks directory with `git rev-parse --git-path hooks` and checking the five managed hook files for the `Entire CLI hooks` marker:

- `prepare-commit-msg`
- `commit-msg`
- `post-commit`
- `post-rewrite`
- `pre-push`

That works when Entire owns the active hook files directly. It misreports health in repositories where another tool owns the active hook wrappers and dispatches to source hooks stored elsewhere.

The result is a false negative or unsafe repair path:

- `entire status` / `entire doctor` can report Entire Git hooks as missing even when the manager's source hook already includes Entire.
- `entire enable --force` can replace active wrapper files owned by Rush, Husky, or another manager.
- Users cannot express that the source of truth lives outside Git's active hooks directory.

**Rush:** Rush generates active wrappers in `.git/hooks/*`, while user-owned hooks live in `common/git-hooks/*`. The generated wrappers should not be treated as the only installation truth.

**Husky (v9, standard `husky init`):** Running `prepare` / `husky` sets `core.hooksPath` to `.husky/_`. Git executes thin stubs in `.husky/_/*` (each sources `.husky/_/h`); Husky then runs the user hook at `.husky/<hook-name>` if it exists. The `_/` tree is generated and usually gitignored—users edit `.husky/pre-commit`, not `.husky/_/pre-commit`.

This differs from Rush in two important ways:

- Entire must **never** write into `.husky/_/*`. Replacing those stubs breaks Husky's dispatch chain; following `core.hooksPath` alone is especially dangerous here.
- Husky repos often start with only `.husky/pre-commit`. Entire's five lifecycle hooks may be missing, partial, or merged with existing commands (lint, tests, etc.)—not five standalone wrapper files like Rush's `common/git-hooks/*`.

Legacy Husky v8 installs may still use `core.hooksPath=.husky` without the `_/` split; external mode should rely on configured `external_dir`, not hard-coded Husky layout detection.

### Examples: input → execution → output

Fictional repo layouts below. Wrapper shapes follow common Rush / Husky v9 behavior; names and scripts are made up.

#### Rush

**Input (on disk after `rush update`, Entire appended to source hooks by hand or `entire enable`):**

```text
my-monorepo/
[...]

PR #2365, opened 2026-09-11
fix(hooks): keep checkpoint delivery working when Lefthook reclaims hooks

### LINKED ISSUE #2237: Re-install discards another tool's newer hook and chains to a stale .pre-entire backup

### LINKED ISSUE #2264: entire status reports "Checkpoints sync to: origin" when the git hooks are not installed

peyton-alt:
https://entire.io/gh/entireio/cli/trails/1301

**Draft.** Replaces the implementation in #2343, which bundled this with an unrelated session-listing change (now #2363).

## Problem

In a repo that also uses Lefthook, Lefthook reclaims Entire's Git hooks and checkpoint capture stops silently. The trigger is routine: any `lefthook.yml` edit — including pulling a teammate's — or any `npm install`, via Lefthook's postinstall running `install -f`.

It is not a narrow window. Lefthook's installed hooks self-sync **on every run**, so the reclaim happens at the top of the very `pre-commit` the agent's own commit triggers — after Entire's turn-start repair has already run. Entire's existing self-heal cannot win that race. Meanwhile `entire status` reports "Checkpoints sync to: origin" (#2264) and `entire doctor` reports "NOT INSTALLED" — the two disagree.

## Approach

Entire owns `entire-lefthook.yml` outright and adds **one** `extends` entry to the clone-local config. Lefthook resolves `extends` at run time, so an already-installed launcher dispatches to Entire immediately — no `lefthook install`, no user action. Entire writes nothing into `.git/hooks/*`, so there is nothing left to reclaim.

Verified end to end against the real lefthook **2.1.10** binary, including with a **`lefthook.toml`** main config — a layout the previous mechanism refused outright.

## Why only Lefthook

I tested the other managers the same way. **pre-commit, Overcommit and hk all overwrite Entire's hooks at install time** — and all three are currently classified as not overwriting, which is fixed here. But none of them reclaims a *different* hook during a later commit. Only Lefthook does, which is why only Lefthook defeats the self-heal and justifies a config-level integration. #1250's general backend is not built: there is now evidence rather than a guess that a second adapter is not imminent.

## What this fixes beyond the headline

- **The recurring Lefthook error.** A repo that already hit #1349 carries `pre-push.old`; Lefthook then refuses to move the hook aside and errors on *every* sync. That outlives the integration unless the leftovers are cleared, which they now are — each only when its contents prove whose it was.
- **The advice was still false.** `entire enable` told users to re-run it after every Lefthook refresh, printed immediately after installing the integration that makes refreshes harmless. That message is what #2263 was filed about.
- **Corrupted scripts were unrecoverable.** Corruption destroys the ownership marker, so the script read as foreign and nothing could repair it — not a turn, not `doctor --force`. The native path had already settled this by backing up and replacing; the two now behave the same.
- **Ownership of `entire-lefthook.yml`.** The path is Entire's choice, the repository is the user's. A file there Entire did not write is displaced to `.pre-entire`, never destroyed.

## What was removed from #2343

- **Byte-exact matching of Lefthook's generated wrapper.** It pinned every package-manager probe branch, so the next Lefthook release would read as a foreign hook and Entire would write native wrappers over Lefthook's own — #1349 in reverse. Its sibling matched strings byte-identical to this repo's own test fixtures, so no test exercised the branch real Lefthook takes.
- **A second, nested rollback layer**, inside a snapshot that already covered the same paths plus the native hooks and backups.
- **The merge into the user's `lefthook-local.yml`**, and with it `validateLefthookMainConfig`, the main-config-derived rejections, and eight near-identical `yaml.Node` traversals. `lefthook_yaml.go` is gone.
- **Proactive turn-start warnings** (~950 lines) — that is #1965's feature, not this one.

## Size

2,145 production / 2,841 test / 155 docs+e2e. Production is down from 3,156 in #2343. The test ratio is 1.32, below the `strategy` package's own 1.71.

## Known issue, not introduced here

The generated pre-push script passes `"$2"` unconditionally, so hooks written by this version fail against an older `entire` on `PATH` (`accepts 1 arg(s), received 2`). The ABI widening came from #2343; flagging it as a real mixed-version hazard.

## Still to do

Rebase onto current main.

Closes #2264, #1349, #2263. Does **not** close #2237 — that backup-chain path is untouched here.


peyton-alt:
Superseded by #2388, which does the same job against `main` in +1,413/−60 instead of +5,025/−298 — no coordinator, no transaction/staging layer, one file for the mechanism. It also fixes two things this branch got wrong: the double dispatch (every hook ran Entire twice, once from its own file and once through Lefthook's launcher) and `LefthookManaged` trusting the `lefthook-local.yml` Entire writes itself.

PR #2370, opened 2026-09-11
fix(strategy): require fetch URLs for checkpoint sync remote election

### LINKED ISSUE #2360: checkpoint_push_remote: election accepts a pushurl-only remote that status then reports as working
`strategy.ResolveCheckpointSyncRemote` elects a `checkpoint_push_remote` that names a remote with only a `pushurl` and no fetch URL, and `entire status` then reports it as the working destination. Such a remote has no configured fetch destination, so checkpoint reads and reconciliation against it cannot work.

The split is between two existing definitions of "is this a configured remote":

- `isConfiguredRemote` (`strategy/manual_commit_push.go:360`) shells out to `git remote get-url <name>`. For a remote with only `remote.<name>.pushurl`, git exits 0 and prints the *remote name* rather than a URL — verified on git 2.50.1 (Apple Git-155): `git config remote.pushonly.pushurl https://example.com/x.git && git remote get-url pushonly` prints `pushonly`, exit 0. This is the function the explicit `checkpoint_push_remote` tier consults, so the fail-closed guard at `checkpoint_sync_remote.go:83` passes.
- `readRemotesInConfigOrder` (`checkpoint_sync_remote.go:230`) reads `remote.*.url` via `git config --get-regexp`, so it correctly omits the same remote. Its own doc comment says so: "Remotes configured with only pushurl are deliberately invisible (spec Unit 1)."

So the automatic tiers exclude a pushurl-only remote and the explicit tier accepts it.

Reproduction, with `{"strategy_options": {"checkpoint_push_remote": "pushonly"}}` in `.entire/settings.local.json` and `remote.pushonly.pushurl` set:

```
$ entire status
  Checkpoints sync to: pushonly (set by checkpoint_push_remote)
```

Surfaced by trail 1286, which adds a `validateCheckpointPushRemote` in the `cli` package so `entire enable` refuses to select such a remote. That is the right call for the picker, but it leaves the two commands disagreeing about the same state — the same trail's `enable` says `Cannot confirm checkpoint uploads to pushonly: remote "pushonly" has no configured fetch URL` for a remote `status` presents as fine.

Fixing it properly means making the eligibility rule part of election rather than of setup, which changes behaviour for any existing repo that has this configured: it would move from "elected" to fail-closed, sync disabled until fixed. That is deliberately out of scope for a setup-picker change, hence this issue.

MuskanPaliwal:
Closes #2360

Trail: https://entire.io/gh/MuskanPaliwal/cli/trails/4

A Git remote can be configured with only a `pushurl` and no fetch URL. Such a remote can receive pushes, but Entire cannot read checkpoint history or reconcile checkpoint state from it.

`ResolveCheckpointSyncRemote` previously treated this configuration inconsistently:

- Explicit and previously captured remote selections were validated with `git remote get-url`.
- Git reports success for a pushurl-only remote—even returning the remote name when no fetch URL exists.
- Automatic election used `remote.*.url` entries and therefore correctly excluded the same remote.

This allowed an explicit `checkpoint_push_remote` to select a destination that automatic election considered unusable. `entire status` then presented that remote as the active checkpoint destination even though checkpoint reads and reconciliation could not work against it.

## Use one eligibility rule throughout election

Remote election now builds the set of remotes with configured fetch URLs once and uses that set for every election tier.

The behavior differs deliberately based on who owns the selection:

| Election source | Previous behavior | New behavior |
| --- | --- | --- |
| Explicit `checkpoint_push_remote` | Elected a pushurl-only remote | Fails closed and explains that the remote has no fetch URL |
| Captured automatic selection | Re-elected a pushurl-only remote | Discards the stale selection and falls through to normal defaults |
| Automatic default election | Already excluded pushurl-only remotes | Unchanged |
| Gated-sync hint | Could recommend a pushurl-only remote | Stays silent because the suggested destination would be unusable |

An explicit setting fails closed because it represents a user decision that should be corrected rather than silently replaced. Captured state remains fail-soft because it was produced automatically and may become stale when remotes are renamed or reconfigured.

## Status and diagnostics

When the explicit remote has no fetch URL, `entire status` now:

- Reports that checkpoints are not syncing.
- Names the invalid remote.
- Explains that it lacks a configured fetch URL.
- Does not also print a misleading `Checkpoints sync to:` destination.

The gated-sync hint uses the same fetchability rule, ensuring it never recommends a destination that checkpoint sync cannot read from.

## Regression coverage

The new tests construct real isolated Git repositories with:

```ini
[remote "pushonly"]
    pushurl = https://example.com/pushonly.git
```

and deliberately omit `remote.pushonly.url`.

They verify that:

- An explicit pushurl-only remote fails closed.
- The error names the remote and its missing fetch URL.
- A captured pushurl-only remote falls back to `origin`.
- Status reports the failure without advertising a sync destination.
- The gated-sync hint remains silent for the unusable remote.

## Verification

The following checks passed:

- `go test ./cmd/entire/cli/strategy -run '^TestHintGatedCheckpointSync$'`
- `mise run check`
  - formatting
  - lint
  - race-enabled unit and integration tests
  - Vogon E2E canary
  - Roger Roger E2E canary
- `mise run lint`

### DIFF
--- a/cmd/entire/cli/strategy/checkpoint_sync_remote.go
+++ b/cmd/entire/cli/strategy/checkpoint_sync_remote.go
@@ -81,9 +81,9 @@ func ResolveCheckpointSyncRemote(ctx context.Context) (CheckpointSyncRemote, err
 	}
 	// Every tier elects from the same fetchable set. `git remote get-url`
 	// accepts a pushurl-only entry even though reads and reconciliation cannot.
-	remotes := configuredRemotesInConfigOrder(ctx)
+	fetchRemotes := configuredRemotesInConfigOrder(ctx)
 	if name := s.GetCheckpointPushRemote(); name != "" {
-		if !hasConfiguredFetchURL(remotes, name) {
+		if !slices.Contains(fetchRemotes, name) {
 			return CheckpointSyncRemote{}, fmt.Errorf(
 				"checkpoint_push_remote %q has no configured fetch URL; checkpoint sync disabled until fixed", name)
 		}
@@ -94,7 +94,7 @@ func ResolveCheckpointSyncRemote(ctx context.Context) (CheckpointSyncRemote, err
 	// is automatic state, so a captured remote that is no longer fetchable
 	// falls through to the default tiers instead of disabling sync.
 	for _, name := range loadCapturedSyncRemotes(ctx) {
-		if hasConfiguredFetchURL(remotes, name) {
+		if slices.Contains(fetchRemotes, name) {
 			return CheckpointSyncRemote{Name: name, Source: SyncRemoteSourceObserved}, nil
 		}
 		logging.Debug(ctx, "captured checkpoint sync remote has no configured fetch URL; falling through",
@@ -102,14 +102,14 @@ func ResolveCheckpointSyncRemote(ctx context.Context) (CheckpointSyncRemote, err
 	}
 
 	switch {
-	case len(remotes) == 0:
+	case len(fetchRemotes) == 0:
 		return CheckpointSyncRemote{}, nil
-	case hasConfiguredFetchURL(remotes, "origin"):
+	case slices.Contains(fetchRemotes, "origin"):
 		return CheckpointSyncRemote{Name: "origin", Source: SyncRemoteSourceDefault}, nil
-	case len(remotes) == 1:
-		return CheckpointSyncRemote{Name: remotes[0], Source: SyncRemoteSourceSole}, nil
+	case len(fetchRemotes) == 1:
+		return CheckpointSyncRemote{Name: fetchRemotes[0], Source: SyncRemoteSourceSole}, nil
 	default:
-		return CheckpointSyncRemote{Name: remotes[0], Source: SyncRemoteSourceFirst}, nil
+		return CheckpointSyncRemote{Name: fetchRemotes[0], Source: SyncRemoteSourceFirst}, nil
 	}
 }
 
@@ -159,7 +159,7 @@ func checkpointSyncAllowedForRemote(ctx context.Context, pushRemote, pendingCapt
 // fact, and committing it to the tracked settings.json would fail-close
 // checkpoint sync for every teammate whose clone lacks that remote name.
 func hintGatedCheckpointSync(ctx context.Context, pushRemote string) {
-	if !hasConfiguredFetchURL(configuredRemotesInConfigOrder(ctx), pushRemote) {
+	if !slices.Contains(configuredRemotesInConfigOrder(ctx), pushRemote) {
 		return
 	}
 	syncRemote, err := ResolveCheckpointSyncRemote(ctx)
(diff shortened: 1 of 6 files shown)


cmd/entire/cli/strategy/checkpoint_sync_remote.go
```diff
@@ -79,35 +79,40 @@ func ResolveCheckpointSyncRemote(ctx context.Context) (CheckpointSyncRemote, err
...
 	}
+	// Every tier elects from the same fetchable set. `git remote get-url`
+	// accepts a pushurl-only entry even though reads and reconciliation cannot.
+	fetchRemotes, err := configuredRemotesInConfigOrderResult(ctx)
+	if err != nil {
+		return CheckpointSyncRemote{}, fmt.Errorf("cannot read git remotes to resolve the checkpoint sync remote: %w", err)
+	}
 	if name := s.GetCheckpointPushRemote(); name != "" {
-		if !isConfiguredRemote(ctx, name) {
+		if !slices.Contains(fetchRemotes, name) {
```
Soph, line 89: **An explicit `checkpoint_push_remote` naming a remote defined through `include.path` now fails closed**

Before this PR, the explicit and captured tiers checked the remote with `git remote get-url`, which reads every config scope, including includes. They now check against `readRemotesInConfigOrder`, which runs `git config --local --get-regexp '^remote\..*\.url$'`. When a specific file option like `--local` is given, git doesn't follow `include.*` by default, so a remote defined in an included file (or in global config) isn't in `fetchRemotes`.

The comment at `cmd/entire/cli/strategy/git_remote_cache.go:70-75` warned about this: "ordered lists only remotes carrying a url key in local config, while isConfiguredRemote runs `git remote get-url`, which also sees pushurl-only remotes and inherited scopes. Deriving one from the other would change election semantics."

Repro: `origin` added normally, plus `fork` defined in an included file:

```ini
# .git/config
[include]
    path = /path/to/remotes.cfg

# /path/to/remotes.cfg
[remote "fork"]
    url = https://example.com/fork.git
```

With `checkpoint_push_remote: "fork"`, `ResolveCheckpointSyncRemote` returns:

- main: `{Name:fork Source:config}`, `err=nil`
- this PR: `checkpoint_push_remote "fork" has no configured fetch URL; checkpoint sync disabled until fixed`

The error message is also wrong here, because `fork` does have a fetch URL. The captured tier and `pendingCaptureCheckpointSyncRemote` hit the same gap, but quietly: a captured include-defined remote falls through to origin, and a push to it never captures.

Possible fixes:

- Add `--includes` to the `git config --local --get-regexp` call in `readRemotesInConfigOrder`. This also makes included remotes eligible for the default tiers, which seems correct.
- Or keep the ordered list as it is, and check the explicit and captured names per remote with an all-scope `git config --get-all remote.<name>.url` that must return a non-empty value.

Whichever you pick, please add a regression test with an include-defined remote next to the pushurl-only cases, and update the `git_remote_cache.go` comment.

MuskanPaliwal: hey @soph,  i used your `--includes` suggestion for the local config read and updated the cache comment. i also added a test with fork defined through `include.path`; it covers push capture and both captured and
explicit selection. `pushurl-only` remotes remain excluded. a remote defined only in global config is still outside the local scan.

MuskanPaliwal:
the review comment is fixed in https://github.com/entireio/cli/pull/2370/commits/fb4d68419b84dbae05ee4c7a637f786e5dde8f0b. the pr is ready for another review. thanks @Soph!

Soph (approved):

Soph:
Thanks!

PR #2395, opened 2026-09-12
fix(checkpoint): surface git-refs push queue failures

### LINKED ISSUE #2393: git-refs writes can return success without durable push bookkeeping
### What happened?

On the git-refs checkpoint backend, an ordinary persistent write updates the
checkpoint ref and then records that ref in
`entire-checkpoint-push-queue.jsonl`. If queue resolution, locking, opening, or
appending fails, `enqueueForPush` logs a warning but the write still returns
success.

I understand the documented intent: enqueue is best-effort so a locally
successful condensation is not failed, and a later write to the same checkpoint
can re-enqueue the ref. The residual problem is that the queue is also the sole
input to the normal git-refs unpushed count and publication path. If there is no
later write, the result is a durable false-success state:

- the checkpoint ref and object exist only in the originating clone;
- the write caller received success;
- the normal unpushed count reports `0`;
- the normal queue publisher pushes `0` refs;
- a fresh clone from the remote cannot recover the checkpoint ref or object.

The local ref remains readable and GC-protected, so this is not immediate local
deletion. It becomes unrecoverable through the normal remote path if that sole
clone is later lost or discarded.

### Expected behavior

A successfully acknowledged checkpoint should not become indefinitely invisible
to both normal pending diagnostics and normal publication. Keeping condensation
fail-soft is reasonable, but the system should retain or reconstruct durable
knowledge that the local ref still needs publication, or explicitly surface a
partial/local-only outcome.

### Actual behavior

The enqueue failure is warning-only. Once the transient obstruction clears,
queue-only status and publication see no pending checkpoint and perform no
reconciliation against local checkpoint refs.

### Steps to reproduce

I reproduced this using a disposable work repository, a local bare remote, and a
fresh recovery clone:

1. Configure `checkpoints.primary.type` as `git-refs`.
2. Create the queue pathname as a directory to deterministically make only the
   JSONL queue open fail.
3. Write a checkpoint through `checkpoint.Open(...).Persistent.Write(...)`.
4. Verify that the write returns `nil` and the checkpoint ref resolves locally.
5. Remove the synthetic obstruction and call
   `strategy.CountUnpushedCheckpoints(...)`; it returns `0`.
6. Call `strategy.PushQueuedCheckpointRefs(...)`; it returns `pushed=0` with
   pushing enabled.
7. Verify the bare remote lacks the checkpoint ref, clone that remote, and
   verify the clone has neither the ref nor the checkpoint commit object.

Observed output:

```text
WARN checkpoint: enqueue checkpoint ref for push failed ... open push queue: ... is a directory
write_success=true local_ref=refs/entire/checkpoints/f6/c1b2c3d4e5f6
count_unpushed=0 pushed=0 push_disabled=false
remote_ref_present=false recovery_ref_present=false recovery_object_present=false
```

The focused end-to-end regression passed while asserting this current behavior;
the control experiment also confirmed that the same write creates both the ref
and queue record when queue persistence is available.

### Entire CLI version

v0.10.6

### OS and architecture

Windows amd64

### Agent

Codex

### Terminal

PowerShell

### Logs / debug output

```bash
=== RUN   TestGitRefsQueueFailure_LosesRemoteRecoverability
WARN checkpoint: enqueue checkpoint ref for push failed ref=refs/entire/checkpoints/f6/c1b2c3d4e5f6 error="open push queue: openat entire-checkpoint-push-queue.jsonl: is a directory"
write_success=true local_ref=refs/entire/checkpoints/f6/c1b2c3d4e5f6 checkpoint_commit=4589427309d230ba0e4bce32bbf4f17913fe7ac0
count_unpushed=0 pushed=0 push_disabled=false
remote_ref_present=false recovery_ref_present=false recovery_object_present=false
--- PASS: TestGitRefsQueueFailure_LosesRemoteRecoverability (1.99s)
```

### Additional context

This is a correctness and checkpoint-durability report, not a security report.
The failure was deterministically injected; real-world trigger frequency is
unknown. Plausible queue persistence failures include filesystem permission,
capacity, locking, or I/O failures, and there is also a structural interruption
window after ref update and before enqueue, but those natural triggers were not
claimed as reproduced here.

The source test ran at `5f09e04de77d627984ff118eb87bd2c16cce8fbd`
and was revalidated on current `main`
`174dca012af81d3d6ebefc2d7cc17f50a5eb37bb`. The ref-write, queue,
unpushed-count, publication, and architecture-document blobs are identical
between those revisions.

Related but distinct work:

- `#1925` fixes cancellation affecting post-ref bookkeeping.
- `#2008` / `#2012` cover removal of a newer same-ref enqueue during an
  in-flight push.
- `#1773` / `#1929` cover logged-out import and later synchronization.
- `#2327` adds explicit checkpoint migration/re-enqueue behavior.
- `#1917` / `#1926` cover a concurrent same-checkpoint ref-update race before
  queue publication.

Potential solution directions include a durable pre-acknowledgement outbox,
surfacing a partial/local-only result, or reconciling local checkpoint refs
against queued/published state before status and publication. I have not assumed
which trade-off is preferred.

ChetanReddyC:
## Summary

- return an explicit partial failure when a git-refs checkpoint is written locally but cannot be added to the push queue
- preserve the local ref and retryable condensation state, then re-enqueue on a later write
- add store-level and condensation-level regressions and update the architecture contract

Closes #2393

## Why this approach

The ref update and queue append cannot be one atomic operation. Rolling back a valid local ref would discard recoverable data, while scanning every local checkpoint ref during pre-push would conflict with the queue's purpose and with fetched-ref semantics. This change therefore makes the partial outcome explicit at the store boundary. The existing fail-soft post-commit caller then preserves session and shadow state so the same checkpoint can be retried safely.

I believe this is the smallest complete fix that fits the current architecture. I would appreciate the maintainers' direction on this approach, and I am happy to adjust it or continue the work under your guidance.

## Verification

- `go test ./cmd/entire/cli/checkpoint ./cmd/entire/cli/strategy -run '^(TestGitRefsStore_WriteReportsQueueFailureAfterRefUpdate|TestCondenseAndUpdateState_QueueFailurePreservesRetryState)$' -count=1`
- `go test ./cmd/entire/cli/checkpoint ./cmd/entire/cli/strategy -run '^$' -count=1`
- `golangci-lint run ./cmd/entire/cli/checkpoint/... ./cmd/entire/cli/strategy/...` (0 issues)
- `git diff --check upstream/main..HEAD`

Broader package test runs on this Windows host encounter existing temporary-repository cleanup failures where `.git` remains locked; the focused regressions and package compilation above pass.

## Session metadata

Codex checkpoint capture was not active for this local Windows linked-worktree session, so no genuine checkpoint trailer was available.

### DIFF
--- a/cmd/entire/cli/checkpoint/refs_store.go
+++ b/cmd/entire/cli/checkpoint/refs_store.go
@@ -196,29 +196,26 @@ func (s *gitRefsStore) refTip(cid id.CheckpointID, ref *plumbing.Reference) (plu
 	return ref.Hash(), tree, nil
 }
 
-// enqueueForPush records refName in the push-discovery queue, logging (never
-// returning) on failure so the local ref write still succeeds.
+// enqueueForPush records refName in the push-discovery queue.
 //
 // The queue is resolved with the cancellation stripped from ctx. By the time we
 // get here the ref is already written locally, and the queue is the ONLY
-// push-discovery mechanism there is — see the
-// pushQueueFileName doc. A ref that misses the queue is never pushed, and the
-// writers above are idempotent, so a re-run skips it as already-present and
-// never re-enqueues it: it stays local-only forever. Queue resolution is now a
-// context-free metadata read, but stripping cancellation here keeps the whole
-// post-ref bookkeeping boundary explicit and prevents a future queue operation
-// from dropping a write that already happened.
-func (s *gitRefsStore) enqueueForPush(ctx context.Context, refName plumbing.ReferenceName) {
+// push-discovery mechanism there is; see the pushQueueFileName doc. A ref that
+// misses the queue is never pushed, so failure is returned as a partial
+// outcome: the checkpoint remains readable locally,
+// but the caller must not treat it as durably queued for publication. Queue
+// resolution is now a context-free metadata read, but stripping cancellation
+// here keeps the whole post-ref bookkeeping boundary explicit and prevents a
+// future queue operation from dropping a write that already happened.
+func (s *gitRefsStore) enqueueForPush(ctx context.Context, refName plumbing.ReferenceName) error {
 	q, err := PushQueueForRepo(context.WithoutCancel(ctx), s.repo)
 	if err != nil {
-		logging.Warn(ctx, "checkpoint: resolve push queue failed; ref not enqueued",
-			slog.String("ref", refName.String()), slog.String("error", err.Error()))
-		return
+		return fmt.Errorf("resolve push queue: %w", err)
 	}
 	if err := q.Enqueue(refName); err != nil {
-		logging.Warn(ctx, "checkpoint: enqueue checkpoint ref for push failed",
-			slog.String("ref", refName.String()), slog.String("error", err.Error()))
+		return fmt.Errorf("enqueue checkpoint ref: %w", err)
 	}
+	return nil
 }
 
 func (s *gitRefsStore) updateCheckpointRef(
@@ -242,7 +239,9 @@ func (s *gitRefsStore) updateCheckpointRef(
 	if err != nil {
 		return err
 	}
-	s.enqueueForPush(ctx, refName)
+	if err := s.enqueueForPush(ctx, refName); err != nil {
+		return fmt.Errorf("checkpoint ref %s was written locally but could not be queued for push: %w", refName, err)
+	}
 	return nil
 }
 
--- a/cmd/entire/cli/checkpoint/refs_store_test.go
+++ b/cmd/entire/cli/checkpoint/refs_store_test.go
@@ -3,6 +3,8 @@ package checkpoint
 import (
 	"context"
 	"fmt"
+	"os"
+	"path/filepath"
 	"testing"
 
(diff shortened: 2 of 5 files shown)


MuskanPaliwal:
Hey @ChetanReddyC, I noticed your note about unresolved Codex hook discovery. I addressed worktree and `.bare` hook discovery in [#2052](https://github.com/entireio/cli/pull/2052), including warnings that show which hooks file Codex reads and what needs fixing.

Which Entire version are you running, and could you share the Codex section of `entire doctor`? Did Entire show a warning when you enabled Codex or started a session? I’d like to understand what’s still failing in your setup.

ChetanReddyC:
Hey @MuskanPaliwal Thanks for checking. To clarify, this was specific to my local Windows linked-worktree session, where Codex checkpoint capture was not active. Inside the Codex sandbox, entire doctor also encountered Access is denied while resolving the hooks path. I’ve confirmed that my original “hook discovery was unresolved” wording was too broad; I don’t currently have evidence of a general issue or regression in #2052. The note was only intended to explain why that PR had no genuine checkpoint trailer.

PR #2399, opened 2026-09-13
fix(strategy): keep non-ASCII staged paths literal so commit linking matches

### LINKED ISSUE #2398: Checkpoint linking fails for non-ASCII filenames
### What happened?

In the interactive/default prompt-mode commit-linking path, `getStagedFiles` runs `git diff --cached --name-only` without `-z` and treats the newline-delimited output as literal repository paths.

Git quotes and C-escapes unusual pathnames in this output mode. For example, with the default `core.quotePath=true`, `café.go` is emitted as `"caf\303\251.go"`. Entire then exact-compares that display string with the literal `café.go` stored in `SessionState.FilesTouched`. The comparison fails, so a related commit can be classified as containing no agent work and receive no `Entire-Checkpoint` trailer.

I reproduced this against current `main` using a focused exact-source test. An ASCII control received its trailer, while an otherwise equivalent UTF-8 canary did not. Parsing the same Git output with `-z` restored the literal path and the expected overlap.

The affected Git commit itself is still created. The demonstrated impact is missing commit-to-session linkage and incomplete provenance, including potentially incomplete `entire explain` attribution. I did not observe source loss, transcript deletion, remote checkpoint loss, or a security boundary violation.

### Steps to reproduce

1. Use the default `core.quotePath=true` setting and default prompt-mode commit linking in an interactive terminal.
2. Start an agent session that records only a non-ASCII file such as `café.go` in `FilesTouched`, with no pending task record that would bypass overlap detection.
3. Stage that file and make a manual commit:

   ```bash
   git config core.quotePath true
   git add "café.go"
   git diff --cached --name-only
   git commit -m "update localized file"
   ```

4. Inspect the commit message.

Expected: the related session is detected and an `Entire-Checkpoint` trailer is added.

Actual: Git's quoted display path does not match the literal `FilesTouched` path, the session is excluded, and the trailer is omitted.

Focused control/canary output:

```text
ASCII control.go:
  literal_match=true overlap=true
  NUL-delimited path="control.go"
  checkpoint_trailer=true

UTF-8 café.go:
  getStagedFiles returned Git's quoted C-escaped display path
  literal_match=false overlap=false
  NUL-delimited path="café.go"
  checkpoint_trailer=false
```

The focused test passed once and then passed three consecutive repeat runs.

### Entire CLI version

v0.10.6

### OS and architecture

Windows 11 amd64 (build 26200.9445)

### Agent

Codex

### Terminal

Codex integrated terminal, PowerShell 7.6.5

### Logs / debug output

```bash
getStagedFiles("control.go") = ["control.go"]
literal_match=true overlap=true
checkpoint_trailer=true

getStagedFiles("café.go") = [Git quoted C-escaped display path]
literal_match=false overlap=false
checkpoint_trailer=false
```

### Additional context

_No response_

wangzhengzhuo05:
**What**

`getStagedFiles` in `cmd/entire/cli/strategy/manual_commit_hooks.go` now parses `git diff --cached --name-only -z` and splits the output on NUL, so staged paths are returned literally instead of as git's quoted/C-escaped display form.

**Why**

With git's default `core.quotePath=true`, the newline-delimited `--name-only` output quotes and C-escapes unusual pathnames — `café.go` comes back as the display string `"caf\303\251.go"`. That string was exact-compared against the literal path stored in `SessionState.FilesTouched`, so a session that touched only a non-ASCII file was classified as contributing no content to the commit: the commit was still created, but it received no `Entire-Checkpoint` trailer and its provenance (`entire explain` attribution) was incomplete. Reported with a focused control/canary reproduction in #2398: an ASCII control file got its trailer, an otherwise equivalent UTF-8 canary did not.

**How**

- Pass `-z` to the git invocation and iterate the NUL-delimited fields, keeping `filepath.ToSlash` on every entry.
- Drop the `\r\n` → `\n` normalization: it only existed to unwrap the newline framing, and `-z` emits NUL-terminated, unquoted names.
- The function's contract is unchanged: `(non-nil empty slice, nil)` when nothing is staged, `(nil, err)` only when git itself fails, and the same error wrapping.
- This follows the precedent already in the same package: `priorAICommitFiles` in `cmd/entire/cli/strategy/telemetry_signals.go` uses `-z` for the same reason, so paths match their `FilesTouched` form.

**Tests**

New `cmd/entire/cli/strategy/staged_files_quoting_test.go`, `TestGetStagedFiles_PreservesNonASCIIPaths`: sets `core.quotePath true` explicitly, stages a literal `café.go` plus an ASCII `control.go`, and asserts both come back literal and that no returned entry is git-quoted or C-escaped.

```
$ gofmt -l cmd/entire/cli/strategy/manual_commit_hooks.go cmd/entire/cli/strategy/staged_files_quoting_test.go
(no output)
$ go test ./cmd/entire/cli/strategy/ -count=1
ok  	github.com/entireio/cli/cmd/entire/cli/strategy	25.187s
$ go build ./...
(exit 0)
```

**Mutation check**

Reverting only the fix (removing `-z`, i.e. the previous behaviour) makes the new test fail with exactly the string from the report — the assertion output is the single staged entry `"\"caf\\303\\251.go\"\ncontrol.go\n"`, which contains neither literal filename. Restoring `-z` returns the package suite to green, so the test guards the defect rather than the implementation.

Fixes #2398.


wangzhengzhuo05:
Closing this out.

It has been waiting 5 days without any maintainer review, and we are winding down this contribution effort — a stale PR sitting in the queue costs maintainers attention without giving them a decision to make, so it is better closed than left open indefinitely.

If the direction is useful, the branch is intact on the fork (`fix/nul-delimited-staged-files`, +50/−7 across 2 file(s)) and can be picked up as-is.

What it contains: *fix(strategy): keep non-ASCII staged paths literal so commit linking matches* — **What** `getStagedFiles` in `cmd/entire/cli/strategy/manual_commit_hooks.go` now parses `git diff --cached --name-only -z` and splits the output on NUL, so staged paths are return

It was verified locally before opening (target test suite green, plus a mutation check that reverts only the fix and confirms the new regression test fails). Thanks for the time.

PR #2403, opened 2026-09-13
checkpoint: accept case-folded shard directories in ParseRef

KC1706:
Ref #2402

On a case-insensitive-but-case-preserving filesystem (macOS APFS, Windows NTFS defaults), a checkpoint ref's shard directory can get case-folded by the filesystem itself — e.g. a ULID checkpoint whose correct shard is `6B` lands inside an already-present `6b` directory left by an unrelated legacy hex checkpoint. `ParseRef` compared the recomputed shard against the ref's actual path component byte-for-byte, so `"6B" != "6b"` caused a well-formed, intact checkpoint to be rejected as malformed and silently disappear from `checkpoint list`/`explain`.

This PR makes that comparison case-insensitive (`strings.EqualFold`), matching what the filesystem already does, and adds two regression tests reproducing both directions of the collision (ULID→lowercase dir, legacy-hex→uppercase dir).

Verified locally: `mise run fmt && mise run lint` (0 issues) and `mise run test:ci` pass.


Soph (approved):
Thanks for doing this. I'll merge it and do a follow up PR addressing the other places where this is an issue in a few minutes. Will close your issue after merging that one.

Thanks!

PR #2493, opened 2026-09-15
Log git's own output when a checkpoint ref push fails

peyton-alt:
https://entire.io/gh/entireio/cli/trails/1342

## Why

When a checkpoint ref push fails, the logs record `git push: exit status 1` and nothing else. The reason is not lost — it is captured and then discarded one layer below the log line.

`remote.Push` runs with `--porcelain` and returns `PushResult.Output` alongside the error, but `batchPushRefs` dropped the result with `_`:

```go
if _, err := remote.PushWithOptions(ctx, remote.PushOptions{Remote: target, RefSpecs: refSpecs}); err != nil {
    return fmt.Errorf("push %d checkpoint refs: %w", len(refs), err)
}
```

So every caller above it — including `.entire/logs` and anything `entire doctor bundle` collects — could only ever say "exit status 1".

Found the hard way: one checkpoint ref in this repo failed on every push across several days, and neither the queue, the logs, nor a support bundle could say why. It later turned out the remote already had the ref (`[up to date]` on a manual push), so the failure was transient — but there was no way to establish that from anything the CLI recorded.

## What changed

- `batchPushRefs` attaches git's output to the error through a typed `pushOutputError`.
- Both git-refs failure sites log it: the per-ref failure, and the **batch** failure, which was not logged at all — the retry banner replaced it, so the first and most informative signal went unrecorded.

## What deliberately did not change

The user-facing retry banner still names no cause. The existing comment at `manual_commit_push.go` is right that a batch fails on an unreachable or unauthorized remote about as often as on divergence, and that telling someone with a dead remote that their refs "diverged" sends them after the wrong problem.

That split is the point of this change: **the user gets a claim the code can support; the log gets what the remote actually said.** It is the same line the codebase already draws elsewhere — "not aborting is correct; not telling is not."

Two smaller judgments:

- **An absent `git_output` is informative**, so no empty attribute is emitted: git produced no output, which means the failure is the transport rather than a verdict from the remote.
- **The error text is unchanged**, so the output cannot leak into user-facing messages — the typed error carries it for logs only.

## Tests

`TestPushOutputAttrs` pins extraction through both wrapping layers the recovery path adds (`fmt.Errorf("push %d checkpoint refs: %w")`, then `fmt.Errorf("sync diverged checkpoint ref %s: %w")`), the two no-attribute cases, and that the error string is untouched.

`mise run lint` clean and `mise run test:ci` exit 0 on this branch.

---

> [!NOTE]
> **Low Risk**
> Observability-only change to error wrapping and warn logs; push behavior and user-facing messages are explicitly unchanged.
> 
> **Overview**
> Checkpoint ref batch pushes used to log only `exit status 1` because `batchPushRefs` discarded porcelain output from `remote.PushWithOptions`. The PR **preserves that output** in a `pushOutputError` wrapper and exposes it to logs via `pushOutputAttrs` as a `git_output` field when git actually printed something.
> 
> **Logging** now records the batch failure before the generic retry banner (previously unlogged), and per-ref failures in `flushCheckpointRefsQueue` include the same attribute. User-facing stderr messages and error strings are unchanged so porcelain details stay in logs/support bundles only.
> 
> `TestPushOutputAttrs` covers wrapped error chains and confirms git output does not leak into `.Error()` text.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b4db25c095876a9004694d640863021f3cd4d0b5. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


peyton-alt:
Superseded by #2488 (trail 1337), which merged this morning and does the same job better.

#2488 folds git's reason into the error so it reaches the user, where this PR deliberately confined it to log attributes — and it elides the *middle* of long output rather than truncating the head. That matters more than it sounds: git prints its verdict last, so head-truncation discards the very reason the output was captured for. A GitHub push-protection block is hundreds of characters of banner and unblock URLs followed by `(push declined due to repository rule violations)` at the end; this PR would have kept the banner and dropped the reason.

#2488 was also rune-aware from the start. The UTF-8 truncation defect the findings gate raised on this branch (trail 1342) had already been fixed there.

Nothing here is worth salvaging over that. Closing rather than reworking it into a delta.

PR #2494, opened 2026-09-15
auth: fall back to the file token store when the Linux keyring is unavailable, and remember the choice

peyton-alt:
https://entire.io/gh/entireio/cli/trails/1343

## Summary

Fixes two customer-reported failures on headless Linux (CLI 0.10.6, Ubuntu VPS with no keyring daemon):

1. `entire login --device` authenticated, then failed on the first keyring write with `The name org.freedesktop.secrets was not provided by any .service files` and discarded the fresh token; the `ENTIRE_TOKEN_STORE=file` hint only appeared after the loss.
2. `ENTIRE_TOKEN_STORE=file` had to be set on every command. With a valid `~/.config/entire/tokens.json`, plain `entire activity` tried the keyring, failed, and reported "Not logged in".

Root cause: backend selection in `internal/entireclient/tokenstore` read only the environment variable, once per process, with no probe and no memory of where tokens were stored.

## What changes

- **Remembered backend.** `token_store.json` next to `contexts.json` records the backend that last received a write. Precedence is now `ENTIRE_TOKEN_STORE` (explicit; `file`, or anything else meaning the keyring) → the marker → the `go test` temp store → the platform default. The marker is never written while `ENTIRE_TOKEN_STORE_PATH` is set, since it cannot carry a path. An explicit `ENTIRE_TOKEN_STORE=keyring entire login` clears it and removes the superseded copy from `tokens.json`.
- **Linux/BSD fallback.** The default keyring is fronted by a fallback: a call that fails for an availability reason (anything but `ErrNotFound` and Ctrl-C, including a timeout and a collection that will not unlock) is retried against `tokens.json`. Once the file store proves it holds the credential, it is adopted for the process, announced once on stderr, and remembered. macOS and Windows never fall back: there a keyring failure is a denied prompt or a locked store, and a plaintext file must not be the silent answer.
- **Honest reporting.** `auth status` and `auth token` report a store read error instead of "Not logged in" (`statusTarget.storeErr`). `logout` still removes the context on a broken store. The login hint says the choice is remembered, gives the way back, tells the truth when `ENTIRE_TOKEN_STORE_PATH` is set, and stays quiet when the file store itself failed (`tokenstore.ErrFileStoreFailed`).
- **Device flow without a display.** `entire login` picks the device-code flow on Linux/BSD when `DISPLAY`, `WAYLAND_DISPLAY`, and `BROWSER` are unset and not under WSL. This is the "login redirects me to 127.0.0.1" report from web terminals that set no SSH variables.
- README and CLAUDE.md describe the selection rules.

## Verification

- `mise run check` green on the final head: lint 0 issues; 11,183 unit tests, 573 integration tests, E2E canary all pass.
- Real binaries built from the final head, run in an `ubuntu:24.04` container under both D-Bus shapes (no session bus; bus with no secrets provider). Before this branch both shapes reproduced the customer's exact error and "Not logged in" with a valid `tokens.json`. After: first command prints the notice and the token and writes the marker; the second command prints the token silently. `entire login` with no display prints the explanation and takes the device flow; with `DISPLAY` set it takes the browser flow.
- macOS binary from the final head: no fallback without a marker (by design), the marker path works, and the no-display rule does not fire on darwin.
- Not run: a real `entire login --device` in the keyring-less container (needs a live account approval). The write path is covered by unit tests against the real file store, and the D-Bus failure it recovers from was reproduced with the real keyring library in the container.

## Behaviour notes for reviewers

- `auth status` now exits non-zero when the store cannot be read; "Not logged in" stays exit 0. `auth token` exits non-zero in both states, as before, but says which.
- A successful fallback **read** adopts and remembers the file store. That is what makes a customer's first command after losing the keyring work with no re-login. The cost, documented in `switchTo`: a stale `tokens.json` can be re-adopted after a transient keyring failure until the file is removed or the user runs `ENTIRE_TOKEN_STORE=keyring entire login`.
- `noLocalDisplay` sends `sudo` shells and display-less `tmux` panes on a Linux desktop to the device flow. Degraded (one typed code), not broken.
- Under `go test`, `resolveBackendLocked` drops an explicit non-`file` `ENTIRE_TOKEN_STORE` so a `keyring` exported in a developer's shell cannot route test writes to the real keyring.
- Comments in `fallback.go` describing `auth status` and the login hint refer to behaviour that lands in later commits of this same branch; a bisect onto the fallback commit alone will find them ahead of the code.

## Open decision

On macOS and Windows the hint after a failed keyring write still leads with "If this machine has no usable OS keyring … `ENTIRE_TOKEN_STORE=file entire login`". That advice predates this branch, but the marker now makes following it stick. Platform-aware wording ("retry and allow the Keychain prompt") is a small follow-up; it is not in this PR pending an owner decision.

## Follow-ups (not in this PR)

- Platform-aware advice in `withHeadlessStoreHint` and `storeReadError` for macOS/Windows (above).
- Route `entire api` and the mirror-create wizard through `storeReadError`; both still print "Not logged in" on a broken store.
- `auth token --jurisdiction` bypasses `statusTarget` and prints the raw store error.
- Single-context `logout` silently skips the server-side revoke when the store cannot be read; `storeErr` now carries the data to warn.
- `runLogoutAll`'s warning repeats the context name after `LoginTokenForContext`'s wrap.

---

> [!NOTE]
> **High Risk**
> Changes authentication persistence, login flow selection, and auth command exit semantics on headless systems, while also rewriting git hook installation/health paths that gate checkpoint sync and push behavior.
> 
> **Overview**
> **Auth and headless login:** Token backend selection now honors `ENTIRE_TOKEN_STORE`, a remembered `token_store.json` marker, and on Linux/BSD a keyring→`tokens.json` fallback when the secret service is unavailable. `entire auth status` / `auth token` surface **store read failures** (exit 1 with remediation) instead of treating them as “not logged in”; `LoginTokenForContext` wraps missing slots with `ErrNotFound`. Login auto-picks the **device flow** when there is no graphical display on Linux/BSD, and login/store error hints reflect remembered vs path-overridden file storage.
> 
> **Git hooks (Lefthook + observability):** Setup, doctor, enable, and uninstall route through **`EnsureGitHookIntegration` / `CheckGitHookIntegration`**, classifying native vs Lefthook integrations as current, degraded, outdated, absent, or error. **`entire status`** (text and `--json`) shares a snapshot: checkpoint storage, hook health, sync state (`ready` / `degraded` / `blocked`), and per-session details without mutating stale sessions (`ListReadOnly`). **Lifecycle** runs hook repair on turn start, dedupes user-visible warnings by fingerprint, merges warnings into a single JSON hook response for Claude/Codex, and uses a **standalone plain-text path for Gemini** so warnings do not break context injection.
> 
> **Tests & docs:** Broad integration coverage for Lefthook refresh, hook repair warnings, and status JSON; README/CLAUDE.md document credential-store rules. `pre-push` CLI now accepts Git’s optional second argument (remote URL).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 718602745d6cf884d58ef51eb492d074ecfa0e21. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/entire/cli/agent/agent.go
+++ b/cmd/entire/cli/agent/agent.go
@@ -409,6 +409,25 @@ type HookResponseWriter interface {
 	WriteHookResponse(message string) error
 }
 
+// StandaloneHookResponseWriter identifies an agent whose user-visible hook
+// response cannot share the same native payload as model context injection.
+// The lifecycle dispatcher emits the warning through WriteHookResponse and
+// retries context injection on a later warning-free turn.
+type StandaloneHookResponseWriter interface {
+	HookResponseWriter
+	RequiresStandaloneHookResponse()
+}
+
+// AsStandaloneHookResponseWriter returns ag when its native protocol requires
+// a standalone user-visible hook response.
+func AsStandaloneHookResponseWriter(ag Agent) (StandaloneHookResponseWriter, bool) {
+	if ag == nil {
+		return nil, false
+	}
+	writer, ok := ag.(StandaloneHookResponseWriter)
+	return writer, ok
+}
+
 // SessionEndBudgeter is implemented by agents whose host enforces a hard
 // wall-clock budget on the session-end hook, because it runs inside the agent's
 // own shutdown sequence rather than between turns.
--- a/cmd/entire/cli/agent/geminicli/lifecycle.go
+++ b/cmd/entire/cli/agent/geminicli/lifecycle.go
@@ -13,10 +13,11 @@ import (
 
 // Compile-time interface assertions for new interfaces.
 var (
-	_ agent.TranscriptAnalyzer = (*GeminiCLIAgent)(nil)
-	_ agent.TokenCalculator    = (*GeminiCLIAgent)(nil)
-	_ agent.HookResponseWriter = (*GeminiCLIAgent)(nil)
-	_ agent.ContextInjector    = (*GeminiCLIAgent)(nil)
+	_ agent.TranscriptAnalyzer           = (*GeminiCLIAgent)(nil)
+	_ agent.TokenCalculator              = (*GeminiCLIAgent)(nil)
+	_ agent.HookResponseWriter           = (*GeminiCLIAgent)(nil)
+	_ agent.StandaloneHookResponseWriter = (*GeminiCLIAgent)(nil)
+	_ agent.ContextInjector              = (*GeminiCLIAgent)(nil)
 )
 
 // WriteHookResponse outputs a hook response message as plain text to stdout.
@@ -37,6 +38,10 @@ func (g *GeminiCLIAgent) WriteHookResponse(message string) error {
 	}
 	return nil
 }
+
+// RequiresStandaloneHookResponse preserves Gemini's plain-text response path;
+// JSON systemMessage is displayed twice by Gemini CLI.
+func (g *GeminiCLIAgent) RequiresStandaloneHookResponse() {}
 
 // InjectionEvent reports that Gemini injects model context at TurnStart (its
 // BeforeAgent hook). Gemini CLI's hook runner merges
--- a/cmd/entire/cli/doctor.go
+++ b/cmd/entire/cli/doctor.go
@@ -622,12 +622,13 @@ func checkGitHooks(cmd *cobra.Command, force bool) error {
 	ctx := cmd.Context()
 	w := cmd.OutOrStdout()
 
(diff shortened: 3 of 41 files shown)


peyton-alt:
Superseded by https://github.com/entireio/cli/pull/2495. This PR was opened from the shared `support` branch and picked up eight unrelated, unmerged hook-repair commits that were already on that branch; the replacement contains only the eight token-store commits on top of `main`.

PR #2503, opened 2026-09-16
fix: embed Windows PE VERSIONINFO in entire.exe and git-remote-entire…

### LINKED ISSUE #2218: Windows: entire.exe reports version 0.0.0.0 in PE metadata
## Problem

On Windows, `Get-Command entire` and File Explorer Properties show version `0.0.0.0`:

```
PS> Get-Command entire

CommandType     Name         Version    Source
-----------     ----         -------    ------
Application     entire.exe   0.0.0.0    C:\Users\Victor\.local\bin\entire.exe
```

`entire version` reports correctly — the issue is that the **Windows PE (Portable Executable) VERSIONINFO resource** is not embedded in the binary. Windows reads version from this resource, not from Go's ldflags. Since Go doesn't embed it by default and the GoReleaser config doesn't generate one, Windows falls back to `0.0.0.0`.

## Root cause

The version is stamped via ldflags into `versioninfo.Version` (line 25 of `.goreleaser.yaml`), which works for `entire version` at runtime. But Windows reads the `VS_FIXEDFILEINFO` / `StringFileInfo` structures from the PE resource section — a completely separate mechanism that requires a compiled `.syso` or `.rc` resource linked into the binary at build time.

### Option A: `goversioninfo` (proven in production)

[`github.com/josephspurrier/goversioninfo`](https://github.com/josephspurrier/goversioninfo) — used by [NetBird](https://github.com/netbirdio/netbird) in production. Their approach (from `.github/workflows/release.yml`):

```yaml
- name: Install goversioninfo
  run: go install github.com/josephspurrier/goversioninfo/cmd/goversioninfo@b66839b

- name: Generate windows syso amd64
  run: goversioninfo -product-name "Entire CLI"
       -copyright "Copyright (c) Entire, Inc."
       -ver-major ${{ steps.semver_parser.outputs.major }}
       -ver-minor ${{ steps.semver_parser.outputs.minor }}
       -ver-patch ${{ steps.semver_parser.outputs.patch }}
       -ver-build 0
       -file-version ${{ steps.semver_parser.outputs.fullversion }}.0
       -product-version ${{ steps.semver_parser.outputs.fullversion }}.0
       -o cmd/entire/resources_windows_amd64.syso

- name: Generate windows syso arm64
  run: goversioninfo -arm -product-name "Entire CLI" ...
       -o cmd/entire/resources_windows_arm64.syso
```

For GoReleaser, this can go in `before.hooks` or a CI step before `goreleaser release`. One invocation per arch.

Pros: proven at scale (NetBird ships it), supports icon embedding, manifest embedding, CLI flags for version override.
Cons: one invocation per arch (not a real cost — two extra lines).

### Option B: `go-winres`

[`github.com/tc-hib/go-winres`](https://github.com/tc-hib/go-winres) — newer tool, convention-based. Place `winres/winres.json` next to `main.go`, run `go-winres make --product-version=X.Y.Z`, and it generates both arch `.syso` files in one pass.

Can be integrated as a GoReleaser `before.hooks` entry:
```yaml
before:
  hooks:
    - go install github.com/tc-hib/go-winres@latest
    - go-winres make --dir cmd/entire --product-version={{.Version}} --file-version={{.Version}}
    - go-winres make --dir cmd/git-remote-entire --product-version={{.Version}} --file-version={{.Version}}
```

Pros: generates both amd64+arm64 `.syso` in one call, cleaner JSON config.
Cons: less widely adopted than goversioninfo.

### Option C: GoReleaser Pro native support

Worth checking if GoReleaser Pro has built-in `resource_compiler` or similar. The project already uses GoReleaser Pro.

## Additional benefit

Both tools can also embed an `RT_MANIFEST` (DPI awareness, UAC `asInvoker` execution level), which is best practice for CLI tools on Windows.

## Affected binaries

- `entire.exe`
- `git-remote-entire.exe`

jaysomani:
fix: embed Windows PE VERSIONINFO in entire.exe (#2218)

Windows showed 0.0.0.0 for entire.exe version in Explorer/Get-Command 
because no .syso generation step existed in the build.

Added a `mise run winres` hook in .goreleaser.yaml that runs go-winres 
for both binaries (amd64 + arm64). Go's linker picks up the .syso files 
automatically, including when cross-compiling from Linux.

Verified: entire.exe FileVersionRaw 0.0.0.0 → 0.99.5.0, native and 
cross-compiled both correct. Nightly version strings degrade gracefully.

Follows existing mise task conventions (same pattern as mise run completions).

Fixes #2218


Soph:
Approach is right and the fix works — I verified it independently. One release-breaking blocker, one dependency call for maintainers.

## Blocker: `mise-tasks/winres` is mode 644

Every other file task is `100755`; this one is `100644`. mise doesn't detect non-executable files as tasks at all (mise 2026.8.5):

```
$ mise run winres 1.2.3
mise ERROR no tasks defined ..., but found 1 non-executable file(s)
exit=1
```

GoReleaser aborts when a `before` hook fails, so the next tagged release and every nightly would die before the build pipe. Nothing catches it first: `lint:shellcheck` selects by shebang not mode (the script passes), `lint:mise` only scans `mise.toml`, and no CI job runs the task.

```
git update-index --chmod=+x mise-tasks/winres
```

Probably a filesystem artifact rather than anything you did — a WSL/DrvFs mount or `core.fileMode=false` won't carry the bit, and mise on Windows can't require it, so this would work locally and break only on the Linux release runner.

## Dependency - sorry, we sent you down the wrong path

#2218 calls go-winres a "newer tool" whose only listed downside is popularity, so following the issue lands you exactly here. That framing is now inverted: go-winres last released 2023-10 (last commit 2024-04), goversioninfo released v1.7.0 in 2026-05 and committed 5 weeks ago.

It matters because of the tree. go-winres v0.3.3 pins `golang.org/x/image v0.12.0`, and `govulncheck -mode=binary` finds one reachable symbol — `GO-2026-5031`, `bmp.Decode` panic, fixed in v0.41.0. Impact here is near nil (no images embedded, repo-controlled inputs, build-time panic at worst), but there's no fix to take: deps were updated on `main` in 2024 and never released. goversioninfo has no `x/image` and the finding disappears; cost is one extra invocation per arch, which #2218 itself calls "not a real cost."

Worth noting this PR already improved on the issue's snippet — pinning `0.3.3` beats the suggested `@latest`. I'd lean goversioninfo since this sits in the release path for signed binaries, but pinned + sumdb-verified + deterministic is defensible. Maintainers' call; I'll fix #2218 either way.

## Nits

- No regression test - `e2e-windows.yml` step asserting `(Get-Item entire.exe).VersionInfo.FileVersion` would cover the rest.
- Commit is authored as `Your Name <jaysomani2016@gmail.com>` (git's default placeholder) — worth setting `user.name` before merge.

jaysomani:
@Soph Thanks for the thorough review! Fixed both issues:

Marked mise-tasks/winres as executable (chmod +x)  was a WSL/DrvFs filesystem artifact as you suspected.
Corrected the git author name.

On the goversioninfo vs go-winres question 
happy to switch if that's the maintainers' preference. And can add the e2e-windows.yml regression test too if wanted. Just let me know.

gtrrz-victor:
> @Soph Thanks for the thorough review! Fixed both issues:
> 
> Marked mise-tasks/winres as executable (chmod +x) was a WSL/DrvFs filesystem artifact as you suspected. Corrected the git author name.
> 
> On the goversioninfo vs go-winres question happy to switch if that's the maintainers' preference. And can add the e2e-windows.yml regression test too if wanted. Just let me know.

Thanks for addressing @Soph's feedback. It'd be great if you could use `goversioninfo` instead of `go-winres`, and also add the e2e test to the Windows workflow as he pointed out.

jaysomani:
Thanks! @gtrrz-victor  will switch to goversioninfo and add the e2e Windows workflow test. Will push shortly.

jaysomani:
Hey @gtrrz-victor Added the PE metadata assertion to ci.yml's test-windows job rather than e2e-windows.yml — the e2e workflow has no build step so entire.exe wouldn't exist there to inspect. Does that work, or would you prefer it somewhere else? please confirm

gtrrz-victor:
Thanks for the goversioninfo switch and the chmod — both landed cleanly, and pinning `v1.7.0` exactly rather than floating is right for something sitting in the release path.

We pulled this onto a branch and tested it on a real Windows 11 machine. The fix works: the task runs fine there, and `Get-Command entire.exe` returns a real version instead of `0.0.0.0`. Four changes before we merge.

## 1. Delete `winres/` and rename the task to `winversion`

`winres` is go-winres's name and `winres/` is its directory convention — both are left over from the first approach, and neither is needed now.

`winres/app.manifest` should go regardless. #2218 suggested it and that suggestion was wrong — we'll fix the issue. We verified it changes no observable behaviour: `requestedExecutionLevel asInvoker` only suppresses UAC installer detection, which applies to 32-bit executables only, and `dpiAware`/`dpiAwareness` do nothing for a console binary. A manifest is a separate shipped behaviour change and belongs in its own PR if we ever want one.

`winres/versioninfo.json` isn't needed either. We diffed both routes on Windows: `-skip-versioninfo` with the flags below produces a `.syso` byte-identical to the template path except for `FileFlagsMask`/`FileOS`/`FileType`, which nothing in our stack reads and neither PowerShell nor Explorer exposes. Every rendered row matched, Language included.

`winversion` rather than `versioninfo`, which would collide with the `cmd/entire/cli/versioninfo` package.

`mise-tasks/winversion` in full:

```bash
#!/bin/bash
#MISE description="Generate Windows PE VERSIONINFO resources (.syso) for entire and git-remote-entire"
#MISE quiet=true

set -euo pipefail

version="${1:?usage: mise run winversion <version>}"

base="${version%%-*}"
IFS='.' read -r major minor patch <<<"$base"
major="${major:-0}"
minor="${minor:-0}"
patch="${patch:-0}"

generate() {
    local pkg="$1" desc="$2" original_name="$3"
    local common_flags=(
        # No versioninfo.json: every field this task sets has a flag. The two
        # below are decimal for 0x0409 / 0x04B0 — omit them and the binary
        # ships an empty translation block and Windows shows no Language.
        -skip-versioninfo
        -translation 1033 -charset 1200
        -company "Entire"
        -copyright "Copyright (c) 2026 Entire Inc."
        -description "$desc"
        -product-name "Entire CLI"
        -internal-name "$original_name"
        -original-name "$original_name"
        -ver-major "$major" -ver-minor "$minor" -ver-patch "$patch"
        -product-ver-major "$major" -product-ver-minor "$minor" -product-ver-patch "$patch"
        -file-version "$version"
        -product-version "$version"
    )
    goversioninfo -64 -o "${pkg}/rsrc_windows_amd64.syso" "${common_flags[@]}"
    goversioninfo -64 -arm -o "${pkg}/rsrc_windows_arm64.syso" "${common_flags[@]}"
}

generate "cmd/entire" "Entire CLI" "entire.exe"
generate "cmd/git-remote-entire" "Git remote helper for entire:// URLs" "git-remote-entire.exe"
```

Changes from your version, all visible above: `-manifest` gone; `-skip-versioninfo` and `-translation`/`-charset` replace the JSON; `-internal-name` added, which was unset (confirmed absent from the built PE); `-copyright` aligned with LICENSE, which says `Copyright (c) 2026 Entire Inc.`; the trailing config-file argument dropped from both calls. Everything else is yours, unchanged.

And the hook:

```diff
 before:
   hooks:
     - go mod tidy
     - mise run completions
-    - mise run winres {{.Version}}
+    - mise run winversion {{.Version}}
```

## 2. Move the check out of `ci.yml` into `nightly-e2e.yml`

The steps in `ci.yml` generate resources and compile on the runner, which tests the task rather than the release — and they add a full `mise install` to a job budgeted `timeout-minutes: 10`. `nightly-e2e.yml` already installs the published nightly with `install.ps1`, puts it on `PATH`, and has just resolved the tag it came from, so the check costs one step there and can compare against that tag rather than merely asserting a version exists.

Remove the added block from `.github/workflows/ci.yml`, restoring `test-windows` to main:

```diff
       # test name so this job selects them without running POSIX-only harnesses.
       - name: Windows unit tests
         run: go test ./cmd/entire/cli/... -run '(Windows|MSYS)' -count=1 -v
-      - uses: jdx/mise-action@f10502fc09dadecfefb962fff68ce77213930204 
-      - name: Generate Windows PE version resources
-        run: mise run winres 9.9.9-ci-verify
-      - name: Build entire.exe with embedded PE resources
-        run: go build -o entire.exe ./cmd/entire/
-      - name: Verify Windows PE version metadata
-        shell: pwsh
-        run: |
-          $info = (Get-Item .\entire.exe).VersionInfo
-          Write-Host "FileVersion: $($info.FileVersion)"
-          Write-Host "ProductName: $($info.ProductName)"
-          $raw = $info.FileVersionRaw
-          if ($raw.Major -eq 0 -and $raw.Minor -eq 0 -and $raw.Build -eq 0 -and $raw.Revision -eq 0) {
-            Write-Error "PE version metadata is still 0.0.0.0 -- .syso generation or linking failed (see #2218)"
-            exit 1
-          }
-          Write-Host "PE version metadata populated correctly ($raw)"
 
   test:
     runs-on: ubuntu-latest
```

and add one step to `.github/workflows/nightly-e2e.yml`, after `Resolve nightly tag and binary path (Windows)`:

```diff
           "NIGHTLY_TAG=$tag" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
           "E2E_ENTIRE_BIN=$((Get-Command entire).Source)" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
 
+      # Windows reads the version from the binary's PE resource section, not
+      # from the ldflags stamp `entire version` prints (#2218). Compared against
+      # the tag this nightly was built from, so resources generated for the
+      # wrong version fail here too, not only missing ones.
+      - name: Verify Windows PE version metadata
+        if: matrix.os == 'windows-latest'
+        shell: pwsh
+        run: |
+          # Stop is what makes the Write-Error calls below terminating.
+          $ErrorActionPreference = "Stop"
+          # goreleaser passes {{.Version}}, which is the tag without its "v".
+          $expected = $env:NIGHTLY_TAG -replace '^v', ''
+          # The numeric tuple carries major.minor.patch only; the prerelease
+          # suffix lives in the string fields.
+          $expectedNumeric = [version]"$($expected -replace '-.*', '').0"
+          foreach ($name in "entire", "git-remote-entire") {
+            $command = Get-Command $name
+            $product = (Get-Item $command.Source).VersionInfo.ProductVersion
+            Write-Host "$name  $($command.Version)  $product"
+            if ($command.Version -ne $expectedNumeric) {
+              Write-Error "$name PE version is $($command.Version), expected $expectedNumeric from $env:NIGHTLY_TAG"
+            }
+            if ($product -ne $expected) {
+              Write-Error "$name PE product version is '$product', expected '$expected'"
+            }
+          }
+
       - name: Checkout the nightly tag
```

Both binaries, because the task generates resources for both and both ship. Two assertions because they fail differently: `Get-Command`'s `Version` is the surface #2218 was reported on and reads `0.0.0.0` when the resource is missing, while `ProductVersion` is the only place the nightly suffix appears and is what catches resources stamped with the wrong version.

## 3. Narrow the gitignore

Go links a `.syso` only from the package directory it compiles — one placed anywhere else produces a binary with no resource section at all, silently — so those two paths are the only ones that can ever receive one.

```diff
-# Generated Windows PE resource objects (mise run winres)
-*.syso
+# Generated Windows PE resource objects (mise run winversion). Go links a .syso
+# only from the package directory it compiles, so these two paths are the only
+# places they can be written.
+cmd/entire/*.syso
+cmd/git-remote-entire/*.syso
```

## 4. Update the PR title and body

Both still describe go-winres and `mise run winres`, and the body becomes the squash commit message. Suggested:

> **fix: embed Windows PE VERSIONINFO in entire.exe and git-remote-entire**
>
> Windows reads a binary's version from the PE VERSIONINFO resource, not from Go's ldflags, and Go embeds none by default — so Explorer and `Get-Command` reported `0.0.0.0` while `entire version` was correct.
>
> Adds a `mise run winversion <version>` task that generates `.syso` resources for both binaries on both architectures with goversioninfo, called from `.goreleaser.yaml`'s before hooks. Go's linker picks them up from the package directory automatically, cross-compiling from Linux included.
>
> Nightlies report the numeric tuple of their patch version — `0.10.7.0` — with the full tag in the Product version string; Explorer's "File version" row and `Get-Command` read only the numeric tuple.
>
> Follows the existing mise task conventions (same pattern as `mise run completions`).
>
> Fixes #2218

jaysomani:
ok will push this fixes and update it

jaysomani:
@gtrrz-victor done

gtrrz-victor:
@jaysomani Thanks for pushing this PR forward! LGTM!

gtrrz-victor (approved):

Issue #2410, opened 2026-09-14
`entire enable --force` writes the hook wrapper through a symlink, clobbering the target file and creating a self-chaining hook


edgar-bestow:
### Summary

When a git hook path is a **symlink** (a common pattern for repo-managed hooks, where `.git/hooks/pre-push` is symlinked to a script checked into the repository) and a `<hook>.pre-entire` backup already exists from a previous install, `entire enable --force` writes its wrapper **through** the symlink instead of replacing it.

This has two consequences:

1. **Data loss** — the file the symlink points at (a version-controlled script in the working tree) is overwritten with the wrapper's contents.
2. **Infinite recursion** — `<hook>` and `<hook>.pre-entire` end up resolving to the same file, so the wrapper's "chain to pre-existing hook" call re-executes itself. Every subsequent commit/push forks shells without bound until the machine runs out of processes.

Verified on Entire CLI 0.10.6 (`go1.26.6`, `darwin/arm64`).

### Reproduction

```sh
mkdir repro && cd repro && git init
mkdir scripts
printf '#!/bin/sh\necho "PROJECT HOOK RAN"\n' > scripts/post-commit.sh
chmod +x scripts/post-commit.sh

# Repo-managed hook, installed as a symlink
ln -nsf "$PWD/scripts/post-commit.sh" .git/hooks/post-commit

# 1st enable — correct behaviour
entire enable --force
# [entire] Backed up existing post-commit to post-commit.pre-entire
#   post-commit             -> regular file (the wrapper)
#   post-commit.pre-entire  -> symlink to scripts/post-commit.sh

# Anything that re-installs the repo's hooks now re-creates the symlink,
# while the .pre-entire backup is still present:
ln -nsf "$PWD/scripts/post-commit.sh" .git/hooks/post-commit

# 2nd enable — the bug
entire enable --force
# [entire] Warning: replacing post-commit (backup post-commit.pre-entire already exists from a previous install)
```

### Actual result

`.git/hooks/post-commit` is still a symlink, and the wrapper was written straight through it:

```
$ cat scripts/post-commit.sh
#!/bin/sh
# Entire CLI hooks
# Post-commit hook: condense session data if commit has Entire-Checkpoint trailer
if command -v entire >/dev/null 2>&1; then entire hooks git post-commit 2>/dev/null || true; else :; fi
# Chain: run pre-existing hook
_entire_hook_dir="$(dirname "$0")"
[...]

PR #2534, opened 2026-09-19
test(strategy): cover symlinked hook with an existing backup (#2410)

KC1706:
## What

Adds a regression test for #2410 — a symlinked hook that also has a pre-existing
`<hook>.pre-entire` backup (the second-`enable` path).

**This is a test-only change.** #2410 is already fixed on `main`: `writeHookFile`
writes via `jsonutil.WriteFileAtomicIn` (temp file + `rename`, which replaces a
leaf symlink instead of following it), and `classifyExistingHook` treats a symlink
as foreign. 0.10.6's bare `os.WriteFile` followed the symlink. So this doesn't
change behaviour — it pins the fixed behaviour so it can't silently regress.

## Why the existing tests don't cover it

Two tests each cover only half of the reported scenario:

- `TestInstallGitHook_SymlinkedHookIsBackedUpNotFollowed` — symlink, but **no**
  backup, so it takes the rename-aside (`!backupExists`) branch.
- `TestInstallGitHook_DoesNotOverwriteExistingBackup` — a backup exists, but the
  hook is a **regular file**, so the write can never follow a link.

Neither exercises the `backupExists` branch **with a symlink at the hook path** —
the exact #2410 state. A regular-file replace behaves identically under both the
buggy (`os.WriteFile`) and fixed (`rename`) write, so that test can't detect
write-through.

## The test

`TestInstallGitHook_SymlinkedHookWithExistingBackupIsNotFollowed` reproduces the
reporter's setup (hook and backup both symlinked to the same repo-managed script)
and asserts both documented consequences are prevented:

- the symlink target is not written through (no data loss);
- `<hook>` and `<hook>.pre-entire` do not resolve to the same file (no
  self-chaining recursion);

plus that the installed hook is a regular file and the existing backup is preserved.

## Verification

Reverting `writeHookFile` to the 0.10.6 `os.WriteFile` write-through shape makes the
test fail on exactly those assertions; on `main` it passes.

- `go test -tags=integration -race ./cmd/entire/cli/strategy/` → ok
- `mise run lint` → 0 issues

Closes nothing on its own (the fix already merged); complements #2410 by locking the
scenario down.


KC1706:
@Soph could you reveiew this?

